PR #3, opened 2016-02-09
Add unit test boilerplate

Mec-iS:
Any good repo needs unit testing :+1:

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,0 +1,2 @@
+# coding=utf-8
+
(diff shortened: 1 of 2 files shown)


vrde:
Hey @Mec-iS, thanks a lot for your pull request.

Unfortunately I have to decline it because we have a test suite that we will push by end of the day (together with the first working code). We still need some help to improve the test coverage, so if you want to give it a look later we the code is online it would be nice :clap:

PR #14, opened 2016-02-11
Remove no longer used requirements.txt

sbellem:
Use `python setup.py develop` instead

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,0 @@
--r requirements/common.txt


vrde:
> Use python setup.py develop instead

or `pip install -e .`, should be the same, right?

sbellem:
> or pip install -e ., should be the same, right?

yes

vrde:
@sbellem: there are some requirements we don't use anymore, like
```
repoze.lru==0.6
fake-factory==0.5.3
tornado==4.3
```

If it's not in the scope of this PR, can you open a new issue please?

PR #22, opened 2016-02-12
Move db exception to the right module

vrde:
Since all the exceptions are in the `bigchaindb.exceptions` module, I moved the one in `bigchaindb.db.utils` there.


sbellem:
:+1: Will merge now since we know all tests passed, as travis is having problems.

PR #30, opened 2016-02-12
Release 0.1.1

sbellem:


r-marques:
Its true that we need to bump the version number but we also need to define what we want to include in the new version.

We could always have a pull request with the new version number and in the pull request reference all issues, and pr that need to be closed for a new release to be made. This way this pull request would also work as a changelog

r-marques:
merging this since we already release the package

PR #33, opened 2016-02-14
Fix some config issues for docker-compose

sbellem:
Since containers are ephemeral, the config file that gets created under `~/.bigchaindb` will be wiped out along with the container each time a new container is started.

We need a way to mount this file, or to only rely on environment variables, which would be probably easier.

For the time being this PR is a workaround, such that it mounts a docker specific config file from the host.

Addresses issue #36.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.cache/
+.coverage
+.eggs/
+.git/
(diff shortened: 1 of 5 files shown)


vrde:
:+1:

PR #40, opened 2016-02-15
Correct Rethinkdb buggy version in setup.py

Mec-iS:
The version of RethinkkDB `post1` in `setup.py` makes impossible to run `bigchaindb-benchmark load` because of error:
```
  File "/usr/lib/python3/dist-packages/pkg_resources.py", line 444, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3/dist-packages/pkg_resources.py", line 725, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3/dist-packages/pkg_resources.py", line 632, in resolve
    raise VersionConflict(dist,req) # XXX put more info here
pkg_resources.VersionConflict: (rethinkdb 2.2.0-1 (/usr/local/lib/python3.4/dist-packages), Requirem
ent.parse('rethinkdb==2.2.0.post1'))
```
I changed to `2.2.0-1` as required by the package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
         ],
     },
     install_requires=[
-        'rethinkdb==2.2.0.post1',
+        'rethinkdb==2.2.0-1',
         'pysha3==0.3',
         'pytz==2015.7',
         'cryptography==1.2.1',


vrde:
@Mec-iS apparently this is related to a bug in `pip`. Which version of `pip` do you use?

sbellem:
The [documentation for packaging](http://python-packaging-user-guide.readthedocs.org/en/latest/requirements/#install-requires) states:

I don't know if that will resolve this issue, but we may consider following the above recommended practice.

Mec-iS:
@sbellem I agree, this PR wasn't meant to be a permanent fix, but a way to make installation working asap.

@vrde You were right. setuptools is too old also.

Mec-iS:
see #41

Issue #46, opened 2016-02-15
Change instructions for installing pip3+ on Ubuntu 14.04


ttmc:
The current instructions for installing pip3+ on Ubuntu 14.04 (in Section 1.3.1 of the docs) do install pip3+, but it's a really old version of pip3 (which causes problems), and there are issue with upgrading it. We found that the following is a much more reliable way to install a recent version of pip3+:
```text
sudo apt-get install python3-setuptools
sudo easy_install3 pip
```
Bonus: Using this, the `pip` command is actually the Python 3 version of pip, which is the version we want to use. (You can check the `pip` version using `pip -V`.)

ttmc:
Resolved by PR #47

PR #48, opened 2016-02-15
Add sudo before pip install bigchaindb sometimes

ttmc:
It turns out that you sometimes must use `sudo pip install bigchaindb` (e.g. when installing on Ubuntu 14.04 on AWS), so that was changed in the installation instructions.


ttmc:
A follow-up to issue #46 and PR #47

PR #52, opened 2016-02-16
Update requirements, ready for v0.1.3

### LINKED ISSUE #41: Make the required version of rethinkdb more general
Currently, setup.py specifies a very specific version of rethinkdb (Python package):

```
'rethinkdb==2.2.0.post1',
```

That's causing some issues already because when I try to install rethinkdb=2.2.0.post1, I get this message: `Requested rethinkdb==2.2.0.post1, but installing version 2.2.0-1`

Then, when I go to run `bigchaindb start`, it fails because it requires `rethinkdb==2.2.0.post1` but 2.2.0-1 is installed.

@vrde suggested changing the setup.py requirement to something more general:

```shell
'rethinkdb>=2.2,<2.3',
```

I guess that `bigchaindb` should also be changed to have the same, looser requirement.

vrde:
Close #41

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ dev_require = [
 
 setup(
     name='BigchainDB',
-    version='0.1.2',
+    version='0.1.3',
     description='BigchainDB: A Scalable Blockchain Database',
     long_description=__doc__,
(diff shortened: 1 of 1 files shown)


r-marques:
:+1:

PR #60, opened 2016-02-17
Add docker setup docs

sbellem:


sbellem:
@ttmc Regarding the `json` document structure I removed the dots (`...`) because they cause the following error when building the documentation:

```bash
docs/source/models.md:: WARNING: Could not parse literal_block as "json". highlighting skipped.
```

Please let me know if you think the proposed example should be improved, or corrected.

ttmc:
@sbellem I understand the reason for removing the dots from the .rst file, but the proposed example gives the impression that the "data" section will be several hashes. A better example would have one "hash" (with value) and then some other JSON. Maybe something like this:
```JavaScript
    "hash": "(SHA3-256 hash hexdigest of the other contents of the other things in data)",
    "title": "The Winds of Plast",
    "creator": "Johnathan Plunkett",
    "IPFS_key": "QmfQ5QAjvg4GtA3wg3adpnDJug8ktA1BxurVqBD8rtgVjP"
```

r-marques:
seems good :+1:

Issue #56, opened 2016-02-16
handle pep8 tool complaint about E712


sbellem:
See pep8 https://www.python.org/dev/peps/pep-0008/#programming-recommendations

> Don't compare boolean values to `True` or `False` using `==` .
> 
> Yes:   `if greeting:`
> No:    `if greeting == True:`
> Worse: `if greeting is True:`

--

**EDIT**

Thanks to @Andygmb for asking clarifications, and to @genecyber for commenting. The above pep8 excerpt may not be the ideal one. This one may be better:

> Comparisons to singletons like None should always be done with is or is not , never the equality operators.

although I am not fully sure, such that if one wishes to insist on referring to pep8, then the section [A Foolish Consistency is the Hobgoblin of Little Minds](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds) is perhaps the best one for this case. E.g.:

> [...] But most importantly: know when to be inconsistent -- sometimes the style guide just doesn't apply. When in doubt, use your best judgment. Look at other examples and decide what looks best. And don't hesitate to ask! [...]

### Context for this issue
We have some tests that check:

`assert x == True`

the pep8 tool (version 1.7.0) will complain with:

```bash
E712 comparison to True should be 'if cond is True:' or 'if cond:'
E712 comparison to False should be 'if cond is False:' or 'if not cond:'
```

The following post is related: http://stackoverflow.com/questions/21019797/pep8-minor-annoyance, along with the answer http://stackoverflow.com/a/21019934/2769475, which helps understand the issue.

So in order to solve this issue, one may run the pep8 tool against the tests directory, and for each E712 error, one may first check whether it is necessary at all to make an explicit verification. That is, a simple

```python
assert x
```

or 

```python
assert not x
```

may be sufficient.

On the other hand, if the case requires an explicit check, such as `y` must exactly be `True`, it seems that the check:

```python
assert y is True
```

would be better ... since `y = 1` should evaluate to `False`. 

Said differently, one should keep in mind what the above mentioned stackoverflow [answer](http://stackoverflow.com/a/21019934/2769475) shows:

```python
>>> assert 1
>>> assert 1 == True
>>> assert 1 is True
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError

>>> assert not 0
>>> assert 0 == False
>>> assert 0 is False
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #63, opened 2016-02-18
Replace `==` with `is` for some tests

vrde:
Replace all occurrences of:
 - `== True` with `is True`
 - `== False` with `is False`
 - `== None` with `is None`


vrde:
Related: #56

sbellem:
:+1:

Issue #58, opened 2016-02-17
Non interactive first start


roderik:
Is there a way to start the bigchaindb for the first time in non interactive mode? "-y" does not work in this case it seems

r-marques:
No. Currently when you start bigchaindb for the first time it checks for the existence of `~/.bigchaindb` config file. It it does not exist it will enter the interactive mode.

One way around it would be to manually create the config file. Then when calling `bigchaindb start` for the first time it will be non interactive.

Another way to create a config file would be using `bigchaindb -c myconfig.json configure` and then call `bigchaindb -c myconfig.json start`

I will propose this an enhancement so that when calling `bigchaindb -y start` for the first time it will create a config file with the default values

roderik:
I'm using a default config to be able to create a docker image https://github.com/Kunstmaan/docker-bigchaindb/blob/master/bigchaindb but i don't like the priv and pub key values in a default config. A simple change could be that it will generate and add it to a config file if the values are not found?

vrde:
@roderik: I made some changes to support the `-y` option when using `start`, check #64. I'll push it in PyPI soon, will ping you when it's done.

vrde:
@roderik the new release is PyPI:
 - https://pypi.python.org/pypi/BigchainDB

Can you give it a try and tell me if it solves your problem?

PR #64, opened 2016-02-18
Allow non interactive first start

vrde:
If no configuration is found, `bigchaindb start` will prompt the user for some configuration values (database host, port, and name) and will generate the keypair for the node. This patch allows to run `bigchaindb start` with the `-y` option to run it in a non-interactive way.

Related: #58.


r-marques:
seems good :+1:

PR #65, opened 2016-02-18
Improve test with xdist

vrde:
This patch allow running tests in parallel, saving ~50% of the time (in my machine I went from 77s to 30s, that's more than 60%!).

Check [xdist documentation](http://pytest.org/latest/xdist.html) for more technical details.


vrde:
On Travis-ci tests went from 73s down to 28s.

Issue #59, opened 2016-02-17
Create a BigchainDB Glossary of Terms


ttmc:
Feel free to comment with terms you want included.

ttmc:
I've started writing a Glossary section in the Bigchain Documentation. I will be away on Thursday and Friday, but don't worry, I won't forget about it!

Mec-iS:
Definitely a good idea.
Let's start from the beginning:
- block
- transaction
- vote (voting system)

ttmc:
A first version of the Glossary has been merged via PR #69

PR #69, opened 2016-02-22
Added v1 of Glossary

ttmc:
This is the first version of the Glossary. Many more terms could be added, but this is a start.

It resolves Issue #59


vrde:
Looks good to me.

PR #72, opened 2016-02-22
Monitoring for BigchainDB

rhsimplex:
Add monitoring/statistics aggregation to BigchainDB.  This PR has essentially two parts:

1. Add statistics gathering to core BigchainDB code with statsd, and
2. Provide a sample analytics environment so you can actually do something with the data.

The first part can be reviewed directly. To the second part, I added a second docker compose file, `docker-compose-monitor.yml`, that sets up the analytics infrastructure locally.  The code for the custom dashboard, for example, lives in a separate repository.

Instruction for setup is in the modified `README.md`, so please make sure it works for you!


r-marques:
@rhsimplex can you put the documentation inside the docs. You can create a new `md` file and add a new entry on the TOC. We want to keep the README small

rhsimplex:
@r-marques ok so add a new .md to `docs/source` -- but I don't know where the TOC is

rhsimplex:
@r-marques done

docs/source/monitoring.md
```diff
@@ -0,0 +1,16 @@
...
+BigchainDB uses [statsd](https://github.com/etsy/statsd) for monitoring.  To fully take advantage of this functionality requires some additional infrastructure: an agent to listen for metrics (e.g. [telegraf](https://github.com/influxdata/telegraf)), a time-series database (e.g. [influxdb](https://influxdata.com/time-series-platform/influxdb/), and a frontend to display analytics (e.g. [Grafana](http://grafana.org/)).
+
+For ease of use, we've provided a docker compose file that sets up all these services for testing. Simply run in the BigchainDB directory:
+
+```text
+$ docker-compose -f docker-compose-monitor.yml build
+$ docker-compose -f docker-compose-monitor.yml up
+```
+
+and point a browser tab to `http://localhost:3000/dashboard/script/bigchaindb_dashboard.js`.  Login and password are `admin` by default.  If BigchainDB is running and processing transactions, you should see analytics—if not, start BigchainDB as above, and refresh the page after a few seconds.
```
r-marques: better to replace `start BigchainDB as above` and provide a link to `installing.html#run-bigchaindb`

rhsimplex: good point!

r-marques:
I am affected by this bug https://github.com/grafana/grafana-docker/issues/27#issuecomment-184927849

Can you update your docker file with the change in the comment?

r-marques:
seems good to me :+1: 
@vrde any thoughts?

rhsimplex:
@sbellem suggests putting the custom docker containers under bigchaindb on dockerhub.  I agree but want to close this PR first.

bigchaindb/core.py
```diff
@@ -10,6 +10,9 @@
 from bigchaindb import config_utils
 from bigchaindb import exceptions
 from bigchaindb.crypto import hash_data, PublicKey, PrivateKey, generate_key_pair
+from bigchaindb.monitor import Monitor
+
+c = Monitor()
```
vrde: Give a more descriptive name to `c`, I think `monitor` is fine.

bigchaindb/block.py
```diff
@@ -4,11 +4,16 @@
...
 
+import bigchaindb
 from bigchaindb import Bigchain
+from bigchaindb.monitor import Monitor
+
 
 
 logger = logging.getLogger(__name__)
 
+c = Monitor()
```
vrde: Give a more descriptive name to `c`, I think `monitor` is fine. (Yup, we will give a more descriptive name also to `b` one day)

bigchaindb/monitor.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if not kwargs:
+            kwargs = {}
+
+        # set prefix, parameters from configuration file
+        if not kwargs.get('prefix'):
+            kwargs['prefix'] = '{hostname}.'.format(hostname=node())
+        if not kwargs.get('host'):
+            kwargs['host'] = bigchaindb.config['statsd']['host']
+        if not kwargs.get('port'):
```
vrde:
```python
if 'prefix' not in kwargs:
    ...
if 'host' not in kwargs:
    ...
if 'port' not in kwargs:
    ...
```

bigchaindb/monitor.py
```diff
@@ -0,0 +1,30 @@
...
+            kwargs['port'] = bigchaindb.config['statsd']['port']
+        super().__init__(*args, **kwargs)
```
sbellem: I wonder, could the above class be simplified to something like:

```python
PREFIX = '{hostname}.'.format(hostname=node())
HOST = bigchaindb.config['statsd']['host']
PORT = bigchaindb.config['statsd']['port']


class Monitor(statsd.StatsClient):
    """Set up statsd monitoring

    """
    def __init__(self, prefix=PREFIX, host=HOST, port=PORT, *args, **kwargs):
        """Overrides statsd client, fixing prefix to messages and loading configuration

        Args:
            *args: arguments (identical to Statsclient)
            **kwargs: keyword arguments (identical to Statsclient)
        """
        config_utils.autoconfigure()

        kwargs.update(dict(prefix=prefix, host=host, port=port))

        # set prefix, parameters from configuration file
        super().__init__(*args, **kwargs)
```

sbellem: pushing the simplification further:

```python
PREFIX = node() + '.'
HOST = bigchaindb.config['statsd']['host']
PORT = bigchaindb.config['statsd']['port']


class Monitor(statsd.StatsClient):
    """Set up statsd monitoring

    """
    def __init__(self, prefix=PREFIX, host=HOST, port=PORT, *args, **kwargs):
        """Overrides statsd client, fixing prefix to messages and loading configuration

        Args:
            prefix: ...
            host: ...
            port: ...
            *args: arguments (identical to Statsclient)
            **kwargs: keyword arguments (identical to Statsclient)
        """
        config_utils.autoconfigure()

        # set prefix, parameters from configuration file
        super().__init__(prefix=prefix, host=host, port=port, *args, **kwargs)
```

vrde: Unfortunately in your code the constants `PREFIX`, `HOST`, and `PORT` are initialized when the module is loaded, and it might be the wrong time to do it because the configuration might not be ready.

sbellem: Yes, I had overlooked that line, and could it be taken out of the class?

vrde: I think it's feasible. Instead of subclassing we can have a function to create the instance, for example.

vrde:
:+1: for me

PR #76, opened 2016-02-23
Bug/71/voter wrong input validation

### LINKED ISSUE #71: Voter is not validating blocks correctly when checking for double spends
When a voter is validating a transaction it checks if the input of a transaction was already spent. To check if the input was already spent it checks if there is any transaction in the bigchain that already used that input. Since when voting the block is already in the bigchain it considers the first transaction using that input has a double spend.

We need to add a check so that the voter does not consider the current block when checking if there is already a transaction in the bigchain with that input.

r-marques:
closes #71

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -269,30 +269,18 @@ class Bigchain(object):
             txid (str): transaction id.
 
         Returns:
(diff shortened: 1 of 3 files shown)


vrde:
:+1: looks good to me.

PR #78, opened 2016-02-23
bigchain -> BigchainDB

gmcmullen:


ttmc:
As @r-marques suggested on Slack, maybe just change bigchain -> BigchainDB in line 506 for now (i.e. in the GENESIS block). We can consider changing other instances of "bigchain" in the future.

ttmc:
PR #82 is the new version of this PR so I'll close this one without merging.

PR #79, opened 2016-02-23
Added protection for discrimination based on species

gmcmullen:
Protects Wrigley!


r-marques:
seems good to me :+1: 
@ttmc thoughts?

diminator:
what about singularities?

ttmc:
Seems okay,

PR #95, opened 2016-02-29
docker images now come from bigchaindb dockerhub account

rhsimplex:
moved docker containers into the bigchaindb dockerhub account -- just reflects those changes

### DIFF
--- a/docker-compose-monitor.yml
+++ b/docker-compose-monitor.yml
@@ -10,7 +10,7 @@ influxdb:
     PRE_CREATE_DB: "telegraf"
 
 grafana:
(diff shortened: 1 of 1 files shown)


r-marques:
since we are at it would it make sense to move the source under bigchaindb organization on github?

rhsimplex:
Good question. Asked @sbellem and @vrde and got different opinions =)

Is it ok to clutter the bigchaindb organization page, which is just the database and whitepaper, with a bunch of docker images?

r-marques:
Lets leave it like this for now. We can then discuss how to organize our docker stuff.

Issue #93, opened 2016-02-29
Improve documentation for non-linux users


r-marques:
We need to state in the documentation that we are currently only supporting linux and point windows users to the vagrant instructions an mac osx users to the docker deloyment

ttmc:
I will do this.

ttmc:
Resolved by PR #96

PR #96, opened 2016-02-29
Improve install docs for non-Linux users

ttmc:
This is to resolve issue #93 

I also added an explanation of _why_ we're starting with Ubuntu 14.04 (for BigchainDB _Server_).


r-marques:
seems good to me :+1:

ttmc:
I just added a comment about how there are no clients/drivers at the moment, but they're coming soon and they'll be able to run on a wide variety of OSes. Assuming the tests pass (!), I'll merge this PR.

PR #97, opened 2016-02-29
Fix bug related to config overwrite

vrde:
Calling `bigchaindb configure` would overwrite the original config even if the user answered `n`.

This PR is to fix this wrong behavior.


r-marques:
:+1:

Issue #90, opened 2016-02-26
Plugin or hook architecture e.g. for validate_transaction


ttmc:
In the [bigchaindb/bigchaindb chat room on Gitter](https://gitter.im/bigchaindb/bigchaindb), @thedoctor asked an interesting question: 

> ...the question I was getting at is whether folks thought a plugin or hook architecture would be of general utility? I'm imagining hooks in key places in `Bigchain` – at least `validate_transaction` and `validate_block`. It would be easy for the community to contribute arbitrary consensus plugins. Make sense?

I know there has been some internal discussion about that idea (i.e. at ascribe GmbH) and said as much on Gitter. I wonder how doable it is, or if there are any big gotchas. I said I'd open an issue about it on GitHub (i.e. _this_ issue). Any thoughts @r-marques @diminator @vrde @andreasATascribe ?

@thedoctor replied:

> Cool – I'll wait for some design discussion there [on GitHub] then I'm happy to implement it and PR.

Great!

r-marques:
The consensus protocol will probably be one of the most difficult parts in bigchaindb so it will be important for us to try to decouple it from the rest of the code as much as possible, so that we can improve it or enhance it without interfering with the rest of the code. So in the end ideally we would have a consensus layer that would be a plug able plugin

ttmc:
Maybe @thedoctor could start by writing a different version of `validate_transaction`, with extra bells and whistles for some particular kind of transaction. The idea is that it's easier to generalize from a couple of examples.

thedoctor:
So if this is a good idea, then some things we should probably discuss:
1. What operations are delegated to the consensus plugin?
  * Are there parts of the consensus process that are considered fundamental and should be kept in core?
    * e.g. Is the requirement that secp256k1 is used for signatures just a consensus rule? (I think so) 
    * On the other hand, what about the serialization scheme? (That seems reasonable to standardize)
2. What data is exposed to a plugin? It's probably unwise to expose rethink itself as that would make database optimization difficult.
3. How are plugins invoked? Naively, I would suggest pulling out the existing functionality into a base plugin from which other plugins could inherit and delegating consensus-related functions to that object.
4. Should we provide a mechanism for chaining plugins like middleware or leave it to the developers to compose their functionality?

ttmc:
@thedoctor It's not clear to me what you have in mind. Maybe if you write some code and submit a pull request (including explanations of various decisions), we could have a more concrete understanding. 

[...]

PR #99, opened 2016-03-01
Modular consensus rules

thedoctor:
This PR is a simple (and incomplete) implementation of the system discussed in issue #90 

I used the setuptools entry_point functionality, which means consensus rule plugins can be distributed over PyPI and declared as dependencies. 
Which rule set to use is determined by a list of names in the bigchain config `consensus_plugins` variable.

This implementation only exposes `Bigchain.validate_block` and `Bigchain.validate_transaction`. Those methods in the `Bigchain` class have been replaced with sequential calls to methods with the same names in each loaded plugin – if any of them raise, the call fails.

Regarding other components to expose in the plugin API, I would want to at least include the signature operations as well, which would make it easy to swap out a chain's crypto system.


ttmc:
Thanks @thedoctor - it may take us a while to get around to looking at this PR, as we're a small team and we've got many irons on the fire. We appreciate your enthusiasm!

In the meantime, we need to get your permission (i.e. a license) to include your contributions in our code, and to sublicense them to others (since you'll retain ownership of the copyright). To do that, we need you to agree to our Contributor License Agreement (CLA). To do that, please read the "Individual Contributor Agreement" linked-to at [https://www.bigchaindb.com/cla/](https://www.bigchaindb.com/cla/), fill in the form, and submit it. We'll get an email notification. We'll send you a long random string, which we'll ask you to copy to these comments, as evidence. Again, the link is:

https://www.bigchaindb.com/cla/

vrde:
@thedoctor thanks, the code looks **really** good, I like your plugin oriented approach. There might be some conflicts with my last PR #102 (that we need to merge soon), but it should be easy to merge. I'll update you tomorrow.

thedoctor:
BEGIN BLOCK

This is to confirm that I agreed to and accepted the BigchainDB Entity Contributor Agreement at https://www.bigchaindb.com/cla/ on behalf of my employer, and to represent and warrant that I have authority to do so.

(3u=rita?oD~=nScx]N!'t2r!^:[h52T4r?`Zg`F"w+og+L7c6Rb\!J`G]hd;$K

END BLOCK

thedoctor:
Cool, once #102 is in develop, I'll merge that in along with some updates to flesh out the feature.

ttmc:
Alberto (@vrde) pointed out another use for this:  having pluggable consensus/validation code will allow us to simulate "faulty/malicious" behavior in a Federation (e.g. for testing purposes).

vrde:
Hey @thedoctor, I just merged my changes. I think you'll get conflicts when you merge, tell me if you need help!

thedoctor:
Cool, I'll resolve this weekend – there'll be conflicts but I don't anticipate any real problems.

bigchaindb/config_utils.py
```diff
@@ -100,3 +102,21 @@ def autoconfigure():
     except FileNotFoundError:
         logger.warning('Cannot find your config file. Run `bigchaindb configure` to create one')
 
+
+def get_plugins(plugin_names):
```
vrde: A couple of improvements:
 - add docstring to describe the function.
 - from line 107 it seems that `plugin_names` can be `None` or an empty list, but when calling the function you must explicitly pass a value. What do you think about changing the signature of the function to `def get_plugins(plugin_names=None)`?

bigchaindb/consensus/__init__.py
```diff
@@ -0,0 +1,43 @@
+# TODO: no real reason to use abc yet, but later we can enforce inheritance from
+#       this class when loading plugins if that's desirable.
+# from abc import ABCMeta
+
+class AbstractConsensusRules:
+
+    # TODO: rather than having plugin-authors inherit and override,
+    #       it'd be cleaner to make a `transactionrule` decorator and etc
```
vrde: Agree, both `validate_transaction` and `validate_block` can be simple functions instead of `classmethod`s. But we can do it later.

setup.py
```diff
@@ -56,13 +56,21 @@
...
+        'bigchaindb.consensus'
+    ],
 
     entry_points={
         'console_scripts': [
             'bigchaindb=bigchaindb.commands.bigchain:main',
             'bigchaindb-benchmark=bigchaindb.commands.bigchain_benchmark:main'
         ],
+        'bigchaindb.plugins': [
+            'base=bigchaindb.consensus.base:ConsensusRules'
```
vrde: `consensus` might be a better name than `base`, it'd be easier to understand IMO. Ideas welcome.

bigchaindb/consensus/silly.py
```diff
@@ -0,0 +1,21 @@
...
+from bigchaindb.consensus import AbstractConsensusRules
+
+
+class SillyConsensusRules(AbstractConsensusRules):
+
+    @classmethod
+    def validate_transaction(cls, bigchain, transaction):
+        # I only like transactions whose timestamps are even.
+        if transaction['transaction']['timestamp'] % 2 != 0:
+            raise StandardError("Odd... very odd indeed.")
```
vrde: :joy:

vrde:
@thedoctor thanks again for the amazing work you are doing!

Further notes:
 - move `bigchaindb/consensus/base.py` to `bigchaindb/consensus.py` (we can always re-introduce the directory `consensus` when we need it)
 - move `bigchaindb/consensus/silly.py` to a document inside `docs/source` (we can take care of it if you want)

bigchaindb/config_utils.py
thedoctor: Ack :+1:

setup.py
thedoctor: The way the change is structured, the name here `base=` needs to be different (and should be descriptive) for each plugin. `bigchaindb.consensus` is the namespace.

thedoctor:
Good notes – I was under the weather this weekend so I didn't get to do much beside the merge; I'll get a new version of this up this week with more polish and an external plugin package to demonstrate how they should be constructed. Since #107 implements `n-of-n` multisig, I still think an `m-of-n` implementation would be a good first plugin. There might be some conflicts with the interface changes in #107 though; notably I think it's probably a good idea to accept `**kwargs` in public methods in case a plugin requires some additional data (e.g. `m_required` in `create_tx`). Not sure how the team feels about that kind of change?

vrde:
> Since #107 implements n-of-n multisig, I still think an m-of-n implementation would be a good first plugin. There might be some conflicts with the interface changes in #107 though; notably I think it's probably a good idea to accept **kwargs in public methods in case a plugin requires some additional data (e.g. m_required in create_tx). Not sure how the team feels about that kind of change?

#107 it's still under development and I'd like to merge your changes first. @diminator can adapt his PR to match the new plugin interface and, if the interface needs changes, we can change it using your idea about `**kwargs`.

It'd be great if you can fix the conflicts with `develop`, after that we can merge your PR.

thedoctor:
Okay, big updates:

1. Added `create_transaction`, `sign_transaction`, and `verify_transaction` to the plugin API. For things like multisig, replicating another protocols' consensus rules, or really anything of interest, changes to the way new transactions are created and verified will be necessary. I couldn't think of a better way to provide this than to bring these functions into the plugin environment.

2. Because of `1.`, I had to move away from the composable plugin system where each plugin's rules are called in series to a model where you only have one consensus plugin active at a time. Functions like `create` and `sign` don't map cleanly to a model where multiple plugins can be executed.
The functionality is equivalent, except that now anyone who wants to compose plugin functionality needs to write a class that inherits and overrides. =/ That sucks, but it's probably for the best – reasoning about chained consensus rules would be difficult, and chances of unexpected interactions seems high.

3. Switched from class to abstract static methods, other misc cleanup.

Things I haven't done:
1. All public methods (e.g. those in `client.py`) that call into `sign_tx` and `create_tx` need to be modified to call into the plugged-in methods.
2. Tests need to be written, but I wanted to see what everyone thought of these changes first.
3. Same with docs.

r-marques:
Not really comfortable in giving the ability to modify the `create_transaction`. We should have a well defined structure for the transaction because it is directly related to the datastore operations. How the data is queried, what secondary indexes are created, ...

Right now I understand the rationale to do this for things like multisig.
We are currently working on defining a generic structure for the signature field so that how a signature is verified is self contained inside the transactions. This would allow us to do simple sig (`1-1`) and multisig (`m-m` and `m-n`).

About `validate_transaction` I would prefer for the plugin to add extra functionality instead of completely overriding the `validate_transaction` similar to what you did with `validate_block` (run the core validator and then run the plugin validator) because there are validations that will always need to happen (checking double spends, checking the schema of the transactions, ...)

vrde:
> Not really comfortable in giving the ability to modify the create_transaction. We should have a well defined structure for the transaction because it is directly related to the datastore operations. How the data is queried, what secondary indexes are created, ...

Agree with @r-marques. Creating, signing, and validating a transaction should not depend on a plugin, they are just "mere" cryptographic utils independent of the protocol. Of course there is a *structure* to follow but, as a client, you should not care too much about the implementation of the consensus.

> 1. All public methods (e.g. those in `client.py`) that call into `sign_tx` and `create_tx` need to be modified to call into the plugged-in methods.

This might be really difficult to manage once we have HTTP clients written in other languages.

@r-marques already said that but yes, it'd be great if, thru the plugin API, you are able to **add** validation rules for `validate_transaction` and `validate_block`.

thedoctor:
> We are currently working on defining a generic structure for the signature field so that how a signature is verified is self contained inside the transactions.

What does that look like? We could avoid the need for a `create_transaction` hook if the idea of `current_owner` and `new_owner` were to be abstracted. e.g. If I want to create a hash-locked transaction, for example, the new owner isn't a pubkey, it's "whomever can produce the value which hashes to this digest" and the `owner` model inhibits those kind of constructs. If we can eliminate that, then the data for pretty much any extension can just be dumped into the `payload`.

> We should have a well defined structure for the transaction because it is directly related to the datastore operations

Agreed, but strict defining the transaction structure and strictly defining the `create_transaction` interface are totally different things.

> I would prefer for the plugin to add extra functionality instead of completely overriding the validate_transaction similar to what you did with validate_block (run the core validator and then run the plugin validator) because there are validations that will always need to happen (checking double spends, checking the schema of the transactions, ...)

Maybe we are imagining different systems here. I agree that many base rules will be really common, but not that will _always_ need to happen and not that will always need to happen in the same way. It's easy to imagine someone writing a plugin for with a different double-spend detection algorithm, or a blockchain where certain transactions can be double spent – e.g. an asset scheme where a root asset can be spent arbitrarily many times without any signature.

> Creating, signing, and validating a transaction should not depend on a plugin, they are just "mere" cryptographic utils independent of the protocol.

So here's my question: if bigchaindb prescribes the transaction model and signature scheme – then what's the point of having a plugin architecture at all? I can't think of many useful or interesting extensions that can be written without affecting the creation, signature, and verification of transactions. For example – how would you implement smart contracts?

> This might be really difficult to manage once we have HTTP clients written in other languages.

Really? I think that's just a matter of defining bigchain's public interface.

Maybe we have different goals – The reason I thought a modular consensus system would be useful was to encourage collaboration and community involvement, but a plugin system that can't alter the ownership model and enforces a signature scheme seems kind of deficient?

Anyway, thanks for taking the time to consider this :smile:

r-marques:
> What does that look like? We could avoid the need for a create_transaction hook if the idea of current_owner and new_owner were to be abstracted. e.g. If I want to create a hash-locked transaction, for example, the new owner isn't a pubkey, it's "whomever can produce the value which hashes to this digest" and the owner model inhibits those kind of constructs. If we can eliminate that, then the data for pretty much any extension can just be dumped into the payload.

We agree that the `new_owner` is a very limited implicit hash lock. What we would love to do is to comply with the interledger protocol that specifies crypto conditions up to `m-n` but not turing complete like smart contracts eg. ethereum. You can find a description of the crypto conditions [here](https://github.com/interledger/five-bells-condition/tree/feature/binary-merkle-v2) and [here](https://www.youtube.com/watch?v=YfBDDWp58po)

Let us know if you agree that this is a generic way forward.

r-marques:
> Agreed, but strict defining the transaction structure and strictly defining the create_transaction interface are totally different things.

Sure as long as the transactions are created in compliance with the crypto conditions and a datastore schema that defines the assets on a specific ledger

r-marques:
> Maybe we are imagining different systems here. I agree that many base rules will be really common, but not that will always need to happen and not that will always need to happen in the same way. It's easy to imagine someone writing a plugin for with a different double-spend detection algorithm, or a blockchain where certain transactions can be double spent – e.g. an asset scheme where a root asset can be spent arbitrarily many times without any signature.

One core rule would be that hash-lock conditions are in sync with the fulfillment. Of course, there might be flexibility on the other validation rules of the ledger such as double-spends. However it will be a challenge to classify what's core and whats plugin.

Out of curiosity, can you give me a use-case where double-spends would be allowed in a ledger?

r-marques:
> Really? I think that's just a matter of defining bigchain's public interface.

as said before, the signing will need to be in sync with the locking conditions.

r-marques:
In general, let's merge this as is. We'll continue to define on what's core and what's plugin and enforce when needed.

Once we implemented the crypto-conditions, these things will become clear

thedoctor:
> ...crypto conditions...
Let us know if you agree that this is a generic way forward.

Yeah! I hadn't investigated ILP; just read their whitepaper and looked at crypto condition spec and it looks great! We can go from 
`create_transaction(current_owner, new_owner, privkey, ...)` to 
`create_transaction(current_condition, new_condition, fulfillment, ...)`. 
I definitely have to think about it more before I'm convinced, but I'm pretty sure we could reduce any desirable system we want to be representable that way – especially since anyone can add their own new crypto condition types. Super cool!

> However it will be a challenge to classify what's core and whats plugin. Probably will depend on the types of assets you want to define on the ledger

I agree, which is kind of why I think everything should be a plugin, including core – which can then just be extended/overridden – although I could probably be convinced that the crypto condition verification and signature code is a notable exception.

> In general, let's merge this as is.

I'll do a little more cleanup and add tests/docs tonight/tomorrow. :)

thedoctor:
> Out of curiosity, can you give me a use-case where double-spends would be allowed in a ledger?

I can try! A (somewhat contrived) example related to #129:
Say you're building a fungible (divisible) asset protocol where
- anyone can create a new asset type
- when creating a new asset type you can choose whether:
  - the `CREATE` transaction explicitly declares all of the units of that asset type that will ever exist OR
  - whether more units can be minted by the `new_owner` of the `CREATE` transaction

This gives us the ability to create both non-inflationary and inflationary currencies. One reasonable way to represent the minting of new units is to say that the `new_owner` of the `CREATE` transaction that initialized an asset type can create as many `TRANSFER`s as he wants that use the origination transaction as the input.

Obviously there are other ways to represent that (especially if you have multiple outputs) that don't require double-spending, but this way has the (small) advantage that you can easily guarantee that the `new_owner` of the `CREATE` transaction will _always_ be able to mint new coins – as opposed to e.g. a protocol where each new minting would spend the previous one to a new transaction, which might have a different `new_owner`. This would make it impossible to steal control of a currency by tricking the originator into signing a transaction that gives minting authority to someone else.

Of course, you could also write a consensus rule that just says that in minting transactions `new_owner == current_owner`, but maybe you want to be able to give other parties minting authority and still be able to revoke that authority. Now we have an interesting protocol with permissioning built-in where the `current_owner` of the original `CREATE` transaction can always make a new transaction capable of disabling any of the `current_owners` of the latest minting transaction.

If you had crypto conditions or a powerful scripting system, the same effect could be achieved with some complex rules, but I think that implementation would probably be more error prone and more difficult to reason about.

That's the best I can come up with off the cuff.

thedoctor:
Sorry about the delay. Finished moving public calls to the consensus methods, added documentation and tests.
Let me know if anything else needs doing here.

diminator:
looks fine to me, great work @thedoctor ! :+1:

thedoctor:
Here's a template for consensus plugin packages: https://github.com/GemHQ/bigchaindb-consensus-template

vrde:
Sorry, was out of the loop for a while. The PR looks good to me, amazing job @thedoctor :)

vrde:
@thedoctor oh great, I was about to ask you to merge with develop :+1:

thedoctor:
Was Travis reconfigured? It's taking a long time and I no longer see links to the status page.

Issue #23, opened 2016-02-12
Handle and test installation on various OSes+distros


ttmc:
The BigchainDB installation requirements will differ slightly from OS to OS, especially the list of OS-level dependencies and how to install them. The setup.py file can take care of the Python dependencies, but not the OS-level dependencies.

How shall we handle OS-specific installation differences? Maybe write a Python script that checks the OS+distro and then installs the appropriate dependencies? Maybe create a Conda installer? Docker? Saltstack?

Would it make sense to test installation on various OS+distro combinations with Travis CI?

* [x] Make list of OSes + versions we want to support now
* [ ] Write an installer (or installers) to install BigchainDB, and all OS-level dependencies, on each of those OSes + versions
* [ ] Add installation tests in Travis CI

andreasATascribe:
List of OSes to support now:
- [X] Ubuntu/Debian
- [ ] Centos/Fedora/RHEL
- [ ] ArchLinux

List of OSes to support later:
- [ ] Alpine Linux
- [ ] Raspian
- [ ] BSD
- [ ] Windows

SohKai:
And OSX, please :innocent:?

r-marques:
Just for reference here are the os level dependencies tested on Fedora 23

`libffi-devel gcc-c++ redhat-rpm-config python3-devel openssl-devel`

ttmc:
For now, we've decided to focus on supporting Linux, with Ubuntu 14.04 as the primary distro & version.

For other OSes, we have [some support for running with Docker](https://bigchaindb.readthedocs.org/en/develop/installing-server.html#run-bigchaindb-with-docker).

PR #101, opened 2016-03-01
Docs: note OS-level dependencies on Fedora

ttmc:
Based on [this comment](https://github.com/bigchaindb/bigchaindb/issues/23#issuecomment-190260765) on Issue #23 :

Added a short section to the "Installing BigchainDB Server" section of the documentation, listing the OS-level dependencies on Fedora 23, and how to install them. (This may also help users of RHEL and CentOS.)

@r-marques I'm not super-familiar with `yum`; did I get the usage right? Is `sudo` needed?

### DIFF
--- a/docs/source/installing.md
+++ b/docs/source/installing.md
@@ -35,9 +35,11 @@ $ sudo apt-get install libffi-dev g++ libssl-dev python3-dev
 
 On Fedora 23, we found that the following was enough (tested in February 2015):
 ```text
-$ yum update
-$ yum install libffi-devel gcc-c++ redhat-rpm-config python3-devel openssl-devel
+$ sudo dnf update
+$ sudo dnf install libffi-devel gcc-c++ redhat-rpm-config python3-devel openssl-devel
 ```
+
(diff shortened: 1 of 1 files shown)


docs/source/installing.md
```diff
@@ -27,12 +27,18 @@ If you don't already have it, then you should [install Python 3.4+](https://www.
...
+On Ubuntu 14.04, we found that the following was enough:
 ```text
 $ sudo apt-get update
 $ sudo apt-get install libffi-dev g++ libssl-dev python3-dev
 ```
 
+On Fedora 23, we found that the following was enough (tested in February 2015):
+```text
+$ yum update
+$ yum install libffi-devel gcc-c++ redhat-rpm-config python3-devel openssl-devel
```
r-marques: yes you need sudo.
And `yum` was replaced in fedora 22.
So for fedora 22+ `sudo dnf install ...`
Before that `sudo yum install`

We can assume the same works for REHL/CentOS. Actually the default installation instructions for rethinkdb in fedora is to use the CentOS packages

ttmc:
I added `sudo` and changed from `yum` to `dnf`.

r-marques:
:+1:

PR #102, opened 2016-03-01
Provide a RESTful api

vrde:
This pull request is divided into two parts:
1. refactor of the `Bigchain` class and creation of the new `Client` class
2. first iteration on the API server
*(and much tests)*

### Refactor
The code in common between `Bigchain` and `Client` has been moved to the `bigchaindb.util` module. Now in `bigchaindb.utl` we have all the functions to create, sign, and verify transactions. The `Client` class is just a wrapper around aforementioned functions plus a key pair.

### API server
I added some basic code to run a Flask server to handle transactions. At the beginning I thought Tornado was the best option, but then I realized it would have been too hard to change all the calls to the database to be asynchronous. We can still optimize things later.
Running `bigchaindb start` from the shell will start a server listening to `localhost:5000` ready to process requests. Check out the documentation (in this pull request) under the **api** section.

Important: the server is now single threaded. I'll plug `gunicorn` in another pull request.


ttmc:
I know this is a minor thing, but in the database world, they call client SDKs "drivers" and a database typically has several available drivers, e.g. a Python driver, a JavaScript driver, etc. For example, RethinkDB has their four officially-supported drivers (Java, JavaScript, Python, Ruby) in subdirectories of a `rethinkdb/drivers/` directory:

https://github.com/rethinkdb/rethinkdb/tree/next/drivers

I wonder if it would make sense to rename and restructure things a bit to follow that example? Maybe in a later pull request, so we can merge this one now?

bigchaindb/core.py
```diff
@@ -75,101 +71,25 @@ def reconnect(self):
...
-
-        Args:
-            signed_transaction (dict): a transaction with the `signature` included.
-
-        Returns:
-            bool: True if the signature is correct, False otherwise.
+    def verify_signature(self, signed_transaction):
+        """Verify the signature of a transaction.
 
+        Refer to the documentation of ``bigchaindb.crypto.verify_signature``
```
r-marques: is this correct? all the logic to verify a signature seems to be here

bigchaindb/core.py
```diff
@@ -365,15 +285,7 @@ def validate_transaction(self, transaction):
                     raise exceptions.DoubleSpend('input `{}` was already spent'.format(
                         transaction['transaction']['input']))
 
-        # Check hash of the transaction
```
r-marques: does it make sense to group this logic? it does not seem to be used anywhere else

bigchaindb/core.py
```diff
@@ -603,26 +510,6 @@ def get_unvoted_blocks(self):
...
-
-        Returns:
-            str: JSON formatted string
-
-        """
-        return json.dumps(data, skipkeys=False, ensure_ascii=False,
-                          separators=(',', ':'), sort_keys=True)
 
     @staticmethod
     def deserialize(data):
```
r-marques: maybe put this in util, just to be consistent

bigchaindb/core.py
```diff
@@ -633,20 +520,9 @@ def deserialize(data):
...
-        Returns:
-            str: UTC timestamp.
 
-        """
-        dt = datetime.utcnow()
-        return "{0:.6f}".format(time.mktime(dt.timetuple()) + dt.microsecond / 1e6)
+        return json.loads(data, encoding="utf-8")
 
     @staticmethod
     def generate_keys():
```
r-marques: this should also go to util or remove it and use crypto.generate_keys

bigchaindb/crypto.py
```diff
@@ -146,6 +144,3 @@ def generate_key_pair():
 
     return (private_value_base58, public_value_compressed_base58)
 
-
-def hash_data(data):
```
r-marques: I think it makes sense to keep all the crypto code together

bigchaindb/processes.py
```diff
@@ -80,3 +81,9 @@ def start(self):
 
         logger.info('starting voter')
         p_voter.start()
+
+        # start the web api
+        webapi = server.create_app()
+        p_webapi = mp.Process(name='webapi', target=webapi.run, kwargs={'host': '0.0.0.0'})
```
r-marques: maybe we should restrict the api access to localhost by default

docs/source/index.rst
```diff
@@ -16,6 +16,7 @@ Table of Contents
    installing
    getting-started
    bigchaindb-cli
+   api
```
r-marques: since we moved a lot of the methods to bigchain.util, we should document that under developer interface

bigchaindb/web/views.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # `force` will try to format the body of the POST request even if the `content-type` header is not
+    # set to `application/json`
+    tx = request.get_json(force=True)
+
+    if tx['transaction']['operation'] == 'CREATE':
+        tx = util.transform_create(tx)
+        tx = util.sign_tx(tx, b.me_private)
+
+    if not util.verify_signature(tx):
```
r-marques: Here we should just validate the transaction before writing and send back whatever exception is thrown by the validator.

docs/source/api.md
```diff
@@ -0,0 +1,53 @@
+# Getting started with the HTTP API
```
ttmc: Maybe name this section http-api-tutorial.md? api.md is ambiguous as there is an HTTP API, and each driver will also have its own API (e.g. the Python API provided by the Python driver, the Java API provided by the Java driver, etc.). If you do change the name, remember to update it in `index.rst` as well.

vrde:
> I wonder if it would make sense to rename and restructure things a bit to follow that example? Maybe in a later pull request, so we can merge this one now?

I agree, we should put the driver in another Python module (and call it something like `bigchaindb-driver`). This would be a major refactoring and I'd like to have it soon. We can work on it and open another PR.

bigchaindb/core.py
vrde: This method is just a proxy to a function in `bigchaindb.crypto.verify_signature` (acutally, it's in `bigchaindb.util.verify_signature`) so I thought it would be nice to refer to the documentation to the original function.

bigchaindb/core.py
vrde: Usually I try to not "over plan", but I'm quite sure we will need this also in the client code to check the validity of the data coming from the Federation.

bigchaindb/core.py
vrde: Agree.

bigchaindb/core.py
vrde: I left it there because some tests rely on that method, but I can remove it (and fix the tests).

bigchaindb/processes.py
vrde: Agree!

bigchaindb/crypto.py
vrde: Yes, agree also on this point, I'm commenting on the main PR, I'd like to move some stuff form `util` to `crypto`, I just noticed my refactoring was inconsistent on some parts.

docs/source/api.md
vrde: Good point, fixing that.

bigchaindb/web/views.py
vrde: After this PR is accepted I'll create some issues so others can contribute in improving the API. It's part of the [roadmap](https://github.com/bigchaindb/bigchaindb/pull/102/files#diff-773a2053af722b60295688c900556bb9R44).

vrde:
I'd like to move basically all the functions in `bigchaindb.util` to `bigchaindb.crypto` since it's crypto related. Another alternative is to create a new module `bigchaindb.transaction` that contains all the basic functions to work with transactions. @r-marques any thoughts?

r-marques:
`bigchain.transactions` (validations, creation, signing ...) seems good but I would still keep all of the crypto primitives in `bigchaindb.crypto`

vrde:
@r-marques OK, I can put everything transaction related in the `transactions` module, while basic crypto stuff under `crypto`.

vrde:
> @r-marques OK, I can put everything transaction related in the transactions module, while basic crypto stuff under crypto.

@r-marques I started refactoring that part and then I realized it's a bit too much for this PR. It's better to do it in another PR.

r-marques:
seems good to me. I created issues #108 #109 #110 to address some things missing in this pull request

:+1:

bigchaindb/util.py
```diff
@@ -22,3 +30,191 @@ def start(self):
...
+
+    """
+    private_key = PrivateKey(private_key)
+    signature = private_key.sign(serialize(transaction))
+    signed_transaction = transaction.copy()
+    signed_transaction.update({'signature': signature})
+    return signed_transaction
+
+
+def create_and_sign_tx(private_key, current_owner, new_owner, tx_input, operation='TRANSFER', payload=None):
```
diminator: maybe misleading, as it is a `operation=TRANSFER` by default and does require a tx_input

maybe remove the `operation` from the kwargs?

bigchaindb/util.py
```diff
@@ -22,3 +30,191 @@ def start(self):
...
+        bool: True if the signature is correct, False otherwise.
+    """
+
+    data = signed_transaction.copy()
+
+    # if assignee field in the transaction, remove it
+    if 'assignee' in data:
+        data.pop('assignee')
+
+    signature = data.pop('signature')
```
diminator: will fail if `signature` not in `data`

```
...
if 'signature' not in data:
    return False

signature = data.pop('signature')
...
```

bigchaindb/util.py
```diff
@@ -22,3 +30,191 @@ def start(self):
...
+    public_key_base58 = signed_transaction['transaction']['current_owner']
+    public_key = PublicKey(public_key_base58)
+    return public_key.verify(serialize(data), signature)
+
+
+def transform_create(tx):
+    """Change the owner and signature for a ``CREATE`` transaction created by a node"""
+
```
diminator: why/where is this used?
seems somewhat weird

bigchaindb/web/views.py
```diff
@@ -0,0 +1,68 @@
...
+        tx = util.transform_create(tx)
+        tx = util.sign_tx(tx, bigchain.me_private)
+
+    if not util.verify_signature(tx):
+        val['error'] = 'Invalid transaction signature'
+
+    val = bigchain.write_transaction(tx)
+
+    return flask.jsonify(**tx)
+
```
diminator: I assume this is a minimal blueprint

retrieving blocks would be a logical extension

bigchaindb/web/views.py
```diff
@@ -0,0 +1,68 @@
...
+    Return:
+        A JSON string containing the data about the transaction.
+    """
+    bigchain = current_app.config['bigchain']
+
+    val = {}
+
+    # `force` will try to format the body of the POST request even if the `content-type` header is not
+    # set to `application/json`
+    tx = request.get_json(force=True)
```
diminator: Is there any payload blueprint/format enforcing in flask?

Now it's quite unclear how the payload should look like and which fields are mandatory/optional.
There is little enforcing... hence people can submit bogus payloads

docs/source/python-api-tutorial.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+
+# Roadmap
+
+The development of the API is still at the beginning and you can follow it on
+[GitHub](https://github.com/bigchaindb/bigchaindb/issues?q=is%3Aissue+is%3Aopen+label%3Arest-api)
+
+There are several key features still missing like:
+ - validating the structure of the transaction
```
diminator: ok, that solves a question I asked above

bigchaindb/util.py
```diff
@@ -1,4 +1,12 @@
...
```
diminator: general thought: what goes in util?
- helper functions
- transaction functions
- singletons/factories
- ...

Currently it looks a bit like a container where we store stuff that doesnt fit into core (node logic). We'll probably need to reshuffle this later

bigchaindb/util.py
vrde: misleading because the function contains the word `create` but the default operation is `TRANSFER`?

diminator: also, but mainly because `tx_input` is mandatory hence implying `operation=TRANSFER`

vrde: Good point, lemme fix that.

bigchaindb/util.py
vrde: Yup, because it's required, and it's totally fine. Data validation must be done **before** calling the function.

If you pass a transaction without `signature`, then it's an **error**, because there is no signature to verify, and the function will raise a `KeyError`.

docs/source/python-api-tutorial.md
vrde: :) yup

bigchaindb/util.py
vrde: yes, totally agree, I tried to refactor this part but was too much for this pull request.

bigchaindb/web/views.py
vrde: yup, all in the roadmap. I just don't want to submit a huge PR.

bigchaindb/web/views.py
vrde: Roadmap :)

bigchaindb/util.py
r-marques: agree. there is this #108 to start addressing that

bigchaindb/util.py
```diff
@@ -22,3 +30,191 @@ def start(self):
...
+    Returns:
+        bool: True if the signature is correct, False otherwise.
+    """
+
+    data = signed_transaction.copy()
+
+    # if assignee field in the transaction, remove it
+    if 'assignee' in data:
+        data.pop('assignee')
```
sbellem: if the value returned by `pop` is not needed, `del` can be used:

> `del d[key]`
> Remove `d[key]` from *d*. Raises a `KeyError` if *key* is not in the map.

ref: https://docs.python.org/3.5/library/stdtypes.html#mapping-types-dict

PR #103, opened 2016-03-02
Use the default ReadTheDocs theme

vrde:
I changed the theme of the docs. In the screenshot, the new theme is the one on the left, the old the one on the right.

I think the new theme is much cleaner and readable.

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ docs_require = [
     'recommonmark>=0.4.0',
     'Sphinx>=1.3.5',
     'sphinxcontrib-napoleon>=0.4.4',
+    'sphinx-rtd-theme>=0.1.9',
 ]
 
 setup(
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,3 @@
+Sphinx==1.3.5
+sphinxcontrib-napoleon==0.4.4
+recommonmark
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
 Sphinx==1.3.5
 sphinxcontrib-napoleon==0.4.4
+sphinx-rtd-theme>=0.1.9
 recommonmark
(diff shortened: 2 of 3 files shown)


ttmc:
Nice! That theme inspires me to look into how to change the color scheme to follow the [BigchainDB style guide](https://www.bigchaindb.com/styleguide/); I can do that in a separate PR though.

setup.py
```diff
@@ -26,6 +26,7 @@
     'recommonmark>=0.4.0',
     'Sphinx>=1.3.5',
     'sphinxcontrib-napoleon>=0.4.4',
+    'sphinx-rtd-theme>=0.1.9',
```
ttmc: One minor thing: could you also update [`bigchaindb/docs/requirements.txt`](https://github.com/bigchaindb/bigchaindb/blob/develop/docs/requirements.txt) to be in sync with the list of docs-related packages in setup.py?

sbellem: I think we could fully really on `setup.py`, perhaps? See https://github.com/bigchaindb/bigchaindb/blob/develop/setup.py#L25

ttmc: Oh, interesting. In that case, we should delete `bigchaindb/docs/requirements.txt` and modify the instructions in [`bigchaindb/docs/README.md`](https://github.com/bigchaindb/bigchaindb/blob/develop/docs/README.md)

vrde: @sbellem yup, great job with `setup.py`.

@ttmc I'm removing it right  now.

vrde: @sbellem I just double checked with @ttmc, ReadTheDocs expects a `requirements.txt` to install dependencies. I'm writing them to ask if they support `extras_require` in `setup.py`, so we can keep everything in the same place.

vrde: I found an old issue related to that https://github.com/rtfd/readthedocs.org/issues/173, I'm asking to reopen it.

vrde:
> Nice! That theme inspires me to look into how to change the color scheme to follow the BigchainDB style guide; I can do that in a separate PR though.

Merging now, can't wait for a theme with customized colors :lipstick:!

PR #107, opened 2016-03-03
Feature multisig

diminator:
This PR overrules #106 (excludes unnecessary files)

The contribution of this fork is to implement n-of-n multisig.

Multisig allows a single `current_owner` to send an asset to multiple `new_owners`.
Moreover, once an asset has multiple `current_owners`, all `current_owners` need to sign the transaction before it validates to `True` on `validate_transaction()`

This PR replaces all occurances of `current_owner` and `new_owner` into `current_owners` and `new_owners`. These are lists of public keys but `create_transaction()` and `sign_transaction()` still supports single strings, not embedded in a list for the single case.

Upon signing a multisig one needs to provide his public key to identify who's signature is being added. Note that this is optional and can be omitted for the single case. For the multisig case, the signature takes the form of a list of dicts (the single case only renders a string, as before):

```
signatures = [
    {
        'public_key': ...,
        'signature': ....
    },
    ...
]
```


r-marques:
Some code related to transactions moved in #102 pr.
Probably a good idea to merge that pr first and them review this one.

thedoctor:
Is there a reason for the choice to implement `n-of-n` instead of `m-of-n`?

r-marques:
Just for simplicity in a first iteration of multisig support we require a signature from all current owners of the digital asset

diminator:
This is a subset of #147, so no longer required

Issue #131, opened 2016-03-14
When validating a transaction, check to ensure it isn't a duplicate


ttmc:
If a transaction is sent by a client (including the timestamp from the client), it's valid, and it's a "create" transaction, then it should eventually end up in a decided_valid block in the bigchain table. It won't be in the backlog table anymore.

If a copy of the same transaction (with the same timestamp and the same hash/id) is sent from the same client a few days later, then the receiver node should check that a transaction with the same hash/id doesn't already exist in a block in the bigchain table. That would be part of the transaction validation logic.

ttmc:
A similar issue may arise once we implement re-assignment of old/stale transactions to new nodes (issue #198 ), so maybe also check-if-duplicate when validating a transaction that has just been read from the backlog... unless we figure out another solution.

ttmc:
When a transaction ends up in an invalid block, it gets copied back to the backlog (see issue #193), so the check for a duplicate transaction in the bigchain table should ignore transactions in invalid blocks.

rhsimplex:
Consolidating discussion under #568

PR #133, opened 2016-03-14
Clarify the meaning of some timestamps in the docs

ttmc:
I clarified the meaning of some timestamps in the documentation (in the Models section). I did the same in the whitepaper.


ttmc:
While I was at it, I also clarified the meaning of the `id` in the transaction model (in the documentation). I also removed the sentence about how that prevents duplicate transactions (because a duplicate transaction _could_ happen, right now, if the second one was sent long after the first one cleared the backlog table). (Issue #131 is to fix that bug.)

PR #138, opened 2016-03-16
Implement ED25519

### LINKED ISSUE #137: implement ED25519 (Schnorr) among ECDSA
part of #127, see description also

diminator:
This PR resolves #137 

- [x] created and abstract class for signing schemes
- [x] factored out ECDSA
- [x] added ED25519
- [x] still use ECDSA for the current signing/validation of messages
- [x] added tests


ttmc:
Should the docs be modified anywhere yet, or is this just preparation for the future?

The signature algorithm is documented in the [Cryptography section](https://bigchaindb.readthedocs.org/en/develop/cryptography.html).

diminator:
> Should the docs be modified anywhere yet, or is this just preparation for the future?

exactly, currently we still use ECDSA, but in the short term we'll be switching to more complex conditions and ED25519. This ofcourse needs to be documented (when we do the actual switchover)

r-marques:
Other then `hash_data` being inside the `asymmetric` module seems good.

We should also check if this is the best implementation of ed25519 to use.

:+1:

Issue #132, opened 2016-03-14
Revisit timestamps?


ttmc:
Right now, there are timestamps in several places:
- Each transaction gets a timestamp from the client who submitted it
- Each block gets a timestamp from the BigchainDB server process (Python) which generated it
- Each vote gets a timestamp from the BigchainDB server process (Python) which generated it

We can't trust the timestamps from clients, so why bother having them at all? If they want to include a timestamp from a [trusted timestamping](https://en.wikipedia.org/wiki/Trusted_timestamping) service (along with a signature from that service), then they can include that in the transaction payload.

Or maybe we could allow clients the option of sending timestamps from a trusted timestamping service (plus associated signature and maybe other metadata, such as the name of the service)?

We should also reconsider the timestamps generated by the BigchainDB servers. It's generally a bad idea to pretend that all the clocks in a distributed system are trustworthy or synchronized (unless you've gone to a lot of trouble, like what Google's [Spanner](https://en.wikipedia.org/wiki/Spanner_%28database%29) does). Unfortunately, users of BigchainDB might not know that, and they might believe the timestamps _mean_ something. Maybe just remove those timestamps?

ttmc:
We may sometimes be able to trust client-provided timestamps, at least for figuring out transaction ordering, because the client has an incentive to get their transactions in the proper order-by-timestamp.

ttmc:
Two more notes on timestamps:
* Even if we don't get rid of timestamps, we should have documentation and UI to help users interpret them properly.
* If we ran a time-synchronization daemon on all BigchainDB servers (and there are many options, e.g. `ntpd`, `tlsdate`), then that would bring them into better (but not perfect) time synchronization with each other and with global time servers. Don't get too excited: this increases cost, complexity and attack surface.

ttmc:
Related: Issue #349 

Some interesting ideas from there:

* The idea of using the median of the vote timestamps (which come from multiple nodes). That could be used to replace transaction and block timestamps (i.e. only votes would have timestamps).
* Assign a timestamp to a transaction when it first arrives at BigchainDB based on the median timestamp from all (or many) federation nodes.

ttmc:
There was some discussion about timestamps on the internal Slack dev channel today. Some of the discussion:

Rodolphe: we can add timestamp verification to the consensus
actually it would be simpler and more performant if we only added timestamp verification for blocks
[...]

PR #141, opened 2016-03-17
Updated ROADMAP.md based on March 16 meeting

ttmc:
I updated the `ROADMAP.md` file based on the March 16 meeting at ascribe GmbH.

I added a top-level heading for **Deployment and Federation Configuration/Management** because it doesn't really fit under **Testing** or anywhere else.

I put "Revisit Timestamps" (Issue #132 ) under the **Transactions / Assets** heading because they're part of the transaction model and they really do need to be reconsidered. Getting transaction ordering "right" is non-trivial. It's not something you can just leave up to RethinkDB/Raft. Transaction order might get messed up before ever getting to RethinkDB/Raft. But you can't trust timestamps... or can you? If so, when? See more explanation in Issue #132


diminator:
looks good for me... lotsa work ahead though :-)

Issue #127, opened 2016-03-09
Crypto-conditions spec according to ILP


r-marques:
Since we are now starting to use multisig it may make sense to consider switching to `ed25519`.
Slightly faster then ecdsa, efficient threshold signatures, no known signature malleability issues.

We would need to find a good implementation with python bindings.

diminator:
The guys from Ripple have a good standard for these things:

https://github.com/interledger/five-bells-condition/blob/feature/binary-merkle-v2/docs/spec.md
https://www.youtube.com/watch?v=YfBDDWp58po

Basically we only need to convert the JavaScript to Python

PR #146, opened 2016-03-18
Crypto conditions following ILP spec (sha256, ed25519, threshold)

diminator:
solves #127 

- [x] conversion of reference ILP javascript library
- [x] SHA256
- [ ] RSA
- [x] ED25519
- [x] Threshold

# Crypto Conditions

This spec is from the [**Interledger Protocol (ILP)**]
(https://interledger.org/five-bells-condition/spec.html)

## Motivation

We would like a way to describe a signed message such that multiple actors in a distributed system can all verify the same signed message and agree on whether it matches the description.

This provides a useful primitive for distributed, event-based systems since we can describe events (represented by signed messages) and therefore define generic authenticated event handlers.

## Terminology

* ##### Condition
  A condition is the hash of a description of a signed message.

* ##### Fulfillment
  A fulfillment consists of a description of a signed message and a signed message that matches the description.

  The description can be hashed and compared to a condition. If the message matches the description and the hash of the description matches the condition, we say that the fulfillment **fulfills** the condition.

* ##### Hashlock
  A tuple consisting of a bytestring and its hash where the hash is published first and the publication of the corresponding bytestring acts as a one-bit, one-time signature.

# Basic Format

## Bitmask

Any system accepting crypto-conditions must be able to state its supported
algorithms. It must be possible to verify that all algorithms used in a certain
condition are indeed supported even if the fulfillment is not available yet.

In order to meet these design goals, we define a bitmask to express the supported primitives.

The following bits are assigned:

|Type Bit |Exp.         |Int.|Condition Type   |
|--------:|------------:|---:|-----------------|
|        1|2<sup>0</sup>|   1|SHA-256          |
|       10|2<sup>1</sup>|   2|RSA-SHA-256      |
|      100|2<sup>2</sup>|   4|THRESHOLD-SHA-256|
|     1000|2<sup>3</sup>|   8|ED25519-SHA-256  |

Conditions contain a bitmask of types they require the implementation to support. Implementations provide a bitmask of types they support.

### ILP Features

Crypto-conditions are a simple multi-algorithm, multi-message, multi-level, multi-signature standard.

* **Multi-algorithm**

  Crypto-conditions can support several different signature and hash algorithms and support for new ones can be added in the future.

  Implementations can state their supported algorithms simply by providing a bitmask. It is easy to verify that a given implementation will be able to verify the fulfillment to a given condition, by verifying that the condition's bitmask `condition` and its own bitmask of supported algorithms `supported` satisfies `condition & ~supported == 0` where `&` is the bitwise AND operator and `~` is the bitwise NOT operator.

  Any new high bit can redefine the meaning of any existing lower bits as long as it is set. This can be used to remove obsolete algorithms.

  The bitmask is encoded as a varint to minimize space usage.

* **Multi-signature**

  Crypto-conditions can abstract away many of the details of multi-sign. When a party provides a condition, other parties can treat it opaquely and do not need to know about its internal structure. That allows parties to define arbitrary multi-signature setups without breaking compatibility.

  Protocol designers can use crypto-conditions as a drop-in replacement for public key signature algorithms and add multi-signature support to their protocols without adding any additional complexity.

* **Multi-level**

  Basic multi-sign is single-level and does not support more complex trust relationships such as "I trust Alice and Bob, but only when Candice also agrees". In single level 2-of-3 Alice and Bob could sign on their own, without Candice's approval.

  Crypto-conditions add that flexibility elegantly, by applying thresholds not just to signatures, but to conditions which can be signatures or further conditions. That allows the creation of an arbitrary threshold boolean circuit of signatures.

* **Multi-message**

  Crypto-conditions can sign not just one, but multiple messages at the same time and by different people. These messages can then be used as inputs for other algorithms.

  This allows resource-controlling systems to perform their functions without knowing the details of the higher-level protocols that these functions are a part of.

## Usage

```python
import binascii
from bigchaindb.crypto.condition import Condition
from bigchaindb.crypto.fulfillment import Fulfillment
from bigchaindb.crypto.fulfillments.sha256 import Sha256Fulfillment

# Parse a condition from a URI
example_condition_uri = 'cc:1:1:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:1'
parsed_condition = Condition.from_uri(example_condition_uri)
print(isinstance(parsed_condition, Condition))
# prints True

print(binascii.hexlify(parsed_condition.hash))
# prints b'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'

# Compile a condition
parsed_condition_uri = parsed_condition.serialize_uri()
print(parsed_condition_uri)
# prints 'cc:1:1:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:1'
print(parsed_condition_uri == example_condition_uri)
# prints True

# Parse a fulfillment
example_fulfillment_uri = 'cf:1:1:AA'
parsed_fulfillment = Fulfillment.from_uri(example_fulfillment_uri)
print(isinstance(parsed_fulfillment, Sha256Fulfillment))
# prints True
# Note: Merely parsing a fulfillment DOES NOT validate it.

# Validate a fulfillment
parsed_fulfillment.validate()
# prints True
```

### Binary types

* **VARUINT**

  Unsigned variable-length integer. Implementation matches [Base128 Varints](https://developers.google.com/protocol-buffers/docs/encoding#varints) in Protocol Buffers. Implementations MAY define different maximum lengths for their varints, as long as that length is long enough to cover their bitmask and their maximum supported fulfillment length. (This is safe, because no larger varuint can appear in a valid crypto-condition.)

* **VARBYTES**

  Consists of a `VARUINT` length field followed by that many bytes.

* **VARARRAY**

  Consists of a `VARUINT` length fields followed by that many bytes filled with elements of the array.

### String types

* **BASE10**

  Variable-length integer encoded as a base-10 (decimal) number. Implementations MUST reject encodings that are too large for them to parse. Implementations MUST be tested for overflows.

* **BASE16**

  Variable-length integer encoded as a base-16 (hexadecimal) number. Implementations MUST reject encodings that are too large for them to parse. Implementations MUST be tested for overflows. No leading zeros.

* **BASE64URL**

  Base64-URL encoding. See [RFC4648 Section 5](https://tools.ietf.org/html/rfc4648#section-5).

### Condition

Conditions are ASCII encoded as:

```
"cc:" BASE10(VERSION) ":" BASE16(TYPE_BITMASK) ":" BASE64URL(HASH) ":" BASE10(MAX_FULFILLMENT_LENGTH)
```

Conditions are binary encoded as:

```
CONDITION =
  VARUINT TYPE_BITMASK
  VARBYTES HASH
  VARUINT MAX_FULFILLMENT_LENGTH
```

The `TYPE_BITMASK` is the boolean OR of the `TYPE_BIT`s of the condition type and all subcondition types, recursively.

### Fulfillment

Fulfillments are ASCII encoded as:

```
"cf:" BASE10(VERSION) ":" BASE16(TYPE_BIT) ":" BASE64URL(FULFILLMENT_PAYLOAD)
```

Fulfillments are binary encoded as:

```
FULFILLMENT =
  VARUINT TYPE_BIT
  FULFILLMENT_PAYLOAD
```

The `TYPE_BIT` is the single bit representing the top level condition type.

# Condition Types

## SHA-256

SHA-256 is assigned the type bit 2<sup>0</sup> = 0x01.

### Notes

This type of condition is also called a hashlock. We can use revealing the preimage as a type of one bit signature.

Bitcoin supports this type of condition via the `OP_HASH256` operator

### Condition

```
HASH = SHA256(PREIMAGE)
```

### Fulfillment

```
FULFILLMENT_PAYLOAD =
  VARBYTES PREIMAGE
```

### Usage

```python
import binascii, hashlib
from bigchaindb.crypto.condition import Condition
from bigchaindb.crypto.fulfillments.sha256 import Sha256Fulfillment

secret = ''
puzzle = binascii.hexlify(hashlib.sha256(secret.encode()).digest())
print(puzzle)
# prints b'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'

# Create a SHA256 condition
sha256condition = Condition()
sha256condition.bitmask = 0x01
sha256condition.hash = binascii.unhexlify(puzzle)
sha256condition.max_fulfillment_length = 1
sha256condition_uri = sha256condition.serialize_uri()
print(sha256condition_uri)
# prints 'cc:1:1:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:1'

# Create a fulfillment
sha256fulfillment = Sha256Fulfillment()

# Create a condition from fulfillment
sha256fulfillment.condition
# raises ValueError: Could not calculate hash, no preimage provided
sha256fulfillment.preimage = secret
print(sha256fulfillment.condition.serialize_uri() == sha256condition_uri)
# prints True

# Compile a fulfillment
print(sha256fulfillment.serialize_uri())
# prints 'cf:1:1:AA'

# Even better: verify that the fulfillment matches the condition
print(sha256fulfillment.validate() and \
    sha256fulfillment.condition.serialize_uri() == sha256condition.serialize_uri())
# prints True
```

## RSA-SHA-256

RSA-SHA-256 is assigned the type bit 2<sup>1</sup> = 0x02.

**Warning:** not (yet) implemented in BigchainDB, for info see the [**ILP specification**](https://interledger.org/five-bells-condition/spec.html)

## ED25519-SHA-256

ED25519-SHA-256 is assigned the type bit 2<sup>3</sup> = 0x08.

### Condition

```
HASH = SHA256(
  VARBYTES PUBLIC_KEY
  VARBYTES MESSAGE_ID
  VARBYTES FIXED_PREFIX
  VARUINT DYNAMIC_MESSAGE_LENGTH
)
```

### Fulfillment

```
FULFILLMENT_PAYLOAD =
  VARBYTES PUBLIC_KEY
  VARBYTES MESSAGE_ID
  VARBYTES FIXED_PREFIX
  VARUINT DYNAMIC_MESSAGE_LENGTH
  VARBYTES DYNAMIC_MESSAGE
  VARBYTES SIGNATURE
```

The `DYNAMIC_MESSAGE_LENGTH` is included to provide a maximum length for `DYNAMIC_MESSAGE` even if the actual message suffix length is different. This value is used to calculate the `MAX_FULFILLMENT_LENGTH` in the condition.

The `MESSAGE_ID` represents an identifier for the message. All messages in a cryptocondition that have a common identifier must match, otherwise the condition is invalid. Implementations may return messages as a map of `MESSAGE_ID` => `MESSAGE` pairs.

The message to be signed is the concatenation of the `FIXED_PREFIX` and `DYNAMIC_MESSAGE`.

The `MESSAGE_ID`, `FIXED_PREFIX`, `DYNAMIC_MESSAGE_LENGTH` and `DYNAMIC_MESSAGE` fields have the same meaning as in the [**RSA-SHA-256 condition type**](https://interledger.org/five-bells-condition/spec.html).

### Usage

```python
from bigchaindb.crypto.ed25519 import Ed25519SigningKey, Ed25519VerifyingKey
from bigchaindb.crypto.fulfillments.ed25519_sha256 import Ed25519Sha256Fulfillment

# We use base58 key encoding
sk = Ed25519SigningKey(b'9qLvREC54mhKYivr88VpckyVWdAFmifJpGjbvV5AiTRs')
vk = sk.get_verifying_key()

# Create an ED25519-SHA256 condition
ed25519_fulfill`nt = Ed25519Sha256Fulfillment()
ed25519_fulfillment.public_key = vk
ed25519_fulfillment.message_prefix = 'Hello world!'
ed25519_fulfillment.max_dynamic_message_length = 32  # defaults to 0
ed25519_condition_uri = ed25519_fulfillment.condition.serialize_uri()
print (ed25519_condition_uri)
# prints 'cc:1:8:qQINW2um59C4DB9JSVXH1igqAmaYGGqryllHUgCpfPU:113'

# ED25519-SHA256 condition not fulfilled
print(ed25519_fulfillment.validate())
# prints False

# Fulfill an ED25519-SHA256 condition
ed25519_fulfillment.message = ' Conditions are here!'
ed25519_fulfillment.sign(sk)
print(ed25519_fulfillment.validate())
# prints True

print(ed25519_fulfillment.serialize_uri())
# prints 
#  'cf:1:8:IOwXK5OtXlY79JMscOEkUDTDVGfvLv1NZOv4GWg0Z-K_DEhlbGxvIHdvcmxkI
#   SAVIENvbmRpdGlvbnMgYXJlIGhlcmUhQENbql531PbCJlRUvKjP56k0XKJMOrIGo2F66u
#   euTtRnYrJB2t2ZttdfXM4gzD_87eH1nZTpu4rTkAx81hSdpwI'
print (ed25519_fulfillment.condition.serialize_uri())

# Parse a fulfillment URI
parsed_ed25519_fulfillment = Ed25519Sha256Fulfillment.from_uri('cf:1:8:IOwXK5OtXlY79JMscOEkUDTDVGfvLv1NZOv4GWg0Z-K_DEhlbGxvIHdvcmxkISAVIENvbmRpdGlvbnMgYXJlIGhlcmUhQENbql531PbCJlRUvKjP56k0XKJMOrIGo2F66ueuTtRnYrJB2t2ZttdfXM4gzD_87eH1nZTpu4rTkAx81hSdpwI')
        
print(parsed_ed25519_fulfillment.validate())
# prints True
print(parsed_ed25519_fulfillment.condition.serialize_uri())
# prints 'cc:1:8:qQINW2um59C4DB9JSVXH1igqAmaYGGqryllHUgCpfPU:113'
```

### Implementation

The exact algorithm and encodings used for `PUBLIC_KEY` and `SIGNATURE` are Ed25519 as defined in [draft-irtf-cfrg-eddsa-04](https://datatracker.ietf.org/doc/draft-irtf-cfrg-eddsa/).

## THRESHOLD-SHA-256

THRESHOLD-SHA-256 is assigned the type bit 2<sup>2</sup> = 0x04.

### Condition

```
HASH = SHA256(
  VARUINT TYPE_BIT
  VARUINT THRESHOLD
  VARARRAY
    VARUINT WEIGHT
    CONDITION
)
```

The `TYPE_BIT` is `0x04`. The reason we need this is because threshold conditions are a structural condition. Structural conditions can have subconditions, meaning their TYPE_BITMASK can have multiple bits set, including other structural conditions. This `TYPE_BIT` prevents the possibility that two different structural fulfillments could ever generate the exact same condition.

The `VARARRAY` of conditions is sorted first based on length, shortest first. Elements of the same length are sorted in lexicographic (big-endian) order, smallest first.

### Fulfillment

```
FULFILLMENT_PAYLOAD =
  VARUINT THRESHOLD
  VARARRAY
    VARUINT WEIGHT
    FULFILLMENT
  VARARRAY
    VARUINT WEIGHT
    CONDITION
```

### Usage

```python
from bigchaindb.crypto.fulfillments.sha256 import Sha256Fulfillment
from bigchaindb.crypto.fulfillments.ed25519_sha256 import Ed25519Sha256Fulfillment
from bigchaindb.crypto.fulfillments.threshold_sha256 import ThresholdSha256Fulfillment

# Parse some fulfillments
sha256_fulfillment = Sha256Fulfillment.from_uri('cf:1:1:AA')
ed25519_fulfillment = Ed25519Sha256Fulfillment.from_uri('cf:1:8:IOwXK5OtXlY79JMscOEkUDTDVGfvLv1NZOv4GWg0Z-K_DEhlbGxvIHdvcmxkISAVIENvbmRpdGlvbnMgYXJlIGhlcmUhQENbql531PbCJlRUvKjP56k0XKJMOrIGo2F66ueuTtRnYrJB2t2ZttdfXM4gzD_87eH1nZTpu4rTkAx81hSdpwI')

# Create a threshold condition
threshold_fulfillment = ThresholdSha256Fulfillment()
threshold_fulfillment.add_subfulfillment(sha256_fulfillment)
threshold_fulfillment.add_subfulfillment(ed25519_fulfillment)
threshold_fulfillment.threshold = 1  # OR gate 
print(threshold_fulfillment.condition.serialize_uri())
# prints 'cc:1:d:9DdkQtOl2m9yjqZzCg6ck5b2zM3tAPLlJMaHsKkszIA:114'

# Compile a threshold fulfillment
threshold_fulfillment_uri = threshold_fulfillment.serialize_uri()
# Note: If there are more than enough fulfilled subconditions, shorter
# fulfillments will be chosen over longer ones.
print(threshold_fulfillment_uri)
# prints 'cf:1:4:AQEBAQABAQggqQINW2um59C4DB9JSVXH1igqAmaYGGqryllHUgCpfPVx'

# Validate fulfillment
print(threshold_fulfillment.validate())
# prints True

# Parse the fulfillment
reparsed_fulfillment = \
    ThresholdSha256Fulfillment.from_uri(threshold_fulfillment_uri)
print(reparsed_fulfillment.validate())
# prints True

# Increase threshold to a 3-port AND gate
threshold_fulfillment.threshold = 3
print(threshold_fulfillment.validate())
# prints False

# Create a nested threshold condition
# VALID = SHA and DSA and (DSA or DSA) 
nested_fulfillment = ThresholdSha256Fulfillment()
nested_fulfillment.add_subfulfillment(ed25519_fulfillment)
nested_fulfillment.add_subfulfillment(ed25519_fulfillment)
nested_fulfillment.threshold = 1  # OR gate 

threshold_fulfillment.add_subfulfillment(nested_fulfillment)
print(threshold_fulfillment.serialize_uri())
# prints 
#  'cf:1:4:AwMBAQABCCDsFyuTrV5WO_STLHDhJFA0w1Rn7y79TWTr-BloNGfivwxIZWxsby
#   B3b3JsZCEgFSBDb25kaXRpb25zIGFyZSBoZXJlIUBDW6ped9T2wiZUVLyoz-epNFyiTDq
#   yBqNheurnrk7UZ2KyQdrdmbbXX1zOIMw__O3h9Z2U6buK05AMfNYUnacCAQQBAQEIIOwX
#   K5OtXlY79JMscOEkUDTDVGfvLv1NZOv4GWg0Z-K_DEhlbGxvIHdvcmxkISAVIENvbmRpd
#   GlvbnMgYXJlIGhlcmUhQENbql531PbCJlRUvKjP56k0XKJMOrIGo2F66ueuTtRnYrJB2t
#   2ZttdfXM4gzD_87eH1nZTpu4rTkAx81hSdpwIBAQggqQINW2um59C4DB9JSVXH1igqAma
#   YGGqryllHUgCpfPVxAA'
```


bigchaindb/crypto/buffer.py
```diff
@@ -0,0 +1,435 @@
...
+import binascii
+
+from six import string_types
+
+MSB = 0x80
+REST = 0x7F
+MSBALL = ~REST
+INT = 2 ** 31
+
+# https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
```
TimDaub: Javascript specs in Python code?
There is `sys.maxint`: https://docs.python.org/2/library/sys.html#sys.maxsize

diminator: > Javascript specs in Python code?
> There is sys.maxint: https://docs.python.org/2/library/sys.html#sys.maxsize

actually, It was intended to keep inline with the js reference library of the interledger protocol
python can handle larger integers than js:
`py:  9223372036854775807`
`js: 9007199254740991`

bigchaindb/crypto/buffer.py
```diff
@@ -0,0 +1,437 @@
...
+        
+        The hasher will pass these bytes into the hashing function. By overriding
+        the Writer class and implementing this method, the Hasher supports any of
+        the datatypes that a Writer can write.
+        
+        Args:
+            in_bytes (Buffer): Bytes to add to the hash.
+        """
+        out = in_bytes
+        if isinstance(out, (list, bytearray)):
```
r-marques: why list?

bigchaindb/crypto/buffer.py
```diff
@@ -0,0 +1,437 @@
...
+    def digest(self):
+        """
+        Return the hash.
+
+        Returns the finished hash based on what has been written to the Hasher so far.
+
+        Return:
+            Buffer: Resulting hash.
+        """
+        return self.hash.digest()
```
r-marques: would it make sense to return the hex representation of the digest?

bigchaindb/crypto/buffer.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        Get digest length for hashing algorithm.
+
+        Args:
+            algorithm (string): Hashing algorithm identifier.
+
+        Return:
+            int: Digest length in bytes.
+        """
+        return len(Hasher(algorithm).digest())
```
r-marques: cleaner to use `self.hash.digest_size`

bigchaindb/crypto/buffer.py
```diff
@@ -0,0 +1,437 @@
...
+        self.buffer = buffer
+        self.cursor = 0
+        self.bookmarks = []
+
+    @staticmethod
+    def from_source(source):
+        """
+        Create a Reader from a source of bytes.
+
+        Currently, this method only allows the creation of a Reader from a Buffer.
```
r-marques: What is a `buffer` in this case?

bigchaindb/crypto/buffer.py
diminator: internally the hasher and other writer/readers work with a bytes representation

bigchaindb/crypto/buffer.py
diminator: the write method supports chunks of bytes in a list `['bytes_1',... , 'bytes_n']`

bigchaindb/crypto/buffer.py
diminator: a `bytearray` that is combined with reading cursor

bigchaindb/crypto/fulfillment.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if not pieces[1] == '1':
+            raise ValueError('Fulfillment must be version 1')
+
+        if not re.match(FULFILLMENT_REGEX, serialized_fulfillment):
+            raise ValueError('Invalid fulfillment format')
+
+        bitmask = int(pieces[2])
+
+        from bigchaindb.crypto.bitmark_registry import BitmaskRegistry
```
r-marques: any particular reason to do the import in here?

bigchaindb/crypto/buffer.py
r-marques: something like `hasher.write([b'a', b'b'])`?

vrde:
No need to raise a `NotImplementedError` when using `abc.ABCMeta`.

```python
from abc import ABCMeta, abstractmethod

class Base(metaclass=ABCMeta):

    @staticmethod
    @abstractmethod
    def foo(val):
        '''Some documentation.'''

    @staticmethod
    @abstractmethod
    def bar(val):
        '''Some documentation.'''


class Concrete(Base):
    pass

# this will raise
# TypeError: Can't instantiate abstract class Concrete with abstract methods foo
c = Concrete()
```

(The plugin code has the same problem, I just created #148 to fix that)

r-marques:
:+1:

bigchaindb/crypto/buffer.py
diminator: note that its a staticmethod, hence no `self`

bigchaindb/crypto/fulfillment.py
diminator: circular imports in the `BitmaskRegistry`
suggestions welcome! @vrde ?

bigchaindb/crypto/asymmetric.py
```diff
@@ -0,0 +1,96 @@
...
+def hash_data(data):
+    """Hash the provided data using SHA3-256"""
+
+    return sha3.sha3_256(data.encode()).hexdigest()
+
+
+from bigchaindb.crypto.ecdsa import EcdsaSigningKey, EcdsaVerifyingKey, ecdsa_generate_key_pair
+
+SigningKey = EcdsaSigningKey
+VerifyingKey = EcdsaVerifyingKey
```
vrde: `SigningKey = ...` and `VerifyingKey = ...` redefine `class SigningKey` and `class VerifyingKey` at the top of the module.

bigchaindb/crypto/fulfillments/ed25519_sha256.py
```diff
@@ -0,0 +1,266 @@
...
+        if not isinstance(value, bytes):
+            value = value.encode()
+        self._message_prefix = value
+
+    @property
+    def max_dynamic_message_length(self):
+        return self._max_dynamic_message_length
+
+    @max_dynamic_message_length.setter
+    def max_dynamic_message_length(self, value):
```
vrde: Why do we need a `@property` and a `@<attrname>.setter`? What about having just the attribute in `self.max_dynamic_message_length`?

bigchaindb/crypto/crypto.md
```diff
@@ -0,0 +1,429 @@
...
+```python
+from bigchaindb.crypto.ed25519 import Ed25519SigningKey, Ed25519VerifyingKey
+from bigchaindb.crypto.fulfillments.ed25519_sha256 import Ed25519Sha256Fulfillment
+
+# We use base58 key encoding
+sk = Ed25519SigningKey(b'9qLvREC54mhKYivr88VpckyVWdAFmifJpGjbvV5AiTRs')
+vk = sk.get_verifying_key()
+
+# Create an ED25519-SHA256 condition
+ed25519_fulfill`nt = Ed25519Sha256Fulfillment()
```
vrde: typo: ``fulfill`nt``

bigchaindb/crypto/fulfillments/ed25519_sha256.py
```diff
@@ -0,0 +1,266 @@
+import base58
+
+from bigchaindb.crypto.ed25519 import Ed25519VerifyingKey
+from bigchaindb.crypto.fulfillments.base_sha256 import BaseSha256Fulfillment
+from bigchaindb.crypto.buffer import Predictor
+
+
+class Ed25519Sha256Fulfillment(BaseSha256Fulfillment):
```
vrde: I noticed that this module uses `@property` extensively. Why not putting everything in the `__init__` and check there the type of the values? [related](https://github.com/bigchaindb/bigchaindb/pull/146/files#r56849311)

diminator: the attributes may be assigned after construction, hence the checks on the setter.
The getters (and non-checked props) are merely to be consistent

vrde:
We are moving the code to another repository:
 - https://github.com/bigchaindb/cryptoconditions

PR #152, opened 2016-03-23
Switch to ED25519 signing scheme

diminator:
see #137, part of #127 

- [x] import keys from cryptoconditions
- [x] switch to ED25519

Note: all crypto tests are covered by cryptoconditions
Question: should we be able to configure the signing scheme?

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -298,7 +298,7 @@ class Bigchain(object):
         # Calculate the hash of the new block
         block_data = util.serialize(block)
         block_hash = crypto.hash_data(block_data)
-        block_signature = crypto.PrivateKey(self.me_private).sign(block_data)
+        block_signature = crypto.SigningKey(self.me_private).sign(block_data).decode()
 
         block = {
             'id': block_hash,
@@ -419,7 +419,7 @@ class Bigchain(object):
         }
 
         vote_data = util.serialize(vote)
-        signature = crypto.PrivateKey(self.me_private).sign(vote_data)
+        signature = crypto.SigningKey(self.me_private).sign(vote_data).decode()
 
         vote_signed = {
             'node_pubkey': self.me,
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -41,8 +41,8 @@ class Bigchain(object):
             host (str): hostname where the rethinkdb is running.
             port (int): port in which rethinkb is running (usually 28015).
             dbname (str): the name of the database to connect to (usually bigchain).
-            public_key (str): the base58 encoded public key for the ECDSA secp256k1 curve.
-            private_key (str): the base58 encoded private key for the ECDSA secp256k1 curve.
+            public_key (str): the base58 encoded public key for the ED25519 curve.
+            private_key (str): the base58 encoded private key for the ED25519 curve.
             keyring (list[str]): list of base58 encoded public keys of the federation nodes.
         """
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,7 +14,7 @@ import pytest
 
 DB_NAME = 'bigchain_test_{}'.format(os.getpid())
 
-CONFIG = {
+CONFIG_ECDSA = {
     'database': {
         'name': DB_NAME
     },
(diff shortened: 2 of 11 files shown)


r-marques:
I guess for now we should just stick to one signing scheme

diminator:
Note that this is a breaking change in the sense that the keypairs need to be regenerated

@vrde any feedback?

tests/conftest.py
```diff
@@ -25,8 +25,22 @@
...
-USER_PUBLIC_KEY = 'r3cEu8GNoz8rYpNJ61k7GqfR8VEvdUbtyHce8u1kaYwh'
+USER_PRIVATE_KEY_ECDSA = 'GmRZxQdQv7tooMijXytQkexKuFN6mJocciJarAmMwTX2'
+USER_PUBLIC_KEY_ECDSA = 'r3cEu8GNoz8rYpNJ61k7GqfR8VEvdUbtyHce8u1kaYwh'
+
+CONFIG_ED25519 = {
+    'database': {
+        'name': DB_NAME
+    },
+    'keypair': {
+        'private': '3wssdnSNsZYLvvQwuag5QNQnSfc5N38KV1ZeAoeHQQVe59N7vReJwXWANf5nncGxW63UzR4qHHv6DJhyLs9arJng',
```
r-marques: after updating cryptoconditions to version 0.1.1 we need to use the 32 byte private key instead

tests/conftest.py
```diff
@@ -37,17 +51,17 @@ def restore_config(request, node_config):
 
 @pytest.fixture(scope='module')
 def node_config():
-    return copy.deepcopy(CONFIG)
+    return copy.deepcopy(CONFIG_ED25519)
```
r-marques: Does this mean that we are only going to test with `ED25519` and not `ECDSA`?
If so we should decide right now that we are only going to use `ed25519` and remove support for `ecdsa` in `crypto.py`

diminator: ok, I'll do that

r-marques:
:+1:

ttmc:
@diminator Please update the section of the docs where the signature algorithm is stated, and give a link or two to references on Ed25519.

That's in `/docs/source/cryptography.md`

http://bigchaindb.readthedocs.org/en/develop/cryptography.html

tests/conftest.py
```diff
@@ -14,7 +14,7 @@
 
 DB_NAME = 'bigchain_test_{}'.format(os.getpid())
 
-CONFIG = {
+CONFIG_ECDSA = {
```
vrde: Why the need to change the constant name?

bigchaindb/core.py
```diff
@@ -298,7 +298,7 @@ def create_block(self, validated_transactions):
         # Calculate the hash of the new block
         block_data = util.serialize(block)
         block_hash = crypto.hash_data(block_data)
-        block_signature = crypto.PrivateKey(self.me_private).sign(block_data)
+        block_signature = crypto.SigningKey(self.me_private).sign(block_data).decode()
```
vrde: What is the `decode` for?

tests/db/test_bigchain_api.py
```diff
@@ -185,11 +185,11 @@ def test_get_previous_block_id(self, b):
 
     def test_create_new_block(self, b):
         new_block = b.create_block([])
-        block_hash = hash_data(util.serialize(new_block['block']))
+        block_hash = crypto.hash_data(util.serialize(new_block['block']))
 
         assert new_block['block']['voters'] == [b.me]
         assert new_block['block']['node_pubkey'] == b.me
-        assert PublicKey(b.me).verify(util.serialize(new_block['block']), new_block['signature']) is True
+        assert crypto.VerifyingKey(b.me).verify(util.serialize(new_block['block']), new_block['signature']) is True
```
vrde: If `VerifyingKey` is used to verify a signature, why it's not just a function? Why do we need an object here?

tests/db/test_bigchain_api.py
```diff
@@ -371,8 +371,8 @@ def test_invalid_transactions_in_block(self, b, user_public_key, ):
         }
 
         block_data = util.serialize(block)
-        block_hash = hash_data(block_data)
-        block_signature = PrivateKey(b.me_private).sign(block_data)
+        block_hash = crypto.hash_data(block_data)
+        block_signature = crypto.SigningKey(b.me_private).sign(block_data)
```
vrde: If `SigningKey` is used to sign data, why it's not just a function? Why do we need an object here?

tests/conftest.py
vrde: Did you remove support for ECDSA? if so, can you please remove `USER_PRIVATE_KEY_ECDSA` and `USER_PUBLIC_KEY_ECDSA`, and all the code related to that?

tests/conftest.py
```diff
@@ -25,8 +25,22 @@
 }
 
 # Test user. inputs will be created for this user. Cryptography Keys
-USER_PRIVATE_KEY = 'GmRZxQdQv7tooMijXytQkexKuFN6mJocciJarAmMwTX2'
-USER_PUBLIC_KEY = 'r3cEu8GNoz8rYpNJ61k7GqfR8VEvdUbtyHce8u1kaYwh'
+USER_PRIVATE_KEY_ECDSA = 'GmRZxQdQv7tooMijXytQkexKuFN6mJocciJarAmMwTX2'
+USER_PUBLIC_KEY_ECDSA = 'r3cEu8GNoz8rYpNJ61k7GqfR8VEvdUbtyHce8u1kaYwh'
+
+CONFIG_ED25519 = {
```
vrde: Since we will have just one configuration, what's the need to have the name of the algo in the name of the constant?

bigchaindb/core.py
diminator: by default, the signature is `bytes`, but that wont serialize

tests/db/test_bigchain_api.py
diminator: because you have to instantiate a key with a certain value first before you can verify

ofc you can create a wrapper function verify(msg, vk) but thats irrelevant I guess

tests/db/test_bigchain_api.py
diminator: see above

PR #155, opened 2016-03-30
Use gunicorn as the application server

vrde:
Right now the API is served by the Flask development server.
This PR is to upgrade the application server to gunicorn, and it's safer to bind to `0.0.0.0`.

Note that the API server is run together with `bigchaindb start`.

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -24,6 +24,10 @@ def e(key, default=None, conv=None):
 
 
 config = {
+    'server': {
+        'bind': ':'.join([e('BIGCHAIN_SERVER_BINDHOST', default='0.0.0.0'),
+                          e('BIGCHAIN_SERVER_BINDPORT', default='5000')])
+    },
     'database': {
         'host': e('BIGCHAIN_DATABASE_HOST', default='localhost'),
         'port': e('BIGCHAIN_DATABASE_PORT', default=28015),
@@ -49,3 +53,4 @@ config = {
 # for more info.
 _config = copy.deepcopy(config)
 from bigchaindb.core import Bigchain  # noqa
+
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -25,8 +25,7 @@ def e(key, default=None, conv=None):
 
 config = {
     'server': {
-        'bind': ':'.join([e('BIGCHAIN_SERVER_BINDHOST', default='0.0.0.0'),
-                          e('BIGCHAIN_SERVER_BINDPORT', default='5000')])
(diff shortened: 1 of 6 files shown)


bigchaindb/__init__.py
```diff
@@ -24,6 +23,9 @@ def e(key, default=None, conv=None):
 
 
 config = {
+    'server': {
+        'bind': e('BIGCHAIN_SERVER_BIND', default='0.0.0.0:5000'),
```
r-marques: should we allow it to bind to all interfaces by default or should we force the user the change the default?

setup.py
```diff
@@ -79,6 +79,7 @@
         'bitcoin==1.1.42',
         'flask==0.10.1',
         'requests==2.9',
+        'gunicorn',
```
r-marques: any reason to omit the version of gunicorn?

r-marques:
tested and seems good :+1:

bigchaindb/__init__.py
vrde: Since it's the public entry point for BigchainDB, it should bind to `0.0.0.0`.

setup.py
vrde: Gunicorn it's quite stable, and I don't think there will be many API changes in the future. In general, I'm reluctant in pinning the version in the `setup.py`, but I'm still not 100% sure how to handle it.

In general, if the project supports semantic versioning, we can stick with the major version. Should I pin the version to `19` then?

r-marques: Not sure what are the best practices but we do it for everything else

diminator:
:+1: after merging develop in here

setup.py
sbellem: this can help: http://python-packaging-user-guide.readthedocs.org/en/latest/requirements/#install-requires

it appears that inequality operators (`<`, `<=`, `>`, `>=`) are preferred

vrde: OK, I won my laziness and actually we can use the [`~` operator](https://www.python.org/dev/peps/pep-0440/#compatible-release). It's totally rad. We can do "gunicorn~=19.0" and it'd install the last version of `gunicorn` with major version `19`.

I also [filed an issue](https://github.com/benoitc/gunicorn/issues/1240) to ask if they use semantic versioning, so we can safely assume that if we pin version `19` we won't encounter compatibility issues.

For now I'll pin version `19` tho.

vrde: @sbellem wrote:
> it appears that inequality operators (<, <=, >, >=) are preferred

check out my last comment, the `~` operator is gooood! :dancers:

sbellem: :+1:

PR #166, opened 2016-04-07
Added script to release allocated-but-unassociated elastic IPs on AWS

ttmc:
Like the title says, plus also some documentation about how to use it.

It's a useful script because allocated-but-unassociated elastic IPs on AWS cost money.


deploy-cluster-aws/release_eips.py
```diff
@@ -0,0 +1,51 @@
...
+    domain = eip['Domain']
+    print('(It has Domain = {}.)'.format(domain))
+    try:
+        if domain == 'vpc':
+            client.release_address(AllocationId=eip['AllocationId'])
+        else:
+            client.release_address(PublicIp=public_ip)
+    except botocore.exceptions.ClientError as e:
+        print('A boto error occurred:')
+        raise
```
vrde: what about removing the `try ... except`? The stack trace of the exception contains information about the error.

ttmc: Oh, yeah, that `try ... except` was left over from a time when I was using it for something else. I'll remove it as suggested.

vrde: great! :santa:

PR #153, opened 2016-03-24
[proposal] Improve configuration by using env variables correctly

vrde:
Hey, this is a proposal to improve the configuration framework. This should make the life of docker users easier (see #36, #49).

You can find most of the information [here](https://github.com/bigchaindb/bigchaindb/blob/improve-config/docs/source/bigchaindb-cli.md).

PR #167, opened 2016-04-07
RethinkDB User Accounts

saschad:


r-marques:
Besides the drivers we also need to check or docker files and documentation and make sure to state that this only works with rethinkdb 2.3+

r-marques:
Also need to check how this fits in the federation model. Who can and cannot create user accounts? Is there an admin/master account? If so we need to rethink (no pun intended) how to handle permissions

ttmc:
Hi @saschad --- I see that you already signed the BigchainDB Individual Contributor Agreement: I got the email. I sent you an email reply explaining the next step. You'll have to do that before we can merge your pull requests.

Over on PR #153, @vrde is making some changes to how configuration works, so he'll probably also have some comments.

saschad:
BEGIN BLOCK

tK11CNNGZymlN3ZACvgnYfVewUQvPMjQaN1XgruIPHQH3HerpfEqSo3XHeTWWAR

END BLOCK

ttmc:
@saschad User accounts and related issues are something on our roadmap, but we need to discuss and think about them more first. BigchainDB is a non-standard system because it's supposed to be decentralized.

Therefore I'm going to close this pull request for now. I created an issue #177 where we can discuss these matters further.

PR #172, opened 2016-04-11
Various AWS deployment improvements

ttmc:
* Deleted the entire file `fab_prepare_chain.py` because it was a hack to make Fabric ssh to only one node instead of all of them. Wrote the `set_hosts()` function in `fabfile.py` as a cleaner way to enable the calling of only one node (using `fab set_hosts:one_node task1 task2 ...`).
* Revamped the launching of BigchainDB into three steps: 1) Run `bigchaindb -y configure` on all nodes, 2) Run `bigchain init` on one node, 3) Run `screen -d -m bigchaindb -y start &` on all nodes
* Removed a bunch of unused Python package/function imports from `fabfile.py`
* In `fabfile.py`, set `env.disable_known_hosts = True` for now (because we're not currently concerned about man-in-the-middle attacks)
* Renamed `hosts_dev` to `public_dns_names`

There's still more to do. For example, we still have to get the keychain set up properly on all the nodes.


deploy-cluster-aws/fabfile.py
```diff
@@ -1,47 +1,58 @@
...
+    Args:
+        hosts (str): 'one_node' or 'two_nodes'
+
+    Example:
+        fab set_hosts:one_node init_bigchaindb
+    """
+    if hosts == 'one_node':
+        env.hosts = public_dns_names[0]
+    elif hosts == 'two_nodes':
+        env.hosts = public_dns_names[0:1]
```
vrde: slices are `[n, m)`, so `[0:1] == [0]`. Did you want to do `public_dns_names[0:2]`?

ttmc: I will change it so:
* the 'one_node' case uses `public_dns_names[:1]` 
* the 'two_node' case uses `public_dns_names[:2]`

Issue #128, opened 2016-03-10
Support for multiple inputs


r-marques:
Ability to transfer multiple digital assets in a single transaction.
The transaction structure will change with this feature.

The current transaction looks like:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "current_owner": "<pub-key>",
        "new_owner": "<pub-key>",
        "input": "<sha3 hash>",
        "operation": "<string>",
        "timestamp": "<timestamp from client>",
        "data": {
            "hash": "<SHA3-256 hash hexdigest of payload>",
            "payload": {
                "title": "The Winds of Plast",
                "creator": "Johnathan Plunkett",
                "IPFS_key": "QmfQ5QAjvg4GtA3wg3adpnDJug8ktA1BxurVqBD8rtgVjP"
            }
        }
    },
    "signature": "<signature of the transaction>"
}
```

The new transaction structure will need to have a inputs and outputs:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "current_owner": "<pub-key>",
        "new_owner": "<pub-key>",
        "inputs": [
            {
                "txid": "<sha3 hash>",
                "number": "input number"
            }
            ],
        "operation": "<string>",
        "timestamp": "<timestamp from client>",
        "data": {
            "hash": "<SHA3-256 hash hexdigest of payload>",
            "payload": {
                "title": "The Winds of Plast",
                "creator": "Johnathan Plunkett",
                "IPFS_key": "QmfQ5QAjvg4GtA3wg3adpnDJug8ktA1BxurVqBD8rtgVjP"
            }
        }
    },
    "signature": "<signature of the transaction>"
}
```

At this point I am not sure if we need to have outputs. If we add multiple outputs to possibly different owners we also need a way to link the inputs to the outputs for non divisible assets.

Its probably best to support multi outputs when addressing divisible assets and/or crypto conditions (where we will need to add a fulfillment to the inputs and a condition to the outputs) 

Right now I will focus on adding only multiple input support to create transactions between a single `current_owner` and a single `new_owner`.

r-marques:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "fulfillments": [
                {
                    "current_owners": ["list of <pub-keys>"],
                    "input": {
                        "txid": "<sha3 hash>",
                        "cid": "condition index"
                    },
                    "fulfillment": "fulfillement of condition cid",
                    "fid": "fulfillment index"
                }
            ],
[...]

PR #174, opened 2016-04-12
WIP Feat/128/multiple input output

### LINKED ISSUE #127: Crypto-conditions spec according to ILP

r-marques:
This is still a work in progress, created so that the review can start.
Missing are still some tests and documentation.

This resolves #127 and #128


bigchaindb/util.py
```diff
@@ -193,16 +316,94 @@ def verify_signature(signed_transaction):
...
+        is_valid = parsed_fulfillment.validate(serialize(fulfillment_message))
+
+        # if not a `CREATE` transaction
+        if fulfillment['input']:
+            is_valid &= parsed_fulfillment.condition.serialize_uri() == \
+                fulfillment_message['condition']['condition']['uri']
+        if not is_valid:
+            return False
+
+    return True
```
TimDaub: `return is_valid`?

bigchaindb/core.py
```diff
@@ -180,36 +178,38 @@ def get_tx_by_payload_hash(self, payload_hash):
             returns `None`
         """
 
-        cursor = r.table('bigchain')\
-            .get_all(payload_hash, index='payload_hash')\
+        cursor = r.table('bigchain') \
+            .get_all(payload_hash, index='payload_hash') \
```
vrde: remove whitespaces before `\`

bigchaindb/core.py
```diff
@@ -180,36 +178,38 @@ def get_tx_by_payload_hash(self, payload_hash):
...
             .run(self.conn)
 
         transactions = list(cursor)
         return transactions
 
-    def get_spent(self, txid):
+    def get_spent(self, inp):
```
vrde: Find a more descriptive name for `inp`.

bigchaindb/util.py
diminator: Nope, you only want to break on False (for loop)

bigchaindb/core.py
diminator: it was added by my built-in pylint, so guess it's legit

bigchaindb/util.py
TimDaub: Oh, missed that it was in one. Nevermind

tests/db/test_bigchain_api.py
```diff
@@ -22,52 +26,66 @@ def test_remove_unclosed_sockets():
...
+        tx = b.create_transaction(b.me, user_sk, None, 'CREATE')
 
-    def test_create_transaction(self, b):
-        tx = b.create_transaction('a', 'b', 'c', 'd')
-
-        assert sorted(tx) == sorted(['id', 'transaction'])
-        assert sorted(tx['transaction']) == sorted(['current_owner', 'new_owner', 'input', 'operation',
-                                                    'timestamp', 'data'])
+        assert sorted(tx) == sorted(['id', 'transaction', 'version'])
+        assert sorted(tx['transaction']) == sorted(['conditions', 'data', 'fulfillments', 'operation', 'timestamp'])
```
sbellem: Just for the sake of simplicity: is it really needed to sort the right hand side? Especially, given that it is already sorted when handed over to `sorted()`.

r-marques: didn't noticed that

sbellem: Ah, ok, it does not matter much in the end. Perhaps the tests can run faster without `sorted`? :smile:

bigchaindb/util.py
```diff
@@ -193,16 +316,94 @@ def verify_signature(signed_transaction):
...
     """
 
-    data = signed_transaction.copy()
+    for fulfillment in signed_transaction['transaction']['fulfillments']:
+        fulfillment_message = get_fulfillment_message(signed_transaction, fulfillment)
+        # verify the fulfillment (for now lets assume there is only one owner)
+        try:
+            parsed_fulfillment = Fulfillment.from_uri(fulfillment['fulfillment'])
+        except Exception:
+            return False
```
sbellem: Would it be possible to use explicit exceptions?

see the section on exceptions in https://www.python.org/dev/peps/pep-0008/#programming-recommendations

rhsimplex:
Comments on the [documentation](https://github.com/bigchaindb/bigchaindb/blob/feat/127/crypto-conditions-ilp-bigchain-integration/docs/source/python-server-api-examples.md), **with annoying bold shouty headers**:

#### The "retrieved" transactions return an `input: None`
rather than `null` as in the documentation
```python
        ...
        "fulfillments":[
            {
                "current_owners":[
                    "3LQ5dTiddXymDhNzETB1rEkp4mA7fEV1Qeiu5ghHiJm9"
                ],
                "fid":0,
                "fulfillment":"cf:4:Iq-BcczwraM2UpF-TDPdwK8fQ6IXkD_6uJaxBZd984yxCGX7Csx-S2FBVe8LVyW2sAtmjsOSV0oiw9-s_9qSJB0dDUl_x8YQk5yxNdQyNVWVM1mWSGQL68gMngdmFG8O",
                "input":null  # <---- this right here. Should be None
            }
       ...
```

#### Input and output are mixed up in the same blocks, 
which could be pretty confusing for someone not familiar with python. For example:

```python
# Sign the transaction
tx_transfer_signed = b.sign_transaction(tx_transfer, testuser1_priv)

# Write the transaction
b.write_transaction(tx_transfer_signed)

# Check if the transaction is already in the bigchain
tx_transfer_retrieved = b.get_transaction(tx_transfer_signed['id'])

{
    "id":"aa11365317cb89bfdae2375bae76d6b8232008f8672507080e3766ca06976dcd",
    "transaction":{
        "conditions":[
            {
                "cid":0,
                "condition":{
                    "details":{
...
```
The output block is not actually the printed output of the line above. Consider the standard:

```python
# Sign the transaction
>>> tx_transfer_signed = b.sign_transaction(tx_transfer, testuser1_priv)

# Write the transaction
>>> b.write_transaction(tx_transfer_signed)

# Check if the transaction is already in the bigchain
>>> tx_transfer_retrieved = b.get_transaction(tx_transfer_signed['id'])
>>> tx_transfer_retrieved
{
    "id":"aa11365317cb89bfdae2375bae76d6b8232008f8672507080e3766ca06976dcd",
    "transaction":{
        "conditions":[
            {
                "cid":0,
                "condition":{
                    "details":{
...
```
OR separate the input and output into different blocks (for instance to make it easier to copy-paste commands). e.g.:

```python
some(input)
```
returns
```python
{ some: output }
```
I see this other places in the documentation (not just on this branch/PR) so maybe it's been settled, but I encourage you (all) to reconsider because it's error-prone. Literally the first output example [here](http://bigchaindb.readthedocs.org/en/master/python-server-api-examples.html) on the master branch docs has a mistake:

```python
# retrieve a transaction from the bigchain
tx_retrieved = b.get_transaction(tx_signed['id'])

   'id': '6539dded9479c47b3c83385ae569ecaa90bcf387240d1ee2ea3ae0f7986aeddd',
   'transaction': {   'current_owner': 'pvGtcm5dvwWMzCqagki1N6CDKYs2J1cCwTNw8CqJic3Q',
                      'data': {   'hash': '872fa6e6f46246cd44afdb2ee9cfae0e72885fb0910e2bcf9a5a2a4eadb417b8',
                                  'payload': {'msg': 'Hello BigchainDB!'}},
                      'input': None,
                      'new_owner': 'ssQnnjketNYmbU3hwgFMEQsc4JVYAmZyWHnHCtFS8aeA',
                      'operation': 'CREATE',
                      'timestamp': '1455108421.753908'}}
```
No open bracket on the output, because of this strange copy-pasted amalgamation _Inputoutputsalat_. Again, from a beginner's standpoint, there is no distinction here between input and output. It's confusing.

#### Not a criticism just curious 
Really both private keys have to be in the same place? Is this the standard?
```python
# Sign with both private keys
tx_multisig_transfer_signed = b.sign_transaction(tx_multisig_transfer, [testuser1_priv, testuser2_priv])
```

#### Maybe remind the user to wait 
```# Check if the transaction is already in the bigchain``` could say ```# Check if the transaction is already in the bigchain (may take a few moments)``` since if you're just blasting through the examples, you might be left wondering what happened here.

### If you're having assert problems I feel bad for ya son, I got 99 problems but a bool ain't one
You can drop the `== True` on the assert statements. In fact this example throws an `AssertionError` because the last line returns a `dict`. If you want to be explicit:
```python
# Optional validation checks
assert threshold_fulfillment.validate(threshold_tx_fulfillment_message)
assert b.verify_signature(threshold_tx_transfer)
assert b.validate_transaction(threshold_tx_transfer) is not None
```

### And again with the output
This is the output of...`b.write_transaction`? A print of `threshold_tx_transfer`?
```python
b.write_transaction(threshold_tx_transfer)

{
    "assignee":"3LQ5dTiddXymDhNzETB1rEkp4mA7fEV1Qeiu5ghHiJm9",
    "id":"a45b2340c59df7422a5788b3c462dee708a18cdf09d1a10bd26be3f31af4b8d7",
    "transaction":{
        "conditions":[
            {
```

### Final remark
Consider including the last example as a separate script with helpful print statements (instead of or in addition to the example documentation), since it's really long.

Other than the minor nitpicks, all the examples work fine for me and it's pretty clear why we would want to do this. Good work!

diminator:
> The "retrieved" transactions return an input: None
> 
> rather than null as in the documentation

yeah, I guess you are using the python dicts and I am representing the actual json
@r-marques @vrde : show python or json version in the docs?

The JSON version would match the rethinkdb dashboard, while the python version would match the code example better

PS: I am actually using the following print statement
`print(json.dumps(tx, sort_keys=True, indent=4, separators=(',', ':')))`

r-marques:
@diminator I would show python dicts because that that's whats actually returned by the methods

bigchaindb/core.py
```diff
@@ -174,36 +174,38 @@ def get_tx_by_payload_hash(self, payload_hash):
...
-            txid (str): transaction id.
+            tx_input (dict): Input of a transaction in the form `{'txid': 'transaction id', 'cid': 'condition id'}`
 
         Returns:
             The transaction that used the `txid` as an input if it exists else it returns `None`
         """
         # checks if an input was already spent
-        # checks if the bigchain has any transaction with input `transaction_id`
+        # checks if the bigchain has any transaction with input {'txid': ..., 'cid': ...}
         response = r.table('bigchain').concat_map(lambda doc: doc['block']['transactions'])\
```
vrde: what's the need of `lamba` here?
`r.table('bigchain').concat_map(r.row['block']['transactions']) ...` is much simpler and should be equivalent.

bigchaindb/core.py
```diff
@@ -174,36 +174,38 @@ def get_tx_by_payload_hash(self, payload_hash):
...
-            .filter(lambda transaction: transaction['transaction']['input'] == txid).run(self.conn)
+            .filter(lambda transaction: transaction['transaction']['fulfillments']
+                    .contains(lambda fulfillment: fulfillment['input'] == tx_input))\
+            .run(self.conn)
 
         # a transaction_id should have been spent at most one time
         transactions = list(response)
         if transactions:
             if len(transactions) != 1:
                 raise Exception('`{}` was spent more then once. There is a problem with the chain'.format(
```
vrde: Do not use generic `Exception`s. There is a `bigchaindb.exceptions.DoubleSpent`. If it's not suitable for this case, create a new one.

bigchaindb/core.py
```diff
@@ -219,17 +221,26 @@ def get_owned_ids(self, owner):
             list: list of `txids` currently owned by `owner`
         """
 
-        response = r.table('bigchain')\
-                    .concat_map(lambda doc: doc['block']['transactions'])\
-                    .filter({'transaction': {'new_owner': owner}})\
-                    .pluck('id')['id']\
-                    .run(self.conn)
+        response = r.table('bigchain') \
+            .concat_map(lambda doc: doc['block']['transactions']) \
```
vrde: `.concat_map(r.row['block']['transactions'])`?

r-marques: Not according to the documentation. Seems like r.row does not work for all cases.
> Note that row does not work within subqueries to access nested documents; you should use anonymous functions to access those documents instead

bigchaindb/core.py
r-marques: not sure if it works. just following the rethinkdb documentation

bigchaindb/core.py
```diff
@@ -219,17 +221,26 @@ def get_owned_ids(self, owner):
...
+                    .contains(lambda c: c['new_owners']
+                              .contains(owner))) \
+            .run(self.conn)
         owned = []
 
-        # remove all inputs already spent
-        for tx_input in list(response):
-            if not self.get_spent(tx_input):
-                owned.append(tx_input)
+        for tx in response:
```
vrde: The following part is a bit difficult to understand, a couple of ideas to make it simpler:
 - add some inline comments, using "natural language" to explain what's happening
 - restructure the code to lower the complexity (since we have an `if`, inside an `if`, inside a `for`, inside a `for`), maybe some of the parts can be split up into different funcs?

bigchaindb/core.py
```diff
@@ -429,37 +440,37 @@ def write_vote(self, block, vote, block_number):
...
         return last_voted[0]
 
     def get_unvoted_blocks(self):
         """Return all the blocks that has not been voted by this node."""
 
-        unvoted = r.table('bigchain')\
-            .filter(lambda doc: doc['votes'].contains(lambda vote: vote['node_pubkey'] == self.me).not_())\
-            .order_by(r.asc((r.row['block']['timestamp'])))\
+        unvoted = r.table('bigchain') \
+            .filter(lambda doc: doc['votes'].contains(lambda vote: vote['node_pubkey'] == self.me).not_()) \
```
vrde: didn't try it, but something like:
`.filter(r.row['votes'].contains(lambda vote: vote['node_pubkey'].eq(self.me)).not_())`

bigchaindb/util.py
```diff
@@ -162,8 +165,52 @@ def create_tx(current_owner, new_owner, tx_input, operation, payload=None):
 
     Raises:
         TypeError: if the optional ``payload`` argument is not a ``dict``.
-    """
 
+    Reference:
+        {
```
vrde: should this JSON be enclosed in a multiline code block?

bigchaindb/util.py
```diff
@@ -175,52 +222,126 @@ def create_tx(current_owner, new_owner, tx_input, operation, payload=None):
...
 
     # serialize and convert to bytes
-    tx_serialized = serialize(tx)
-    tx_hash = crypto.hash_data(tx_serialized)
+    tx_hash = get_hash_data(tx)
 
     # create the transaction
     transaction = {
         'id': tx_hash,
+        'version': 1,
```
vrde: good.

bigchaindb/util.py
```diff
@@ -175,52 +222,126 @@ def create_tx(current_owner, new_owner, tx_input, operation, payload=None):
...
         'transaction': tx
     }
 
     return transaction
 
 
-def sign_tx(transaction, private_key):
+# TODO: Change sign_tx to populate the fulfillments
+def sign_tx(transaction, sks):
```
vrde: `sks` is difficult to understand, be more explicit.

bigchaindb/util.py
```diff
@@ -175,52 +222,126 @@ def create_tx(current_owner, new_owner, tx_input, operation, payload=None):
...
     """Sign a transaction
 
     A transaction signed with the `current_owner` corresponding private key.
 
     Args:
         transaction (dict): transaction to sign.
-        private_key (str): base58 encoded private key to create a signature of the transaction.
+        sk (base58 str): base58 encoded private key to create a signature of the transaction.
```
vrde: Also, change this.

vrde: not sure if the Google style docstrings support the syntax `sk (base58 str)`

bigchaindb/util.py
```diff
@@ -175,52 +222,126 @@ def create_tx(current_owner, new_owner, tx_input, operation, payload=None):
...
+        key_pairs[vk] = signing_key
+
+    tx = copy.deepcopy(transaction)
+
+    for fulfillment in tx['transaction']['fulfillments']:
+        fulfillment_message = get_fulfillment_message(transaction, fulfillment)
+        parsed_fulfillment = cc.Fulfillment.from_json(fulfillment_message['condition']['condition']['details'])
+        parsed_fulfillment_copy = copy.deepcopy(parsed_fulfillment)
+        # single current owner
+        if isinstance(parsed_fulfillment, cc.Ed25519Fulfillment):
```
vrde: the branches in the `if` ... `else` can be put in smaller functions/methods to limit the complexity of the code and simplify tests.

bigchaindb/core.py
r-marques: I understand using `r.row` to eliminate one `lambda` function but why move away from the python syntax when `eq` an `==` are the same?

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
         bool: True if the signature is correct, False otherwise.
     """
 
-    data = signed_transaction.copy()
+    for fulfillment in signed_transaction['transaction']['fulfillments']:
+        fulfillment_message = get_fulfillment_message(signed_transaction, fulfillment)
+        # verify the fulfillment (for now lets assume there is only one owner)
+        try:
+            parsed_fulfillment = cc.Fulfillment.from_uri(fulfillment['fulfillment'])
+        except Exception:
```
vrde: **NEVER** use

```
try:
    ...
except Exception:
    ...
```

bigchaindb/util.py
r-marques: I added it to the docstring as a reference while implementing the new transaction structure. It can maybe be removed

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
...
+            return False
+        is_valid = parsed_fulfillment.validate(serialize(fulfillment_message))
+
+        # if not a `CREATE` transaction
+        if fulfillment['input']:
```
vrde: why not checking if the transaction is a `CREATE` then?

bigchaindb/util.py
vrde: > You only want to break on False (for loop)

Not saying you have to do it, but it's totally fine to use `break` in Python to exit a `for` loop.

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
...
+    return True
+
+
+def get_fulfillment_message(transaction, fulfillment, serialized=False):
+    """Get the fulfillment message for signing a specific fulfillment in a transaction
+
+    Args:
+        transaction (dict): a transaction
+        fulfillment (dict): a specific fulfillment (for a condition index) within the transaction
+        serialized (bool): False returns a dict, True returns a serialized string
```
vrde: `serialized (Optional[bool])`

vrde: It's weird to change the return value of a function using a flag. The *serialization* can be added by the developer composing `get_fulfillment_message` with another function that just serializes.

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
...
+
+    common_data = {
+        'operation': transaction['transaction']['operation'],
+        'timestamp': transaction['transaction']['timestamp'],
+        'data': transaction['transaction']['data'],
+        'version': transaction['version'],
+        'id': transaction['id']
+    }
+
+    fulfillment_message = common_data.copy()
```
vrde: what's the need for `common_data.copy()`? `common_data` is already a new object. Is it because you don't want side effects on `transaction['transaction']['data']`? Also, `dict.copy` returns a shallow copy of the dictionary, so there might be side effects on `transaction['transaction']['data']` anyway.

AFAIK `transaction['transaction']['data']` can be a nested dictionary, if there is a chance to have side effects there then you can just copy only that dictionary using `copy.deepcopy`.

The question here is: do we need to copy or  not?

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
...
+    if fulfillment['input']:
+        # get previous condition
+        previous_tx = b.get_transaction(fulfillment['input']['txid'])
+        conditions = sorted(previous_tx['transaction']['conditions'], key=lambda d: d['cid'])
+        fulfillment_message['condition'] = conditions[fulfillment['input']['cid']]
+    # if `CREATE` transaction
+    # there is no previous transaction so we need to create one on the fly
+    else:
+        current_owner = transaction['transaction']['fulfillments'][0]['current_owners'][0]
+        condition = json.loads(cc.Ed25519Fulfillment(public_key=current_owner).serialize_json())
```
vrde: Why encode to json then decode from json?

bigchaindb/util.py
```diff
@@ -251,16 +373,87 @@ def verify_signature(signed_transaction):
...
+    fulfillment_message.update({
+        'input': fulfillment['input'],
+        'condition': None,
+    })
+
+    # if `TRANSFER` transaction
+    if fulfillment['input']:
+        # get previous condition
+        previous_tx = b.get_transaction(fulfillment['input']['txid'])
+        conditions = sorted(previous_tx['transaction']['conditions'], key=lambda d: d['cid'])
```
vrde: Do we need to sort conditions? Aren't they already in an array?

bigchaindb/util.py
r-marques: Not sure I understand. The flag is just to indicate if you want to return the result serialized or not.
We can remove that option and force however uses this method to always serialize afterwards if needed

bigchaindb/util.py
r-marques: They are already in an list but there are no guarantees on the order of the list.
I need to sort it so that I can access the correct condtion (or output) in a multi output transaction.
`conditions[fulfillment['input']['cid']]`

bigchaindb/util.py
r-marques: this bit of code was executed inside a for loop before it was isolated in a method. the copy is unnecessary now. will remove it

bigchaindb/util.py
diminator: ofc, but that doesnt apply here

bigchaindb/util.py
diminator: an update to the docs will suffice

bigchaindb/util.py
r-marques: We want a python dictionary and cryptoconditions has no method to return a python dict so we retrieve the json of the condition and then we decode it into a python dict

bigchaindb/util.py
r-marques: The `verify_signature` method doesn't really care what type of transaction it is. It first verifies if the fulfillments are correct and then it checks if the fulfillments actually match the conditions in the input, if a input exists

Issue #156, opened 2016-03-31
Refactor monitoring code


vrde:
I'd like to move the monitoring code from `bigchaindb.core.Bigchain` to the modules `bigchaindb.block` and `bigchaindb.voter`.

Why? For three main reasons:
- to keep the `Bigchain` class cleaner
- the call to `monitor = Monitor()` ([core.py:15](https://github.com/bigchaindb/bigchaindb/blob/develop/bigchaindb/core.py#L15)) might trigger `bigchaindb.config_utils.autoconfigure` **before** other configuration functions
- because `bigchaindb.block` and `bigchaindb.voter` are the places where the computation happens.

Unfortunately #153 partially depends on this issue. I can use a workaround to have #153 merged on `develop` but this issue should be fixed relatively soon.

vrde:
@rhsimplex any thoughts on that?

rhsimplex:
Oh hi just saw this, sorry. Yeah I don't see why not. Do you need anything from me?

PR #181, opened 2016-04-13
Move monitoring code from the main class to processes

vrde:
This PR is related to #156.

I might need some help from @rhsimplex to check if the new setup is compatible to the one we had before.

Note also we don't monitor `create_transaction` anymore, because this action will be done on the client side, not on the server side.

I moved `write_transaction` to the API server, because this is where the write happens. An idea for the future: we can add `statsd` on every API endpoint to measure how the HTTP Server behaves.

### DIFF
--- a/bigchaindb/block.py
+++ b/bigchaindb/block.py
@@ -11,8 +11,6 @@ from bigchaindb.monitor import Monitor
 
 logger = logging.getLogger(__name__)
 
-monitor = Monitor()
-
 
 class Block(object):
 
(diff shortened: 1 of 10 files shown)


rhsimplex:
Transaction Write Time/Rate are not showing up properly in the dashboard. I'll try to find specifically what the failure is.

r-marques:
the code seems good :+1:

vrde:
@rhsimplex moar tests, moar coverage. Everything should work fine now :ok_hand:.

rhsimplex:
:+1:

vrde:
@rhsimplex quick fix to remove the logging from `requests`.

PR #183, opened 2016-04-15
More solid webserver

vrde:
This PR is to fix a mistake I made on sharing the connection to the database between different threads.

I've added a context manager to manage a pool of instances (`Bigchain` instances, in our case), and installed it in the views for the API.

I've also changed the default bind address from `0.0.0.0` to `localhost`.


bigchaindb/__init__.py
```diff
@@ -9,7 +9,7 @@
     'server': {
         # Note: this section supports all the Gunicorn settings:
         #       - http://docs.gunicorn.org/en/stable/settings.html
-        'bind': '0.0.0.0:9984',
+        'bind': 'localhost:9984',
```
sbellem: @vrde:

1. Could you briefly comment on why this change is necessary?
2. Is there a reason for not parametrizing the host?

bigchaindb/web/server.py
```diff
@@ -79,8 +80,8 @@ def create_server(settings):
     if not settings.get('threads'):
         settings['threads'] = (multiprocessing.cpu_count() * 2) + 1
 
-    debug = settings.pop('debug', False)
-    app = create_app(debug)
+    app = create_app(settings)
+    settings.pop('debug', False)
```
sbellem: I wonder why popping the value it, with a default, but not assigning it. Would simply deleting `debug` do the work?

```python
try:
    del settings['debug']
except KeyError:
    pass
```

bigchaindb/util.py
```diff
@@ -31,6 +34,53 @@ def start(self):
             self.processes.append(proc)
 
 
+# Inspired by:
+# - http://stackoverflow.com/a/24741694/597097
+def pool(builder, size, timeout=None):
```
r-marques: document timeout arg

bigchaindb/util.py
```diff
@@ -31,6 +34,53 @@ def start(self):
...
+        statement.
+    """
+
+    lock = threading.Lock()
+    local_pool = queue.Queue()
+    current_size = 0
+
+    @contextlib.contextmanager
+    def pooled():
+        nonlocal current_size
```
diminator: Q: how is the state of `current_size` maintained?
Is it some weak-global var stored in `app.config['bigchain_pool']`?

bigchaindb/util.py
```diff
@@ -31,6 +34,53 @@ def start(self):
...
+                # We need to check again if we have slots available, since
+                # the situation might be different after acquiring the lock
+                if current_size < size:
+                    current_size += 1
+                    instance = builder()
+
+        # Watchout: current_size can be equal to size if the previous part of
+        # the function has been executed, that's why we need to check if the
+        # instance is None.
+        if instance is None and current_size == size:
```
r-marques: `current_size == size` this check seems redundant since `instance` will only be `None` if `current_size == size`

bigchaindb/web/server.py
r-marques: just out of curiosity what is the reason to pop this?

bigchaindb/util.py
```diff
@@ -31,6 +34,53 @@ def start(self):
...
+            instance = local_pool.get(timeout=timeout)
+
+        yield instance
```
diminator: Just to be clear: everything before `yield` is called on the `__enter__` and everything after `yield` id called on the `__exit__`? So upon yield, `instance` binds to the `as` target and the body is executed?

r-marques:
other then my comments :+1:

bigchaindb/util.py
vrde: good point! Removing :dancer:

bigchaindb/util.py
diminator: Shouldnt you retrieve `current_size` from the `local_pool` size

bigchaindb/util.py
```diff
@@ -31,6 +34,53 @@ def start(self):
...
+                if current_size < size:
+                    current_size += 1
+                    instance = builder()
+
+        # Watchout: current_size can be equal to size if the previous part of
+        # the function has been executed, that's why we need to check if the
+        # instance is None.
+        if instance is None and current_size == size:
+            instance = local_pool.get(timeout=timeout)
+
```
diminator: so `instance` is reused from the pool if it's full?

it looks like connections are added untill `size` is reached and then are reused. But connections aren't closed?

diminator:
Other than some of my Q's for a better understanding (maybe update the comments/docs a bit) :+1:

bigchaindb/__init__.py
vrde: > Could you briefly comment on why this change is necessary?
Sure, you can take a look to the changes I made [in the docs](https://github.com/bigchaindb/bigchaindb/pull/183/files?short_path=af33c3d#diff-af33c3d6099d055484e6f6471b741cac)

> Is there a reason for not parametrizing the host?
Yes, the `server` section is compatible to the [Gunicorn settings](http://docs.gunicorn.org/en/stable/settings.html), so I'm just using their stuff. I thought about splitting `host` and `port` but this would require to add code. In this case I preferred to keep things simple and compatible with Gunicorn. This information is in the `bigchaindb/__init__.py` file but not in the docs, I'm adding it now!

bigchaindb/web/server.py
vrde: > using the EAFP approach:

Usually I like the EAFP approach, but in this case I just used the `pop` method with a default value to avoid the exception.


> just out of curiosity what is the reason to pop this?

It's to clean up the dictionary before passing it to `StandaloneApplication`, but I can remove it since `load_config` takes care of ignoring keys that are not expected.

bigchaindb/__init__.py
sbellem: @vrde  Ok, thanks.

bigchaindb/web/server.py
sbellem: ok

bigchaindb/util.py
vrde: sure

bigchaindb/util.py
vrde: > Q: how is the state of current_size maintained?

In the context of the enclosing function.

vrde: > Shouldnt you retrieve current_size from the local_pool size

No, because `local_pool.size()` is just the number of instances available in the pool. We need to keep reference of the total number of instances created (that will converge to `size` after some time). I think [this test](https://github.com/bigchaindb/bigchaindb/pull/183/files#diff-5832adf374920d75d4bf48e546367f53R40) should help understanding the logic.

bigchaindb/util.py
vrde: > Just to be clear: everything before yield is called on the `__enter__` and everything after `yield` id called on the `__exit__`? So upon yield, `instance` binds to the `as` target and the body is executed?

yup, isn't it beautiful? :nail_care:

bigchaindb/util.py
vrde: > it looks like connections are added untill size is reached and then are reused. But connections aren't closed?

yes, that's the point of having a pool. Opening a connection to the DB every time the Server gets a request is really intense for the system.

PR #186, opened 2016-04-18
Added CLI command to export public key

### LINKED ISSUE #179: Add command-line tools to export & import public keys
The idea is to add a `keys` subcommand to the `bigchaindb` command. For starters, it would have two capabilities:

To export a node's public key to `stdout`, the idea is to have the command:
```text
bigchaindb keys --export-my-pubkey
```

To import another node's public key into a node's keyring (list of public keys of all _other_ nodes in the federation), the idea is to have the command:
```text
bigchaindb keys --import-to-keyring KEY
```

where `KEY` is the public key to add to the keyring, e.g. `nJq6EmdUkvFjQRB5hFvDmvZtv1deb3W3RgmiAq6dyygC`

ttmc:
The main thing that this pull request does is add a new command to the BigchainDB command-line interface (CLI):
```text
bigchaindb export-my-pubkey
```

That command prints the node's public key to standard output (or exits and gives an error message on stderr if nothing set the node's public key).

This PR originally included another new command to import a public key to a node's keychain, but then I realized that our approach to managing the keyring(s) needs to be rethought. I'll write more about that in a new issue. UPDATE: See issue #187 

Also in this PR:
* More thorough documentation for the `bigchaindb` command.
* Lots of minor changes to code docstrings and syntax (i.e. not changing functionality, just helping the next person reading that code for the first time)

Resolves #179

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -13,10 +13,7 @@ import json
 import bigchaindb
 import bigchaindb.config_utils
 from bigchaindb import db
-from bigchaindb.exceptions import (
-    DatabaseAlreadyExists,
-    KeypairNotFoundException
-)
+from bigchaindb.exceptions import DatabaseAlreadyExists, KeypairNotFoundException
 from bigchaindb.commands.utils import base_parser, start
 from bigchaindb.processes import Processes
 from bigchaindb import crypto
@@ -43,8 +40,7 @@ def run_configure(args, skip_if_exists=False):
     """Run a script to configure the current node.
 
     Args:
-        skip_if_exists (bool): skip the function if a config file already
-                               exists
+        skip_if_exists (bool): skip the function if a config file already exists
     """
     config_path = args.config or bigchaindb.config_utils.CONFIG_DEFAULT_PATH
     config_file_exists = os.path.exists(config_path)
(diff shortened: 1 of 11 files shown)


bigchaindb/commands/bigchain.py
```diff
@@ -11,7 +13,10 @@
 import bigchaindb
 import bigchaindb.config_utils
 from bigchaindb import db
-from bigchaindb.exceptions import DatabaseAlreadyExists, KeypairNotFoundException
+from bigchaindb.exceptions import (
```
r-marques: We should decide the format to handle multiple imports and add it to the style guide.
 I personally prefer
`from bigchaindb.exceptions import DatabaseAlreadyExists, KeypairNotFoundException`
since its less verbose and easier to read

bigchaindb/commands/bigchain.py
```diff
@@ -38,7 +43,8 @@ def run_configure(args, skip_if_exists=False):
     """Run a script to configure the current node.
 
     Args:
-        skip_if_exists (bool): skip the function if a conf file already exists
```
r-marques: we agreed in using 120 ch per line

bigchaindb/config_utils.py
```diff
@@ -22,6 +22,7 @@
 import bigchaindb
 from bigchaindb.consensus import AbstractConsensusRules
 
+# logging.basicConfig(level=logging.DEBUG)
```
r-marques: This should be removed if not used

PR #188, opened 2016-04-18
Revised the Cryptography section of the docs

ttmc:
@diminator I added some more information and links to the section of the documentation on **Signature Algorithm and Keys**.

I also removed the last paragraph because it's too technical for someone wanting an overview, but not technical enough for someone wanting full details (and someone wanting full details can read the source code).


diminator:
Maybe you can add something like:
All keys are represented with the base58 encoding by default.

ttmc:
@diminator I added that sentence. Do you want to merge this PR?

diminator:
merging

PR #208, opened 2016-04-21
Feat/204/print version number

### LINKED ISSUE #70: Unified place to set the version number
Right now, the BigchainDB version number is stored in at least two places:

1. /setup.py inside the setup(), there is e.g. `version='0.1.3'`
2. /docs/source/conf.py in the variables named version (e.g. "0.1") and release (e.g. "0.1.3-rc4").

Maybe there will be more places in the future.

It would be nice if there was one central place (file?) where the long version name could be stored, and then a script could read that central place to modify conf.py and setup.py accordingly, so the version number stays in sync across all files.

### LINKED ISSUE #204: Add a command line argument to print the version with `bigchaindb` cli
Add a command line argument to print the version

```bash
$ bigchaindb --version
bigchaindb 0.1.5
$ bigchaindb -v
bigchaindb 0.1.5
```
Maybe this issue is a good place to resolve #70

r-marques:
This adds a cli flag to print the version number. Resolves #204 
This is a proposal to have a unified place so set the version number which resolves #70 

Looked at https://python-packaging-user-guide.readthedocs.org/en/latest/single_source_version/ which contains several solutions

Also looked at how rethinkdb https://github.com/rethinkdb/rethinkdb/blob/next/drivers/python/setup.py and requests do it https://github.com/kennethreitz/requests/blob/master/setup.py

After trying several approaches I think the best one is
```python
version = {}
with open("...sample/version.py") as fp:
    exec(fp.read(), version)
```

With this approach we parsing a source file with regex which and it does not require for the package (in this case `bigchaindb`) to be installed.


diminator:
really good:
- a single source of truth
- no circular imports
- no regex imposed
- no self-install for installing docs
- checked with setup, and CLI (`bigchaindb` and `bigchaindb-benchmark`)
- docs should also be fine

ttmc:
:+1:

sbellem:
:+1:

bigchaindb/version.py
```diff
@@ -0,0 +1,2 @@
+__version__ = '0.1.5'
+__short_version__ = '0.1'
```
vrde: PS: if you want to be more DRY:

```python
__short_version__ = '.'.join(__version__.split('.')[:2])
```
or

```python
__short_version__ = __version__[:__version__.rfind('.')]
```

vrde:
:+1:

diminator:
@vrde @r-marques lets merge!

PR #212, opened 2016-04-22
update to cryptoconditions 0.1.6

diminator:
removes the need of the cryptography dependency

can potentialy drop the `libffi-dev` dependency and more?


r-marques:
seems good :+1: 
once the checks pass we can merge in master

PR #219, opened 2016-04-25
Small fix to the documentation

r-marques:
`get_tx_by_payload_hash` returns an empty list if there are no transactions instead of None


ttmc:
:+1:

Issue #193, opened 2016-04-20
Transactions from an invalid block should be put back into the backlog


r-marques:
When a block is marked invalid the transactions in the block need to be re-assigned and put back into the backlog.

The difficult part is to decide which node should do it. A simple solution would be to have a changefeed listening to updates on the `bigchain` table (the only updates permitted should be appending new votes) and everytime a new vote comes in we check if a majority decision was reached. If so and if the decision is invalid put the transactions back into the `backlog`. Since the primary key in the `backlog` table is the `id` of the transaction even if several nodes try to do this at the same time there will be no duplicated transactions in the `backlog`.

rhsimplex:
Ok, trying to get some understanding here.  It seems the functions involved from `core.py` is `write_vote`.
`write_vote` occasionally looks up a block by primary key, and appends a new `vote` to the `votes` field at that record.

So the changefeed will look something like 

```python
feed = r.table('bigchain').filter(r.row['votes']).changes().run(self.conn)
```
Which I suppose will go somewhere in `processes.py` (somewhere in the `map_bigchain` function)?  When a majority invalid vote happens (btw, how does bigchain know what a majority is...from the number of keys?) then all transactions are put back into `q_new_transactions` or inserted directly? I think inserted directly.  The question is where does the function which does that go...in `block.py`?

Do the votes currently trigger anything at all? It looks like the votes get written but are never used again, correct?

Finally, should an invalidated block be deleted once all of its transactions are liquidated?

r-marques:
> So the changefeed will look something like 
```python
feed = r.table('bigchain').filter(r.row['votes']).changes().run(self.conn)
```
Which I suppose will go somewhere in `processes.py` (somewhere in the `map_bigchain` function)?  

We can use the already existing changefeed `r.table('bigchain').changes().run(b.conn)` and just act on updates assuming that the only updates in bigchain table will be new votes.
You can put the `new_val` in a queue and then create a process that will read from the queue, check if the block was decided invalid and if so write the transactions back to the backlog using `write_transactions` in every transaction in the block.

PR #221, opened 2016-04-25
add election class, process starts

rhsimplex:
Addresses #193.


r-marques:
oh look a sunburst

rhsimplex:
@r-marques that's how you know this is a quality PR

r-marques:
hum not sure if that yellow and red should be there

bigchaindb/voter.py
```diff
@@ -190,3 +190,69 @@ def start(self):
...
+            next_block = self.q_block_new_vote.get()
+            n_voters = len(next_block['block']['voters'])
+
+            vote_list = [vote['vote']['is_block_valid'] for vote in next_block['votes']]
+
+            n_invalid_votes = vote_list.count(False)
+
+            if n_invalid_votes > int(n_voters/2):
+                self.q_invalid_blocks.put(next_block)
+            else:
```
r-marques: I guess we can remove this.

rhsimplex:
I don't know what it means. I ran the test suite locally and it seemed ok...

bigchaindb/voter.py
```diff
@@ -190,3 +190,69 @@ def start(self):
...
+        """
+        Initialize the class with the needed queues.
+
+        Initialize a queue where blocks with new votes will be held
+        """
+        self.q_block_new_vote = q_block_new_vote
+        self.q_blocks_with_quorum = mp.Queue()
+        self.q_invalid_blocks = mp.Queue()
+
+    def check_for_quorum(self):
```
r-marques: I would move most of the logic in this method to `bigchaindb.core.Bigchain`.
It would be handy to have a method in which you give it a block (any block) and it returns either `valid`, `invalid`, or undecided.
We will need to use this functionality a lot.

bigchaindb/voter.py
```diff
@@ -190,3 +190,69 @@ def start(self):
...
+            b = Bigchain()
+            # FIXME: this is unsafe and could lose transactions
+            #   Since queue items are removed immediately, there is no guarantee
+            #   all transactions will be rewritten. Imagine if node 5/5 casts a
+            #   deciding invalid vote.  This block will never show up in the
+            #   "update" changefeed again.  Suppose the below loop is writing
+            #   transactions and the worker crashes.  Those un-written transactions
+            #   will never be reviewed again.
+            #
+            #   Ideally, queue item should be removed on completion, a la celery or
```
r-marques: This would be possible to do but we would need a persistent queue instead of an in memory queue.

On the other hand remember that this is a distributed system and all nodes will be doing this for the same block so that would only be a problem if the all nodes crashed

bigchaindb/voter.py
```diff
@@ -190,3 +190,69 @@ def start(self):
...
+        """
+        while True:
+            next_block = self.q_block_new_vote.get()
+            n_voters = len(next_block['block']['voters'])
+
+            vote_list = [vote['vote']['is_block_valid'] for vote in next_block['votes']]
+
+            n_invalid_votes = vote_list.count(False)
+
+            if n_invalid_votes > int(n_voters/2):
```
r-marques: We need to be careful with a split vote situation. A block is valid if the majority `n/2 + 1` says its valid and we may have a even number of nodes in the federation.

Image the case in which the federation has 10 nodes, 5 vote valid and 5 vote invalid. We have a split vote situation.
To be on the safe side in such a situation we should consider the block invalid, reassign the transactions and let the system try again.

r-marques:
The code looks really nice. Just some remarks and some edge cases that need to be addressed.
And then we need some tests. You can check the `block.py` or `voter.py` unit tests

bigchaindb/voter.py
r-marques: Another important thing that needs to happen here is that it is not enough to count votes. We also need to validate the signature in the vote.

tests/db/test_voter.py
```diff
@@ -324,9 +324,40 @@ def test_quorum(self, b):
...
+
+        # construct invalid block
+        test_block_2['block']['votes'] = [invalid_vote_2, invalid_vote_2, valid_vote_2, valid_vote_2]
+        q_block_new_vote.put(test_block_2)
+
+        election.start()
+        time.sleep(1)
+        election.kill()
 
+        # TEST GOES HERE??
```
rhsimplex: Not sure how to test if a particular transaction made it back into the backlog. If someone can show me, then I can finish this test.

bigchaindb/core.py
```diff
@@ -485,3 +485,25 @@ def get_unvoted_blocks(self):
...
+        vote_list = [vote['vote']['is_block_valid']
+                     for vote in block['block']['votes']
+                     if self.consensus.verify_vote_signature(vote)]
+
+        # validate votes here
+
+        n_invalid_votes = vote_list.count(False)
+        n_valid_votes = vote_list.count(True)
+
+        if n_invalid_votes >= int(n_voters/2):
```
vrde: In Python 3 you can use the `//` operator:
```python
3 // 2 == int(3 / 2)
```

bigchaindb/core.py
```diff
@@ -485,3 +485,25 @@ def get_unvoted_blocks(self):
...
+            return 'invalid'
```
vrde: Should we export `invalid`, `valid`, and `undecided` as constants? It'd be cleaner.

bigchaindb/core.py
rhsimplex: oh that is neat

vrde: Also, `int` is equivalent to `math.floor` for positive numbers.

In this scenario:
 - `n_voters` is odd, let's say `5`
 - `vote_list` is `[True, True, False, False]`
 - `n_invalid_votes` is `2`
 - `int(n_voters/2)` is `2`

the function would return `invalid`, is that what we want?

An alternative approach is to add a new variable, called `quorum` (so it's more explicit), that is `math.ceil(n_votes / 2)` or `round(n_votes / 2)` and then branch against it.

bigchaindb/core.py
rhsimplex: I don't know! Is there an example somewhere else?

bigchaindb/core.py
```diff
@@ -485,3 +485,25 @@ def get_unvoted_blocks(self):
...
+        vote_list = [vote['vote']['is_block_valid']
+                     for vote in block['block']['votes']
+                     if self.consensus.verify_vote_signature(vote)]
+
+        # validate votes here
+
+        n_invalid_votes = vote_list.count(False)
+        n_valid_votes = vote_list.count(True)
+
+        if n_invalid_votes >= n_voters // 2:
```
vrde: Also, `int` is equivalent to `math.floor` for positive numbers.

In this scenario:
 - `n_voters` is odd, let's say `5`
 - `vote_list` is `[True, True, False, False]`
 - `n_invalid_votes` is `2`
 - `int(n_voters/2)` is `2`

the function would return `invalid`, is that what we want?

An alternative approach is to add a new variable, called `quorum` (so it's more explicit), that is `math.ceil(n_votes / 2)` or `round(n_votes / 2)` and then branch against it.

bigchaindb/util.py
```diff
@@ -501,6 +501,26 @@ def get_hash_data(transaction):
...
+    Args:
+        signed_vote (dict): a vote with the `signature` included.
+
+    Returns:
+        bool: True if the signature is correct, False otherwise.
+    """
+
+    data = signed_vote.copy()
+
+    signature = data.pop('signature')
```
r-marques: I guess there is not need for the `copy` or the `pop`.
`public_key.verify(serialize(signed_vote['vote']), signed_vote['signature'])` should be enough.

bigchaindb/util.py
```diff
@@ -501,6 +501,26 @@ def get_hash_data(transaction):
     return crypto.hash_data(serialize(tx))
 
 
+def verify_vote_signature(signed_vote):
```
r-marques: Verifying the signature is not enough. We also need to check if a signature was created by a valid voter, so we need to check if the `node_pubkey` is in the `voters`.

Maybe have another method called `validate` block that checks if the `node_pubkey` is valid and then check the signature.

bigchaindb/core.py
vrde: Usually constants are at a [module level](https://www.python.org/dev/peps/pep-0008/#constants) but in this case I think it's fine to keep them on a class level.

bigchaindb/util.py
rhsimplex: You mean `voters` as in the voter list that comes with the vote? Couldn't that be faked too?

bigchaindb/voter.py
```diff
@@ -196,3 +196,69 @@ def start(self):
...
+class Election(object):
+
+    def __init__(self, q_block_new_vote):
+        """
+        Initialize the class with the needed queues.
+
+        Initialize a queue where blocks with new votes will be held
+        """
+        self.q_block_new_vote = q_block_new_vote
+        self.q_blocks_with_quorum = mp.Queue()
```
r-marques: This queue does not seem to be used.

tests/db/test_voter.py
```diff
@@ -297,6 +297,120 @@ def test_updating_block_number_must_be_atomic(self):
...
+        test_block['block']['votes'] = [invalid_vote, invalid_vote]
+        assert b.block_election_status(test_block) == 'undecided'
+
+        test_block['block']['votes'] = [valid_vote, valid_vote, valid_vote]
+        assert b.block_election_status(test_block) == 'valid'
+
+        test_block['block']['votes'] = [invalid_vote, invalid_vote, invalid_vote]
+        assert b.block_election_status(test_block) == 'invalid'
+
+
```
r-marques: too many blank lines

tests/db/test_voter.py
```diff
@@ -297,6 +297,120 @@ def test_updating_block_number_must_be_atomic(self):
...
+        #
+        test_block['block']['votes'] = [invalid_vote, invalid_vote]
+        assert b.block_election_status(test_block) == 'invalid'
+
+        # test unanimously valid block with one improperly signed vote -- should still succeed
+        test_block['block']['votes'] = [valid_vote, valid_vote, valid_vote, improperly_signed_valid_vote]
+        assert b.block_election_status(test_block) == 'valid'
+
+        # test unanimously valid block with two improperly signed votes -- should fail to have quorum
+        test_block['block']['votes'] = [valid_vote, valid_vote,
```
r-marques: This is a situation in which quorum will never be reached and we cannot have that situation.

To be on the safe side we should define that vote is invalid if:
- the vote is correct and it voted the block invalid
- the vote is not correct (invalid signature, ...)

Better to just requeue the transactions and try again then having the block forever stuck in an undecided state

tests/db/test_voter.py
```diff
@@ -297,6 +297,120 @@ def test_updating_block_number_must_be_atomic(self):
...
+
+        # test split vote
+        test_block['block']['votes'] = [invalid_vote, invalid_vote, valid_vote, valid_vote]
+        assert b.block_election_status(test_block) == 'invalid'
+
+        # test undecided
+        test_block['block']['votes'] = [valid_vote, valid_vote]
+        assert b.block_election_status(test_block) == 'undecided'
+
+        # test partial quorum situations for odd numbers of voters
```
r-marques: since we are recreating a block here maybe we can put this rest into a new unittest just to keep them smaller and more readable. Also easier to pinpoint the failure if the test fails

bigchaindb/consensus.py
```diff
@@ -89,6 +89,18 @@ def verify_signature(signed_transaction):
...
+        """Verify a cast vote.
+
+        Args:
+            signed_vote (dict): signed vote to verify
+
+        Returns:
+            bool: True if the votes's required signature data is present
+                and correct, False otherwise.
+        """
+        raise NotImplementedError
```
diminator: I guess the abstractmethods don't require `NotImplementedError` when using `ABC`.

same goes for the other methods in this abstract class

bigchaindb/consensus.py
```diff
@@ -89,6 +89,18 @@ def verify_signature(signed_transaction):
         """
         raise NotImplementedError
 
+    @abstractmethod
```
diminator: maybe also include:

`@staticmethod` on top of `@abstractmethod`

bigchaindb/core.py
```diff
@@ -485,3 +486,25 @@ def get_unvoted_blocks(self):
...
+        vote_list = [vote['vote']['is_block_valid']
+                     for vote in block['block']['votes']
+                     if self.consensus.verify_vote_signature(vote)]
+
+        # validate votes here
+
+        n_invalid_votes = vote_list.count(False)
+        n_valid_votes = vote_list.count(True)
+
+        if n_invalid_votes >= math.ceil(n_voters / 2):
```
diminator: maybe a line of comment on why using the ceil/floor

bigchaindb/util.py
```diff
@@ -501,6 +501,26 @@ def get_hash_data(transaction):
...
+        signed_vote (dict): a vote with the `signature` included.
+
+    Returns:
+        bool: True if the signature is correct, False otherwise.
+    """
+
+    data = signed_vote.copy()
+
+    signature = data.pop('signature')
+    public_key_base58 = signed_vote['node_pubkey']
```
diminator: in the latest code we use `sk` and `vk`  (signing key/verifying key) as it stands closer to the crypto jargon

so `public_key_base58 => vk_base58`

tests/db/test_voter.py
rhsimplex: **Any invalid vote make a block invalid**, am I understanding that correctly?

bigchaindb/consensus.py
rhsimplex: This is an issue for the refactor then?

bigchaindb/consensus.py
rhsimplex: Why?

diminator: Pylint likes to have the signatures of inherited methods to be the same

bigchaindb/consensus.py
diminator: sure

bigchaindb/consensus.py
rhsimplex: I'm not sure I agree, but it should go in the refactor anyway.  It's probably out of the scope of this PR for me to change all the function signatures in `consensus.py`.

diminator: put in refactor: of course!

any reason why you would disagree on this?

Not specifying the `@staticmethod` decorator might trick child classes that the first argument is a `self` but that's not the case

rhsimplex: I don't disagree, I'm `undecided` =)

r-marques:
:+1:

bigchaindb/core.py
```diff
@@ -485,3 +487,30 @@ def get_unvoted_blocks(self):
             unvoted.pop(0)
 
         return unvoted
+
+    def block_election_status(self, block):
+        """
+        Tallies the votes on a block, and returns the status: valid, invalid, or undecided
```
vrde: Minor but important for consistency, docstrings should be *imperative*.

bigchaindb/core.py
```diff
@@ -485,3 +487,30 @@ def get_unvoted_blocks(self):
...
+        n_valid_votes = sum(vote_list)
+        n_invalid_votes = len(vote_list) - n_valid_votes
+
+        # The use of ceiling and floor is to account for the case of an
+        # even number of voters where half the voters have voted 'invalid'
+        # and half 'valid'. In this case, the block should be marked invalid
+        # to avoid a tie. In the case of an odd number of voters this is not
+        # relevant, since one side must be a majority.
+        if n_invalid_votes >= math.ceil(n_voters / 2):
+            return 'invalid'
```
vrde: would be nice to put `invalid`, `valid`, and `undecided` under `Bigchain.BLOCK_INVALID = 'invalid'`, `Bigchaindb.BLOCK_VALID = 'valid'', so it's easier for a developer to use the return value of the method. Also, IDEs and editors with code completion will make use of it.

tests/db/test_voter.py
```diff
@@ -297,6 +297,153 @@ def test_updating_block_number_must_be_atomic(self):
...
+                      for member in test_federation]
+
+        # fake "no" votes
+        invalid_vote = [member.vote(test_block, 'abc', False)
+                        for member in test_federation]
+
+        # fake "yes" votes with incorrect signatures
+        improperly_signed_valid_vote = [member.vote(test_block, 'abc', True) for
+                                        member in test_federation]
+        [vote['vote'].update(this_should_ruin_things='lol')
```
vrde: lol

ttmc: Maybe `lol` is an emoticon for a person with their hands up.

r-marques:
@rhsimplex maybe update the tests to use the class constants in the asserts. Other then that I think its ready to merge :+1:

Issue #199, opened 2016-04-20
Improve core functionality


r-marques:
## Abstract

Implement core functionality from whitepaper.
## Goals

This issue is meant to be a **todo** of all the issues related to core functionality that needs to be implemented for the core algorithm to match the description in the whitepaper and also tasks that to not add any features and are sometimes pushed back but that increase the robustness and are required for a quality production ready product.

### General

- [x] What to do if a tx gets copied from an undecided or decided_valid block to the backlog? - #220 (unassigned)
- [x] Set RethinkDB read mode to 'majority' to improve consistency - #158 (ryan)
- [x] Validate transaction before writing it to the backlog #109 (closed, won't do)
- [x] Set durability == `hard` for all writes #504 
- [x] Transfer tx validation must check to ensure all referenced outputs are in valid blocks - #526
- [x] Ensure no same-asset transactions get put in the same block - #527 (ryan)
- [x] Double spend shouldn't be included in a block - #592
- [x] Rewriting transactions from invalid block may lead to DoubleSpends - #774

### Voting

- ~~Skip/abort voting on a block if/when it is already decided: #190~~
- [x] Handling the case of votes pointing to different blocks as the previous block: #165 (ryan), PR #565 
### Algorithmic/WP
- [x] Transactions from invalid block should be put back in the backlog - #193  (ryan), PR #221

### Querying

- [x] All backend implementations of `get_transactions_by_asset_id()` should ignore invalid blocks: #332
- [x] All backend implementations of `get_transactions_by_metadata_id()` should ignore invalid blocks: #792
- [x] Queries to the `bigchain` table should ignore invalid blocks - issue #192 (ryan), PR #324

### Mutability

The wire protocol firewall (#594) will prevent deletes & updates on the `bigchain` & `votes` tables. If a node deletes a transaction from the `backlog` table, then the client can just resend it, so such deletions don't have to be reverted. As for updates to the `backlog` table, those must be allowed, for example, to reassign stale transactions (PR #359).
 
- [x] Voting as separate table: #368 (ryan)
- ~~Prevent deletes on the `bigchain` table - #197 (ryan), PR #330~~
- ~~Handle deletes on the `backlog` table - #196 (ryan), PR #337~~
- ~~Handle updates on the `bigchain` table - #195 (unassigned)~~
- ~~Prevent updates on the `backlog` table - #194 (unassigned)~~
[...]

PR #224, opened 2016-04-26
Refactor Docker setup

vrde:
This PR heavily refactors the setup for Docker (might break the monitoring, but I'm confident we can have it running in the next few days).

The new `Dockerfile`, in order to **embed** the RethinkDB process, extends the [`rethinkdb:2.3`](https://hub.docker.com/_/rethinkdb/) image on Docker hub adding the BigchainDB code. This allow us to drop the `docker-compose` configuration and have a single image easy to deploy.

To simplify the integration with Docker, some new options have been added. The major one is the possibility to start the RethinkDB process before starting BigchainDB. Moreover, the env variables set during the `configure` command are used to create the configuration file that will be persisted to the `$HOME/bigchaindb_docker` directory.

For simplicity reasons the `bigchaindb-benchmark` CLI has been merged with `bigchaindb`.


bigchaindb/commands/bigchain.py
```diff
@@ -122,8 +146,12 @@ def run_drop(args):
 
 def run_start(args):
     """Start the processes to run the node"""
-    # run_configure(args, skip_if_exists=True)
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
+
+    if args.start_rethinkdb:
```
r-marques: should we check if rethinkdb is already running?

docs/source/installing-server.md
```diff
@@ -122,41 +122,41 @@ $ git clone git@github.com:bigchaindb/bigchaindb.git
...
+You can see BigchainDB processing your transactions running:
+```text
+$ docker logs -f $BIGCHAIN_CID
+```
+
+
 If you're on Linux, you can probably view the RethinkDB dashboard at:
 
-[http://localhost:58080/](http://localhost:58080/)
+[http://localhost:8080/](http://localhost:8080/)
```
r-marques: This will break if rethinkdb is running on the host. It is a good idea to expose in the host ports that are not the default.

docs/source/installing-server.md
```diff
@@ -122,41 +122,41 @@ $ git clone git@github.com:bigchaindb/bigchaindb.git
...
-$ docker-compose up -d
+$ BIGCHAIN_CID=$(docker run -d -v "$HOME/bigchaindb_docker:/data" local-bigchaindb)
 ```
 
-then you can load test transactions via:
+then you can load test transactions (using the `-m` option to enable using all
+the cores in the host machine) via:
 ```text
-$ docker-compose run --rm bigchaindb bigchaindb-benchmark load
+$ docker exec -d $BIGCHAIN_CID load -m
```
r-marques: I don't understand. If we are running the loader in a running container why do we bind the sever to '0.0.0.0'

docs/source/installing-server.md
```diff
@@ -122,41 +122,41 @@ $ git clone git@github.com:bigchaindb/bigchaindb.git
...
 ```
 
-As shown above, make sure that you set the database and statsd hosts to their
-corresponding service names (`rethinkdb`, `statsd`), defined in`docker-compose.yml`
-and `docker-compose-monitor.yml`.
-
 You can then start it up (in the background, as a daemon) using:
 ```text
-$ docker-compose up -d
+$ BIGCHAIN_CID=$(docker run -d -v "$HOME/bigchaindb_docker:/data" local-bigchaindb)
```
r-marques: I found that you can just name your container https://docs.docker.com/compose/compose-file/#container-name
This removes the need for an env

bigchaindb/commands/bigchain.py
vrde: no need, `start_rethinkdb` will fail because [this](https://github.com/bigchaindb/bigchaindb/pull/224/files#diff-8f049242225a1f66c973d24911b5179aR22) won't happen.

docs/source/installing-server.md
vrde: looks nice but I think it's only for `docker-compose`, will dig more later to see if the option is on `docker` itself.

docs/source/installing-server.md
vrde: Good point, anyway it's just something we can fix using `docker` with `-P` or `-p 58080:8080`. I'll add more stuff to the doc to make it clearer.

docs/source/installing-server.md
vrde: To make it accessible from the host :)

bigchaindb/commands/utils.py
```diff
@@ -20,6 +20,10 @@ def start_rethinkdb():
 
     for line in proc.stdout:
         if line.startswith('Server ready'):
+            # FIXME: seems like tables are not ready when the server is ready,
```
r-marques: https://rethinkdb.com/api/python/wait/

`database.wait` with `all_replicas_ready` should be enough but you can also run the wait on the tables just to be sure

vrde:
> database.wait with all_replicas_ready should be enough but you can also run the wait on the tables just to be sure

Yay, that's super good! I can remove the **experimental** label now :stuck_out_tongue: 

(no, just kidding, will keep it for a while)

bigchaindb/commands/utils.py
vrde: I just tried it, unfortunately it doesn't work because the database does not exits at all.
It seems to be equivalent to waiting on a non-existing database:
```
r.db('foobar').wait().run(c)
...
ReqlOpFailedError: Database `foobar` does not exist in:
r.db('foobar').wait()
^^^^^^^^^^^^^^       
```

Even if it's not the best solution, I'll keep a `time.sleep` for now. The right place to address the problem is in the workers, and to make them more solid on connecting to the database, so they reconnect when the connection to the DB is lost. This should be solved by #199 basically.

docs/source/installing-server.md
vrde: OK, I just need to add `--name` when running a container, so I can reference it later. With this I can get rid of the env variable.

bigchaindb/commands/utils.py
vrde: OK, tried again, seems to work if the table has been created (and it's not yet ready) :+1:

bigchaindb/commands/bigchain.py
```diff
@@ -9,12 +9,20 @@
...
+
 
 import bigchaindb
 import bigchaindb.config_utils
+from bigchaindb.util import ProcessGroup
+from bigchaindb.client import temp_client
 from bigchaindb import db
-from bigchaindb.exceptions import DatabaseAlreadyExists, KeypairNotFoundException
-from bigchaindb.commands.utils import base_parser, start
+from bigchaindb.exceptions import (StartupError,
```
r-marques: This is something that I saw in another pull request - how to do multi line imports.
Now its a good time to agree on a best practice and add it to the python style guide.

Another solution is:
```python
from bigchaindb.exceptions import StartupError, DatabaseAlreadyExists
from bigchaindb.exceptions import KeyPairNotFoundException
```
This approach seems simpler to read but if there is code that is executed at a module level, it will be executed multiple times.

bigchaindb/commands/bigchain.py
```diff
@@ -43,7 +59,13 @@ def run_configure(args, skip_if_exists=False):
         skip_if_exists (bool): skip the function if a config file already exists
     """
     config_path = args.config or bigchaindb.config_utils.CONFIG_DEFAULT_PATH
-    config_file_exists = os.path.exists(config_path)
+    bigchaindb.config_utils.autoconfigure(filename=False, force=True)
+
+
```
r-marques: to many blank lines

bigchaindb/commands/bigchain.py
```diff
@@ -43,7 +59,13 @@ def run_configure(args, skip_if_exists=False):
...
+    config_file_exists = False
+    # if the config path is `-` then it's stdout
```
r-marques: can you explain this `config_path` thing

bigchaindb/commands/bigchain.py
```diff
@@ -137,10 +170,47 @@ def run_start(args):
     processes.start()
 
 
+def _run_load(tx_left, stats):
+    logstats.thread.start(stats)
+    client = temp_client()
+    # b = bigchaindb.Bigchain()
```
r-marques: can we maybe remove this commented line?

bigchaindb/commands/utils.py
```diff
@@ -4,10 +4,59 @@
...
+                raise StartupError('Error waiting for the database `{}` '
+                                   'to be ready'.format(dbname)) from exc
+
+            return proc
+
+    # We are here when we exhaust the stdout of the process.
+    # The last `line` contains info about the error.
+    raise StartupError(line)
+
+    proc.kill()
```
r-marques: this line is never reached

bigchaindb/commands/bigchain.py
r-marques: nevermind. found the reason in the help of the command

bigchaindb/config_utils.py
```diff
@@ -91,7 +91,11 @@ def file_config(filename=None):
               file at CONFIG_DEFAULT_PATH, if filename == None)
     """
     logger.debug('On entry into file_config(), filename = {}'.format(filename))
-    if not filename:
+
```
r-marques: Can you explain the code below?
It seems that `filename` should only be a file name or `None`.
Why do we check if it is false?

docs/source/installing-server.md
```diff
@@ -111,52 +111,116 @@ If it's the first time you've run `bigchaindb start`, then it creates the databa
...
+In a terminal shell, pull the latest version of the BigchainDB Docker image
 ```text
-$ docker-compose build
+docker pull bigchaindb/bigchaindb:latest
 ```
 
-then do a one-time configuration step to create the config file; it will be
-stored on your host machine under ` ~/.bigchaindb_docker/config`:
+then do a one-time configuration step to create the config file; we will use
+the `-y` option to accept all the default values. The configuration file will
```
r-marques: The command below does not use the `-y`

ttmc: I fixed that.

bigchaindb/commands/bigchain.py
vrde: Yup, we need to decide what's the best way to do it. I personally like using parenthesis, but it's up to the team to decide.

> This approach seems simpler to read but if there is code that is executed at a module level, it will be executed multiple times.

It would execute just once, on the first import.

`mod.py`:
```python
greeting = 'hello'
animal = 'cat'
print(greeting)
```

```
>>> from mod import greeting
hello
>>> from mod import animal
>>> 
```

bigchaindb/commands/utils.py
vrde: Oh right, I totally ignored my `[unreachable] Unreachable code`.

bigchaindb/config_utils.py
vrde: Just reviewed my code, that part was smelling. I removed that thing. It was to support `autoconfigure` but without loading the default file configuration.

r-marques:
:+1:

PR #225, opened 2016-04-26
run the loader inside a running container

### LINKED ISSUE #213: docker problem when running the bigchaindb-benchmark
when running `docker-compose run --rm bigchaindb bigchaindb-benchmark load` it fails with:
```bash
  File "/usr/local/lib/python3.5/multiprocessing/process.py", line 254, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.5/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/src/app/bigchaindb/commands/bigchain_benchmark.py", line 25, in _run_load
    tx = client.create()
  File "/usr/src/app/bigchaindb/client.py", line 67, in create
    return self._push(signed_tx)
  File "/usr/src/app/bigchaindb/client.py", line 102, in _push
    res = requests.post(self.api_endpoint + '/transactions/', json=tx)
  File "/usr/local/lib/python3.5/site-packages/requests/api.py", line 107, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/requests/api.py", line 53, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.5/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/requests/adapters.py", line 437, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=9984): Max retries exceeded with url: /api/v1/transactions/ (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x7f85bb181128>: Failed to establish a new connection: [Errno 111] Connection refused',))
```

The `Max retries exceeded with url` can be misleading and there are several issues in the requests library repo about this but it usually means that the remote server for some reason is not accepting connections.

If I access the bash of the bigchaindb container and run `bigchaindb-benchmark load` from there it works fine.

The only change in the code that I can think of that broke the docker deployment was adding gunicorn.

r-marques:
This resolves #213 
Its not a perfect solution but it works and will allow us to do a new release.
A better solution is on #224 but it still needs some work

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ rethinkdb-data:
 
 bigchaindb:
   build: .
+  container_name: docker-bigchaindb
   volumes:
     - ./bigchaindb:/usr/src/app/bigchaindb
     - ./tests:/usr/src/app/tests
--- a/docs/source/installing-server.md
+++ b/docs/source/installing-server.md
@@ -135,7 +135,7 @@ API Server bind? (default `localhost:9984`):
 Database host? (default `localhost`): rethinkdb
 Database port? (default `28015`): 
 Database name? (default `bigchain`): 
-Statsd host? (default `localhost`): statsd
+Statsd host? (default `localhost`):
 Statsd port? (default `8125`): 
 Statsd rate? (default `0.01`): 
 Ready to go!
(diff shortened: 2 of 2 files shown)


diminator:
:+1: 
:100: / :100:

PR #226, opened 2016-04-26
Updated version number and changelog

r-marques:


bigchaindb/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.1.5'
+__version__ = '0.2.0'
 __short_version__ = '0.1'
```
diminator: maybe bump this too?

diminator: or make it a fn of

Issue #187, opened 2016-04-18
Managing the "current keyring" in the federation


ttmc:
I wrote a command-line command `bigchaindb add-to-keyring KEY` to add another node's public key (`KEY`) to the node's keyring. (I was working on resolving issue #179.) Then I realized that there are problems with the current keyring model...

**What is the keyring?**

For a given node, the keyring a list of the public keys of all the other nodes in the federation. It determines who is in the federation (and not). Every node has its own keyring. It's one of the node's configuration settings.

The keyring is used in two situations:
1. When a node assigns a transaction to another node, it randomly picks one from its keyring.
2. When a node creates a new block, it includes a list of all nodes which can vote on that block. That list is its keyring.

**The problems**

Each node has its own keyring. Those keyrings might not agree (i.e. no consensus on the list of current federation members). The nodes should all get the current list of federation nodes from the same standard source that they can all agree upon.

Secondly, we may want to know more than the public keys of the current federation nodes. We may want additional information about them, such as the associated organization name.

**Proposal**

Store all **proposed lists of federation nodes** in JSON documents in a new `fednodes` table in the RethinkDB database. (Not all will be accepted, but they'll all be kept, hence the word "proposed.")

Data model for a proposed list of federation nodes ("prop_list"?)

```javascript
{
    "id": "<hash of the serialized proposal>",
    "proposal": {
        "prop_num": // <an integer>,
        "previous_id": "<id of previous proposed list>",
        "timestamp": "<proposal-creation timestamp>",
        "creator_pubkey": "<public key of node that created this proposal>",
        "nodes": [
            {"node_owner": "Organization Name 253",
             "public_key": "29tgo20IIUG857..."},
            {"node_owner": "Orgname A23",
             "public_key": "HUi93isohFFI56..."}
        ],
    },
    "signature": "<signature of the proposal>",
    "votes": // <a list of signed votes, something like what blocks have>
}
```

As with blocks, a proposed list can be `undecided`, `decided_valid`, or `decided_invalid`. It depends on the votes that the proposed list has gotten so far. (It can be determined by querying.) The rules are the same as for blocks?

The first proposed list (i.e. the genesis list) is always `decided_valid`.

The "current list of federation nodes" is the latest `decided_valid` list. (This can be determined by querying. `prop_num` is there to make this query a bit easier: you can start by looking at the proposed list with `prop_num` == "number of documents in `fednodes`" and then follow the links backwards until you get to a `decided_valid` proposed list.)

[...]

PR #227, opened 2016-04-26
Further AWS deployment improvements

ttmc:
This pull request adds a bunch of new stuff to the AWS deployment scripts:
* An easier way to make Fabric send the `bigchaindb init` command to only one node: use the `@hosts(hostname)` decorator.
* The `make_confiles.sh` script to generate a set of default BigchainDB config files. You have to do this before calling `./startup.sh` as explained in the updated documentation for AWS deployment.
* The `clusterize_confiles.py` script to make all the config files for a cluster consistent with each other and with the hostnames of the nodes in the cluster
* Updated `startup.sh` and `fabfile.py` to clusterize the set of config files and send a different config file to each node in the cluster

I also figured out a way to make sure the nodes are all ready to start getting commands from Fabric (via ssh), rather than waiting 45 seconds. I just create a local socket and then try to connect it to port 22 on all the nodes, and keep trying those connections until they all work.

Note: These scripts are just to help us get a test cluster up and running now, before we implement a better way of handling the list of public keys in the federation, like what's outlined in Issue #187.

### DIFF
--- a/deploy-cluster-aws/clusterize_confiles.py
+++ b/deploy-cluster-aws/clusterize_confiles.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+"""Given a directory full of default BigchainDB config files,
+transform them into config files for a cluster with proper
+keyrings, API endpoint values, etc.
+
+Note: This script assumes that there is a file named hostlist.py
+containing public_dns_names = a list of the public DNS names of
+all the hosts in the cluster.
+
+Usage:
+    python clusterize_confiles.py <dir> <number_of_files>
+"""
+
+from __future__ import unicode_literals
+import os
+import json
+import argparse
+
+from hostlist import public_dns_names
+
+
+# Parse the command-line arguments
+parser = argparse.ArgumentParser()
+parser.add_argument('dir',
+                    help='Directory containing the config files')
+parser.add_argument('number_of_files',
+                    help='Number of config files expected in dir',
+                    type=int)
+args = parser.parse_args()
+
+conf_dir = args.dir
+numfiles_expected = int(args.number_of_files)
+
+# Check if the number of files in conf_dir is what was expected
+conf_files = os.listdir(conf_dir)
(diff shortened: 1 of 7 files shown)


deploy-cluster-aws/clusterize_confiles.py
```diff
@@ -0,0 +1,72 @@
...
+        pubkeys.append(pubkey)
+
+# Rewrite each config file, one at a time
+for i, filename in enumerate(conf_files):
+    file_path = os.path.join(conf_dir, filename)
+    with open(file_path, 'r') as f:
+        conf_dict = json.load(f)
+        # The keyring is the list of *all* public keys
+        # minus the config file's own public key
+        keyring = list(pubkeys)
```
r-marques: smart

deploy-cluster-aws/fabfile.py
```diff
@@ -138,13 +133,26 @@ def configure_bigchaindb():
...
 
+# Send the specified configuration file to
+# the remote host and save it there in
+# ~/.bigchaindb
+# Use in conjunction with set_host()
+# No @parallel
+@task
+def send_confile(confile):
+    put('confiles/' + confile, 'tempfile')
+    sudo('mv tempfile ~/.bigchaindb')
```
r-marques: do we need sudo here?
I am afraid that the bigchaindb process may not have the right read/write permissions

r-marques:
some comments. other then that seems :+1:

deploy-cluster-aws/fabfile.py
ttmc: Good question. Originally, I tried to do something like `put('confiles/' + confile, '~/.bigchaindb')` but I was getting an error about not having sufficient permissions. So I just did it in two steps, using sudo to move the file into ~.

I was curious to see if sudo() was necessary, using run() instead. It works with run()!

I'll push the new code that uses run() instead.

Issue #228, opened 2016-04-27
Rogue peer can vote


perlinvoice:
Hello
I have a set up with two federation nodes 1 and 2 with key pairs pub1,pri1 and pub2,pri2
Both keyrings have been configured accordingly.

Assume this scenario: node 1 is down and node 3 is a rogue node with key pair pub1, pri3

Node 3 transactions are not voted by Node 1 and remain in the backlog, however Node 2 transactions are voted by (rogue) Node 3 and move from the backlog to the blockchain.

Is it the expected behavior? 

Thanks in advance

ttmc:
I don't fully understand your question because voting doesn't happen to transactions in the backlog table, it happens to blocks in the bigchain table.

That said, you do raise some valid concerns:

* What's stopping a rogue node from connecting to the RethinkDB database to do stuff? We still have some work to do to prevent that, such as firewalls and using the new RethinkDB permissions.
* When validating a transaction or block, nodes should check the public key of the node that signed it, to make sure that it's a "known" public key.

r-marques:
@perlinvoice thanks for noticing this.
No, this is not the intended behavior.

What is happening here is that when a transaction reaches a federation node that federation node randomly assigns the transaction to another federation node and stores it in the backlog. Each federation node is then listening to events on the backlog for transactions assigned to them in order to validate them, include them in a block and write the block to the bigchain table.

Since you configured node3 with the same public key as node2, node3 will start processing the transactions assigned to node2.

The problem is that node3 will not be able to create valid blocks because it does not have node2 private key to correctly sign the block.
Looking at the code it seems that we are not checking the signature when validating the block.

Can you confirm that after node3 takes over, node1 still votes the blocks created by node3 as valid?

This is a worst case scenario - a malicious node being able to join the cluster. We are still working on the correct deployment of the federation but in a production environment the rethinkdb cluster should be secure with TLS for intra cluster communication https://rethinkdb.com/docs/security/#using-tls and firewall restrictions on the machines running the cluster instances.

r-marques:
closing this. It should have been fixed in #233

PR #233, opened 2016-04-28
validate_block now checks if a federation node created the block.

### LINKED ISSUE #229: Block validation does not check for the signature in the block
Currently `validate_block` does not check if the signature in the block is correct. This would allow a malicious node that gains access to the cluster to create valid blocks without a valid private key.

relates to #228

r-marques:
resolves #229 
related to #228 

This pull request adds code to check if a block was created by a federation node.
To do so it checks if `block['block']['node_pubkey']` is a federation node and also checks if the signature is correct.

The validation seems to be split between `core` and `consensus`. I decided to add the code to the `consensus` module but during the planned refactor we should address this code split.

### DIFF
--- a/bigchaindb/consensus.py
+++ b/bigchaindb/consensus.py
@@ -167,7 +167,6 @@ class BaseConsensusRules(AbstractConsensusRules):
 
         return transaction
 
-    # TODO: Unsure if a bigchain parameter is really necessary here?
     @staticmethod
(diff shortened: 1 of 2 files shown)


rhsimplex:
Almost done here.  Not quite sure how to check if a particular transaction ended up back in the backlog

vrde:
:+1:

Issue #236, opened 2016-04-30
python setup.py install fail.


nexusz99:
I tried to install bigchaindb from source and i got an error like this.

Environment : OS X (10.11.04), python 3.5

`error: Setup script exited with error in BigchainDB setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers`

Fortunately, 'pip3 install bigchaindb' works good.

nexusz99:
PR #237 can solve this error

vrde:
Hi @nexusz99, unfortunately some Python distributions come with a very old version of `setuptools`.

Can you try to run:
```
$ pip install --upgrade setuptools
```
and tell me if it resolves the problem? It should install `setuptools-20.10.1`.

Related to #237, the `~` is to pin a [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) of gunicorn.

nexusz99:
My pip environment was " pip (1.5.6) setuptools (3.6)". 

And after run "pip install --upgrade setuptools", the problem solved!!

Thanks a lot. It was my fault. 

Thanks.

nexusz99:
How about check the version of setuptools before install?

vrde:
> How about check the version of setuptools before install?

That's an interesting idea, will think about it and update this issue soon.

ttmc:
Resolved by PR #276 and PR #279

PR #237, opened 2016-04-30
Fix install_requires option. The '~=' is wrong operation.

nexusz99:


vrde:
Hi @nexusz99, thanks for you PR.

The `~` operator is to pin gunicorn to a [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release).

An alternative approach is to use `gunicorn>=19.0,==19.*`, but I find the `~` operator much easier to read. Anyway, this problem is related to an old version of `setuptools`, so upgrading `setuptools` should fix the problem you reported.

Before changing `setup.py` I'd like to understand if it's a common problem or not.

Can you please tell me the installed versions for `setuptools` and `pip`? You can run `pip freeze --all | grep -E "^(setuptools|pip)"`.

nexusz99:
@vrde thanks!

I solve this problem by  your help

Relate to "https://github.com/bigchaindb/bigchaindb/issues/236"

Thanks

vrde:
OK, I'm closing this one. We can continue talking on #236.

Issue #246, opened 2016-05-04
Project Minky (Benchmarking)


ttmc:
## Abstract

This is a project to do some more benchmarking of **BigchainDB core** (i.e. servers only, not using the HTTP API). All servers will be in one AWS data center.
## Goals
- Use tests to find out where the current bottlenecks are in the current BigchainDB core
- Develop more tools as necessary, for deployment, test definition, analysis, etc.
- For each test, produce a document to describe the setup and outcomes
## Steps
- [x] AWS monitoring server deployment scripts & docs: PR #222 
- [x] Merge PR #254 (i.e. get the latest AWS deployment scripts into `master`)
- [x] Change AWS deployment so `bigchaindb start` doesn't happen by default: PR #257 
- [x] Add new `bigchaindb set-shards` command and use it to set the number of shards: PR #258 & PR #259 
- [x] Create a new subfolder to contain benchmarking stuff: benchmarking utilities and subfolders for different tests. See PR #261 
- [x] Make it possible to deploy servers with the same set of keypairs: PR #263 
- [x] Create a command to get some simple statistics on the blocks written to the bigchain table over time: PR #266 . Note: It seems this method of getting statistics (from the changefeed) isn't consistent with the RethinkDB web interface / graph.
- [x] Better support for New Relic Server Monitoring: PR #316 
- [ ] Do some tests and document them.

For each test:
- Create a test document, describing the setup, the input transactions, how to run the test (including how to set up measurement / monitoring)
- Analyze measurements and report them somewhere, maybe in a GitHub issue?

We will do more tests, and other kinds of tests, in other projects.

ttmc:
I changed the description of this issue, to bring it in line with our current benchmarking goals, which will focus on servers for now, i.e. not using the HTTP API.

ttmc:
I'm closing this project because it's very old and we'll create a new issue or project when necessary.

PR #254, opened 2016-05-06
Deploy clients on AWS

ttmc:
* Renamed `startup.sh` to `awsdeploy.sh`
* Modified that script and the fabfile to enable deployment of servers _or_ clients
* The "tag" is now auto-generated, rather than being something that's specified on the command line
* Updated the docs to explain how to use `awsdeploy.sh`

This is part of Project Minky: PR #246


vrde:
:+1:

PR #258, opened 2016-05-09
Add command to configure number of shards.

r-marques:
Changed aws deployment script to automatically set the number of shards.
Created tests.

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -13,6 +13,7 @@ import builtins
 
 import logstats
 
+import rethinkdb as r
 
 import bigchaindb
 import bigchaindb.config_utils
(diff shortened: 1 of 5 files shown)


r-marques:
@vrde can you take a look at this pr.

bigchaindb/commands/bigchain.py
```diff
@@ -243,6 +250,13 @@ def main():
     subparsers.add_parser('start',
                           help='Start BigchainDB')
 
+    # parser for configuring the number of shards
+    sharding_parser = subparsers.add_parser('sharding',
```
vrde: It's more consistent if we keep all the commands are *verbs*.

vrde:
Looks good to me.

PR #263, opened 2016-05-10
AWS server deployment with keypairs from a file of keypairs

ttmc:
* Added a new script `write_keypairs_file.py` to generate a static file named `keypairs.py` containing the specified number of keypairs
* Modified `clusterize_confiles.py`: added a new optional `-k` argument. If you include that argument, then all keypairs will be read from `keypairs.py`
* Added some documentation about how to use the new features


r-marques:
I think we should commit keypairs.py. If we are going to create datasets we don't want to lose those test keys

ttmc:
Okay I committed a 128-keypair `keypairs.py` file.

r-marques:
:+1:

PR #264, opened 2016-05-10
rename verify_signature to validate_fulfillments

diminator:
title says it all

reasoning:
- `verify_signature` was limited to signature conditions
- `cryptoconditions.Fulfillment` has a method `validate()`
- bigchain has a method `validate_transaction` which does additional checks

Also

- cleaned up abstract class of consensus
- updated some docstring types

### DIFF
--- a/bigchaindb/consensus.py
+++ b/bigchaindb/consensus.py
@@ -15,6 +15,7 @@ class AbstractConsensusRules(metaclass=ABCMeta):
     All methods listed below must be implemented.
     """
 
+    @staticmethod
     @abstractmethod
     def validate_transaction(bigchain, transaction):
         """Validate a transaction.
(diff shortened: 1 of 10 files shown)


bigchaindb/consensus.py
```diff
@@ -31,8 +32,8 @@ def validate_transaction(bigchain, transaction):
             Descriptive exceptions indicating the reason the transaction failed.
             See the `exceptions` module for bigchain-native error classes.
         """
-        raise NotImplementedError
```
r-marques: Why do we remove this. Isn't this to make sure that you implement all methods when you sublclass this?

r-marques:
:+1:

bigchaindb/consensus.py
diminator: Using ABC classes you dont need to
They return a

```
TypeError: Can't instantiate abstract class B with abstract methods f
```

if method `f` wasnt concrete

PR #266, opened 2016-05-11
Added a command to gather metrics from a cluster

r-marques:


benchmarking-tests/benchmark_utils.py
```diff
@@ -36,6 +44,38 @@ def run_set_statsd_host(args):
...
+    logger.info('Starting gathering metrics. {} transasctions in the backlog'.format(num_transactions))
+
+    # listen to the changefeed
+    try:
+        for change in r.table('bigchain').changes().run(conn):
+            # check only for new blocks
+            if change['old_val'] is None:
+                block_num_transactions = len(change['new_val']['block']['transactions'])
+                time_now = time.time()
+                logger.info('{} {}'.format(time_now, block_num_transactions))
```
vrde: Not crucial here, but as a side note is better to use:
```python
logger.info('%s %s', time_now, block_num_transactions)
```
In this way you postpone the *formatting* to the logger function. Inside the logger function there is some logic to decide if showing the log or not, if the answer is yes then the string is constructed, otherwise no, saving time.

(I'll note it down and I can review the BigchainDB code later, since I'm not sure if we are doing the right thing there, and it's much more important than here)

vrde:
:+1:

benchmarking-tests/benchmark_utils.py
```diff
@@ -36,6 +44,38 @@ def run_set_statsd_host(args):
...
+                csv_writer.writerow([str(time_now), str(block_num_transactions)])
+
+                num_transactions -= block_num_transactions
+                if num_transactions == 0:
```
ttmc: There's an implicit assumption here that all the transactions are valid, so each one will never-again be seen in the backlog for inclusion in a future block. A more general check would be to query the backlog to check how many transactions are in it (i.e. `r.table('backlog').count().run(conn)`)... but that adds some querying overhead.

r-marques: True, this needs to be improved. That's is why I added the option to stop gathering metrics with Ctrl-C

benchmarking-tests/benchmark_utils.py
ttmc: @vrde Maybe something should be added to [our Python Style Guide](https://github.com/bigchaindb/bigchaindb/blob/master/PYTHON_STYLE_GUIDE.md#using-the--operator-or-format-to-format-strings)? Right now, it says to use format() always.

r-marques:
regarding the logging I will leave format for now to make it easier to format the output since now I am using tabular data to print statistics

benchmarking-tests/benchmark_utils.py
```diff
@@ -63,11 +71,22 @@ def run_gather_metrics(args):
...
 
-                num_transactions -= block_num_transactions
-                if num_transactions == 0:
+                # log statistics
+                if initial_time is None:
+                    initial_time = time_now
+
+                num_transactions_received += block_num_transactions
+                elapsed_time = time_now - initial_time
+                elapsed_time = elapsed_time if elapsed_time != 0 else 1
```
ttmc: This is going to make the first `transactions_per_second` equal to the number of transactions in the first block, which is misleading. What to do instead? Maybe arrange things so that the first `transactions_per_second` is `float('nan')` (i.e. Not a Number)? http://stackoverflow.com/questions/19374254/assigning-a-variable-nan-in-python-without-numpy

benchmarking-tests/benchmark_utils.py
```diff
@@ -36,6 +44,57 @@ def run_set_statsd_host(args):
...
+                logger.info('\t{:<20} {:<20} {:<20} {:<20}'.format(time_now, block_num_transactions,
+                                                                   transactions_per_second, percent_complete))
+
+                if (num_transactions - num_transactions_received) == 0:
+                    break
+    except KeyboardInterrupt:
+        logger.info('Interrupted. Exiting early...')
+
+    # close files
+    csv_file.close()
```
ttmc: Maybe the `csv_file.close()` should go inside [a `finally` clause](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) (clean-up actions that must be executed under all circumstances), i.e.
```python
    finally:
        csv_file.close()
```

ttmc:
:+1:

PR #267, opened 2016-05-11
Change the block processes to use GroupProcess

r-marques:
The block.py module uses several subprocesses to perform different kind of tasks. This pull request replaces multiprocessing.Process with bigchaindb.uitl.ProcessGroup. This way all of the subprocesses are able to use all the cores in the machine.

In some preliminary benchmark tests this speeds up the number of transactions per second by 5x.


vrde:
:+1:

PR #271, opened 2016-05-11
Add lru_cache to load_consensus_plugin

vrde:
Creds to @r-marques for finding the bottleneck. Here is a fix :dancer:


r-marques:
perfect :-1:

PR #273, opened 2016-05-11
More ways to add tx to backlog

ttmc:
This is a work in progress, but I wanted to share what I did so far:

* Added a `--payload-size` option to the `benchmark_utils.py add-backlog` command. It takes an argument which (currently) can be one of `minimal`,  `small`, `medium`, or `large`. The idea is to have an easy way to test how the size of the transactions affects performance.

Please don't merge yet.


ttmc:
I was going to make something more sophisticated than this, but it's okay for now and can probably be merged.

vrde:
:+1:

PR #276, opened 2016-05-12
Install docs: Suggest upgrading pip wheel & setuptools

ttmc:
The AWS deployment fabfile was already upgrading pip, wheel, and setuptools to their latest versions. I added a line to do the same in the BigchainDB Installation documentation.

I also edited the AWS deployment fabfile, changing `-U` to `--upgrade`; it [means the same thing](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-U) but it's easier to guess what it's doing if you're new to pip.


ttmc:
Related to PR #279

r-marques:
:+1:

Issue #274, opened 2016-05-12
I cannot install bigchaindb..


jindragon:
I installed many times with these commands.
-----------------------------------------------------------------------------------------------------------
source /etc/lsb-release && echo "deb http://download.rethinkdb.com/apt $DISTRIB_CODENAME main" | sudo tee /etc/apt/sources.list.d/rethinkdb.list

wget -qO- https://download.rethinkdb.com/apt/pubkey.gpg | sudo apt-key add -

sudo apt-get update

sudo apt-get install rethinkdb

sudo apt-get update

sudo apt-get install g++ python3-dev

sudo apt-get install python3-setuptools

sudo easy_install3 pip

sudo pip install bigchaindb

sudo pip install --upgrade bigchaindb
-------------------------------------------------------------------------------------------------------
But recently, I tried to install and there was an error message.
![image](https://cloud.githubusercontent.com/assets/16792324/15203138/b783aad4-183b-11e6-9626-8431b607e41e.png)

So i tried with
pip install bigchaindb==0.1.5
but also failed.

![image](https://cloud.githubusercontent.com/assets/16792324/15203166/f3e4733c-183b-11e6-9ccf-707ad09fd877.png)

How can I fix it?

r-marques:
BigchainDB v0.1.5 required libffi-dev so you would need to also do `apt-get install libffi-dev` before installing bigchaindb.

For the latest version it seems to be because the setuptools installed from the repos is a really old version. Can you try doing `pip install --upgrade pip wheel setuptools` before installing bigchaindb.

We updated the docs in this pull request #276

vrde:
@jindragon before updating `setuptools`, can you please run this in a terminal:
```bash
 $ python3 -c 'import setuptools; print(setuptools.__version__)'
```

vrde:
@jindragon would be amazing if you can try #279 before upgrading `setuptools`. The pull request checks if the version of `setuptools` you are using supports the `~` operator, and throws a human-friendly error if not.

jindragon:
I didn't installed setuptools..
But I installed BigchainDB successfully.
I will install setuptools, latest version.

vrde:
hey @jindragon, after applying @r-marques suggestions, are you able to install BigchainDB?

jindragon:
Yes! Thanks very much!

PR #279, opened 2016-05-12
Display useful message when setuptools is outdated

vrde:
This should help users trying to install BigchainDB using Pip and an old version of `setuptools`.

Related: #236 #274

On Ubuntu 14:04, the old error was:
```
root@139a61af2e8d:/bigchaindb# python3 setup.py install

Installed /tmp/easy_install-_nf2audp/pytest-runner-2.7.1/setuptools_scm-1.11.0-py3.4.egg
your setuptools is too old (<12)
setuptools_scm functionality is degraded
zip_safe flag not set; analyzing archive contents...

Installed /bigchaindb/pytest_runner-2.7.1-py3.4.egg
error in BigchainDB setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
```

while the new error is:
```
root@139a61af2e8d:/bigchaindb# python3 setup.py install
Your Python distribution comes with an incompatible version of `setuptools`. Please run:
 $ pip3 install --upgrade setuptools
and then run this command again
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,10 +6,27 @@ For full docs visit https://bigchaindb.readthedocs.org
 """
 from setuptools import setup, find_packages
 
+
 # get the version
 version = {}
 with open('bigchaindb/version.py') as fp:
     exec(fp.read(), version)
+
+
+# check if setuptools is up to date
+def check_setuptools_features():
+    import pkg_resources
+    try:
+        list(pkg_resources.parse_requirements('foo~=1.0'))
+    except pkg_resources.RequirementParseError:
+        exit('Your Python distribution comes with an incompatible version '
+             'of `setuptools`. Please run:\n'
+             ' $ pip install --update setuptools\n'
+             'and then run this command again')
+
+
+check_setuptools_features()
+
 
 tests_require = [
     'pytest',
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ def check_setuptools_features():
     import pkg_resources
     try:
         list(pkg_resources.parse_requirements('foo~=1.0'))
-    except pkg_resources.RequirementParseError:
+    except ValueError:
         exit('Your Python distribution comes with an incompatible version '
              'of `setuptools`. Please run:\n'
              ' $ pip install --update setuptools\n'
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -6,11 +6,28 @@
...
 with open('bigchaindb/version.py') as fp:
     exec(fp.read(), version)
 
+
+# check if setuptools is up to date
+def check_setuptools_features():
+    import pkg_resources
+    try:
+        list(pkg_resources.parse_requirements('foo~=1.0'))
+    except pkg_resources.RequirementParseError:
```
r-marques: This creates another problem because with older versions of setuptools `pkg_resources.RequirementsParseError` may not exists. I think this is the case with ubuntu 14.04 which from the repos installs setuptools v3.0.0

vrde: > This creates another problem because with older versions of setuptools pkg_resources.RequirementsParseError may not exists. I think this is the case with ubuntu 14.04 which from the repos installs setuptools v3.0.0

doing some more tesing now

setup.py
```diff
@@ -18,7 +18,7 @@ def check_setuptools_features():
     import pkg_resources
     try:
         list(pkg_resources.parse_requirements('foo~=1.0'))
-    except pkg_resources.RequirementParseError:
+    except ValueError:
         exit('Your Python distribution comes with an incompatible version '
              'of `setuptools`. Please run:\n'
              ' $ pip install --update setuptools\n'
```
r-marques: shouldn't it be `--upgrade`?

r-marques: and maybe its safer so use `pip3` instead of `pip` for users not using a virtualenv. even inside a python3 virtualenv there is a pip3

ttmc: It should be `--upgrade` but I'm not sure `pip3` is necessary. The BigchainDB installation documentation and the AWS deployment fabfile both use `pip` to do the upgrade and it doesn't seem to cause any problems. If we do decide to go with `pip3` then I can just change the documentation and AWS deployment fabfile.

r-marques:
:+1:

ttmc:
I will change the documentation and the AWS deployment fabfile to use `pip3` for the upgrade.

PR #281, opened 2016-05-12
Fix #260 transaction malleability

### LINKED ISSUE #260: `get_fulfillment_message` doesnt include conditions of transaction, only of input transaction
## Problem

The fulfillment message is the message that needs to be signed in order to seal the transaction with a signature.

Currently, for a `TRANSFER`, this looks like:

```
fulfillment_message = {
        'operation': transaction['transaction']['operation'],
        'timestamp': transaction['transaction']['timestamp'],
        'data': transaction['transaction']['data'],
        'version': transaction['version'],
        'id': transaction['id'],
        'condition': conditions[fulfillment['input']['cid']]
    }
```

A malicious node might now change the current condition of the transaction without going noticed

## Solution

- [ ] Test by creating a scenario where a malicious node changes the transaction condition, when he is assignee
- [ ] Include the condition of the current transaction (and not only the input transaction) to the message
- [ ] Optionally: refactor and document better

diminator:
solves #260

### DIFF
--- a/bigchaindb/util.py
+++ b/bigchaindb/util.py
@@ -276,7 +276,7 @@ def create_tx(current_owners, new_owners, inputs, operation, payload=None):
                 'new_owners': new_owners,
                 'condition': {
                     'details': json.loads(condition.serialize_json()),
-                    'uri': condition.condition.serialize_uri()
+                    'uri': condition.condition_uri
                 },
                 'cid': fulfillment['fid']
             })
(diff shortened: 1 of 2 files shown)


r-marques:
:+1:

bigchaindb/util.py
```diff
@@ -436,19 +439,22 @@ def validate_fulfillments(signed_transaction):
...
         is_valid = parsed_fulfillment.validate(serialize(fulfillment_message))
 
         # if transaction has an input (i.e. not a `CREATE` transaction)
-        if fulfillment['input']:
-            is_valid &= parsed_fulfillment.condition.serialize_uri() == \
-                fulfillment_message['condition']['condition']['uri']
+        # TODO: avoid instantiation, pass as argument!
+        bigchain = bigchaindb.Bigchain()
+        input_condition = get_input_condition(bigchain, fulfillment)
+        is_valid &= parsed_fulfillment.condition.serialize_uri() == \
```
vrde: dunno the API, but can you use `condition_uri` [like here](https://github.com/bigchaindb/bigchaindb/pull/281/files#diff-618d4617455ed1678f5bef7f9a492fd7R279) also in this context?

bigchaindb/util.py
```diff
@@ -436,19 +439,22 @@ def validate_fulfillments(signed_transaction):
...
         is_valid = parsed_fulfillment.validate(serialize(fulfillment_message))
 
         # if transaction has an input (i.e. not a `CREATE` transaction)
-        if fulfillment['input']:
-            is_valid &= parsed_fulfillment.condition.serialize_uri() == \
-                fulfillment_message['condition']['condition']['uri']
+        # TODO: avoid instantiation, pass as argument!
+        bigchain = bigchaindb.Bigchain()
+        input_condition = get_input_condition(bigchain, fulfillment)
+        is_valid &= parsed_fulfillment.condition_uri == \
```
vrde: I'd use the logical operator `and` instead of the bitwise operator `&`, so:
```python
is_valid = is_valid and parsed_fulfillment.condition_uri == input_condition['condition']['uri']
```
because we are doing a logical operation, and not working on a bit level.
AFAIK your solution with `&=` and my proposal with `and` are totally compatible, it's just to be a bit (pun intended) more precise :santa:

vrde:
:+1:

PR #308, opened 2016-05-13
Update ROADMAP: add table

trentmc:
The table will make it easy to quickly slice & dice to specific tickets, based on topic (row) & timing (column).


vrde:
Nice job with the labels, like the matrix.

PR #311, opened 2016-05-13
[Do not merge] Add schema validation

vrde:
This is a WIP with some ideas on how to do schema validation.
You can read more in the `README.md` in the `schemas/` directory.

The proposal will be improved with a more layered approach on transactions validation, e.g.:
 1. check if the *shape* (or structure) of the transaction is valid
 2. check if signatures are valid
 3. [and more]

If one of the steps fails, the TX is invalid.

### DIFF
--- a/schemas/README.md
+++ b/schemas/README.md
@@ -1,0 +1,15 @@
+# Introduction
+
+This directory contains the schemas for the different JSON documents BigchainDB uses.
+
+The aim is to provide:
+ - a strict definition/documentation of the data structures used in BigchainDB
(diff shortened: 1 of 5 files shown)


diminator:
@vrde, I think this PR is close to adaption

can you verify?

We could add some tests on the definition side, but thats also covered by the other tests ofcourse

diminator:
@vrde would be cool if we can refer to the uri schemas of interledger:
https://github.com/interledger/five-bells-condition/tree/master/schemas

PR #316, opened 2016-05-18
Better Support for New Relic Server Monitoring

ttmc:
* Tested the existing code for installing the New Relic agent on all AWS instances, and made some improvements. For example, the New Relic license key is now read from an environment variable named `NEWRELIC_KEY` and there's a helpful error message if that isn't set.
* Added documentation about what New Relic Server Monitoring is and how to use it to monitor a BigchainDB cluster on AWS.

### DIFF
--- a/deploy-cluster-aws/fabfile.py
+++ b/deploy-cluster-aws/fabfile.py
@@ -6,7 +6,7 @@ BigchainDB, including its storage backend (RethinkDB).
 from __future__ import with_statement, unicode_literals
 
 from os import environ  # a mapping (like a dict)
-from sys import exit as sys_exit
+import sys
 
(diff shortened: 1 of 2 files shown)


deploy-cluster-aws/fabfile.py
```diff
@@ -5,6 +5,9 @@
 
 from __future__ import with_statement, unicode_literals
 
+from os import environ  # a mapping (like a dict)
+from sys import exit as sys_exit
```
vrde: you can just do:
```python
import sys
```
and then in the code:
```python
sys.exit('...')
```

it's simpler

ttmc:
@vrde Thanks, I changed it to `import sys` and `sys.exit('...`

Issue #105, opened 2016-03-03
Consider using secondary indexes in some queries


r-marques:
Consider using secondary indexes for
- [x] `get_spent`
- [x] `get_owned_ids`
- [x] `get_transaction`

Issue #192, opened 2016-04-20
Queries to the bigchain table should ignore invalid blocks


r-marques:
When querying the `bigchain` table the queries should ignore invalid blocks.

related to #134 

TODO:
- [ ] Change read queries to ignore invalid blocks
- [ ] Implement a query to check if a block is valid or not

The queries that we do at this moment are `get_owned_ids`, `get_spent`, `get_transaction`, and `get_tx_by_payload_hash`.

- `get_owned_ids`
Returns the inputs that a given public key owns. We need the query to ignore invalid blocks.
The tricky part is if the block is still undecided. One solution would be to also return the inputs of blocks that are undecided (we can assume that most blocks will be valid). The worst case scenario for this solution is that if the block ends up being invalid, the transaction using the input will also be invalid.

- `get_spent`
Checks if a given txid was already spent. We need the query to ignore invalid blocks (if the input was spent in an invalid block then it was not spent).
Same as above. If the block is undecided we should set it as spent (assume most blocks will be valid)

- `get_transaction`
The query should ignore invalid blocks. Still return if the block is undecided.

- `get_tx_by_payload_hash`
Ignore invalid blocks

For this we also need the ability to query the database for a block and check whether it is undecided or what was the majority decision

rhsimplex:
It seems what we really need is a way to get the block given the transaction.  The current RethinkDB call in `get_transaction` looks like:

```python
    response = r.table('bigchain').concat_map(lambda doc: doc['block']['transactions'])\
        .filter(lambda transaction: transaction['id'] == txid).run(self.conn)
```

The blocks are abstracted away by `concat_map`. Is there an obvious way to preserve the hierarchy in the result (without returning the entire transaction block)? I'll look into it. Would be cool to do this in one DB query rather than two.

Once #221 is merged, we'll have the functionality to check election results. Then it should be easy to query the status of a block: `get_block` could be the analogue of 'get_transaction', and could include a threshold (defaulting to `undecided` or better).

r-marques:
I guess you could filter `bigchain` before the `concat_map`. You can basically run any code on the filter.

something like:
```python
r.table('bigchain').filter(lambda block: b.check_election_status(block) == 'valid').concat_map...
```

rhsimplex:
I tried that, and it doesn't work as-is. Not sure why, perhaps because `b.check_election_status` is an instance method.

ttmc:
[...]

PR #320, opened 2016-05-18
Use secondary index on get_transaction

vrde:
This PR partially solve #105.

The new query makes use of the secondary index `transaction_id`, and should make the look up from `O(n)` to `O(1)`.

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -145,8 +145,11 @@ class Bigchain(object):
             If no transaction with that `txid` was found it returns `None`
         """
 
-        response = r.table('bigchain').concat_map(lambda doc: doc['block']['transactions'])\
-            .filter(lambda transaction: transaction['id'] == txid).run(self.conn)
+        response = r.table('bigchain')\
+                    .get_all(txid, index='transaction_id')\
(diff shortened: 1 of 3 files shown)


r-marques:
:+1:

diminator:
:+1: Optimize all the things!

rhsimplex:
I'm currently working on #192 which duplicates and extends this functionality.  In fact most query methods benefit from a secondary index, and all require a check on block validity anyway.  This seems to require some restructuring to the query methods. I haven't prepared the PR yet but you can see the changes on branch https://github.com/bigchaindb/bigchaindb/tree/core/192/ignore-invalid-blocks .  So maybe we can wait on merging this PR until I'm done?

vrde:
> So maybe we can wait on merging this PR until I'm done?

Yes, it's better. Even if we don't merge my PR, I did some extra work profiling the old and the new query. The new one is much faster, and less CPU intense for RethinkDB (and this is for your queries in your PR as well). :clap:

vrde:
Will close this in favor of #192. There are also some other optimizations I have in mind, will discuss it in another issue.

PR #322, opened 2016-05-19
Use pip3 instead of pip to upgrade pip, wheel & setuptools

ttmc:
I tested this by deploying a 3-node cluster on AWS and it worked fine.


r-marques:
:+1:

PR #324, opened 2016-05-19
Ignore invalid blocks

rhsimplex:
Addresses #192 except for checking blocks on `get_tx_by_payload_hash`, because no tests exist for this query and I understand we're going to change it soon anyway.

May also duplicate #320.

All queries now have one extra query for block validity.  I haven't found a more compact/elegant way to do it.


bigchaindb/core.py
```diff
@@ -151,16 +150,74 @@ def get_transaction(self, txid):
...
+            # If the transaction is in a valid or any undecided block, return it. Does not check
+            # if transactions in undecided blocks are consistent, but selects the valid block before
+            # undecided ones
+            for _id in validity:
+                target_block_id = _id
+                if validity[_id] == 'valid':
+                    break
+
+            # Query the transaction in the target block and return
+            response = r.table('bigchain').filter(lambda block: block['id'] == target_block_id).nth(0)\
```
vrde: can we use just `get`? The `id` is the primary index, so it's fast and the code is more compact. Also, we can get rid of the `nth(0)` operator.

rhsimplex: What? No! If anything I need to chain more methods on there. jk, i have no idea what I was thinking with that one.

tests/db/test_bigchain_api.py
```diff
@@ -1124,6 +1180,42 @@ def test_get_spent_single_tx_single_output(self, b, user_sk, user_vk):
...
+
+        # create a transaction and block
+        tx = b.create_transaction(user_vk, user2_vk, owned_inputs_user1, 'TRANSFER')
+        tx_signed = b.sign_transaction(tx, user_sk)
+        block = b.create_block([tx_signed])
+        b.write_block(block, durability='hard')
+
+        # vote the block invalid
+        vote = b.vote(block, b.get_last_voted_block()['id'], False)
+        b.write_vote(block, vote, 2)
```
vrde: should the `block_number` be `3` instead of `2`? Should work anyway in the context of the test.

PR #325, opened 2016-05-19
Update ROADMAP.md and Links to it

ttmc:
The BigchainDB roadmap has been moved to https://github.com/bigchaindb/org/blob/master/ROADMAP.md so the ROADMAP.md file in this repo was edited to link there.

Deleted the roadmap.jpg file.

Some documentation links to the ROADMAP.md file were also updated.


vrde:
Should we add also the Waffle.io badge in `README.md`? We can add also a direct link to the waffle board in `ROADMAP.md` IMO

ttmc:
I'd rather not add the waffle.io badge for now. We did some tests earlier today and it didn't seem to be reporting the correct number.

PR #337, opened 2016-05-27
Core/196/handle backlog deletes

### LINKED ISSUE #196: Handle deletes on the backlog table
We need to create a changefeed that listens to deletes in the `backlog` table. Deletes in the `backlog` should only be allowed if the transactions were included in a block in the `bigchain` table.

The voter code right now deletes the transactions as it processes them which means that there will a some time gap between the time a transaction is deleted and the time the block makes it to the `bigchain`. I can think of two solutions for this:

1. When a transaction is deleted cache it and check the `bigchain` table for the transaction after a timeout period (e.g. 30 seconds). If after the timeout period the transaction is not in the `bigchain` re-insert the transaction in the backlog.
2. The node creating the block first waits for the block to be written to the `bigchain` and them deletes the transactions from the `backlog`.

The second solution seems more robust since it does not rely on a timeout period but it requires changes to the voter process

rhsimplex:
Ready for review.  Resolves #196.


bigchaindb/block.py
```diff
@@ -229,3 +232,89 @@ def _start(self):
...
+                self.q_tx_delete.put('stop')
+                self.q_transaction_to_revert.put('stop')
+                return
+
+            # check if tx is in a (valid) block
+            validity = b.get_blocks_status_containing_tx(tx['id'])
+
+            if validity and list(validity.values()).count(Bigchain.BLOCK_VALID) == 1:
+                # tx made it into a block, and can safely be deleted
+                self.q_tx_delete.put(tx['id'])
```
r-marques: why do we need to store the transactions here?

rhsimplex: As opposed to just the `txid`? No I don't think so, I will make the changes.

r-marques: It was more what do we do with `q_tx_delete`?

bigchaindb/block.py
```diff
@@ -229,3 +232,89 @@ def _start(self):
...
+            # poison pill
+            if tx == 'stop':
+                self.q_tx_delete.put('stop')
+                self.q_transaction_to_revert.put('stop')
+                return
+
+            # check if tx is in a (valid) block
+            validity = b.get_blocks_status_containing_tx(tx['id'])
+
+            if validity and list(validity.values()).count(Bigchain.BLOCK_VALID) == 1:
```
r-marques: This will not work like this. There is some period of time since a block is created (and transactions are deleted) and the block is voted valid.

Example timeline:
1. Block1 is createad with tx1 / tx1 is  deleted
2. Receive the delete change for tx1
3. Block is not valid yet / re-insert tx1
4. Block1 is voted valid
5. Block2 is created with tx1

bigchaindb/block.py
rhsimplex: It's there because `BacklogDeleteRevert` is a derived class of block, so I didn't have to rewrite the `validate_transactions`. It expects a queue to take invalid transactions, but yes it's functionally useless.

bigchaindb/block.py
rhsimplex: I added a tunable delay with the latest commit that should fix the problem. The default is 30s.  Can you confirm this fixes the problem? If so I will add the delay to the configurable settings, documentation, etc

r-marques: Do we need to check the validity of the block? This seems to be doing overlapping work with #193 

The way I see it there are 3 different possibilities:
1. if a transaction was deleted but made it into a block regardless of validity, its ok (if the block is later invalid #193 should take care of the rest).
2. If the transaction did not made it into a block and is invalid, its ok
3. If the transaction did not made it into a block and is valid, revert

bigchaindb/block.py
r-marques: It should not be a problem. You never start the `delete_transactions` process

rhsimplex: Yikes, my mistake...but couldn't the queue fill up and start throwing exceptions?

bigchaindb/block.py
rhsimplex: I only meant to check the validity of transactions -- am I missing something big here?

r-marques: Yes that is true.
I think that it would be better to instead of subclassing `Block` just duplicate the `validate_transaction` in `BacklogDeleteRevert`.

This way we don't need to add extra functionality to empty the delete queue and it becomes easier to read the code. Looking at the code it took me some time to figure out how the transactions went from `q_tx_to_validate` to `q_tx_validated`

rhsimplex:
Should I close this? Will backlog also be append only?

r-marques:
This is not that simple because the backlog is used as a temporary storage for unprocessed transactions

vrde:
Stale PR, closing it.

Issue #340, opened 2016-05-30
Remove "operation" from transactions: not needed


ttmc:
We don't need the "operation" (which can be "CREATE" or "TRANSFER") in the [transaction data model](http://bigchaindb.readthedocs.io/en/latest/models.html#the-transaction-model). We can already figure out if a transaction is a creation or transfer transaction by checking to see if it has any inputs. (No inputs means it's a creation transaction.)

For the most part, I think the code already uses the number of inputs to figure out the transaction operation already, so this wouldn't require much change to the code.

sonkesun:
I am trying to think of the best way to do this. We could easily make use of the checks that already exist in the transaction generation function (create_tx) in order to decide whether its a CREATE or a TRANSFER. But what about the "GENESIS" operation (transaction) required for the genesis block? One could think of generating functions (copy & paste) exclusively for this operation. I wonder what makes the GENESIS operation unique given the information provided in the transaction generation? Anyone have any thoughts?

vrde:
In #444 we discuss the idea of removing the `GENESIS` operation. An idea to solve this issue is to focus just on the `CREATE` and `TRANSFER` operation, ignoring (for now) the `GENESIS` operation. We use RethinkDB, that is schemaless, so I think you can easily remove the `operation` key from the default transaction model, but keep it just for the `GENESIS` op.

Does this help you?

sonkesun:
One way to solve this would be to overwrite the 'CREATE' operation only for the 'GENESIS' transaction and to update the transaction id (hash) with the new hash in the prepare_genesis_block function of util.py. 
The 'CREATE' and 'TRANSFER' operations can be set in create_tx depending on the input.  By doing this the operation would no longer be required for the create_transaction function and we keep the 'GENESIS' operation. What do you think?

ttmc:
It seems we're keeping the `operation` in the transaction model for the foreseeable future, so I will 1) label this issue as `wontfix` and 2) close it.

PR #343, opened 2016-05-30
Documentation on Transaction Concepts, and some other stuff

ttmc:
There was no documentation explaining the basics of transactions (i.e. what they are, the kinds, how they work), so I wrote some documentation about that. I put it in **The Transaction, Block and Vote Models** section of the docs.

Note: This section of the docs will also strongly influence (or become) revisions to the BigchainDB whitepaper, so it's important for it to have minimal errors.

I also made a some other minor changes to the docs.

@diminator I assigned you to this PR because I may have made some technical errors and you're among the people most familiar with the technicalities of transactions.


docs/source/models.md
```diff
@@ -1,53 +1,77 @@
 # The Transaction, Block and Vote Models
 
-Transactions, blocks and votes are represented using JSON documents with the following models (schemas). See [the section on cryptography](cryptography.html) for more information about how we calculate hashes and signatures.
+BigchainDB stores all its records in JSON documents.
+
+The three main kinds of records are transactions, blocks and votes. 
+_Transactions_ are used to register, issue, create or transfer things (e.g. assets). Multiple transactions are combined with some other metadata to form _blocks_. Nodes append _votes_ to blocks. This section is a reference on the details of transactions, blocks and votes.
```
r-marques: maybe we should chose only one between register, issue, create

docs/source/models.md
```diff
@@ -1,53 +1,77 @@
...
+
+Later, when we get to the models for the block and the vote, we'll see that both include a signature (from the node which created it). You may wonder why transactions don't have signatures... The answer is that they do! They're just hidden inside the `fulfillment` string of each fulfillment. A creation transaction is signed by the node that created it. A transfer transaction is signed by whoever currently controls or owns it.
+
+What gets signed? For each fulfillment in the transaction, the "fullfillment message" that gets signed includes the `operation`, `timestamp`, `data`, `version`, `id`, corresponding `condition`, and the fulfillment itself, except with its fulfillment string set to `null`. The computed signature goes into creating the `fulfillment` string of the fulfillment.
+
 
 ## Conditions and Fulfillments
 
+An aside: In what follows, the list of `new_owners` (in a condition) is always who owned the asset at the time the transaction completed, but before the next transaction started. The list of `current_owners` (in a fulfillment) is always equal to the list of `new_owners` in that asset's previous transaction.
```
r-marques: > the list of `new_owners` (in a condition) is always who owned the asset at the time the transaction completed

It should be "the list of `new_owners` (in a condition) is always who ownsthe asset at the time the transaction completed".

> The list of `current_owners` (in a fulfillment) is always equal to the list of `new_owners` in that asset's previous transaction.

Hum, not sure if this is true. In the case of a m-n multisig transaction there may be less `current_owners` then the previous transaction `new_owners`.

docs/source/models.md
```diff
@@ -83,7 +107,7 @@ If there is only one _new owner_, the condition will be a single-signature condi
 
 #### Multiple New Owners
 
-If there are multiple _new owners_, we can create a ThresholdCondition requiring a signature from each new owner in order
+If there are multiple _new owners_, they can create a ThresholdCondition requiring a signature from each of them in order
```
r-marques: requiring a signature from each of them or a subset

docs/source/models.md
ttmc: This was added after a discussion with Bruce...  A transaction with no inputs can have many uses, and they can't all be captured with one verb, or at least I don't think they can.

In any case, if we remove the "operation" from the transaction data model (as suggested in issue #340), we're free to use whatever verb we like when describing what one of these input-free transactions is doing.

docs/source/models.md
ttmc: On the first point, I'm pretty sure the two versions mean the same thing.

On the second point, I'm not sure. @diminator can you comment?

docs/source/models.md
ttmc: Yes, in general, a ThresholdCondition can require signatures from only a subset of `new_owners`, but in the example that follows, all of them must sign. There's a comment at the end of this section about the subset option.

docs/source/models.md
diminator: > In the case of a m-n multisig transaction there may be less current_owners then the previous transaction new_owners.

not true, actually the multisig only refers to the number of people to sign off on the transaction. So Troy's statement is correct in my opinion

r-marques: Just not sure what type of validation we do in this case. Do we check this? Should we?

diminator: probably should check it...

Issue #198, opened 2016-04-20
Handle stale transactions on the `backlog` table


r-marques:
We need to handle the case in which a transactions in the `backlog` is assigned to nodeA and nodeA becomes unavailable (network connection, the node went down, couldn't handle the awesomeness of BigchainDB and caught on fire)

The solution would be a process that would check the backlog periodically for stale transactions and reassign them.

ttmc:
For reference, this is the routine `reassignOldTransactions()` in **Listing 3** in the [whitepaper](https://www.bigchaindb.com/whitepaper/bigchaindb-whitepaper.pdf) (around page 23, although the page number may change as we edit the whitepaper).

PR #359, opened 2016-06-06
Core/198/handle stale transactions

rhsimplex:
addresses #198

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -144,7 +144,7 @@ class Bigchain(object):
         response = r.table('backlog').insert(signed_transaction, durability=durability).run(self.conn)
         return response
 
-    def reassign_transaction(self, transaction, durability='soft'):
+    def reassign_transaction(self, transaction, durability='hard'):
         """Assign a transaction to a new node
 
         Args:
@@ -156,16 +156,17 @@ class Bigchain(object):
 
         if self.nodes_except_me:
             try:
-                index_current_assignee = self.nodes_except_me.index(transaction['assignee'])
-                new_assignee = random.choice(self.nodes_except_me[:index_current_assignee] +
-                                             self.nodes_except_me[index_current_assignee + 1:])
+                federation_nodes = self.nodes_except_me + [self.me]
+                index_current_assignee = federation_nodes.index(transaction['assignee'])
+                new_assignee = random.choice(federation_nodes[:index_current_assignee] +
+                                             federation_nodes[index_current_assignee + 1:])
             except ValueError:
                 # current assignee not in federation
                 new_assignee = random.choice(self.nodes_except_me)
 
         else:
-            # There is no other node to assign to, nothing to do
-            return None
+            # There is no other node to assign to
+            new_assignee = self.me 
 
         response = r.table('backlog')\
             .get(transaction['id'])\
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -136,6 +136,7 @@ class Bigchain(object):
         signed_transaction = deepcopy(signed_transaction)
         # update the transaction
         signed_transaction.update({'assignee': assignee})
+        signed_transaction.update({'assignment_timestamp': util.timestamp()})
 
         # write to the backlog
         response = r.table('backlog').insert(signed_transaction, durability=durability).run(self.conn)
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -2,6 +2,7 @@ import random
 import math
 import collections
 from copy import deepcopy
+from time import time
 
 from itertools import compress
 import rethinkdb as r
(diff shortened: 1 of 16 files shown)


bigchaindb/processes.py
```diff
@@ -70,6 +70,7 @@ def start(self):
         p_map_bigchain.start()
         logger.info('starting block')
         block.start()
+        stale.start()
```
vrde: add a `logger.info`

bigchaindb/core.py
```diff
@@ -127,11 +129,54 @@ def write_transaction(self, signed_transaction, durability='soft'):
...
+
+    def get_stale_transactions(self):
+        """Get a RethinkDB cursor of stale transactions
+
+        Transactions are considered stale if they have been assigned a node, but are still in the
+        backlog after some amount of time specified in the configuration
+        """
+
+        return r.table('backlog')\
+            .filter(lambda tx: time() - tx['assignment_timestamp'] >
```
vrde: Should we index `assignment_timestamp`?

r-marques: there is already a secondary index for it https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/db/utils.py#L68

rhsimplex: @rodolphe i added a second timestamp not used by consensus. It's just to keep track of how long a tx has been sitting since last reassignment

bigchaindb/core.py
```diff
@@ -127,11 +129,54 @@ def write_transaction(self, signed_transaction, durability='soft'):
 
         # update the transaction
         signed_transaction.update({'assignee': assignee})
+        signed_transaction.update({'assignment_timestamp': time()})
```
vrde: If we use `r.now` we might change `time()` to `r.now()`.

tests/pipelines/test_stale_monitor.py
```diff
@@ -0,0 +1,112 @@
...
+
+    stm = stale.StaleTransactionMonitor(timeout=0.001,
+                                        backlog_reassign_delay=0.001)
+    tx_stale = stm.check_transactions()
+
+    for _tx in tx_stale:
+        assert tx == _tx
+
+
+def test_reassign_transactions(b, user_vk):
```
vrde: To avoid depending on the system clock you can:
- put a `monkeypatch.setattr(time, 'time', lambda: 1)` before L:26
- change `timeout=0.001` to `timeout=0`, and `backlog_reassign_delay=1`
- put a `monkeypatch.setattr(time, 'time', lambda: 10)` before L:32
- [and do the same with the rest of the test]

(if you are using `r.now` then you should monkeypatch something different)

rhsimplex:
Ok i don't know how to fix the test coverage.  The only two untested lines left are the configuration.  I've tried adding `backlog_reassign_delay` in `test_commands.py` and `test_config_utils.py` in the last two commits but apparently that's not helping. Suggestion?

bigchaindb/core.py
r-marques: We need to be careful when using r.now because this is actually a query to the database to get the database server time. Besides the extra delay for the query I don't know how this values differs from time(). Not sure if rethinkdb uses the clock of the machine it is working on or if uses some kind of synchronized clock when in a cluster

bigchaindb/core.py
```diff
@@ -136,11 +138,54 @@ def write_transaction(self, signed_transaction, durability='soft'):
         signed_transaction = deepcopy(signed_transaction)
         # update the transaction
         signed_transaction.update({'assignee': assignee})
+        signed_transaction.update({'assignment_timestamp': time()})
```
r-marques: Instead of using `time.time()` maybe we should use `bigchaindb.util.timestamp()`
It should not make any difference in this case but I feel it would be more consistent and this would not break if in the future we change how federation nodes get their time

bigchaindb/core.py
```diff
@@ -136,11 +138,54 @@ def write_transaction(self, signed_transaction, durability='soft'):
...
 
         # write to the backlog
         response = r.table('backlog').insert(signed_transaction, durability=durability).run(self.conn)
         return response
 
+    def reassign_transaction(self, transaction, durability='hard'):
```
r-marques: I am not sure about the logic of this method.
I agree that we should assign it to someone other then the current assignee, but I don't understand why it should return `None`. Worst case scenario the node would assign it to itself

bigchaindb/core.py
```diff
@@ -136,11 +138,54 @@ def write_transaction(self, signed_transaction, durability='soft'):
...
+
+        if self.nodes_except_me:
+            try:
+                federation_nodes = self.nodes_except_me + [self.me]
+                index_current_assignee = federation_nodes.index(transaction['assignee'])
+                new_assignee = random.choice(federation_nodes[:index_current_assignee] +
+                                             federation_nodes[index_current_assignee + 1:])
+            except ValueError:
+                # current assignee not in federation
+                new_assignee = random.choice(self.nodes_except_me)
```
r-marques: why don't we use `self.me` here also?

bigchaindb/pipelines/stale.py
```diff
@@ -0,0 +1,76 @@
...
+        self.timeout = timeout
+
+    def check_transactions(self):
+        """Poll backlog for stale transactions
+
+        Returns:
+            txs (list): txs to be re assigned
+        """
+        sleep(self.timeout)
+        for tx in list(self.bigchain.get_stale_transactions()):
```
r-marques: `get_stale_transactions` returns a rethinkdb cursor. There is no need for a `list` here.

bigchaindb/core.py
rhsimplex: Don't think it will work -- for stale transactions, we need to look within a window of time.  A string time stamp may not order correctly (e.g. "100" comes before "20").  For instance, you can't call RethinkDBs `between` function on "0" and "140000000" and get the right answer, I believe.

bigchaindb/core.py
rhsimplex: We can do this, but it means a single node cluster (which is already timing out) will spend a lot of effort reassigning transactions to itself.

bigchaindb/core.py
r-marques: timestamp returns a fixed length string. You can always convert to int before writing it to the database. I am just concerned of problems that may arise if we start getting timestamps from different sources in different parts of the code

bigchaindb/core.py
r-marques: Makes sense.

There is also the problem that since the assignment_timestamp is not updated then the stale monitor of that node will be constantly retrieving the same transactions. But I guess that if this is a problem then the node should increase the reassign_timeout

bigchaindb/core.py
rhsimplex: Ok, then this shouldn't be too difficult

r-marques:
Another thing that is missing is that the block creation pipeline now needs to also listen for updates,

bigchaindb/core.py
r-marques: After thinking a bit more about this it would make sense for a single node to reassign the transaction back to itself. Lets say the node lost one of the notifications from the changefeed by reassigning the transaction to itself it will fire up a new changefeed notification.

r-marques:
cc @vrde 
This is failing right now because the UPDATE from the changefeed returns `{"old_val": ..., "new_val": ...}` instead of the transaction

Two solutions would be:
- change the ChangeFeed class to return only `new_val`
- in `block.filter_tx` check if its a transaction or update notification and to `tx = update['new_val']`

r-marques:
works on my computer so I think we are good to go :+1:

vrde:
> works on my computer so I think we are good to go

:joy:

Issue #129, opened 2016-03-10
Support for divisible assets


r-marques:
relates to #128 #99 #125 

The ability to divide assets is common in crypto currency ledgers but its difficult to envision on a digital asset ledger where there is no notion of value or amount.

There are use cases where this may be desirable:
- Building a crypto currency with BigchainDB
- Track the supply chain of physical goods. A raw material that is transformed at some point in the supply chain.

I propose that we define a generic template for digital assets that would have a field that specifies if the digital asset is divisible or not and add a pluggable consensus rule to and the case of divisible assets. The consensus rule should extend the core consensus instead of completely overriding it.

A generic divisible asset should provide:
- Value
- Minimum divisible unit eg 1/10^8

The consensus rules need to:
- Respect the minimum divisible unit
- Enforce the conservation law. The sum of all the parts should never with greater then the initial value.

**TODO**:
- [x] Define digital asset #125 
- [x] Implement digital asset template #564
- [x] Document digital asset template #566 
- [x] Implement divisible assets #488 
- [x] Documentation and examples for divisible assets #486

TimDaub:
I think I'm in line with what was proposed here.

What I'd like to have is define on the creation of a digital asset is it's divisibility.
The use case for this is simple:

You want to enable users to create artificially limited assets (essentially "limited editions").
In that case, you'd only be able to divide and "spend" the asset by the number of times it was defined in its registration. In turn this allows economics where a digital asset is only spendable in a limited quantity to gain actual value.

r-marques:
The simplest way to go about it would be to assume that the minimum divisible unit is always `1`. And in the definition of the digital asset you specify an amount. For example if you create a divisible digital asset with `amount=10` this means that there will be at most 10 shares of that digital asset.

lucidfrontier45:
Hi guys.

Any updates on this issue? I want to make a bitcoin like financial system with numeric payload and I think divisible asset is necessary for this usage.

ttmc:
Support for divisible assets is not currently a high priority for us. If you want to make a "bitcoin like financial system" then why not use one of the existing cryptocurrencies?

furiousOyster:
@ttmc because transactional thruput !  bigchaindb holds great promise in tx/sec and we, the naive onlookers, feel that a numeric payload would be easy* and super beneficial.

ttmc:
[...]

PR #364, opened 2016-06-06
Add some notes to docs on transaction concepts

ttmc:
I added some more documentation about transaction concepts, based on the old (internal) FAQ documents.

* BigchainDB only supports indivisible assets, but issue #129 proposes to change that
* The client creating a tx may not be involved in the tx in any way
* What does a node check when validating a transaction?

Note: Technically, nodes don't do schema validation today, but they will soon; see PR #311

### DIFF
--- a/docs/source/models.md
+++ b/docs/source/models.md
@@ -11,19 +11,30 @@ Below we often refer to cryptographic hashes, keys and signatures. The details o
 
 Transactions are the most basic kind of record stored by BigchainDB. There are two kinds: creation transactions and transfer transactions.
 
-A creation transaction can be used to register, issue, create or otherwise initiate the history of a single thing (or asset) in BigchainDB. For example, one might register an identity or a creative work. The things are often called "assets" but they might not be literal assets. A creation transaction also establishes the initial owner or owners of the asset. Only a federation node can create a valid creation transaction (but it's usually made based on a message from a client).
+A _creation transaction_ can be used to register, issue, create or otherwise initiate the history of a single thing (or asset) in BigchainDB. For example, one might register an identity or a creative work. The things are often called "assets" but they might not be literal assets. A creation transaction also establishes the initial owner or owners of the asset. Only a federation node can create a valid creation transaction (but it's usually made based on a message from a client).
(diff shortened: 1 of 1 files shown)


docs/source/models.md
```diff
@@ -11,20 +11,31 @@ Below we often refer to cryptographic hashes, keys and signatures. The details o
...
+A _creation transaction_ can be used to register, issue, create or otherwise initiate the history of a single thing (or asset) in BigchainDB. For example, one might register an identity or a creative work. The things are often called "assets" but they might not be literal assets. A creation transaction also establishes the initial owner or owners of the asset. Only a federation node can create a valid creation transaction (but it's usually made based on a message from a client).
 
-A transfer transaction can transfer one or more assets to new owners.
+Currently, BigchainDB only supports indivisible assets. You can't split an asset apart into multiple assets, nor can you combine several assets together into one. This means BigchainDB doesn't support currency, at least not in the usual sense. [Issue #129](https://github.com/bigchaindb/bigchaindb/issues/129) is an enhancement proposal to support divisible assets.
```
r-marques: Not comfortable in mentioning currencies on bigchaindb. Its a different beast that requires much more then just divisible assets and bigchaindb was never designed with crypto currencies in mind.

ttmc: Okay I will remove the sentence, "This means BigchainDB doesn't support currency..."

r-marques: I wouldn't state that either. We are not working towards adding support for crypto currencies on bigchaindb but people may still want to experiment so see how feasible it would be.

What I am saying is that there is a lot more that needs to be done for bigchaindb to support crypto currencies. Like for instance all the economics that would need to be added to the consensus protocol (if federation nodes have no limitation on the amount of new "coins" it can mint a crypto currency would always be worthless)

ttmc: Indeed, it's even worse than that, because once a ledger starts supporting the movement of "currency," there are all sorts of laws and regulations that one must comply with. I'll defer a discussion of currency for a to-be-written section of the docs.

PR #366, opened 2016-06-07
Documentation reorganization + new sections on decentralization & immutability

ttmc:
The right sidebar of the BigchainDB Documentation was looking full and there is still a lot more documentation to write. I didn't want us to end up with a right sidebar having 100 items! I looked at what big Python projects like Django do. The basic idea is to use subdirectories, subsubdirectories, etc, each with its own index file. I reorganized the docs so that the top level nav has these six items:

* Introduction
* Topic Guides
* BigchainDB Servers (Nodes) & Clusters
* BigchainDB Drivers & Clients
* Release Notes
* Appendices

Some of those are files and some are directories with their own index file. Neat, eh? I updated all the links (internal and in `README.md`).

I wrote some new sections based on stuff that used to be in the old internal FAQ documents:

* How BigchainDB is Decentralized
* Kinds of Node Diversity
* How BigchainDB is Immutable / Tamper-Resistant

Two other minor things:

* I got rid of the stand-alone section about contributing because there's already something about that in the Introduction.
* I renamed the "Developer Interface" section as "The Bigchain class".


ttmc:
I'm going to merge this now because it's either:
* reorganization, or
* based substantially on internal documents which have already been reviewed.

PR #367, opened 2016-06-07
Add informative JSON message to the API root

vrde:
Expose some basic data about the node when visiting the API root.

Example:

```
{
  "api_endpoint": "http://localhost:9984/api/v1",
  "keyring": [],
  "public_key": "9kqcpibHQza5RxMsvpPpviB2B1gaohn44oPa48UND1zE",
  "software": "BigchainDB",
  "version": "0.4.0"
}
```

### DIFF
--- a/bigchaindb/web/server.py
+++ b/bigchaindb/web/server.py
@@ -62,6 +62,7 @@ def create_app(settings):
     app.config['bigchain_pool'] = util.pool(Bigchain, size=settings.get('threads', 4))
     app.config['monitor'] = Monitor()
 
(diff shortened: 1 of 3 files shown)


ttmc:
The [transaction model](https://bigchaindb.readthedocs.io/en/latest/models.html#the-transaction-model) has a different "version": the version of the transaction data model (or is it the API version?). I wonder if that should be reported as well.

r-marques:
This seems good enough for now. Since we have only one version for the transactions at this point I don't think there is a need for it.

:+1:

PR #369, opened 2016-06-08
Return 404 when transaction not found

vrde:
The Web API now returns `404` instead of a `500` when a transaction is not found.


r-marques:
:+1:

TimDaub:
:+1:

PR #370, opened 2016-06-09
In the docs, added some words of caution regarding timestamps etc.

ttmc:
Suggested by @TimDaub 

In the BigchainDB documentation, in the section about transaction, block, and vote models, I added some words of caution about using timestamps.

Related issue: Issue #132


TimDaub:
:+1:

PR #378, opened 2016-06-13
ensure timestamp is since epoch in UTC

diminator:
the `mktime` used local time not UTC, hence times weren't in global UTC

### DIFF
--- a/bigchaindb/util.py
+++ b/bigchaindb/util.py
@@ -133,7 +133,7 @@ def timestamp():
         str: UTC timestamp.
 
     """
-    return "{0:.6f}".format(time.time())
+    return "{:d}".format(int(round(time.time())))
 
 
 # TODO: Consider remove the operation (if there are no inputs CREATE else TRANSFER)
--- a/bigchaindb/util.py
+++ b/bigchaindb/util.py
@@ -133,8 +133,7 @@ def timestamp():
         str: UTC timestamp.
 
     """
-    dt = datetime.utcnow()
-    return "{0:.6f}".format(time.mktime(dt.timetuple()) + dt.microsecond / 1e6)
+    return "{0:.6f}".format(time.time())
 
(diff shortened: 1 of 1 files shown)


r-marques:
:+1:

ttmc:
Apparently `time.time()` [is timezone-independent](http://stackoverflow.com/questions/11845803/is-pythons-time-time-timezone-specific). Good!

However, from [the Python 3.4 docs for `time.time()`](https://docs.python.org/3.4/library/time.html#time.time):

> Return the time in seconds since the epoch as a floating point number. Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second.

I suggest that we also use this PR to round the timestamps off to the nearest second, since that's all the precision they have anyway.

diminator:
> I suggest that we also use this PR to round the timestamps off to the nearest second, since that's all the precision they have anyway.

@r-marques @ttmc for me that's ok, :+1: for rounding to nearest second?

diminator:
> I suggest that we also use this PR to round the timestamps off to the nearest second, since that's all the precision they have anyway.

@r-marques @ttmc any idea if there is a `format()` expression to `round` (and not only `floor` or `ceil`)?

ttmc:
I looked around but even if there is one, I think we'd end up with easier-to-understand code if we just use a proper rounding function, rather than some fancy-pants, impossible-to-read expression for `format()`.

diminator:
> I suggest that we also use this PR to round the timestamps off to the nearest second, since that's all the precision they have anyway.

It seems that rounding the `timestamp` will yield duplicate transaction hashes, which are rejected.
In order to do this properly we could change the `payload.hash` by `payload.uuid`

I suggest to do this rounding in another issue?

r-marques:
closing this pull request in favor of #384

PR #379, opened 2016-06-13
Votes table

### LINKED ISSUE #368: Voting as separate table
### Problem

Per the white paper &sect; 4.7.3, votes are appended directly to the blocks.  While the votes are signed, and can thus be validated, there is no notion of when a vote is written or if its value has been changed.  This makes it impossible to revert Byzantine changes to the block table as long as changes are handled by reacting to a changefeed -- a changed vote could be changed back, but that would trigger a change and so on.  This was recognized while trying to resolve #195 and #194.

### Solution

@ttmc proposed the following solution: write votes to a table separate from blocks. This has a few advantages (see initial discussion on #195):

1. All updates to the block and voting tables (except appends) can be treated as incorrect. This is conceptually simpler.

2. Votes are hashed like blocks now, providing a verifiable chain(?)

3. The "update causes an update causes an update..." quagmire is avoided.  Updates to vote and block tables need only verify the new value fits in the chain (e.g. hash of next/previous block is correct).  This doesn't protect against the last vote being altered by the voting node, but once another vote is cast, it's locked.

### Tasks:

- [ ] BigchainDB setup initializes third table for votes -- should include a "genesis" vote

- [ ] New votes append to votes table instead of block table. Primary key should be something like `block_id + voting_node_id` -- this is unique, and another insert with this id will be rejected.

- [ ] New vote should include hash of previous vote. (Is this necessary?) 

- [ ] Queries will need to be updated, particularly those which depend on block validity.  Block validity will be decided via query to the votes table rather than the block table

- [ ] Some processes (block, voter) will need modification

- [ ] Update documentation and WP

rhsimplex:
Ready to review.  Aims to resolve #368.

What's missing? I haven't done anything with the way the processes handle updates/deletes yet.  There are still some open questions on that @ttmc and I discussed at some length.

Also the handling of block numbers is clunky (currently the only old test that is still failing).  In the spirit of append-only, the block numbers should be in the votes.  This makes for some convoluted resolution of the block order though.

Test coverage decreased because we removed one test that is no longer meaningful


ttmc:
The `bigchaindb set-shards` command should also set the number of shards on the 'votes' table, so I fixed that and its test.

ttmc:
The documentation of the "block model" and "vote model" (data models) will also have to be updated. See http://bigchaindb.readthedocs.io/en/latest/topic-guides/models.html#the-block-model

bigchaindb/core.py
```diff
@@ -20,6 +21,10 @@ class ImproperVoteError(Exception):
     pass
 
 
+class MultipleVotesError(Exception):
```
r-marques: all exceptions should go into `exceptions.py`

bigchaindb/core.py
```diff
@@ -449,15 +454,22 @@ def has_previous_vote(self, block):
...
-                if vote['node_pubkey'] == self.me:
-                    if util.verify_vote_signature(block, vote):
-                        return True
-                    else:
-                        raise ImproperVoteError('Block {block_id} already has an incorrectly signed vote '
-                                                'from public key {me}').format(block_id=block['id'], me=self.me)
-        return False
+        votes = r.table('votes').get_all([block['id'], self.me], index='block_and_voter').run(self.conn)
+
+        votes = list(votes)
```
r-marques: `list()` and `query` can be collapsed into a single line

bigchaindb/core.py
```diff
@@ -449,15 +454,22 @@ def has_previous_vote(self, block):
...
+
+        if len(votes) > 1:
+            raise MultipleVotesError('Block {block_id} has {n_votes} votes from public key {me}'
+                                     .format(block_id=block['id'], n_votes=str(len(votes)), me=self.me))
+        has_previous_vote = False
+        for vote in votes:
```
r-marques: at this point there should be only one vote, don't think there is a need for a `for` loop

bigchaindb/core.py
```diff
@@ -560,44 +572,44 @@ def write_vote(self, block, vote, block_number):
...
-        last_voted = list(last_voted)
         if not last_voted:
             return list(r.table('bigchain')
                         .filter(r.row['block_number'] == 0)
                         .run(self.conn))[0]
 
-        return last_voted[0]
+        res = r.table('bigchain').get(last_voted[0]['vote']['voting_for_block']).run(self.conn)
+
+        if return_block_number:
```
r-marques: doesn't the block already contains the `block_number`?

bigchaindb/core.py
```diff
@@ -560,44 +572,44 @@ def write_vote(self, block, vote, block_number):
...
 
     def get_unvoted_blocks(self):
         """Return all the blocks that has not been voted by this node."""
 
         unvoted = r.table('bigchain') \
-            .filter(lambda doc: doc['votes'].contains(lambda vote: vote['node_pubkey'] == self.me).not_()) \
-            .order_by(r.asc((r.row['block']['timestamp']))) \
+            .filter(lambda block: r.table('votes').get_all([block['id'], self.me], index='block_and_voter')
+                    .is_empty()) \
+            .order_by(r.desc('block_number')) \
```
r-marques: if a block hasn't been voted yet it will have no `block_number` that is why we need to order by timestamp.
`r.asc((r.row['block']['timestamp']))` instead of `r.desc('block_number')`. This is the reason the test is failing

vrde:
If we want to remove the `block_number`, should we do it in this PR?

bigchaindb/core.py
rhsimplex: This is a self-check on the node to see if it has more than one vote on a particular block.  In normal operation, no this should not happen.

If you still think I should remove it, no problem.

bigchaindb/core.py
rhsimplex: Nope, and this is kind of an awkward fix -- the genesis block contains block a block number==0, but otherwise block numbers are in the votes now (so the block never has to be updated).

`return_block_number` is a flag because only the bootstrapping voter process needs it -- all other queries by the consensus algorithm don't use the block number, so there's no need to read the voter table for it.

bigchaindb/core.py
r-marques: Its just that there is a `raise` before so at this point votes is never `> 1`

tests/db/test_voter.py
```diff
@@ -330,8 +330,10 @@ def test_voter_chains_blocks_with_the_previous_ones(self, b):
             .run(b.conn)
 
         assert blocks[0]['block_number'] == 0  # genesis block
-        assert votes[0]['block_number'] == 1 and votes[0]['vote']['voting_for_block'] == blocks[1]['id']
-        assert votes[1]['block_number'] == 2 and votes[1]['vote']['voting_for_block'] == blocks[2]['id']
+        assert votes[0]['vote']['voting_for_block'] == blocks[1]['id'] or \
+               votes[0]['vote']['voting_for_block'] == blocks[2]['id']
```
vrde: IMO more readable:
```python
assert votes[0]['vote']['voting_for_block'] in (blocks[1]['id'], blocks[2]['id'])
```

bigchaindb/core.py
```diff
@@ -444,15 +444,20 @@ def has_previous_vote(self, block):
...
+        if len(votes) > 1:
+            raise exceptions.MultipleVotesError('Block {block_id} has {n_votes} votes from public key {me}'
+                                     .format(block_id=block['id'], n_votes=str(len(votes)), me=self.me))
+        has_previous_vote = False
+        for vote in votes:
+            if util.verify_vote_signature(block, vote):
+                has_previous_vote = True
+            else:
+                raise exceptions.ImproperVoteError('Block {block_id} already has an incorrectly signed vote '
+                                        'from public key {me}').format(block_id=block['id'], me=self.me)
```
r-marques: the `.format` needs to come before closing the parenthesis

bigchaindb/core.py
r-marques: Should we make the genesis block behave like any other blocks.
- Create the genesis block with voters = [b.me]
- Write the block and write the vote in the votes table

bigchaindb/core.py
```diff
@@ -444,15 +444,20 @@ def has_previous_vote(self, block):
...
-                        ).format(block['id'], self.me)
-                    return True
-        return False
+        votes = list(r.table('votes').get_all([block['id'], self.me], index='block_and_voter').run(self.conn))
+
+        if len(votes) > 1:
+            raise exceptions.MultipleVotesError('Block {block_id} has {n_votes} votes from public key {me}'
+                                     .format(block_id=block['id'], n_votes=str(len(votes)), me=self.me))
+        has_previous_vote = False
+        for vote in votes:
```
r-marques: I still don't understand this for loop. You already raise an exception before `if len(votes) > 1` and at this point you either have 0 or 1 vote

bigchaindb/core.py
```diff
@@ -555,44 +560,44 @@ def write_vote(self, block, vote, block_number):
...
 
-        r.table('bigchain') \
-            .get(vote['vote']['voting_for_block']) \
-            .update(update) \
+        r.table('votes') \
+            .insert(vote) \
             .run(self.conn)
 
-    def get_last_voted_block(self):
+    def get_last_voted_block(self, return_block_number=False):
```
r-marques: Do we need the optional argument `return_block_number`?
This method seems to be called only from `Voter.bootstrap` with `return_block_number=True`
Anyway if a block was voted then it already has a block number.

rhsimplex: That's the problem, blocks never have block numbers...they're stored in the votes themselves.  So I made this kind of ugly compromise to avoid rewriting the voter module (if you want the block number, that's an additional database read -- but as you said, this is the only function that uses it).

Yet another reason to get rid of them, right?

r-marques: There is no additional database read. You make a read to the votes table and then a read to the bigchain table regardless

rhsimplex: yes you're right.  i guess it changes the spec of returned block, but maybe it should just be on by default?

r-marques: I guess so. Its outside of the `block` structure so it shouldn't interfere with anything and right now its just being used for the voter

r-marques:
:+1:

PR #381, opened 2016-06-14
[Do not merge] Use IPLD for hashing of transactions and payloads

TimDaub:
This PR is somewhat a WIP.

A little summary on what has been done:

- serialize all objects using IPLD (that internally uses [CBOR](http://cbor.io/) and SHA2-256 for hashing)
- hash serialized objects using [multihash](https://github.com/jbenet/multihash) ([pymultihash](https://github.com/ivilata/pymultihash))

To be determined:

- This is probably not an option we'd like to have turned on by default, but a feature that can optionally be enabled, right?
- Is it worth the same effort to do this with linked data? I'd say so.

Anyways. For anyone interested, this is a first cut on how IPLD could look like in BigchainDB.

TODOs:

- [x] See mostly TODOs for: https://github.com/bigchaindb/py-ipld/issues/1
- [ ] Depending on the TBDs, we'll have to adjust the current documentation

For everyone trying this:

Please note that the py-ipld currently is a WIP.


.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
 
 install:
   - sudo apt-get install rethinkdb
-  - pip install -e .[test]
+  - pip install --process-dependency-links -e .[test]
```
TimDaub: This is only temporary. When py-ipld is on pipy, this will be gone.

bigchaindb/crypto.py
```diff
@@ -1,12 +1,12 @@
...
-import sha3
 from cryptoconditions import crypto
+from ipld import multihash
 
 
-def hash_data(data):
-    """Hash the provided data using SHA3-256"""
-    return sha3.sha3_256(data.encode()).hexdigest()
+def hash_data(serialized_data):
+    """Hash the provided data using IPLD-specific multihash SHA2-256"""
```
TimDaub: IPLD implementations use SHA2-256.

ttmc:
I guess the docs on serialization should also be modified? See https://bigchaindb.readthedocs.io/en/latest/appendices/json-serialization.html

ttmc:
A proposal: Switch to multihash for all hashes in BigchainDB, with SHA3-256 as the default hashing function used by BigchainDB. (Multihash lets you use other hashing functions if you want; that's kind of the whole point; if someone wants to use SHA-256, they can.) That lets us continue using SHA3-256 yet opens us up to playing nicely with IPLD etc.

I'm less sure how to manage/select/choose from the various methods of serialization. It seems IPLD wants everyone to use CBOR. That seems strange alongside the flexibility of Multihash.

@TimDaub Maybe you could share your thoughts about how to proceed (here)?

setup.py
```diff
@@ -104,6 +104,7 @@ def check_setuptools_features():
         'flask==0.10.1',
         'requests==2.9',
         'gunicorn~=19.0',
+        'py-ipld==0.0.1',
```
r-marques: this is now called just `ipld`, right?

TimDaub: Oh yes. I actually have a PR ready for changing this.

TimDaub:
@troy @diminator @r-marques:

I discovered another problem with including IPLD into BigchainDB.
Currently, when creating a transaction, its id is [being created](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/util.py#L293) before the transaction is signed.

Now, when requesting a transaction using for example the included REST API, its requested using it's hash/id: `GET /v1/transactions/txid`.
If a client now wants to integrity check the transaction they'd just take the data + proposed transaction id hash they've received and do the following comparison:

`hash(data) == hash_used_for_request`

Now, the problem here is that this statement would yield `false`, since the client wasn't aware of the actual hashing implementation of bigchaindb, which is located [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/util.py#L493).
The secret sauce: The fulfillment's signature is not hashed since it itself is the hash we're looking for in the first place

For a client that's only communicating with bigchaindb, this might not be a problem since the cryptographic-integrity checking can be specialized to the transaction body of bigchaindb, but for a client that is ledger-generic (like an IPLD-resolver), this is problematic.

For a ledger-generic client, the following requirements must be fulfilled:

- Specific is requested/queried by addressing content. Specifically this means that a hash is requested. All of the returned data must yield the requested hash again, meaning  `hash(data) == hash_used_for_request` ==> True`
- Hashes for data elements are unnecessary, as a client integrity-checks the data against it's `hash_used_for_request` anyways.


The implementation of the following recommendations would be helpful (and also probably difficult to implement):

- To a client, a transaction has no explicit `id` anymore. The client computes the `id` of a transaction simply by hashing its contents themselves
- To a client, a transaction has no explicit `tx.data.hash` anymore. The client computes the `id` of a transaction simply by hashing its contents themselves
- The returned data from a e.g. REST API call GET'ing a transaction by it's id returns the data that exactly yields a corresponding hash. `hash(data) == hash_used_for_request` ==> True` is fulfilled all the time.
- BigchainDB internally is indexing and addressing transaction ids somehow, but this concept is completely hidden and abstracted to the client. The client keeps track of an index of transactions they're interested.


BTW.: I'm aware that this could cause transaction malleability. The point is to fix the problem without doing that :)

troy:
@TimDaub wrong Troy :)

TimDaub:
This will be much more clean to implement using the new transaction model proposed in #641.

PR #384, opened 2016-06-15
Feat/380/round timestamp add uuid

### LINKED ISSUE #380: Round timestamps to the nearest second & avoid duplicate tx hashes
Timestamps now come from calling `time.time()` (see `bigchaindb/util.py`). According to [the Python 3.4 docs for `time.time()`](https://docs.python.org/3.4/library/time.html#time.time), it will:

> Return the time in seconds since the epoch as a floating point number. Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second.

I suggested we round timestamps off to the nearest second, since that's all the precision they have anyway. @diminator already did this in an earlier commit (reverted):

https://github.com/bigchaindb/bigchaindb/pull/378/commits/1fa35f068faf76838afab7287c984d428574cdc6

but then he wondered if that might lead to transactions with the same hash (= id = database primary key). If so, the duplicate transactions would be dropped. Maybe that's okay? Do we want to allow transactions so similar that their only difference is the timestamp? Why not just tell clients that if they don't want their transactions dropped, each one should have a different payload or conditions?

If we _do_ allow transactions to be duplicates in everything-but-timestamp, then @diminator suggested using a UUID instead of the payload hash inside the payload (in a comment on Pull Request #378). That would work, although we'd lose the ability to verify the payload integrity directly (by calculating its hash). However, we could still verify the payload integrity indirectly from the signature(s) in the fulfillment(s).

A variation would be to have both a payload UUID _and_ a payload hash, with the hash including the UUID.

r-marques:
resolves #380

Changes done in this pull request:
- Round the timestamp to have second precision.
- Replaced the payload hash with an uuid4. This allows us to index the payload even if it is a duplicate
- Updated secondary index `payload_hash -> payload_uuid`
- Added and updated tests

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -386,6 +386,10 @@ class Bigchain(object):
             dict: created block.
         """
 
+        # Prevent the creation of empty blocks
+        if len(validated_transactions) == 0:
+            raise exceptions.OperationError('Empty block creation is not allowed')
+
         # Create the new block
         block = {
             'timestamp': util.timestamp(),
--- a/bigchaindb/db/utils.py
+++ b/bigchaindb/db/utils.py
@@ -54,7 +54,7 @@ def init():
         .run(conn)
     # secondary index for payload hash
     r.db(dbname).table('bigchain')\
-        .index_create('payload_hash', r.row['block']['transactions']['transaction']['data']['hash'], multi=True)\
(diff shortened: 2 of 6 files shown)


ttmc:
I guess the docs for the transaction model (data model) should be updated:  https://bigchaindb.readthedocs.io/en/latest/topic-guides/models.html#the-transaction-model

ttmc:
:+1:

PR #390, opened 2016-06-16
Renamed Bigchain.federation_nodes as Bigchain.nodes_except_me

ttmc:
Someone new to the BigchainDB code might look at the list named `Bigchain.federation_nodes` and think that it includes all the federation nodes, but it doesn't. (It doesn't include the public key of the current node; that's stored in `Bigchain.me`.)

I figured it would be a good idea to rename it so there's less ambiguity. I changed it to `Bigchain.nodes_except_me`. That name makes a good pair with `Bigchain.me`.

An alternative would be `Bigchain.keyring` but that name is also somewhat ambiguous. I suppose BigchainDB developers should know what "keyring" means (i.e. that it excludes the public key of the current node). I prefer to assume as little as possible.


sbellem:
:+1:

Issue #389, opened 2016-06-16
New command-line command to set replication factor


ttmc:
We need a command-line command to set the replication factor of all the tables. Right now there isn't one, but we need it. [RethinkDB failover](https://rethinkdb.com/docs/failover/) won't work unless there are three or more replicas.

For the command syntax, how about:
```text
bigchaindb set-replicas replication_factor
```

similar to: 
```text
bigchaindb set-shards num_shards
```

I will do this now. I don't expect it to take long.

ttmc:
Writing the `bigchaindb set-replicas` command was straightforward. Testing it turns out to be tricky because if you want to set the number of replicas to 2, say, then you need at least two nodes in the RethinkDB cluster... but we normally only do tests with a one-node "cluster."

I asked @vrde what he'd do and he said that the unit test doesn't have to make sure that RethinkDB's reconfigure command works (i.e. we can assume it does); all we need to test is that the RethinkDB reconfigure command gets the correct argument. I'm not sure how to do that.

If someone could help with that test, I'd be very happy. I pushed the branch to this repository. It's named `feat/389/cli-set-replication-factor`. (I didn't make it into a PR.) The relevant code is in:

* `/bigchaindb/commands/bigchain.py`
* `/tests/test_commands.py` (at the very end)

ttmc:
Resolved by Pull Request #392

PR #392, opened 2016-06-17
[Needs test] Added bigchaindb set-replicas command

ttmc:
This is to resolve issue #389. Copying from the comment I made on that issue:

Writing the `bigchaindb set-replicas` command was straightforward. Testing it turns out to be tricky because if you want to set the number of replicas to 2, say, then you need at least two nodes in the RethinkDB cluster... but we normally only do tests with a one-node "cluster."

(There's a dummy test in the code I pushed, but it doesn't work.)

I asked @vrde what he'd do and he said that the unit test doesn't have to make sure that RethinkDB's reconfigure command works (i.e. we can assume it does); all we need to test is that the RethinkDB reconfigure command gets the correct argument. I'm not sure how to do that.


tests/test_commands.py
```diff
@@ -248,7 +248,7 @@ def test_set_shards(b):
 
 
 def test_set_replicas(b):
-    import rethindb as r
```
ttmc: Doh! I guess this test was failing for more than one reason.

ttmc:
:+1:

PR #393, opened 2016-06-20
More docs improvements

ttmc:
* More notes on node requirements:  memory, storage (RethinkDB can take advantage of SSDs), file systems supported
* New short section on BigchainDB and Smart Contracts
* Added a high-level explanation of conditions (with help from @diminator )
* Added short subsection on BigchainDB Integration with Other Blockchains
* Added note about how one can't currently do queries via the HTTP API. I suggested they use the Python Server API for prototypes, for now
* Moved "Topic Guides" later in the Table of Contents
* Added a Quickstart page with instructions to setup a stand-alone node for learning or experimenting
* Added a new short page about Example Apps (pointing to the `bigchaindb-examples` repo)
* Added the setting of the replication factor to the instructions for cluster node setup
* Refactored the setup instructions into four cases (all linked from the Introduction page):
    1. stand-alone node for learning & experimenting (Quickstart)
    2. set up and run a federation (points to the page about how to set up a cluster node, which got simplified quite a bit)
    3. set up to contribute to the development of BigchainDB Server (points to the CONTRIBUTING.md file, which also got revised)
    4. deploy a cluster on AWS

### DIFF
--- a/docs/source/drivers-clients/http-client-server-api.md
+++ b/docs/source/drivers-clients/http-client-server-api.md
@@ -2,14 +2,13 @@
 
 Note: The HTTP client-server API is currently quite rudimentary. For example, there is no ability to do complex queries using the HTTP API. We plan to add querying capabilities in the future. If you want to build a full-featured proof-of-concept, we suggest you use [the Python server API](../nodes/python-server-api-examples.html) for now.
 
-When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at:
+When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at the address stored in the BigchainDB node configuration settings. The default is for that address to be:
 
 [http://localhost:9984/api/v1/](http://localhost:9984/api/v1/)
 
-For security reasons, the server binds to `localhost:9984`.
-If you want to bind the server to `0.0.0.0` we recommend you to read
-[Deploying Gunicorn](http://docs.gunicorn.org/en/stable/deploy.html) and
-follow the instructions to deploy it in production.
+but that address can be changed by changing the "API endpoint" configuration setting (e.g. in a local config file). There's more information about setting the API endpoint in [the section about Configuring a BigchainDB Node](../nodes/configuration.html).
+
+There are other configuration settings related to the web server (serving the HTTP API). In particular, the default is for the web server socket to bind to `localhost:9984` but that can be changed (e.g. to `0.0.0.0:9984`). For more details, see the "server" settings ("bind", "workers" and "threads") in [the section about Configuring a BigchainDB Node](../nodes/configuration.html).
 
 The HTTP API currently exposes two endpoints, one to get information about a
 specific transaction id, and one to push a transaction to the BigchainDB
--- a/docs/source/drivers-clients/http-client-server-api.md
+++ b/docs/source/drivers-clients/http-client-server-api.md
@@ -1,11 +1,12 @@
 # The HTTP Client-Server API
 
+Note: The HTTP client-server API is currently quite rudimentary. For example, there is no ability to do complex queries using the HTTP API. We plan to add querying capabilities in the future. If you want to build a full-featured proof-of-concept, we suggest you use [the Python server API](../nodes/python-server-api-examples.html) for now.
+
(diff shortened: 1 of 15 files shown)


docs/source/drivers-clients/http-client-server-api.md
```diff
@@ -1,11 +1,12 @@
...
-When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at:
+Note: The HTTP client-server API is currently quite rudimentary. For example, there is no ability to do complex queries using the HTTP API. We plan to add querying capabilities in the future. If you want to build a full-featured proof-of-concept, we suggest you use [the Python server API](../nodes/python-server-api-examples.html) for now.
 
-- [http://localhost:9984/api/v1/](http://localhost:9984/api/v1/)
+When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at:
 
+[http://localhost:9984/api/v1/](http://localhost:9984/api/v1/)
 
-Please note that for security reasons the server binds to `localhost:9984`.
+For security reasons, the server binds to `localhost:9984`.
```
r-marques: There may also be reasons to bind it to `0.0.0.0` in development (we do this in docker). We only need to change the api endpoint in the configuration file

r-marques:
:+1:

docs/source/index.rst
```diff
@@ -11,9 +8,10 @@ Table of Contents
    :maxdepth: 1
 
    introduction
-   topic-guides/index
+   quickstart
```
diminator: :+1:

docs/source/topic-guides/models.md
```diff
@@ -90,13 +90,35 @@ One other note: Currently, transactions contain only the public keys of asset-ow
...
+
+The (single) output of a threshold condition can be used as one of the inputs of other threshold conditions. This means that one can combine threshold conditions to build complex logical expressions, e.g. (x OR y) AND (u OR v).
+
+Aside: In BigchainDB, the output of an m-of-n threshold condition can be inverted on the way out, so an output that would have been TRUE would get changed to FALSE (and vice versa). This enables the creation of NOT, NOR and NAND gates. At the time of writing, this “inverted threshold condition” is BigchainDB-specific (i.e. not part of the Interledger specs). It should only be used in combination with a timeout condition.
+
+When one creates a condition, one can calculate its fulfillment length (e.g. 96). The more complex the condition, the larger its fulfillment length will be. A BigchainDB federation can put an upper limit on the allowed fulfillment length, as a way of capping the complexity of conditions (and the computing time required to validate them).
```
diminator: When one creates a condition, one can simply extract the maximum length of the fulfillment URI by inspecting the last field of the condition URI

diminator: condition URI: `cc:<typebit>:<bitmask>:<hash>:<max_fulfillment_length>`

docs/source/topic-guides/models.md
```diff
@@ -90,13 +90,35 @@ One other note: Currently, transactions contain only the public keys of asset-ow
...
+
+If someone tries to make a condition where the output of a threshold condition feeds into the input of another “earlier” threshold condition (i.e. in a closed logical circuit), then their computer will take forever to calculate the “condition URI”, at least in theory. In practice, their computer will run out of memory or their client software will timeout after a while.
```
diminator: to calculate an "infinite condition URI"....

BigchainDB and other practical systems only accept finite condition URI's

diminator:
:+1:

docs/source/topic-guides/models.md
ttmc: While this is true, I think it's too much detail for this high-level explanation.

diminator: Yeah, I was only pointing out that the calculate can be simplified to mere inspection of the URI.
But indeed, you may omit

docs/source/topic-guides/models.md
ttmc: Okay I added the idea that the condition URI is infinite in the closed-loop case.

While it's true that "BigchainDB and other practical systems only accept finite condition URIs", a client can't construct an infinite one in the first place.

PR #395, opened 2016-06-21
Change config precedence so that env variables can override config.

### LINKED ISSUE #394: Proposal to change bigchaindb configuration system
##### Change the order of precedence of the configuration system:

The current order of precedence is:
1. default config
2. env config
3. file config

Proposed precedence order:
1. default config
2. file config
3. env config (mostly for overriding settings at runtime)

It would be helpful to change just a few settings at runtime (e.g. run several bigchaindb instances that share the same settings except for the db name and api port). Right now the easier way to do it would be to create a config file for each instance.

##### Add a method to programmatically update the configuration.

When calling bigchaindb programmatically it may be useful to change some of the configuration. Some settings can be passed to `Bigchain.__init__()` but not all the configuration can be changed this way (e.g. statds or api settings)

`config_utils` provides `set_config` that takes the default config and updtes the keys passed. It would be helpful to have a method `update_config` to take the current configuration and update some keys.

__TODO__:
- [x] change config precedence
- [x] implement update config method
- [x] update the docs
- [x] update / write tests

r-marques:
resolves #394 

Changed precedence of the config system so that env variables can override config settings.
Added `update_config` method to programmatically update the config retaining any previous changes.
Created tests

### DIFF
--- a/bigchaindb/config_utils.py
+++ b/bigchaindb/config_utils.py
@@ -183,6 +183,20 @@ def set_config(config):
     bigchaindb.config['CONFIGURED'] = True
 
 
+def update_config(config):
+    """Update bigchaindb.config with whatever is in the provided config dict,
+    and then set bigchaindb.config['CONFIGURED'] = True
+
+    Args:
+        config (dict): the config dict to read for changes
+                       to the default config
+    """
+
+    # Update the default config with whatever is in the passed config
+    update(bigchaindb.config, update_types(config, bigchaindb.config))
+    bigchaindb.config['CONFIGURED'] = True
+
+
 def write_config(config, filename=None):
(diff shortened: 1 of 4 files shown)


ttmc:
:+1:

diminator:
Ok, so the `update_config` can dynamically change the config?

vrde:
Looks good to me!
One question: what is the use case for updating the config at run time (aka `update_config`)?

r-marques:
@vrde don't remember an exact use case. @rhsimplex add a problem writing a test where he wanted to change a config parameter using `set_config` but it did not work because after that the key pair was not set.

For bigchaindb-examples it also helps because we want to programmatically spawn multiple ledgers.

Issue #401, opened 2016-06-23
Failed when I test with "The Python Driver API by Example"


guozhongxin:
I meet errors when I follow the guide of "The Python Driver API by Example" http://bigchaindb.readthedocs.io/en/latest/drivers-clients/python-driver-api-examples.html.

> \>>> from bigchaindb.client import temp_client
> \>>> c1 = temp_client()
> \>>> tx = c1.create({'value':'233333', 'msg':'test client'})
> \>>> tx
> {'transaction': {'conditions': [{'new_owners': ['Ghve9ecC9gekwWh7LHMdrTDthVcD5LUiA3uLHcsHcRdG'], 'cid': 0, 'condition': {'uri': 'cc:4:20:6VswEYH3o-jW8AkaY_-SiCEibRQGDkFdeGF9ZzvdPw8:96', 'details': {'type': 'fulfillment', 'signature': None, 'type_id': 4, 'bitmask': 32, 'public_key': 'Ghve9ecC9gekwWh7LHMdrTDthVcD5LUiA3uLHcsHcRdG'}}}], 'operation': 'CREATE', 'fulfillments': [{'fid': 0, 'input': None, 'current_owners': ['DNh9fJwqfh8rMgxi2uZ8znCra1xhEYczFfcp6HKii4zu'], 'fulfillment': 'cf:4:t9rWS08U6nMrazXOwBhZvPUGmB4HU59_rQYBM4Z370Lne68F5Ph2TE_ZFQy_bVHVpMum52JJWe3OQ8EAGd1i4eUtF6J_5_2XSs5UFbIYTWOC1n1Hg6bmRC3LnhmcM7cM'}], 'timestamp': '1466667154.303503', 'data': {'hash': '28e538ae3672b171a33286901221a5ba5856aca42127cb29ee7f50bd480765b5', 'payload': {'value': '233333', 'msg': 'test client'}}}, 'version': 1, 'assignee': 'DNh9fJwqfh8rMgxi2uZ8znCra1xhEYczFfcp6HKii4zu', 'id': 'faeda120b633bb9c828a6981c4ed2047280b993153131278e3bf41915a90a2ce'}
>
> \>>> c1.transfer(c2.public_key, tx['id'], {'value':'33333', 'msg':'test client transder'})
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/usr/local/lib/python3.4/dist-packages/bigchaindb/client.py", line 91, in transfer
[...]

PR #402, opened 2016-06-23
Updated 'The Python Driver API by Example' documentation

ttmc:
The second argument to the `transfer()` call, and all the outputs, were updated to be consistent with how things work today.

Thanks to @guozhongxin who pointed this out in Issue #401


sbellem:
:+1:

PR #403, opened 2016-06-23
Docs: Revised CLI and configuration settings docs

ttmc:
* Moved some documentation for the `bigchaindb` command (and sub-commands) from the "Configuring a BigchainDB Node" page to the page about the `bigchaindb` command.
* Renamed the "Configuring a BigchainDB Node" page to "Node Configuration Settings". That page is now mostly a list of configuration settings, in logical groupings. Each one has an explanation, default values, examples, and how to set it via environment variables or a local config file.
* The page "Node Configuration Settings" opens with a concise explanation of how each setting's value is determined.
* Removed some complex examples in favor of a bunch of small examples and snippets.

@vrde I would especially appreciate your comments and suggestions on this PR. It's not urgent though.


r-marques:
I do like a lot the changes to the configuration settings docs but I still think that we should have a subsection with all the env variables like we add before for reference

ttmc:
Okay, I will add a list of all environment variables near the top.

r-marques:
:+1:

PR #406, opened 2016-06-27
First draft of Docs on Backup

ttmc:
* A first draft of the section "Backing Up & Restoring Data"
* Moved the "Nodes, Clusters & Federations" (terminology) section from under "Topic Guides" to under "BigchainDB Clusters & Federations"
* Split the section "Set Up and Run a Federation" into separate sections under "BigchainDB Clusters & Federations", because it was going to be too long once all the documentation-to-come came. One of those new separate sections is the new backup section.


r-marques:
:+1:

ttmc:
I still want to add some notes about:

* Using RAID or similar on each node so that if one of the drives fails, the node keeps working as if nothing happened, and the failed hard drive can be replaced.
* Using RethinkDB's existing replication and server tags capabilities to create a set of dedicated "backup-only nodes" --- which would be functionally very similar to continuous backup.

so maybe this won't get merged until after the upcoming version-release.

r-marques:
:+1:

PR #409, opened 2016-06-29
Updated CHANGELOG.md for the v0.5.0 release

ttmc:
I'm increasing the version number to 0.5.0 because the change to the way configuration settings get set might change how existing setups work:  if someone is using both environment variables and a local config file, then updating to this version will change the values of their config settings.


r-marques:
:+1:

PR #410, opened 2016-06-29
Pass a custom bigchain instance to util.sign_tx

r-marques:
`util.sign_tx` requires a `bigchaindb.Bigchain` instance to be able to retrieve the details of the outputs of the previous transaction in order to create the fulfillment.

In the previous implementation `util.sign_tx` was instantiating a `Bigchain` instance with `bigchaindb.Bigchain()`. This does not work if you are signing a transaction created with a custom Bigchain instance.

Example:
```python
b = bigchaindb.Bigchain(dbname='custom_bigchain')
tx = b.create_transaction(.....)
tx_signed = b.sign_transaction(tx, sk)
```

This would not work because sign_transaction would use `b = bigchaindb.Bigchain()` and so would look in the `bigchain` database for the previous transaction by default instead of the `custom_bigchain`

To solve this `Bigchain.sign_transaction` passes `self` by default to `util.sign_tx`

### DIFF
--- a/bigchaindb/util.py
+++ b/bigchaindb/util.py
@@ -298,7 +298,7 @@ def create_tx(current_owners, new_owners, inputs, operation, payload=None):
     return transaction
 
 
-def sign_tx(transaction, signing_keys):
+def sign_tx(transaction, signing_keys, bigchain=None):
     """Sign a transaction
 
     A transaction signed with the `current_owner` corresponding private key.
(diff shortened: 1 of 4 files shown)


bigchaindb/util.py
```diff
@@ -324,10 +326,11 @@ def sign_tx(transaction, signing_keys):
 
     tx = copy.deepcopy(transaction)
 
+    bigchain = bigchain if bigchain is not None else bigchaindb.Bigchain()
```
ttmc: Would this work instead?

`bigchain = bigchain or bigchaindb.Bigchain()`

r-marques: I guess that would work in certain situations but it seems risky and not fun to debug
For instance
```python
bigchain = bigchain or bigchaindb.Bigchain()
```
is different from
```python
bigchain = bigchaindb.Bigchain() or bigchain
```

PR #411, opened 2016-06-30
Complete move from payload hashes to payload UUIDs

shauns:
- [x] Add `get_tx_by_payload_uuid`
- [x] Add tests for asset lookup functionality

~~`get_tx_by_payload_hash` still relies on this index being present (see https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/core.py#L220 )~~

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -234,10 +234,9 @@ class Bigchain(object):
             A list of transactions containing that payload. If no transaction exists with that payload it
             returns an empty list `[]`
         """
-
         cursor = r.table('bigchain') \
             .get_all(payload_uuid, index='payload_uuid') \
-            .concat_map(lambda block: block['block']['transactions']) \  # concat all transaction lists
+            .concat_map(lambda block: block['block']['transactions']) \ 
             .filter(lambda transaction: transaction['transaction']['data']['uuid'] == payload_uuid) \
             .run(b.conn)
 
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -238,7 +238,7 @@ class Bigchain(object):
             .get_all(payload_uuid, index='payload_uuid') \
             .concat_map(lambda block: block['block']['transactions']) \
             .filter(lambda transaction: transaction['transaction']['data']['uuid'] == payload_uuid) \
-            .run(b.conn)
+            .run(self.conn)
 
         transactions = list(cursor)
         return transactions
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -237,7 +237,9 @@ class Bigchain(object):
 
         cursor = r.table('bigchain') \
             .get_all(payload_uuid, index='payload_uuid') \
-            .run(self.conn)
+            .concat_map(lambda block: block['block']['transactions']) \  # concat all transaction lists
+            .filter(lambda transaction: transaction['transaction']['data']['uuid'] == payload_uuid) \
+            .run(b.conn)
 
         transactions = list(cursor)
         return transactions
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -236,7 +236,7 @@ class Bigchain(object):
         """
         cursor = r.table('bigchain') \
             .get_all(payload_uuid, index='payload_uuid') \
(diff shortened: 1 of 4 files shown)


r-marques:
hi @shauns 

It was an oversight by our part not changing `get_tx_by_payload_hash`.
Payload hash no longer exists. In #384 we changed the payload hash do a uuid. This would allows us to be able to index duplicate payloads and prevent txid collisions if several transactions are created at the same time with the same payload (or no payload).

So instead of re-adding the secorndary index for payload hash which does not exist we need to change `get_tx_by_payload_hash` to `get_tx_by_payload_uuid` and use the right secondary index in the query

shauns:
Ok - I follow. I'll update the PR if you want it?

r-marques:
Sure, that would be great. I will assign #412 to you

tests/db/test_bigchain_api.py
```diff
@@ -56,6 +56,19 @@ def test_create_transaction_payload(self, b, user_vk):
         assert len(tx['transaction']['data']['uuid']) == 36
         assert tx['transaction']['data']['payload'] == payload
 
+    def test_get_transactions_for_payload(self, b, user_vk):
+        payload = {'msg': 'Hello BigchainDB!'}
+        tx = b.create_transaction(b.me, user_vk, None, 'CREATE', payload=payload)
+        payload_uuid = tx['transaction']['data']['uuid']
+
+        matches = b.get_tx_by_payload_uuid(payload_uuid)
```
r-marques: This method will check the `bigchain` table for transactions with that matching payload. At this point you created the transactions but the transaction was not written to the the database. To do so you need to create and write a block.

```python
block = b.create_block([tx])
```

shauns: Thanks for pointer, figured it would be something along these lines.

bigchaindb/core.py
```diff
@@ -217,26 +217,26 @@ def get_blocks_status_containing_tx(self, txid):
...
+            payload_uuid (str): the UUID for this particular payload.
 
         Returns:
             A list of transactions containing that payload. If no transaction exists with that payload it
             returns an empty list `[]`
         """
 
         cursor = r.table('bigchain') \
-            .get_all(payload_hash, index='payload_hash') \
+            .get_all(payload_uuid, index='payload_uuid') \
```
r-marques: Ok, we have another problem here. The secondary index will return all matching documents in the table and the documents in the bigchain table are blocks not transactions. So this will return the block where the matching transaction exists.

We need to improve the query a bit to return only the matching transactions:
1. Once we get the matching block(s) we need to flatten all the transactions on those blocks
2. Filter the transactions based on the criteria

```python
cursor = r.table('bigchain') \
               .get_all(payload_uuid, index='payload_uuid') \
               .concat_map(lambda block: block['block']['transactions']) \  # concat all transaction lists
               .filter(lambda transaction: transaction['transaction']['data']['uuid'] == payload_uuid) \
               .run(b.conn)
```

r-marques:
Looks good, thanks.

I guess its ready to merge

ttmc:
Hi @shauns 

Before we can merge this pull request, which may contain your intellectual property in the form of copyright or patents, our lawyers say we need you or your organization to agree to one of our contributor agreements. If you are contributing on behalf of yourself (and not on behalf of your employer or another organization you are part of) then you should:

1. Go to: https://www.bigchaindb.com/cla/
2. Read the Individual Contributor Agreement
3. Fill in the form "For Individuals"
4. Check the box to agree
5. Click the SEND button

If you're contributing as an employee, and/or you want all employees of your employing organization to be covered by our contributor agreement, then someone in your organization with the authority to enter agreements on behalf of all employees must do the following:

1. Go to: https://www.bigchaindb.com/cla/
2. Read the Entity Contributor Agreement
3. Fill in the form "For Organizations”
4. Check the box to agree
5. Click the SEND button

We will then email you (or your employer) with further instructions.

shauns:

sUdeBAfxvFP3T8GOh55dGZ5WdIWiOg6vJDotd1MWNVWQIeCSXw3taSELqnnNqpY

ttmc:
:+1:

r-marques:
resolves #412

Issue #348, opened 2016-05-31
Rename current_owners and new_owners field in the transaction


diminator:
A transaction now contains `current_owners` in the condition field and `new_owners` in the fulfillment field. 

However, this is ambiguous when reasoning about `bigchaindb.get_owned_ids()`:
- The resulting transactions include these where one has been included in the condition, hence the `new_owners` field
- There might be co-ownership

Solution: rename the `current_owners` and `new_owners` field into `source` and `destination`

ttmc:
I'm not sure `source` and `destination` are an improvement. They obscure the fact that they are lists, and that they are lists of owners (i.e. public keys of owners).

The purpose of a transaction is, in part, to change the owners of one or more assets. Each asset has a fulfillment-condition pair, with the "owners just before this transaction happened" listed in the fulfillment, and the "owners just after this transaction happened" listed in the condition.

It would be nice if we could just call the two lists `owners_just_before_this_transaction_happened` and `owners_just_after_this_transaction_happened` but those names are clearly too long.

How about `owners_before` and `owners_after` (where "this transaction" is implicit)?

Those names aren't very different from what we have now (`current_owners` and `new_owners`). The problem with the current names is that you have to take a very particular point of view to reason about them properly (namely the view of the world _before_ the transaction happened), and that's not obvious... especially for someone trying to understand the transaction in the future.

PR #415, opened 2016-07-05
Docs fix: an asset can't have zero owners

ttmc:
The **Assets** section under **Topic Guides** used to say "An asset can have zero, one, or several owners." but that is wrong. An asset can't have zero owners. This was pointed out by @MiracleB on Gitter. I fixed it.

What I wrote on Gitter, when @MiracleB asked about the zero owners case:

> When I wrote that, I was thinking about the case of a simple hashlock condition --- anyone who knows (or guesses) the secret preimage can transfer the asset. The current owner(s) don't have to do anything to create the transfer (fulfillment). If the current owner(s) lose the secret preimage, then it becomes effectively impossible to transfer the asset, but someone still owns the asset. I guess there must always be at least one asset owner.

### DIFF
--- a/docs/source/topic-guides/assets.md
+++ b/docs/source/topic-guides/assets.md
@@ -4,7 +4,7 @@ BigchainDB can store data of any kind (within reason), but it's designed to be p
 
 * The fundamental thing that one submits to a BigchainDB federation to be checked and stored (if valid) is a _transaction_, and there are two kinds: creation transactions and transfer transactions.
 * A creation transaction can be use to register any kind of indivisible asset, along with arbitrary metadata.
-* An asset can have zero, one, or several owners.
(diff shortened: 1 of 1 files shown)


ttmc:
I was chatting with @r-marques and he noted that it is possible for an asset to have zero owners. Doh! Here's how: The owners would create a transfer transaction fulfilling the asset's current conditions, and creating a new simple hashlock condition in which the list of owners is empty.
Anyone with the secret preimage could use it to claim the asset for themselves.

The language around "current owners" and "new owners" can be very confusing. There's another issue #348 to address that. The general rule is: to determine who owns an asset now, trace it to the last transfer transaction involving that asset, and then look in the _conditions_.

I will close this PR and not merge it.

Issue #331, opened 2016-05-25
Deployment & Operations issues


ttmc:
Tools and documentation to help with deployment and day-to-day operations of BigchainDB nodes and clusters/federations.

- [x] Initial documentation on federations - Issue #388 
- [x] New and revised timestamp docs (including notes on NTP daemon setup) - Issue #397
- [x] Initial docs on RethinkDB storage setup - Issue #416 
- [x] Initial docs on firewall setup - Issue #417 
- [x] Ability to deploy a cluster on AWS using EBS-only instances - Issue #461 and Pull Request #469 

Trent said this Project Issue was too general so it was split apart into more specific ones:

* Secure management of private keys (on nodes) - Issue #189
* Cluster Configuration Management system - Issue #436
* Production Node Deployment & Management - Issue #498
* More...

**Historical Notes**

This Project-Issue started out mixed up with benchmarking on the physical planning board (Project Minky, issue #246 and Project Norbert, issue #321). It got separated out into its own issue (this issue), which originally was just about _documenting_ how to set up and run a federation. Over time, it became clear that there were missing deployment tools, so creating those also became part of this issue. It's now a real "Project-Issue," linking to many sub-issues.

This issue also replaces issue #126 (Define how a federation is managed).

ttmc:
This project includes issue #126 ("Define how a federation is managed").

PR #417, opened 2016-07-05
Docs on firewall setup

ttmc:
I added a list of the default ports where inbound traffic is expected, for a BigchainDB node. This is to help with firewall setup.

This is part of resolving issue #331.


docs/source/nodes/setup-run-node.md
```diff
@@ -10,13 +10,18 @@ The first step is to get a server (or equivalent) which meets [the requirements
...
 
-The steps that you must take to secure your server depend on your server OS and where your server is physically located. There are many articles and books about how to secure a server. Ask a search engine.
+The steps that you must take to secure your server depend on your server OS and where your server is physically located. There are many articles and books about how to secure a server. Here we just cover special considerations when securing a BigchainDB node.
 
-Here we just cover special considerations when securing a BigchainDB node.
+When you set up the firewall, here is a list of the ports where inbound traffic is expected:
 
-TODO: Special security considerations for BigchainDB nodes.
+22 - The default SSH port (TCP) <br>
+123 - The default NTP port (**UDP**) <br>
```
vrde: Can you double check if NTP needs the port open for inbound traffic?

ttmc: Good question... I looked into it. Indeed, port 123 doesn't need to be open to all inbound traffic. A firewall rule to allow "established and related" inbound traffic is sufficient (e.g. allow responses to requests sent to external NTP servers). I updated the documentation accordingly (now moved to a new page in the Appendices).

docs/source/appendices/firewall-notes.md
```diff
@@ -0,0 +1,23 @@
+# Notes on Firewall Setup
+
+When you set up the firewall (or security group on AWS) for a BigchainD node, here is a list of the ports where _unsolicited_ inbound traffic is expected:
+
+22 - The default SSH port (TCP) <br>
+8080 - The default port for the RethinkDB web interface (TCP) <br>
```
r-marques: port `8080` and `28015` should probably be closed to the outside. The web ui should only be accessed by the admin of the node and the client driver port should only be accessed by the bigchaindb layer which is running in the same node

ttmc: Good points. I'll add some notes along those lines.

docs/source/appendices/firewall-notes.md
```diff
@@ -0,0 +1,60 @@
...
+
+This page summarizes the ports used by BigchainDB and the traffic they should accept or reject, e.g. using a firewall or AWS security group.
+
+
+## Inbound at Port 22
+
+Port 22 is the default SSH port (TCP). It should expect unsolicited inbound traffic from arbitrary IP addresses.
+
+
+## Inbound at Port 123
```
vrde: We don't need a rule to allow `NEW` packets going through `iptables`. I'd move this section to **Outbound Traffic**, saying that if your firewall policy is to explicitly drop packets going through the `OUTPUT` chain, then you must remember to enable the port `123`.

docs/source/appendices/firewall-notes.md
```diff
@@ -0,0 +1,60 @@
...
+
+Port 29015 is the default port for RethinkDB intracluster connections (TCP). It should only accept incoming traffic from other RethinkDB servers in the cluster (a list of IP addresses that you should be able to find out).
+
+
+## Inbound at Other Ports
+
+Other ports you might need to consider include:
+
+53 - The default DNS port (UDP) <br>
+161 - The default SNMP port (usually UDP, sometimes TCP)
```
vrde: I'd include those ports in the **outbound traffic** section as well.

docs/source/appendices/firewall-notes.md
```diff
@@ -1,23 +1,72 @@
...
+
+If Gunicorn and the reverse proxy are running on the same server, then you'll have to tell Gunicorn to listen on some port other than 9984 (so that the reverse proxy can listen on port 9984). You can do that by setting `server.bind` to 'localhost:PORT' in the [BigchainDB Configuration Settings](../nodes/configuration.html), where PORT is whatever port you chose (e.g. 9983).
+
+You may want to have Gunicorn and the reverse proxy running on different servers, so that both can listen on port 9984. That would also help isolate the effects of a denial-of-service attack.
+
+
+## Port 28015
+
+Port 28015 is the default port used by RethinkDB client driver connections (TCP). If your BigchainDB node is just one server, then Port 28015 only needs to listen on localhost, because all the client drivers will be running on localhost. Port 28015 doesn't need to accept inbound traffic from the outside world.
```
r-marques: > If your BigchainDB node is just one server, then Port 28015 only needs to listen on localhost

Don't understand this sentence. Assuming that a bigchaindb instance will be running on the same machine as rethinkdb this port only needs to be available on the `localhost` interface, regardless of the number of nodes in the federation

ttmc: I'm just trying to keep the documentation future-proof, because in the future, it's possible that a BigchainDB node may consist of several servers running different processes. For example, one server might be running RethinkDB and another might be doing validation only, but validation requires making queries to RethinkDB (over port 28015).

r-marques:
:+1:

Issue #283, opened 2016-05-13
Support snapshots / archival backup


trentmc:
# Abstract
Modern database management systems (DBMSs) include a means to periodically back up ("snapshot") the actual data state as well as the log (which can be replayed to recover the state). This typically goes to magnetic drives or other offline storage. Leverage the tech & infrastructure of other DBMSs to support snapshots / archival.

Also to consider: increase the replication factor of the tables. Replicas are like "live backups" (of a shard).

# Motivation
If primary and all replicas of a shard are deleted, then data is lost. There's a similar problem with _changes_ to all nodes associated with a shard.

Backups help with security, and feelings of security. Will probably help to solve some of #284. Customers ask about it.

# Challenges
If there is no currently-available and _scalable_ RethinkDB backup option, then developing one might be challenging.

# Tasks

* [x] Document existing backup options offered by RethinkDB and third parties: PR #406 (Troy)
* [x] If necessary, design, develop and document additional backup options

ttmc:
I just finished writing the documentation about existing backup options and I think those options are probably good enough for now. I'm going to close this issue. If we hear from users that they want more backup functionality, then we can open a new issue.

PR #418, opened 2016-07-05
Docs: Added two concepts to backup strategies

ttmc:
I added two concepts to the section on backup strategies:

1. Replication isn't backup; you still need to consider having a normal, "cold" backup.
2. Added the idea of disconnecting the `backup` set, waiting for it to settle, and snapshotting all its file systems.

Helps with issue #283

### DIFF
--- a/docs/source/clusters-feds/backup.md
+++ b/docs/source/clusters-feds/backup.md
@@ -73,9 +73,11 @@ Yes, in principle, but it would be difficult to know if you've recovered every b
 
 ## Backup by Copying RethinkDB Data Files
 
-It's _possible_ to back up a BigchainDB database by creating a point-in-time copy (or "snapshot") of the RethinkDB data files (on all nodes, at roughly the same time). It's not a very practical approach to backup: the resulting set of files will be much larger (collectively) than what one would get using `rethinkdb dump`, and there are no guarantees on how consistent that data will be, especially for recently-written data.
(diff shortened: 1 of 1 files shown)


docs/source/clusters-feds/backup.md
```diff
@@ -102,3 +113,11 @@ Although it's not advertised as such, RethinkDB's built-in replication feature i
 * Send a RethinkDB reconfigure command to the RethinkDB cluster to make it so that the `original` set has the same number of replicas as before (or maybe one less), and the `backup` set has one replica. Also, make sure the `primary_replica_tag='original'` so that all primary shards live on the `original` nodes.
 
 The [RethinkDB documentation on sharding and replication](https://www.rethinkdb.com/docs/sharding-and-replication/) has the details of how to set server tags and do RethinkDB reconfiguration.
+
+Once you've set up a set of backup-only RethinkDB servers, you could make a point-in-time snapshot of their storage devices, as a form of backup.
+
+You might want to disconnect the `backup` set from the `original` set first, and then wait for reads and writes in the `backup` set to stop. That way, all the data in the `backup` set will be consistent before you take the snapshot.
```
r-marques: I guess this would have the same problem as above. There are no guarantees that all the backup nodes will have the same data

ttmc: It seems to me that the `backup` set would have just one copy of each shard (i.e. but no primary replicas). I don't see how there could be inconsistency between the servers of the `backup` set.

The servers in the `backup` set could be made inaccessible from clients (i.e. they only communicate with other RethinkDB servers, never with the outside world). If you disconnect the `backup` set from the `original` set, it becomes isolated, with no reads or writes coming from outside. It becomes a static island of data, so you can take a snapshot.

I should add these ideas to the text, assuming I didn't make a mistake in my reasoning. Did I?

r-marques:
:+1:

PR #420, opened 2016-07-06
Docs: Amazon Linux instances already have NTP set up

ttmc:
This is a small pull request. I just added a small subsection (in "Notes on NTP Daemon Setup") to note that Amazon Linux instances already have an NTP daemon installed, configured, and running.


r-marques:
:+1:

PR #422, opened 2016-07-08
Docs: Moved AWS Setup notes to the Appendices

ttmc:
Because the "AWS Setup" instructions are the same (more or less) regardless of whether you want to deploy a single node or a test cluster, I moved the "AWS Setup" notes from the "Deploy a Cluster on AWS" page to a new page in the Appendices. I also made some minor changes to note the differences when deploying a single node.


r-marques:
:+1:

PR #423, opened 2016-07-08
Docs: rm page re local multi-node rethinkdb cluster

ttmc:
We don't really need the documentation page "Deploying a Local Multi-Node RethinkDB Cluster" because 1) it's an unusual setup and 2) all the information on that page can already be found elsewhere online.


r-marques:
:+1:

PR #425, opened 2016-07-11
updated documentation about the configuration

### LINKED ISSUE #272: Allow environment variables to specify location of config
It'd be nice to also let users use an environment variable to specify their configuration locations (for `bigchaindb configure`, `bigchaindb init`, `bigchaindb start`, and any others?) so that they don't need to keep specifying it in every command. Especially useful when using docker.

r-marques:
resolves #272

The environment variable `BIGCHAINDB_CONFIG_PATH` was not documented. It allows us to specify the path of the config file using an environment variable


ttmc:
:+1:

@r-marques I will give you the satisfaction of merging it.

PR #428, opened 2016-07-11
document --experimental-start-rethinkdb

### LINKED ISSUE #206: Automatically start rethinkdb if not running
Looking at the [horizon.io](http://horizon.io/) project one of the things I like was that when running `hz serve --dev` it automatically starts a rethinkdb instance in the current folder if one is not running.

It would be nice to have something similar for bigchaindb. I don't think its a good idea to do it by default because the user may already have an existing rethinkdb database and maybe he just forgot to start rethinkdb. But maybe add a flag to `bigchaindb start` to automatically start a rethinkdb instance if none is running.

r-marques:
resolves #206


ttmc:
:+1:

PR #430, opened 2016-07-12
Docs: Notes on mutable assets & determining asset owners

ttmc:
Based on some questions from @MiracleB on Gitter, I added some more documentation about assets:

* How to determine the current owner(s) of an asset.
* Mutable and immutable assets, and how to change a mutable asset.

### DIFF
--- a/docs/source/topic-guides/models.md
+++ b/docs/source/topic-guides/models.md
@@ -33,6 +33,11 @@ Every create transaction contains exactly one fulfillment-condition pair. A tran
 
 ![Tracking the stories of three assets](../_static/stories_3_assets.png)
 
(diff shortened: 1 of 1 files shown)


sbellem:
:+1:

docs/source/topic-guides/models.md
```diff
@@ -33,6 +33,11 @@ Every create transaction contains exactly one fulfillment-condition pair. A tran
 
 ![Tracking the stories of three assets](../_static/stories_3_assets.png)
 
+To determine the current owner(s) of an asset, find the most recent valid transaction involving it, and then look at the list of owners in the _conditions_ (not the fulfillments).
```
r-marques: `Bigchain.get_owned_ids` also returns all the current unspent outputs

ttmc: While that's true, it answers a different question: What assets have this public key listed as an owner (or co-owner) right now?

PR #431, opened 2016-07-13
Docs: more notes on RAM requirements

ttmc:
* Increasing the replication factor increases the required RAM (for associated metadata)
* What happens if a RethinkDB process runs out of RAM


r-marques:
:+1:

PR #434, opened 2016-07-13
Docs: removed page about HTTP API

### LINKED ISSUE #232: Update or replace apiary documentation
relates to #230 

The apiary documentation is out of date. The keys used in the examples are keys generate by the old `ecdsa` signing algorithm instead of `ed25519`.

The apiary documentation is a bit useless because the api endpoints do not call any running server. We should decide if we want to keep using apiary or move the api documentation to readthedocs

ttmc:
resolves #232

### DIFF
--- a/docs/source/drivers-clients/index.rst
+++ b/docs/source/drivers-clients/index.rst
@@ -7,7 +7,6 @@ BigchainDB Drivers & Clients
 .. toctree::
    :maxdepth: 1
 
-   http-client-server-api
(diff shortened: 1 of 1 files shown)


r-marques:
The api is still there and the documentation must reflect that. We can just remove the link to apiary and improve a bit the the documentation in readthedocs

ttmc:
Oh, I misunderstood. I'll close this pull request and let a different one resolve #232

PR #440, opened 2016-07-14
remove block update

rhsimplex:
old logic for handling updates to blocks, no longer needed now that votes are a separate table. this should have been part of #379 and makes sure #221 still works correctly


r-marques:
:+1:

rhsimplex:
@r-marques actually this was a little more involved than I thought. Thanks @vrde for catching this.

r-marques:
good catch

vrde:
Closing since we moved to _pipelines_.

PR #442, opened 2016-07-14
Moved the HTTP API docs from Apiary to /docs

ttmc:
I moved the documentation of the client-server HTTP API from Apiary to our source tree (in `/docs`, i.e. our documentation using Sphinx). I used the [sphinxcontrib-httpdomain extension](https://pythonhosted.org/sphinxcontrib-httpdomain/#).

The HTTP API seems a bit broken right now, so the example GET request is made up and I wasn't able to figure out a working new-transaction POST. Those are left as TODO items. When someone fixes/improves the current HTTP API, they can fix and augment these docs accordingly.


r-marques:
It renders nicely :+1:

PR #451, opened 2016-07-14
Remove block_number

vrde:
`block_number` before #379 was used in few queries to order blocks and to retrieve the Genesis block (the one with `block_number = 0`).

After #379 `block_number` has been moved to the `votes` table, losing part of it's value.

This PR is to remove `block_number` from the code–base and replace it with a `timestamp`–based ordering. In case more votes are cast during the same `timestamp` (the resolution is one second) then the attributes `previous_block` and `voting_for_block` are used to determine the order of the votes.

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -584,9 +584,6 @@ class Bigchain(object):
 
         res = r.table('bigchain').get(last_voted[0]['vote']['voting_for_block']).run(self.conn)
 
-        if 'block_number' in last_voted[0]:
-            res['block_number'] = last_voted[0]['block_number']
-
         return res
 
     def get_unvoted_blocks(self):
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -598,9 +598,11 @@ class Bigchain(object):
             .order_by(r.desc(r.row['block']['timestamp'])) \
             .run(self.conn)
 
+        # FIXME: I (@vrde) don't like this solution. Filtering should be done at a
+        #        database level. Solving issue #444 can help untangling the situation
         unvoted = filter(lambda block: not util.is_genesis_block(block), unvoted)
 
-        return unvoted
+        return list(unvoted)
 
     def block_election_status(self, block):
         """Tally the votes on a block, and return the status: valid, invalid, or undecided."""
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -577,11 +577,11 @@ class Bigchain(object):
             # get the latest value for the vote timestamp (over all votes)
             max_timestamp = r.table('votes') \
                 .filter(r.row['node_pubkey'] == self.me) \
-                .max(lambda x: x['vote']['timestamp']) \
+                .max(r.row['vote']['timestamp']) \
                 .run(self.conn)['vote']['timestamp']
 
             last_voted = list(r.table('votes') \
-                .filter(lambda x: x['vote']['timestamp'] == max_timestamp) \
+                .filter(r.row['vote']['timestamp'] == max_timestamp) \
                 .filter(r.row['node_pubkey'] == self.me) \
                 .run(self.conn))
 
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -598,8 +598,7 @@ class Bigchain(object):
             .order_by(r.desc(r.row['block']['timestamp'])) \
             .run(self.conn)
 
-        if unvoted and util.is_genesis_block(unvoted[0]):
-            unvoted.pop(0)
+        unvoted = filter(lambda block: not util.is_genesis_block(block), unvoted)
 
         return unvoted
 
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -492,6 +492,16 @@ class Bigchain(object):
         response = r.table('bigchain').get_all(transaction_id, index='transaction_id').run(self.conn)
         return True if len(response.items) > 0 else False
 
+    def prepare_genesis_block(self):
+        """Prepare a genesis block."""
+
+        payload = {'message': 'Hello World from the BigchainDB'}
+        transaction = self.create_transaction([self.me], [self.me], None, 'GENESIS', payload=payload)
+        transaction_signed = self.sign_transaction(transaction, self.me_private)
+
+        # create the block
+        return self.create_block([transaction_signed])
+
     # TODO: Unless we prescribe the signature of create_transaction, this will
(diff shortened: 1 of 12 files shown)


bigchaindb/core.py
```diff
@@ -553,41 +556,47 @@ def vote(self, block, previous_block_id, decision, invalid_reason=None):
...
-
         r.table('votes') \
             .insert(vote) \
             .run(self.conn)
 
     def get_last_voted_block(self):
         """Returns the last block that this node voted on."""
 
+
         last_voted = r.table('votes') \
```
r-marques: This is returning every vote that this node has casted. At some point we may want to optimize this

bigchaindb/core.py
```diff
@@ -597,13 +606,14 @@ def get_unvoted_blocks(self):
         unvoted = r.table('bigchain') \
             .filter(lambda block: r.table('votes').get_all([block['id'], self.me], index='block_and_voter')
                     .is_empty()) \
-            .order_by(r.desc('block_number')) \
+            .order_by(r.desc(r.row['block']['timestamp'])) \
```
r-marques: We should probably use `r.asc` to retrieve a list starting from the oldest block. This way the oldest blocks are voted first.

tests/db/conftest.py
```diff
@@ -108,13 +108,13 @@ def inputs(user_vk, amount=1, b=None):
...
-    block = b.create_block(transactions)
-    b.write_block(block, durability='hard')
-    return block
+    for block in range(blocks):
+        transactions = []
+        for i in range(amount):
+            tx = b.create_transaction(b.me, user_vk, None, 'CREATE')
+            tx_signed = b.sign_transaction(tx, b.me_private)
+            transactions.append(tx_signed)
+            b.write_transaction(tx_signed)
```
r-marques: This may create problems. We are writing transactions to the backlog and then we are already creating blocks with the same transactions.

bigchaindb/core.py
rhsimplex: Easy optimzation: [use the ReQL `min` expression](https://rethinkdb.com/api/python/min/) after filtering for node and block ID. Then you will get the minimum timestamp associated with that filter, and you can just pull those.

`r.expr(['a', 'a', 'b', 'c']).min().run(r.connect())` returns `'a'`

r-marques: The problem is we want to return all newest timestamps so that we can then ordering following the chain. In your example with min we would want it to return `['a', 'a']`

tests/conftest.py
```diff
@@ -8,6 +8,7 @@
 
 import os
 import copy
+from functools import partial
```
sbellem: Is this needed?

tests/db/conftest.py
```diff
@@ -98,7 +98,7 @@ def fin():
 
 
 @pytest.fixture
-def inputs(user_vk, amount=1, b=None):
+def inputs(user_vk, amount=4, b=None, blocks=4):
```
sbellem: Just a note regarding the passing of arguments to a pytest fixtures. It is not really possible. At least not in the way that one would expect it to be with a normal function. 

There are at least two alternatives, both relying on the usage of the [`request`](http://docs.pytest.org/en/latest/builtin.html#_pytest.python.FixtureRequest) object:

* [Apply indirect on particular arguments](http://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments)
* [Fixtures can introspect the requesting test context](http://docs.pytest.org/en/latest/fixture.html#fixtures-can-introspect-the-requesting-test-context)



stackoverflow post: [py.test: Pass a parameter to a fixture function](http://stackoverflow.com/questions/18011902/py-test-pass-a-parameter-to-a-fixture-function)

tests/db/conftest.py
vrde: Oh, right, I was reproducing the previous behavior and I didn't notice that.

tests/conftest.py
vrde: :+1:

tests/db/conftest.py
vrde: Good catch!
I think the parameters were there because the fixture before was a function.

bigchaindb/core.py
```diff
@@ -553,41 +556,54 @@ def vote(self, block, previous_block_id, decision, invalid_reason=None):
...
-            .limit(1) \
-            .run(self.conn)
-
-        # return last vote if last vote exists else return Genesis block
-        if not last_voted:
+        try:
+            # get the latest value for the vote timestamp (over all votes)
+            max_timestamp = r.table('votes') \
+                .filter(r.row['node_pubkey'] == self.me) \
+                .concat_map(lambda x: [x['vote']['timestamp']]) \
```
r-marques: Is the concat_map operation really required? `.max(r.row['vote']['timestamp'])` should be enough, no?

bigchaindb/core.py
```diff
@@ -553,41 +556,54 @@ def vote(self, block, previous_block_id, decision, invalid_reason=None):
...
+                .max() \
+                .run(self.conn)
+
+            last_voted = list(r.table('votes') \
```
r-marques: There should be less results for a given `timestamp` then for a `node_pubkey`. It is probably more efficient to first filter by `timestamp` and then by `node_pubkey`

Also `r.row['vote']['timestamp']` and `lambda x: x['vote']['timestamp']` seem to be equivalent, maybe we should stick to one notation.

bigchaindb/core.py
```diff
@@ -553,41 +556,54 @@ def vote(self, block, previous_block_id, decision, invalid_reason=None):
...
+            # return last vote if last vote exists else return Genesis block
             return list(r.table('bigchain')
-                        .filter(r.row['block_number'] == 0)
+                        .filter(util.is_genesis_block)
                         .run(self.conn))[0]
 
-        res = r.table('bigchain').get(last_voted[0]['vote']['voting_for_block']).run(self.conn)
+        # since the resolution of timestamp is a second,
+        # we might have more than one vote per timestamp
+        if len(last_voted) > 1:
```
r-marques: the inner algorithm seems to work fine if the length `last_voted` is 1 so we may just remove the `if else` statement

vrde: :+1:

ttmc:
@rhsimplex You asked me to remind you to remove the "block_number" from the vote model in the documentation:  at the end of the file /docs/source/topic-guides/models.md

r-marques:
once @vrde pushes his changes :+1:

PR #468, opened 2016-07-19
Fix and remove duplicated tests for Voter

vrde:
Tests for the voter were moved in `db/test_voter.py`. This PR is to remove duplicated tests.


sbellem:
:+1:

r-marques:
:+1:

tests/db/test_bigchain_api.py
```diff
@@ -609,119 +609,6 @@ def test_invalid_node_pubkey(self, b):
             b.validate_block(block)
 
 
-class TestBigchainVoter(object):
```
rhsimplex: Just for completeness' sake, can you point out which tests these duplicate?

PR #469, opened 2016-07-19
Added ability to deploy test cluster on AWS using EBS for storage

### LINKED ISSUE #461: Ability to deploy a cluster on AWS using EBS-only instances
Some [AWS instance types](https://aws.amazon.com/ec2/instance-types/) don't come with any "instance store" (included storage); they're "EBS-only." Examples are the T2 and M4 families. The goal of this issue is to modify the AWS deployment scripts to it becomes possible to deploy a cluster using EBS-only instance types. There will probably be one or more new configuration settings, such as how much EBS storage to attach.

This is a sub-issue under Issue #331

ttmc:
Resolves #461 
(Ability to deploy a cluster on AWS using EBS-only instances)

* Three new AWS deployment config file settings:
    * `USING_EBS`
    * `EBS_VOLUME_SIZE`
    * `EBS_OPTIMIZED`
* Explanations and example values are provided in `example_deploy_conf.py`
* Their values are displayed on stdout when deploying
* Their values are checked for type and size (where applicable) when deploying
* Their values are used to determine what gets deployed on AWS!

Also:
* Broke apart the "install_rethinkdb" Fabric command into a bunch of separate commands like "prep_rethinkdb_storage" and "configure_rethinkdb"
* Updated the associated documentation
* Added a new Fabric command: set_replicas. It's not really related but it was trivial so I just added it in this PR


deploy-cluster-aws/fabfile.py
```diff
@@ -97,45 +97,94 @@ def upgrade_setuptools():
...
+    # Convert USING_EBS from a string to a bool
+    USING_EBS = (USING_EBS.lower() == 'true')
+
+    # Make the /data directory for RethinkDB data
+    sudo("mkdir -p /data")
+
+    # OLD: with settings(warn_only=True):
+    if USING_EBS:  # on /dev/xvdp
+        # See https://tinyurl.com/h2nut68
+        sudo("mkfs -t ext4 /dev/xvdp")
```
r-marques: The EBS volume device will always be `xvdp`?

deploy-cluster-aws/example_deploy_conf.py
```diff
@@ -49,3 +49,19 @@
...
+# If USING_EBS=False, EBS_VOLUME_SIZE is irrelevant and not used
+EBS_VOLUME_SIZE=30
+
+# EBS_OPTIMIZED is True or False, depending on whether you want
+# EBS-optimized instances. See:
+# https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
+# Not all instance types support EBS optimization.
+# Setting EBS_OPTIMIZED=True may cost more, but not always.
+# If USING_EBS=False, EBS_OPTIMIZED is irrelevant and not used
+EBS_OPTIMIZED=False
```
r-marques: What happens if I set EBS_OPTIMIZED to `true` and select an INSTANCE_TYPE that does not support it? will it just be ignored?

deploy-cluster-aws/fabfile.py
ttmc: Yeah, when the instance is launched by `launch_ec2_nodes.py`, an associated EBS volume is also launched, and I set it to have the "p" device name (on the line `'DeviceName': '/dev/sdp',`).

From the comment:  Why /dev/sdp? See https://tinyurl.com/z2zqm6n

deploy-cluster-aws/example_deploy_conf.py
ttmc: Yes, it will be ignored.

deploy-cluster-aws/fabfile.py
r-marques: I didn't know that you could define the device names. I though the OS did that

r-marques:
:+1:

deploy-cluster-aws/fabfile.py
ttmc: I tried a few different device names and then used ssh to login to the instance and see what the device names were (using `lsblk`). I also checked on the AWS EC2 web console. The EBS volume always had the name I'd assigned. That might not always work for all operating systems or even all Linux distros, but apparently it does work for the version of Ubuntu that we use by default for AWS deployments.

PR #471, opened 2016-07-20
Split db.utils.init into multiple functions - Issue #458

### LINKED ISSUE #458: Extract multiple functionalities of db.utils.init into functions to ease its reusability
This will help to address issue #249

Assignee: @lluminita

lluminita:
Resolves #458 
The functionality of the db.utils.init was extracted into separate functions. New tests were added into tests.db.test_utils.

### DIFF
--- a/bigchaindb/db/utils.py
+++ b/bigchaindb/db/utils.py
@@ -26,19 +26,18 @@ def create_database(conn, dbname):
     if r.db_list().contains(dbname).run(conn):
         raise exceptions.DatabaseAlreadyExists('Database `{}` already exists'.format(dbname))
 
-    logger.info('Create:')
-    logger.info(' - database `%s`', dbname)
+    logger.info('Create database `%s`.', dbname)
     r.db_create(dbname).run(conn)
 
 
 def create_table(conn, dbname, table_name):
-    logger.info(' - %s table', table_name)
+    logger.info('Create `%s` table.', table_name)
     # create the table
     r.db(dbname).table_create(table_name).run(conn)
 
 
 def create_bigchain_secondary_index(conn, dbname):
-    logger.info(' - bigchain indexes')
+    logger.info('Create `bigchain` secondary index.')
     # to order blocks by timestamp
     r.db(dbname).table('bigchain')\
         .index_create('block_timestamp', r.row['block']['timestamp'])\
@@ -59,7 +58,7 @@ def create_bigchain_secondary_index(conn, dbname):
 
 
 def create_backlog_secondary_index(conn, dbname):
-    logger.info(' - backlog indexes')
+    logger.info('Create `backlog` secondary index.')
     # to order transactions by timestamp
     r.db(dbname).table('backlog')\
         .index_create('transaction_timestamp',
@@ -76,7 +75,7 @@ def create_backlog_secondary_index(conn, dbname):
 
 
 def create_votes_secondary_index(conn, dbname):
-    logger.info(' - indexes')
+    logger.info('Create `votes` secondary index.')
     # compound index to order votes by block id and node
     r.db(dbname).table('votes')\
         .index_create('block_and_voter',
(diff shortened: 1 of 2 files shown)


bigchaindb/db/utils.py
```diff
@@ -18,54 +18,91 @@ def get_conn():
...
-    r.db(dbname).table('backlog').index_wait().run(conn)
-    r.db(dbname).table('bigchain').index_wait().run(conn)
     r.db(dbname).table('votes').index_wait().run(conn)
 
+
+def init():
+    # Try to access the keypair, throws an exception if it does not exist
+    b = bigchaindb.Bigchain()
+
+    conn = get_conn()
```
r-marques: I don't remember why we created `get_conn()`. It his the same as `b.conn`.
Since we are refactoring this maybe we can get rid off `get_conn`

@vrde thoughts?

vrde: `get_conn` is a generic function that returns a connection to the DB using the data in the configuration. I created it because sometimes you need to just get a connection to the database without using a `Bigchain` instance.

I think the plan was to have the method `core.py::Bigchain::reconnect` use the `get_conn` function. I'd prefer to see `get_conn` used by `reconnect` instead of weed out `get_conn`.

bigchaindb/db/utils.py
```diff
@@ -18,54 +18,91 @@ def get_conn():
...
+    return bigchaindb.config['database']['name']
 
-    conn = get_conn()
-    dbname = bigchaindb.config['database']['name']
 
+def create_database(conn, dbname):
     if r.db_list().contains(dbname).run(conn):
         raise exceptions.DatabaseAlreadyExists('Database `{}` already exists'.format(dbname))
 
     logger.info('Create:')
```
vrde: @sbellem, @r-marques: since we are refactoring, I'm not sure the usage of `log` here is the best. There is no "best practice" for that, but my gut tells me that we are not logging single messages, but a group of messages that together makes sense, separated no. A log message should be about an event, and should make sense by itself, while here we have sparse information, so instead of:
```
 - database `%s`
```
something like:
```
Create database `%s`
```
is better IMO.

What do you think?

lluminita: Yeah initially it was logging a group of messages. I wasn't sure if I should move the `log` inside the function or leave it in the `init` before the function call, as a group of messages.
I noticed now that also in the table creation and creation of the indexes functions (like: [here](https://github.com/bigchaindb/bigchaindb/pull/471/files#diff-31c0e86720248cb39d3f98060112153fR35) and [here](https://github.com/bigchaindb/bigchaindb/pull/471/files#diff-31c0e86720248cb39d3f98060112153fR62) ) it should be something like `Create %s table` or `Create index`, if we leave the `log` in the function definition.

r-marques: agree

bigchaindb/db/utils.py
r-marques: Yes they are actually different, `reconnect` specifies the database to use while `get_conn` doesn't which makes sense since there is no database before the `init`

lluminita:
@vrde and @r-marques I fixed the logger messages. please have another look. thanks

bigchaindb/db/utils.py
vrde: we can still change the signature of `get_conn`. If we can keep all DB related functions in the `bigchaindb.db` module would be nice IMO. Not a huge priority tho.

vrde:
> @vrde and @r-marques I fixed the logger messages. please have another look. thanks

looks good to me! Thanks for the refactoring :dancers:

bigchaindb/db/utils.py
lluminita: I am not sure I understand what needs to be changed here... I see `reconnect` is in `bigchain.core` so if all DB related functions should be in `bigchain.db` than probably some more refactoring should be done and maybe is better to create another issue for this?

bigchaindb/db/utils.py
sbellem: :+1:

PR #472, opened 2016-07-20
Docs: moved The Python Server API by Example under Drivers & Clients

ttmc:
@TimDaub suggested that I move the documentation page "The Python Server API by Example" from under "BigchainDB Nodes" to under "BigchainDB Drivers & Clients", so I did.


r-marques:
:+1:

PR #475, opened 2016-07-21
Refactor method signature

vrde:
The method in `bigchaindb/core.py::Bigchain::vote` accepts as first argument a `block` dictionary but in the code only the `block['id']` is used.

Having simpler requirements for the method will also help another refactoring I'm doing.


rhsimplex:
:+1:

PR #476, opened 2016-07-21
Fix log message for voter

vrde:
Cleanup voting logging message.

### DIFF
--- a/bigchaindb/voter.py
+++ b/bigchaindb/voter.py
@@ -156,7 +156,8 @@ class Voter(object):
                 return
 
             block, vote = elem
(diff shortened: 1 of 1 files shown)


ttmc:
It should work but why use the old string formatting rather than .format()?

vrde:
> It should work but why use the old string formatting rather than .format()?

It's best practice. If you use `.format` the "string composition" happens *before* the method call (in this case `.info`), while if you use positional arguments the composition happens later, after knowing if the log message will be actually displayed or not (depends on the logging level).

r-marques:
:+1:

PR #478, opened 2016-07-21
Minimized base software installed on AWS instance

ttmc:
I don't know why, but AWS deployment seems more flaky lately, even before I made this change. When it does fail, it usually fails / aborts at some point during the installation of the base software. It's not always clear why the fail happened.

We were installing a bunch of base software that wasn't necessary, so I cut that out. That sped up installation and minimizes the amount of things that can go wrong during the installation of base software.


r-marques:
Its difficult to review this unless I actually deploy an instance. But if the deployment works correctly :+1:

ttmc:
I still wasn't very happy with how flaky and unpredictable the installation of the base software was, so I made more revisions and now it seems to be less flaky. I also added more comments to the steps of the base software installation. I'll merge this PR once the checks pass.

PR #481, opened 2016-07-25
Docs: make pip3 installation instructions consistent with Docker and AWS

ttmc:
The documentation's instructions for installing `pip3` had become inconsistent with how we do it in the AWS deployment scripts and the Dockerfile. We now use `sudo apt-get install python3-pip` rather than:

```text
sudo apt-get install python3-setuptools
sudo easy_install3 pip
```

### DIFF
--- a/docs/source/nodes/setup-run-node.md
+++ b/docs/source/nodes/setup-run-node.md
@@ -115,27 +115,35 @@ With OS-level dependencies installed, you can install BigchainDB Server with `pi
 
 ### How to Install BigchainDB with pip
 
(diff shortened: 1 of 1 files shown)


vrde:
:+1: 

minor note: if you are working on a virtual env based on Python3, then usually `pip` and `pip3` are the same commands.

ttmc:
I think that's covered by this sentence: "In the following instructions, we call it `pip3` but you may be able to use `pip` if that refers to the same thing."

I leave it up to them to determine if they refer to the same thing.

PR #484, opened 2016-07-25
Refactor multiprocessing for block

vrde:
This is the first PR of a series to refactor the multiprocessing logic we have.
Please note that this PR introduces a new module ([multipipes]((https://github.com/bigchaindb/multipipes)) that encloses all the multiprocessing logic. The module will be improved with more test coverage and docs in parallel with the development of the next PRs.

Related: #493


bigchaindb/pipelines/block.py
```diff
@@ -0,0 +1,80 @@
...
+    def delete_tx(self, tx):
+        r.table('backlog')\
+         .get(tx['id'])\
+         .delete(durability='hard')\
+         .run(self.bigchain.conn)
+
+        return tx
+
+    def validate_tx(self, tx):
+        tx = self.bigchain.is_valid_transaction(tx)
```
ttmc: You asked me to make a note about this minor thing, so you can fix it as part of this PR:  The docstring for `is_valid_transaction()` says that it returns "bool: `True` if the transaction is valid, `False` otherwise" but that's wrong: it returns the transaction if the transaction is valid, and `False` otherwise.

bigchaindb/pipelines/block.py
```diff
@@ -0,0 +1,80 @@
...
+        if tx:
+            return tx
```
ttmc: If `tx` is `False` at this point, then `False` won't be returned; instead the default Python return value will be returned, which is `None`. Is that okay? I guess when you use `None` as a boolean expression, then it will be evaluated as if it were `False`? Maybe add a comment or docstring about this? i.e. This method returns either the transaction or `None`.

bigchaindb/pipelines/block.py
```diff
@@ -0,0 +1,80 @@
...
+    def create(self, tx, timeout=False):
+        if tx:
+            self.txs.append(tx)
+        if self.txs and (len(self.txs) == 1000 or timeout):
+            block = self.bigchain.create_block(self.txs)
+            self.txs = []
+            return block
+
+    def write(self, block):
+        self.bigchain.write_block(block)
```
ttmc: The default durability in `write_block()` is currently `soft` so it would seem so... unless we change the default durability of `write_block()` to `hard`

bigchaindb/processes.py
```diff
@@ -92,18 +90,16 @@ def start(self):
         # initialize the processes
         p_map_bigchain = mp.Process(name='bigchain_mapper', target=self.map_bigchain)
         p_map_backlog = mp.Process(name='backlog_mapper', target=self.map_backlog)
```
r-marques: `map_backlog` is not doing anything now

bigchaindb/pipelines/block.py
```diff
@@ -25,11 +56,35 @@ def delete_tx(self, tx):
...
+        - a timeout happened.
+
+        Args:
+            tx (dict): the transaction to validate, might be None if
+                a timeout happens.
+            timeout (bool): ``True`` if a timeout happened
+                (Default: ``False``).
+
+        Returns:
+            The block.
```
ttmc: It returns None if `self.txs` is an empty list, or (`self.txs` doesn't have 1000 elements and no timeout), I think.

bigchaindb/pipelines/block.py
```diff
@@ -25,11 +56,35 @@ def delete_tx(self, tx):
...
+        """
         if tx:
             self.txs.append(tx)
         if self.txs and (len(self.txs) == 1000 or timeout):
```
ttmc: I find this a bit hard to read. I wonder if this would be more clear:
```python
if len(self.txs) == 1000 or (timeout and self.txs):
```

vrde: :+1:

bigchaindb/processes.py
```diff
@@ -30,7 +31,6 @@ class Processes(object):
     def __init__(self):
         # initialize the class
         self.q_new_block = mp.Queue()
-        self.q_new_transaction = mp.Queue()
         self.q_block_new_vote = mp.Queue()
         self.q_revert_delete = mp.Queue()
```
r-marques: we can delete the `map_backlog` method

Issue #158, opened 2016-03-31
Proposal: set RethinkDB read mode to 'majority' to improve consistency


AnthonyAkentiev:
Hello.
As mentioned here - https://rethinkdb.com/docs/consistency/ default read mode is 'single'.
Default write mode is 'majority' and default durability is 'hard'. It's OK. 

Setting read mode to 'majority' will improve consistency but with read latencies penalised:

https://aphyr.com/posts/329-jepsen-rethinkdb-2-1-5 (warning: old post!)
"There is no significant availability impact to choosing majority reads, though latencies rise significantly."
We don't want TXs to be lost in Bigchain DB)

What do you think?

p.s. 
And read that too:
https://aphyr.com/posts/330-jepsen-rethinkdb-2-2-3-reconfiguration
"However, long-running tests, involving randomized network partitions and reconfiguring the cluster membership, resulted in nonlinearizable histories: stale reads, illegal compare-and-sets, and the loss of acknowledged operations."

r-marques:
thanks @AnthonyAkentiev 

We do need some more stress testing to decide on the correct read modes. I agree that read with `majority` should be set while performing the consensus phase but it should not be needed afterwards for the querying of the data since the data will be immutable.

Regarding aphyr problem it as been addressed and fixed by the rethinkdb guys https://github.com/rethinkdb/rethinkdb/releases/tag/v2.2.4 and https://github.com/rethinkdb/rethinkdb/issues/5289#issuecomment-175394540

rhsimplex:
Closed by #497

PR #497, opened 2016-07-26
Majority read

rhsimplex:
Addresses #158. Should `read_mode` be a configurable setting?


vrde:
:+1: 

> Should read_mode be a configurable setting?

If this is the default (safe) behavior and no one needs to customize it I'd say no.

Issue #437, opened 2016-07-13
bigchaindb drop raises exception if database does not exist


r-marques:
If the bigchain database does not exist, running `bigchaindb drop` raises and exception:

``` bash
$ bigchaindb drop
INFO:bigchaindb.config_utils:Configuration loaded from `/home/rodmar/.bigchaindb`
Do you want to drop `bigchain` database? [y/n]: y
INFO:bigchaindb.db.utils:Drop database `bigchain`
Traceback (most recent call last):
  File "/home/rodmar/code/code-ascribe/bigchaindb/bigchaindb/db/utils.py", line 88, in drop
    r.db_drop(dbname).run(conn)
  File "/home/rodmar/.venvs/bigchaindb/lib/python3.4/site-packages/rethinkdb/ast.py", line 118, in run
    return c._start(self, **global_optargs)
  File "/home/rodmar/.venvs/bigchaindb/lib/python3.4/site-packages/rethinkdb/net.py", line 625, in _start
    return self._instance.run_query(q, global_optargs.get('noreply', False))
  File "/home/rodmar/.venvs/bigchaindb/lib/python3.4/site-packages/rethinkdb/net.py", line 468, in run_query
    raise res.make_error(query)
rethinkdb.errors.ReqlOpFailedError: Database `bigchain` does not exist in:
r.db_drop('bigchain')
^^^^^^^^^^^^^^^^^^^^^

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/rodmar/.venvs/bigchaindb/bin/bigchaindb", line 9, in <module>
    load_entry_point('BigchainDB', 'console_scripts', 'bigchaindb')()
  File "/home/rodmar/code/code-ascribe/bigchaindb/bigchaindb/commands/bigchain.py", line 306, in main
    utils.start(parser, globals())
  File "/home/rodmar/code/code-ascribe/bigchaindb/bigchaindb/commands/utils.py", line 95, in start
    func(args)
  File "/home/rodmar/code/code-ascribe/bigchaindb/bigchaindb/commands/bigchain.py", line 149, in run_drop
    db.drop(assume_yes=args.yes)
  File "/home/rodmar/code/code-ascribe/bigchaindb/bigchaindb/db/utils.py", line 91, in drop
    raise exceptions.DatabaseDoesNotExist('Database `{}` does not exist'.format(dbname))
bigchaindb.exceptions.DatabaseDoesNotExist: Database `bigchain` does not exist
[...]

PR #508, opened 2016-07-29
attempted drop to nonexistent database prints a friendly message

mickhalsband:
Implemented a simple except clause to print a human readable error message if 'drop' is called for a non-existent database.

Update:
Tests have been properly implemented
~~No tests have been updated due to conformance with the existing unit-test scheme (test suite does not address failures)~~

Related: #437

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -22,7 +22,7 @@ from bigchaindb.client import temp_client
 from bigchaindb import db
 from bigchaindb.exceptions import (StartupError,
                                    DatabaseAlreadyExists,
-                                   KeypairNotFoundException)
+                                   KeypairNotFoundException, DatabaseDoesNotExist)
 from bigchaindb.commands import utils
 from bigchaindb.processes import Processes
 from bigchaindb import crypto
@@ -146,7 +146,13 @@ def run_init(args):
 def run_drop(args):
     """Drop the database"""
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
-    db.drop(assume_yes=args.yes)
+    try:
+        db_name = 'Unknown database'
+        if db != None: db_name = db.get_database_name()
+        db.drop(assume_yes=args.yes)
+    except DatabaseDoesNotExist:
+        print('Cannot drop \'{name}\'. The database does not exist.'.format(name=db_name), file=sys.stderr)
+        print('Consider initializing the database before attempting drop.', file=sys.stderr)
 
 
 def run_start(args):
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -22,7 +22,8 @@ from bigchaindb.client import temp_client
 from bigchaindb import db
 from bigchaindb.exceptions import (StartupError,
                                    DatabaseAlreadyExists,
-                                   KeypairNotFoundException, DatabaseDoesNotExist)
+                                   KeypairNotFoundException,
+                                   DatabaseDoesNotExist)
 from bigchaindb.commands import utils
 from bigchaindb.processes import Processes
 from bigchaindb import crypto
@@ -147,10 +148,9 @@ def run_drop(args):
     """Drop the database"""
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
     try:
-        db_name = 'Unknown database'
-        if db != None: db_name = db.get_database_name()
         db.drop(assume_yes=args.yes)
     except DatabaseDoesNotExist:
+        db_name = db.get_database_name()
         print('Cannot drop \'{name}\'. The database does not exist.'.format(name=db_name), file=sys.stderr)
         print('Consider initializing the database before attempting drop.', file=sys.stderr)
 
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -47,6 +47,17 @@ def mock_rethink_db_drop(monkeypatch):
 
 
 @pytest.fixture
+def mock_rethink_db_drop_not_exist(monkeypatch):
+    def mockreturn(dbname):
+        class MockDropped(object):
+            def run(self, conn):
+                from bigchaindb.exceptions import DatabaseDoesNotExist
+                raise DatabaseDoesNotExist
+        return MockDropped()
+    monkeypatch.setattr('rethinkdb.db_drop', mockreturn)
+
+
+@pytest.fixture
 def mock_generate_key_pair(monkeypatch):
     monkeypatch.setattr('bigchaindb.crypto.generate_key_pair', lambda: ('privkey', 'pubkey'))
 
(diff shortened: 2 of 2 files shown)


mickhalsband:
BEGIN BLOCK

IdhhhCRRagkZCRa4GcMu69x6vlmutEINgfDIdUFjTER8m1zrROPMINoFYnrNier

END BLOCK

bigchaindb/commands/bigchain.py
```diff
@@ -146,7 +146,13 @@ def run_init(args):
 def run_drop(args):
     """Drop the database"""
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
-    db.drop(assume_yes=args.yes)
+    try:
+        db_name = 'Unknown database'
+        if db != None: db_name = db.get_database_name()
```
ttmc: I think you could replace the last two lines with this one:
```python
db_name = db.get_database_name() if db else 'Unknown database'
```
and put it before the `try:`

ttmc:
Thanks @mickhalsband --- I looked at your changes and made a comment, but at least one other member of our team should look at them before we merge this pull request. It's now late Friday night in Berlin (where we are based), so the review might not happen until Monday.

ttmc:
It seems that the Codecov (code coverage) tests are failing because there's no test covering the `except:` case.

bigchaindb/commands/bigchain.py
```diff
@@ -22,7 +22,7 @@
 from bigchaindb import db
 from bigchaindb.exceptions import (StartupError,
                                    DatabaseAlreadyExists,
-                                   KeypairNotFoundException)
+                                   KeypairNotFoundException, DatabaseDoesNotExist)
```
vrde: Can you put `DatabaseDoesNotExist` in a new line?

bigchaindb/commands/bigchain.py
vrde: `db` is defined in [line 22](https://github.com/mickhalsband/bigchaindb/blob/32fcbc0bbf2a8965aec160b42c8194d2737fad8c/bigchaindb/commands/bigchain.py#L22) (it's an import) so it cannot be `None`.

The name of the database is defined even if there is no configuration available (see [`bigchaindb/__init__.py`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/__init__.py#L20)), so you can assume `db.get_database_name()` to always return a value.

bigchaindb/commands/bigchain.py
```diff
@@ -146,7 +146,13 @@ def run_init(args):
...
     """Drop the database"""
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
-    db.drop(assume_yes=args.yes)
+    try:
+        db_name = 'Unknown database'
+        if db != None: db_name = db.get_database_name()
+        db.drop(assume_yes=args.yes)
+    except DatabaseDoesNotExist:
+        print('Cannot drop \'{name}\'. The database does not exist.'.format(name=db_name), file=sys.stderr)
+        print('Consider initializing the database before attempting drop.', file=sys.stderr)
```
vrde: You can remove this line since it's redundant (see line above).

vrde:
> @mickhalsband: No tests have been updated due to conformance with the existing unit-test scheme (test suite does not address failures)

I don't fully understand what you mean here. Code coverage decreased by a 0.20%, and we cannot accept this PR until the new `try... except` is covered by a test.

Thanks for your PR!

bigchaindb/commands/bigchain.py
mickhalsband: @vrde which line did you mean is redundant? the `if db != None` line?
(I agree - since it's an imported module and not a class - as I previously assumed - this is a redundant check)

vrde: I'm referring to L155 `print('Consider initializing the database before attempting drop.', file=sys.stderr)`, you can remove it since it's not adding much information to the previous message.

bigchaindb/commands/bigchain.py
```diff
@@ -146,7 +147,12 @@ def run_init(args):
 def run_drop(args):
     """Drop the database"""
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
-    db.drop(assume_yes=args.yes)
+    try:
+        db.drop(assume_yes=args.yes)
+    except DatabaseDoesNotExist:
+        db_name = db.get_database_name()
+        print('Cannot drop \'{name}\'. The database does not exist.'.format(name=db_name), file=sys.stderr)
```
vrde: [Double quotes](https://github.com/bigchaindb/bigchaindb/blob/master/PYTHON_STYLE_GUIDE.md#single-or-double-quotes) are allowed in this context.

tests/test_commands.py
```diff
@@ -169,6 +180,12 @@ def test_drop_existing_db(mock_rethink_db_drop):
     run_drop(args)
 
 
+def test_drop_non_existing_db(mock_rethink_db_drop_not_exist):
+    from bigchaindb.commands.bigchain import run_drop
+    args = Namespace(config=None, yes=True)
+    run_drop(args)
```
vrde: This test is similar to `test_bigchain_run_init_when_db_exists` and `test_drop_existing_db`, but I don't like those two test much because they are missing `assert`ions.
Fixing those two tests is out of the scope of this PR, but do you have any suggestions on how to improve your new `test_drop_non_existing_db` test?

mickhalsband: @vrde I was looking for return values which could be propagated from run_drop() for example, but there are none.
One approach would be to propagate the underlying rethink db status through run_drop (perhaps translating it to some bigchaindb format) and then to the testing code. This could then easily be asserted (looking at the returned `tables_dropped` field, for example).
Regardless, refactoring the wrapper layer at bigchaindb/db/utils.py into a wrapper class could allow a native bigchaindb database status which could be more easily observed by tests (and other api users).
What kind of improvements did you have in mind?

vrde: > One approach would be to propagate the underlying rethink db status through run_drop (perhaps translating it to some bigchaindb format) and then to the testing code. This could then easily be asserted (looking at the returned tables_dropped field, for example).

Those `run_*` functions are just used as a bridge from the command line to the actual code to run, so they prepare/set up the state for BigchainDB to work.

> What kind of improvements did you have in mind?

In the test you can check if `stderr` contains the error message you expect. You can use `capsys`, check it out here:
 - https://github.com/bigchaindb/bigchaindb/blob/master/tests/test_commands.py#L111

tests/test_commands.py
```diff
@@ -47,6 +47,17 @@ def run(self, conn):
 
 
 @pytest.fixture
+def mock_rethink_db_drop_not_exist(monkeypatch):
```
vrde: Did you consider using `unittest.object.patch`? Might be simpler, you can see an example here:
 - https://github.com/bigchaindb/bigchaindb/blob/master/tests/pipelines/test_utils.py#L21

vrde:
Will close it for now and reopen if there is a follow up.

PR #509, opened 2016-07-30
Feat/439/connect multiple nodes docker

carloscorreia94:
Assuming that:

We're using the flag **--link** when creating the docker containers that will link to the first one.

Lacking tests for the two commands i've created as I didn't see a easy way through connecting the test tools to the creation of docker containers (I could've tested it natively but was having a lot of trouble with conflicts)

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -156,7 +156,11 @@ def run_start(args):
 
     if args.start_rethinkdb:
         try:
-            proc = utils.start_rethinkdb()
+            # For sake of simplicity, args.start_rethinkdb holds either true or a join address
+            if args.start_rethinkdb is True:
+                proc = utils.start_rethinkdb()
+            else:
+                proc = utils.start_rethinkdb(args.start_rethinkdb)
         except StartupError as e:
             sys.exit('Error starting RethinkDB, reason is: {}'.format(e))
         logger.info('RethinkDB started with PID %s' % proc.pid)
@@ -238,6 +242,11 @@ def main():
                         dest='start_rethinkdb',
                         action='store_true',
                         help='Run RethinkDB on start')
+
+    parser.add_argument('--experimental-join',
+                        dest='start_rethinkdb',
+                        action='store',
+                        help='Join to an existing BigChainDB Node')
 
     # all the commands are contained in the subparsers object,
     # the command selected by the user will be stored in `args.command`
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -158,7 +158,7 @@ def run_start(args):
         try:
             if args.start_rethinkdb is True:
                 proc = utils.start_rethinkdb()
-            elif args.start_rethinkdb == '-auto':
+            elif args.start_rethinkdb == 'auto':
                 proc = utils.start_rethinkdb(utils.find_rethinkdb_host())
             else:
                 proc = utils.start_rethinkdb(args.start_rethinkdb)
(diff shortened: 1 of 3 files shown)


carloscorreia94:

ajVTwGWr7NeVBvnpJHa8E2lb6MBMq43ErzR2m7ku9vMDb8qrpGKNlEWcLxRCn0o

rhsimplex:
Hi @carloscorreia94,

Your PR is failing two tests.  You can run those tests individually with:
```bash
py.test -k test_bigchain_run_start
```

and 

```bash
py.test -k test_bigchain_run_start_with_rethinkdb
```

as long as you have pytest installed. Can you fix them? You can find the test code in `tests/test_commands.py`.

carloscorreia94:
Hello @rhsimplex ,

The tests are now passing, I'm not sure how but they seemed to pass when I've first run them.
So I've refactored my logic a little bit, for better. Only thing that's failing is the test coverage, for which at this point I can't do anything. Thanks a lot :-)

bigchaindb/commands/bigchain.py
```diff
@@ -239,6 +244,20 @@ def main():
                         action='store_true',
                         help='Run RethinkDB on start')
 
+    parser.add_argument('--experimental-join',
+                        dest='start_rethinkdb',
+                        action='store',
+                        help='Join to an existing BigChainDB Node '
```
rhsimplex: should be `BigchainDB`

bigchaindb/commands/utils.py
```diff
@@ -14,16 +14,46 @@
...
+    linked by docker with ``--link <container_name>`
+    DISCLAIMER: This is a really dummy function, but should work as long as --link
+    option in docker maintains its consistency
+
+    Returns:
+        IP Address of the host machine plus the default port to connect
+    """
+    default_port = '29015'
+
+    proc = subprocess.Popen(['cat', '/etc/hosts'],
```
rhsimplex: why not just open the file here? what's the reasoning behind starting a subprocess?

bigchaindb/commands/utils.py
```diff
@@ -14,16 +14,46 @@
...
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT,
+                            universal_newlines=True)
+    lines = proc.stdout.readlines()
+    if len(lines) < 6:
```
rhsimplex: Why 6?

rhsimplex:
Hi @carloscorreia94 thanks for the changes.  We can't decrease test coverage with a PR.  You don't have to test the docker integration, but you should test the command works the way you expect. 

You can see some examples of testing commands (including mocking the input and output) in `tests/test_commands.py`.

```python
@patch('subprocess.Popen')
def test_start_rethinkdb_returns_a_process_when_successful(mock_popen):
    from bigchaindb.commands import utils
    mock_popen.return_value = Mock(stdout=['Server ready'])
    assert utils.start_rethinkdb() is mock_popen.return_value


@patch('subprocess.Popen')
def test_start_rethinkdb_exits_when_cannot_start(mock_popen):
    from bigchaindb import exceptions
    from bigchaindb.commands import utils
    mock_popen.return_value = Mock(stdout=['Nopety nope'])
    with pytest.raises(exceptions.StartupError):
        utils.start_rethinkdb()
```
Give it a try, and ask if you need help.

carloscorreia94:
Ok, I've settled the function in better terms without the need to open a subprocess. About the tests, I've kinda replicated the ones you sent me because I couldn't get rid of the "Error waiting for the database" on my Macbook. So there I've played it safe for the Travis CI build to see if the test coverage would suffice. Thanks @rhsimplex

rhsimplex:
Looking good, but unfortunately still not enough coverage. [Can you see this link?](https://codecov.io/gh/bigchaindb/bigchaindb/compare/ff042b5954abe48c7264d43128d52584eab2a806...da51f2df68da2e417b1e37824e8b02e22a92b59e)  It shows you which parts of your diff are not covered by tests.  80% of a new PR must be covered by tests, and you can't decrease the overall coverage.

An easy fix would be to write a test for the `find_rethinkdb_hosts` function you added -- it currently has no test coverage.

carloscorreia94:
Hello @rhsimplex,
I've successfully managed to pass the codecov/patch check by adding some more tests, but can't seem to overcome the codecov/project check. I've read something on it, but can't figure out why the project coverage will decrease compared to the patch coverage (line by line via pull request) 

Thank you in advance!

carloscorreia94:
All right @rhsimplex , I think now should be good to go :-)

Issue #122, opened 2016-03-08
Clean exit for the bigchaindb-benchmark process


r-marques:
Implement a clean exit for `bigchaindb-benchmark load` process.

Issue #43, opened 2016-02-15
BigchainDB container doesn't exit gracefully


rhsimplex:
The docker container works! Thank you!

```python
$ docker-compose up
Creating bigchaindb_rethinkdb-data_1
Creating bigchaindb_rethinkdb_1
Creating bigchaindb_bigchaindb_1
Attaching to bigchaindb_rethinkdb-data_1, bigchaindb_rethinkdb_1, bigchaindb_bigchaindb_1
rethinkdb_1      | Recursively removing directory /data/rethinkdb_data/tmp
rethinkdb_1      | Initializing directory /data/rethinkdb_data
rethinkdb_1      | Running rethinkdb 2.2.4~0jessie (GCC 4.9.2)...
rethinkdb_1      | Running on Linux 3.13.0-68-generic x86_64
rethinkdb_1      | Loading data from directory /data/rethinkdb_data
bigchaindb_rethinkdb-data_1 exited with code 0
rethinkdb_1      | Listening for intracluster connections on port 29015
rethinkdb_1      | Listening for client driver connections on port 28015
rethinkdb_1      | Listening for administrative HTTP connections on port 8080
rethinkdb_1      | Listening on addresses: 127.0.0.1, 172.17.0.3, ::1, fe80::42:acff:fe11:3%1640
rethinkdb_1      | Server ready, "643d16d5c461_92r" 140afe00-9284-40b0-8c35-a6b5c291106f
bigchaindb_1     | INFO:bigchaindb.config_utils:Configuration loaded from `/root/.bigchaindb_docker/config`
bigchaindb_1     | INFO:bigchaindb.db.utils:Create:
bigchaindb_1     | INFO:bigchaindb.db.utils: - database `bigchain`
bigchaindb_1     | INFO:bigchaindb.db.utils: - tables
bigchaindb_1     | INFO:bigchaindb.db.utils: - indexes
bigchaindb_1     | INFO:bigchaindb.db.utils: - genesis block
bigchaindb_1     | INFO:bigchaindb.db.utils:Done, have fun!
bigchaindb_1     | INFO:bigchaindb.commands.bigchain:Start bigchaindb main process
bigchaindb_1     | INFO:bigchaindb.processes:starting bigchain mapper
bigchaindb_1     | INFO:bigchaindb.processes:starting backlog mapper
bigchaindb_1     | INFO:bigchaindb.processes:starting block
bigchaindb_1     | INFO:bigchaindb.processes:starting voter
[...]

PR #510, opened 2016-08-01
Refactor multiprocessing for election

rhsimplex:
per issue #312 #495

PR #515, opened 2016-08-02
Refactor multiprocessing for vote

### LINKED ISSUE #494: Port voter to the new pipeline architecture
The goal is to untangle the logic of the voter from the multiprocessing part (that is managing the processes, connect them using queues, and more).

vrde:
Closes #494, #43, #122


bigchaindb/pipelines/vote.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+
+        self.bigchain.write_vote(vote)
+        return vote
+
+
+def initial():
+    """Return unvoted blocks."""
+    b = Bigchain()
+    initial = b.get_unvoted_blocks()
```
r-marques: confusing assigning a variable with the same name as the method

bigchaindb/pipelines/vote.py
```diff
@@ -0,0 +1,138 @@
...
+    """This class encapsulates the logic to vote on blocks.
+
+    Note:
+        Methods of this class will be executed in different processes.
+    """
+
+    def __init__(self):
+        """Initialize the Block voter."""
+
+        last_voted = Bigchain().get_last_voted_block()
```
r-marques: cleaner if you call this after instantiating `Bigchain`, no need to instantiate it twice

bigchaindb/pipelines/vote.py
```diff
@@ -0,0 +1,138 @@
...
+            tx_validity (bool): the validity of the transaction
+            block_id (str): the id of block containing the transaction
+            num_tx (int): the total number of transactions to process
+
+        Returns:
+            None, or a vote if a decision has been reached.
+        """
+
+        self.counters[block_id] += 1
+        self.validity[block_id] = tx_validity and self.validity.get(block_id,
```
r-marques: Checking that all the transactions are valid is not enough to validate the block.
You also need to check the fields in the block itself, basically call `consensus.validate_block`

r-marques:
This pr is related to #510 
After both are merged there will need to be some cleanup.
I guess `bigchaindb/voter.py` and `tests/db/test_voter.py` could be deleted

bigchaindb/pipelines/vote.py
vrde: You are right, will update my PR to check it!

bigchaindb/pipelines/vote.py
vrde: It's to overcome a limitation with `multiprocessing`. Will add a comment to explain that :)

r-marques:
:+1:

Issue #419, opened 2016-07-05
Confusing error message for config file.


nsgeorgi:
I manually updated the config file which by default is located at ~/.bigchaindb and I received the below error message after I tried this command in python API "b = Bigchain()". After the help of @vrde I realised that I had some typo in the config file and I got the error cause it was no longer valid json. In the future you could print a better error message.
![screenshot-from-2016-07-05-04_24_32](https://cloud.githubusercontent.com/assets/8840982/16603317/f0f17434-4312-11e6-9fca-91fe36cec34f.png)

vrde:
Thanks for reporting the issue @nsgeorgi!

Maybe we can fix this by reviewing the configuration system and use something more classical than a JSON file. If we use [ConfigObj](http://www.voidspace.org.uk/python/configobj.html) we can even add schema validation. It's a bit old (~6 years) but if someone knows anything newer please tell us.
@sbellem: any thoughts on that?

PR #531, opened 2016-08-08
Bug/419/config syntax error handling

d01phin:
This is a WIP for config file error handling, addressing the issue #419


d01phin:
The question here is how should `bigchaindb start` command handle invalid configuration file exception?

bigchaindb/config_utils.py
```diff
@@ -97,8 +98,11 @@ def file_config(filename=None):
         filename = CONFIG_DEFAULT_PATH
 
     logger.debug('file_config() will try to open `{}`'.format(filename))
-    with open(filename) as f:
-        config = json.load(f)
+    try:
+        with open(filename) as f:
+            config = json.load(f)
+    except Exception as err:
```
vrde: Exception handling here is too broad. Usually a `try ... except Exception` is an [antipattern](https://realpython.com/blog/python/the-most-diabolical-python-antipattern/) to avoid. Also, the `FileNotFoundError` is used in `autoconfigure` at L215.

My suggestion here is to catch only the exceptions you expect.

d01phin: Thank you for pointing out! That's why I was asking. Just worth noting that catching `Exception` is not really an antipattern as described in the blog post you have mentioned, as it was rethrowing new type of exception with details about the location of problem, and just silently suppressing it.

However, I agree that mixing file open and JSON parsing exceptions was a bad idea, because `FileNotFound` (or any other IO exceptions?) are handled from the calling functions. 

So again, my question is, addressing the original issue, should we replace the `ValueError` raised by `json.load()` by custom exception or let calling function to handle `ValueError` as they do with `FileNotFound`?

vrde:
:+1:

ttmc:
:+1:

bigchaindb/config_utils.py
```diff
@@ -98,7 +99,12 @@ def file_config(filename=None):
 
     logger.debug('file_config() will try to open `{}`'.format(filename))
     with open(filename) as f:
-        config = json.load(f)
+        try:
+            config = json.load(f)
+        except ValueError as err:
+            raise exceptions.ConfigurationError(
+                'Failed to load configuration from `{}`, reason: {}'.format(filename, err)
```
vrde: I tried your branch locally, and the error is like:
```
Failed to load configuration from `/home/vrde/.bigchaindb`, reason: Expecting value: line 1 column 59 (char 58)
```

A clearer wording would be:
```
Failed to parse the JSON configuration from `{}`. {}
```
in this way we highlight that the problem is on **parsing** a **JSON** file.

PR #532, opened 2016-08-08
Many revisions to the docs laying groundwork for some big upcoming changes

ttmc:
This PR includes a bunch of documentation revisions to prepare for some big upcoming changes:

* Clarified the difference between a minimal BigchainDB node for development and testing purposes, and a production node. There's a short new page about that too.
* Set up a new section where the instructions for setup & management of production nodes will go.
* BigchainDB Server and BigchainDB clients will eventually be different things, each with its own configuration settings, CLI, etc. Added some notes and changed some names to reflect that.
* The existing page about setting up a production node is now gone. It will be replaced by several cloud-provider-specific pages (and automation)

This PR also includes many cosmetic changes: 

* Moved the Docker page to the Appendices
* Rearranged some sections and fixed the links
* Moved the page titled **Nodes, Clusters & Federations** up front because it's basic terminology that must be understood before the documentation makes sense
* Moved the instructions for installing OS-level dependencies, and upgrading pip, to the Appendices (and linked to them from the appropriate places)


docs/source/appendices/install-os-level-deps.md
```diff
@@ -0,0 +1,17 @@
...
+
+BigchainDB Server has some OS-level dependencies that must be installed.
+
+On Ubuntu 14.04, we found that the following was enough:
+```text
+sudo apt-get update
+sudo apt-get install g++ python3-dev
+```
+
+On Fedora 23, we found that the following was enough (tested in February 2015):
```
r-marques: I recently did the same in a new installation of Fedora 24 at home so I guess we could change:
- `Fedora 23 -> Fedora 23/24`
- remove `(tested in February 2015)`

r-marques:
:+1:

PR #539, opened 2016-08-10
Documentation Revisions 

ttmc:
* In the new section on production nodes, wrote an **Overview** page about how a production node is different from a dev/test node, and how there are tools to help
* Moved "AWS Setup" instructions that were specific to deploying a testing cluster from the Appendices back to the page about deploying a testing cluster on AWS
* Fixed some broken links
* (This PR originally added some docs about installing Chef Server, but it's not being used anymore.)


vrde:
:+1:

PR #330, opened 2016-05-25
Core/197/revert bigchain deletes

rhsimplex:
Resolves #197, ready for review.

PR #544, opened 2016-08-12
remove block delete reverter

rhsimplex:
Removes block delete reverter introduced in #330, in accordance with future architecture changes.

RIP my first BigchainDB PR :skull:


vrde:
:+1:

vrde:
PS: I didn't noticed that `processes.py` still has the queue to manage deletes. Will remove it in #515.

Issue #554, opened 2016-08-16
Server API: POST and GET transactions


diminator:
See #362 

### Post a transaction 
```
POST /transactions/

PAYLOAD
    <TX>

RETURNS
    201 + <ID>
    {4xx, 5xx} + <reason>
``` 
BigchainDB foresees `bigchaindb.write_transaction(tx)` to make the call to the DB

#### Questions:
- Does this post do a multicast or simply `bigchaindb.write_transaction(tx)`?
- see #347: `CREATE` tx's require a fulfillment from the federation: ie.  `bigchaindb.sign_transaction(tx, private_key=bigchain.me_private)` is called in the `POST /transactions/` endpoint. In the future we want to split this up: 
  1. retrieve the permission first (ie a separate call to get the fulfillment from the federation)
  2. POST the signed tx.

### Retrieve a transaction
```
GET /transactions/<ID>

RETURNS
    200 + <TX> 
    404
```
BigchainDB foresees `bigchaindb.get_transaction(id)` to make the call to the DB

#### Questions: 

- does the call only looks to the `bigchain` blocks or also in `backlog`?
- does it also look into `bigchain` blocks that are `UNDECIDED`?

TimDaub:
> Does this post do a multicast or simply bigchaindb.write_transaction(tx)?

I think for now we don't need to do a multicast.

sohkai:
> Does this post do a multicast or simply bigchaindb.write_transaction(tx)?

Sylvain and I had a brief discussion about this before and I'm sure you guys have had similar ones. Supporting multi-cast and multi-node connections should be their own projects; their scopes are way too big.

> #### Questions:
    does the call only looks to the bigchain blocks or also in backlog?
    does it also look into bigchain blocks that are UNDECIDED?

From the discussion with the `/status` endpoint, I think we should only return transactions that are DECIDED and in the chain. Otherwise, a 404 (and maybe a url link to check the transaction's status, depending how hyperlinked we want to get).

r-marques:
> Does this post do a multicast or simply bigchaindb.write_transaction(tx)?

This will probably be done in the future but it needs to be done on the client side. The client should contact multiple servers so that a rogue server does not simply drop the transactions. We need a protocol for this.

Agree with the second point. We can even have an explicit threshold fulfillment in a `CREATE` transaction so it requires a signature from both the `federation node` and the `new_owner` so that you cannot create assets to random pub keys.

The flow would be:
[...]

PR #563, opened 2016-08-17
#554 transactions GET/POST endpoints

diminator:
This PR solves #554

- [x] split up views per resource
- [x] add error cases to get/post transaction
- [x] update docs/test

### Questions:
- upon `POST /transactions/` should the reply contain more than the transaction (augmented with `assignee`)? 

We could also provide the output of `bigchaindb.write_transaction(tx)`, which amounts to:
```
def write_transaction(self, signed_transaction, durability='soft'):
    ....
    # write to the backlog
    response = r.table('backlog').insert(signed_transaction, durability=durability).run(self.conn)
    return response 
```

and `response` has the form:

```
{'errors': 0, 'inserted': 1, 'unchanged': 0, 'deleted': 0, 'skipped': 0, 'replaced': 0}
```

### DIFF
--- a/bigchaindb/web/views/info.py
+++ b/bigchaindb/web/views/info.py
@@ -1,7 +1,7 @@
 """This module provides the blueprint for some basic API endpoints.
 
-For more information please refer to the documentation in Apiary:
- - http://docs.bigchaindb.apiary.io/
+For more information please refer to the documentation on ReadTheDocs:
+ - https://bigchaindb.readthedocs.io/en/latest/drivers-clients/http-client-server-api.html
 """
 
 import flask
@@ -12,6 +12,7 @@ from bigchaindb import version
 
 
 info_views = Blueprint('info_views', __name__)
+
 
 @info_views.route('/')
 def home():
--- a/bigchaindb/web/views/info.py
+++ b/bigchaindb/web/views/info.py
@@ -1,0 +1,25 @@
+"""This module provides the blueprint for some basic API endpoints.
+
+For more information please refer to the documentation in Apiary:
+ - http://docs.bigchaindb.apiary.io/
+"""
+
+import flask
+from flask import Blueprint
+
+import bigchaindb
+from bigchaindb import version
+
+
+info_views = Blueprint('info_views', __name__)
+
+@info_views.route('/')
+def home():
+    return flask.jsonify({
+        'software': 'BigchainDB',
+        'version': version.__version__,
+        'public_key': bigchaindb.config['keypair']['public'],
+        'keyring': bigchaindb.config['keyring'],
+        'api_endpoint': bigchaindb.config['api_endpoint']
+    })
+
--- a/docs/source/drivers-clients/http-client-server-api.rst
+++ b/docs/source/drivers-clients/http-client-server-api.rst
@@ -13,17 +13,21 @@ There are other configuration settings related to the web server (serving the HT
 
 The HTTP API currently exposes two endpoints, one to get information about a specific transaction, and one to push a new transaction to the BigchainDB cluster.
 
-.. http:get:: /transactions/(tx_id)
+.. http:get:: /transactions/{tx_id}
 
-   The transaction with the transaction ID `tx_id`.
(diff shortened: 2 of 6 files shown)


r-marques:
> upon POST /transactions/ should the reply contain more than the transaction (augmented with assignee)? 

I think not. This is just a temporary field related to the consensus protocol

> We could also provide the output of bigchaindb.write_transaction(tx)

I think we should abstract from the database replies. Internally we check if the write was successful and just return a 2XX with the transaction

bigchaindb/web/views/info.py
```diff
@@ -0,0 +1,25 @@
+"""This module provides the blueprint for some basic API endpoints.
+
+For more information please refer to the documentation in Apiary:
+ - http://docs.bigchaindb.apiary.io/
```
sbellem: I thought we were documenting in read the docs (https://bigchaindb.readthedocs.io/en/latest/drivers-clients/http-client-server-api.html)?

diminator: yes, need to change that

docs/source/drivers-clients/http-client-server-api.rst
```diff
@@ -66,20 +96,129 @@ The HTTP API currently exposes two endpoints, one to get information about a spe
...
+        }
+      }
 
    :statuscode 201: A new transaction was created.
+   :statuscode 400: The transaction was invalid and not created.
+
+   **Disclaimer**
+
+   ``CREATE`` transactions are treated differently from ``TRANSFER`` assets.
+   The reason is that a ``CREATE`` transaction needs to be signed by the federation nodes and not by the client.
```
sbellem: > ...  signed by the federation nodes ...

Do multiple nodes sign, or only one?

diminator: yeah, that's a separate issue probably, I assume now there is the one node you are connecting to

docs/source/drivers-clients/http-client-server-api.rst
```diff
@@ -66,20 +96,129 @@ The HTTP API currently exposes two endpoints, one to get information about a spe
...
+
+   The following python snippet at the client can be used to generate ``CREATE`` transactions before they can be pushed to the API server:
+
+   .. code-block:: python
+
+       from bigchaindb import util
+       tx = util.create_and_sign_tx(my_privkey, my_pubkey, my_pubkey, None, 'CREATE')
+
+   When POSTing ``tx`` to the API, the ``CREATE`` transaction will be signed by the federation nodes.
```
sbellem: same question as in https://github.com/bigchaindb/bigchaindb/pull/563/files#r75122213

the federation nodes

OR

a federation node

docs/source/drivers-clients/http-client-server-api.rst
```diff
@@ -66,20 +96,129 @@ The HTTP API currently exposes two endpoints, one to get information about a spe
...
 
    :statuscode 201: A new transaction was created.
+   :statuscode 400: The transaction was invalid and not created.
+
+   **Disclaimer**
+
+   ``CREATE`` transactions are treated differently from ``TRANSFER`` assets.
+   The reason is that a ``CREATE`` transaction needs to be signed by the federation nodes and not by the client.
+
+   The following python snippet at the client can be used to generate ``CREATE`` transactions before they can be pushed to the API server:
```
sbellem: about the wording: ".. in a client ..." maybe instead?

For client code examples, it may be preferable to avoid including/requiring server code.

sbellem:
:+1:

docs/source/drivers-clients/http-client-server-api.rst
diminator: > For client code examples, it may be preferable to avoid including/requiring server code.

yeah, but this is the server ofc, and there is a special case on how to push transactions...
you mean dropping this entire disclaimer?

docs/source/drivers-clients/http-client-server-api.rst
sbellem: ok, thanks!

PR #569, opened 2016-08-17
Add support for multiple ops in ChangeFeed

vrde:
This PR is to add support to listen to multiple operations in the ChangeFeed class.

(PR #359 depends on this one)


r-marques:
:+1:

PR #567, opened 2016-08-17
AWS users now generate their own SSH key pair

ttmc:
I made it so the instructions are for AWS users to generate their own SSH key pair (i.e. the one used to SSH to instances on AWS), rather than having Amazon do it for them. The user sends the public key to AWS, and they put the name of the SSH key pair in the AWS deployment configuration file.

This resolves #556

PR #572, opened 2016-08-17
Using Terraform to provision a one-machine BigchainDB node on AWS

### LINKED ISSUE #540: Python scripts to provision minimal production node on AWS
Write Python scripts using boto3 (the AWS Python SDK) to provision AWS instances, AWS settings, and AWS services for a minimal production node. Each AWS instance will be based on an AMI; many will be just an operating system (e.g. Ubuntu 14.04) but some might include more (e.g. a RethinkDB AMI or an nginx AMI).

Instances for:
- nginx
- BigchainDB
- RethinkDB proxy
- RethinkDB
- Monitoring?

Other AWS services:
- Elastic storage for RethinkDB on Amazon EBS

This issue doesn't include installing software or doing system configuration beyond what came with the AMIs. That will be handled by a configuration management system (probably Ansible) in a cloud-provider-independent way, as much as possible.

ttmc:
This pull request assumes that #567 is already merged into master, so check if that's true first.

Probably the easiest way to review this pull request is to follow the steps of the documentation in the section **Production Node Setup & Management**. That section has three pages, but you may also have to generate an SSH key pair as explained in the **Appendices - Basic AWS Setup**.

Main stuff in this pull request:

* Wrote some `.tf` Terraform files to provision the resources for a one-machine BigchainDB node on AWS
* Wrote some documentation about how to install Terraform
* Wrote some documentation about how to use Terraform to provision the one-machine node using the above-mentioned `.tf` files
* Note that it's assumed that the user/operator generated their own SSH key pair for use with AWS

This resolves #540


ttmc:
I don't know why Codecov is saying that "Merging #572 into master will decrease coverage by 1.65%".

At first, I thought it was because I forgot to update `codecov.yml` to tell it to ignore `ntools/*`. I did a commit to fix that and Codeov is still saying that coverage will go down. :confused:

r-marques:
Seems good :+1:
Just looked at the documentation and terraform files but did not actually deploy an instance

PR #573, opened 2016-08-18
Fixed docstring in is_valid_transaction()

### LINKED ISSUE #438: A little mis-discription
"is_valid_transaction(transaction)[source]
Check whether a transacion is valid or invalid.
Similar to validate_transaction but does not raise an exception if the transaction is valid."

The description above is from “http://bigchaindb.readthedocs.io/en/latest/appendices/the-Bigchain-class.html”.   And the last letter should be "invalid".

ttmc:
Fixes #438


r-marques:
:+1:

Issue #568, opened 2016-08-17
remove tx from backlog `after` written in bigchain


diminator:
see @ttmc comment in #555 

cc @r-marques @rhsimplex @sbellem @TimDaub @sohkai 

>I realized there's another state that we forgot about in the planning meeting yesterday: when a node pulls a transaction out of the backlog table (to consider putting it in its next block), it deletes the transaction from the backlog right away. For a short time, the transaction isn't in any table; it's just in the memory of the assigned node. As far as the RethinkDB cluster is concerned, that transaction doesn't exist. It's in a state that one might call "IN_MEMORY_OF_NODE_i".

>One solution would be to delete the transaction from the backlog later. The question becomes when. That's not so easy! You either end up with a time interval where the transaction isn't in any table (same as now, just shorter), or a time interval where the transaction is both in the backlog and in a block in the bigchain table.

>Dolo, the table clown, suggests adding a 23rd table. Good one, Dolo!

>Maybe just leave everything as is and add yet-another interpretation to the 404 case: the transaction might have been in a node's memory at query time.

The `bigchaindb.get_status(txid)` yields some `None` values over the lifetime of a transaction

```
undecided
None
None
valid
valid
```
This is probably due to the fact that the tx is temporarily in-memory. Can be easily resolved by removing tx from `backlog` after being written to `bigchain` and not before

ttmc:
Whoever makes this change should also make sure that the node assigned to the transaction deletes it from the backlog if it's invalid.

rhsimplex:
Ok, I have a plan for this.

![I'm Eisenhower!](http://pavetech.olhblogspace.com/wp-content/uploads/2012/09/Planning-Mtg.jpg)

I have a pending PR to reassign transactions that have been "stale" for too long #359.  One feature of my approach is to add a second timestamp at the time of assignment for internal bookkeeping.  That way, a separate process can periodically check that no transaction is getting too old.  If a transaction has been assigned for too long--e.g. a node has gone down--it gets reassigned.

We can use this functionality, and delete transactions after they are added to bigchain.  As @ttmc points out, we need to watch for:

1. There WILL be a moment when the `tx` is in the backlog and bigchain.  When/if another node picks it up before the node which wrote it can delete it from the backlog, it must be rejected.  Of course, it will eventually be rejected when the `tx` is validated during block construction, but this is a lot of wasted effort.  Maybe we should check for duplicate transactions immediately? Not full validation just duplicate checking.

[...]

PR #574, opened 2016-08-18
return block status

### LINKED ISSUE #557: Improvement proposal for get_transaction
relates to #555 

Right now `bigchaindb.get_transaction` returns a transaction that exists in the `bigchain` table that is not in an invalid block. However it does not return any information about the status of the transaction.

We could improve the logic of `get_transaction` so that it would return every possible state of a transaction. This would require also querying the `backlog` if the transaction is not on `bigchain`. If the transaction exists instead of returning the transaction we would also return the status of the transaction.

Proposed status:
- UNDECIDED_BACKLOG
- UNDECIDED_BIGCHAIN
- DECIDED
- None (transaction is not in the backlog or bigchain, worst case its only in an invalid block in bigchain)

rhsimplex:
Aims to resolve #557. Ready for review.

When a transaction is retrieved by `get_transaction`, a `validity` field is added with a value one of `valid`, `undecided`, or `backlog`. I chose this naming convention to be consistent with the block labels we already have.
```python
       # return if a block has been voted invalid                                                                                                                                                                                                                               
        BLOCK_INVALID = 'invalid'                                                                                                                                                                                                                                                
        # return if a block is valid, or tx is in valid block                                                                                                                                                                                                                    
        BLOCK_VALID = TX_VALID = 'valid'                                                                                                                                                                                                                                         
        # return if block is undecided, or tx is in undecided block                                                                                                                                                                                                              
        BLOCK_UNDECIDED = TX_UNDECIDED = 'undecided'                                                                                                                                                                                                                             
        # return if transaction is in backlog                                                                                                                                                                                                                                    
        TX_IN_BACKLOG = 'backlog'
```


r-marques:
Not sure if we should embed the the status in the dict itself or if we should return a tuple `(status, tx)`, because `tx` is the actual document that is stored in the database and the status is not stored anywhere. I think this may lead to confusion.

r-marques:
Also for this to work correctly #568 needs to be addressed or else there will be a period of time when the transaction is being processed where it does not exist in the backlog or bigchain

diminator:
> return a tuple (status, tx)

@r-marques @rhsimplex : updated to use a tuple when `include_status=True` (`False` by default)

bigchaindb/core.py
```diff
@@ -132,46 +137,59 @@ def write_transaction(self, signed_transaction, durability='soft'):
...
-                if validity[_id] == Bigchain.BLOCK_VALID:
-                    break
-
-            # Query the transaction in the target block and return
-            response = r.table('bigchain', read_mode=self.read_mode).get(target_block_id)\
-                .get_field('block').get_field('transactions')\
-                .filter(lambda tx: tx['id'] == txid).run(self.conn)
-
-            return response[0]
+            if validity:
```
r-marques: This `if` block can be simplified. If you enter the block the transaction is at least undecided
```python
if validity:
    tx_status = self.TX_UNDECIDED
    for _id in validity:
        if validity[_id] == Bigchain.BLOCK_VALID:
            tx_status = self.TX_VALID
            break
```

tests/db/test_bigchain_api.py
```diff
@@ -120,8 +120,22 @@ def test_read_transaction(self, b, user_vk, user_sk):
         block = b.create_block([tx_signed])
         b.write_block(block, durability='hard')
 
-        response = b.get_transaction(tx_signed["id"])
+        response, status = b.get_transaction(tx_signed["id"], include_status=True)
+        # add validity information, which will be returned
+        assert util.serialize(tx_signed) == util.serialize(response)
+        assert status == 'undecided'
```
r-marques: Better to use variable names if we decide to change the strings later.
`assert status == b.TX_UNDECIDED`

bigchaindb/core.py
diminator: not sure if that will work with `target_block_id`

diminator: > not sure if that will work with `target_block_id`

yeah, it will I think, since `target_block_id == _id` all the time

tests/db/test_bigchain_api.py
```diff
@@ -143,6 +157,26 @@ def test_read_transaction_invalid_block(self, b, user_vk, user_sk):
         assert response is None
 
     @pytest.mark.usefixtures('inputs')
+    def test_read_transaction_valid_block(self, b, user_vk, user_sk):
+        input_tx = b.get_owned_ids(user_vk).pop()
+        tx = b.create_transaction(user_vk, user_vk, input_tx, 'TRANSFER')
+        tx_signed = b.sign_transaction(tx, user_sk)
+        b.write_transaction(tx_signed)
```
r-marques: I don't think there is a need to write the transaction since you are manually creating and voting on the block

r-marques:
:+1:

PR #582, opened 2016-08-22
Docs: Put back the 'Set Up and Run a Cluster Node' page, for now

ttmc:
I had removed this page but I realized it's probably a good idea to keep it around for now. It can be removed in the future, once the new section on production node setup & management is more complete.


r-marques:
:+1:

PR #584, opened 2016-08-22
Api/555/transactions status endpoint

diminator:
follows up on #574
- endpoint on `/transactions/<ID>/status`
- docs/tests


r-marques:
@diminator can you sync up with master before starting review

tests/web/test_basic_views.py
```diff
@@ -62,3 +62,29 @@ def test_post_invalid_transfer_transaction_returns_400(b, client, user_vk, user_
 
     assert res.status_code == 400
 
+
+@pytest.mark.usefixtures('inputs')
+def test_post_invalid_transfer_transaction_returns_400(b, client, user_vk, user_sk):
+    to_keypair = crypto.generate_key_pair()
+    input_valid = b.get_owned_ids(user_vk).pop()
+    transfer = b.create_transaction(user_vk, to_keypair[0], input_valid, 'TRANSFER')
```
r-marques: `to_keypair[0]` is a private key. It doesn't matter in this case but we should do it correctly

r-marques:
:+1:

r-marques:
:+1:

Issue #587, opened 2016-08-23
server with class based views


diminator:
- [ ] will be easier to port to other http frameworks
- [ ] will also address #585 

proposal: use [flask restful](https://flask-restful.readthedocs.io)

diminator:
solved by #588

Issue #585, opened 2016-08-22
server api: to slash or not slash?


sbellem:
This ticket aims to reconsider the current position on trailing slashes as it is inconvenient for future client implementations.

See related comment: https://github.com/bigchaindb/bigchaindb/issues/362#issuecomment-241419134

diminator:
solved by #588

PR #588, opened 2016-08-23
flask restful with class views

diminator:
solves #585, #587

### Note:
need to `pip install` due to new dependencies

### DIFF
--- a/bigchaindb/web/views/transactions.py
+++ b/bigchaindb/web/views/transactions.py
@@ -65,8 +65,8 @@ class TransactionStatusApi(Resource):
             tx_id (str): the id of the transaction.
 
         Return:
-            A JSON string containing the status of the transaction.
-            Possible values: "valid", "invalid", "undecided", "backlog"
+            A ``dict`` in the format ``{'status': <status>}``, where ``<status>``
+is one      of "valid", "invalid", "undecided", "backlog".
         """
 
         pool = current_app.config['bigchain_pool']
@@ -85,7 +85,7 @@ class TransactionListApi(Resource):
         """API endpoint to push transactions to the Federation.
 
         Return:
-            A JSON string containing the data about the transaction.
+            A ``dict`` containing the data about the transaction.
         """
         pool = current_app.config['bigchain_pool']
         monitor = current_app.config['monitor']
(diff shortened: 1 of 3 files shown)


bigchaindb/web/views/transactions.py
```diff
@@ -35,74 +36,83 @@ def record(state):
...
+        Args:
+            tx_id (str): the id of the transaction.
 
-    Return:
-        A JSON string containing the data about the transaction.
-    """
-    pool = current_app.config['bigchain_pool']
-    monitor = current_app.config['monitor']
+        Return:
+            A JSON string containing the status of the transaction.
```
vrde: Since the conversion to JSON is happening outside the method, I'd change the description to:
```
A ``dict`` in the format ``{'status': <status>}``, where ``<status>``
is one of "valid", "invalid", "undecided", "backlog".
```

setup.py
```diff
@@ -102,6 +102,7 @@ def check_setuptools_features():
         'logstats==0.2.1',
         'base58==0.2.2',
         'flask==0.10.1',
+        'flask-restful==0.3.5',
```
vrde: I think we can safely use: `flask-restful~=0.3.0` in this way we pin the `0.3.x` version of `flask-restful`. They should use semantic versioning. (Yes, we should apply the same pattern to the other requirements, we can do it later).

tests/web/test_basic_views.py
```diff
@@ -13,15 +13,22 @@ def test_get_transaction_endpoint(b, client, user_vk):
     input_tx = b.get_owned_ids(user_vk).pop()
     tx = b.get_transaction(input_tx['txid'])
     res = client.get(TX_ENDPOINT + input_tx['txid'])
+    assert res.status_code == 200
     assert tx == res.json
+
+    res = client.get(TX_ENDPOINT + input_tx['txid'] + '/')
```
vrde: Do we need the trailing `/`?

diminator: we support both using `strict_slashes=False`

vrde:
@diminator nice work in upgrading to `flask-restful`!

tests/web/test_basic_views.py
sohkai: Just a thought: what if we had parameterized tests for each of these API tests that made sure both trailing and no-trailing slash endpoints work?

PR #589, opened 2016-08-23
Feat/551/ansible run ntp daemon

### LINKED ISSUE #551: Minimal Ansible thing to install, configure & run an NTP daemon
This is a sub-issue of issue #498.

ttmc:
* Clarified (in the documentation) how the Terraform/Ansible template is just an example of how one might provision and deploy a production BigchainDB node (but they can use whatever tools they prefer)
* New docs about how to install Ansible
* New Ansible stuff to ensure base software is installed, and to ensure an NTP daemon is installed/running on a remote host (a one-machine BigchainDB production node)
* Documentation on how to use that Ansible stuff, i.e. how to run the Ansible playbook
* Updated `.gitignore`
* Fixed a bunch of broken links

Resolves #551


docs/source/prod-node-depl-tplt/start-one-m-node.md
```diff
@@ -0,0 +1,45 @@
...
+
+The next step is to run the Ansible playbook `one-m-node.yml`:
+```text
+# cd to the directory .../bigchaindb/ntools/one-m/ansible
+ansible-playbook -i hosts --private-key ~/.ssh/<key-name> one-m-node.yml
+```
+
+where `<key-name>` should be replaced by the name of the SSH private key you created earlier (for SSHing to the host machine at your cloud hosting provider).
+
+What did you just do? Running that playbook ensures all the software necessary for a one-machine BigchainDB node is installed, configured, and running properly. You can run that playbook on a regular schedule to ensure that the system stays properly configured. If something is okay, it does nothing; it only takes action when something is not as-desired.
```
r-marques: It is not installed, will ansible install it?

r-marques:
:+1:

docs/source/prod-node-depl-tplt/start-one-m-node.md
ttmc: If an NTP daemon isn't installed, then one will be installed, yes. As the next line notes, "At the time of writing, the playbook only installs, configures and runs an NTP daemon, but more is coming soon."

Issue #500, opened 2016-07-27
Remove plugin architecture


vrde:
# Reasons to remove it:
- code difficult to navigate
- we are not using it
- too much abstraction

# Reasons to keep it:
- might be useful in the future to simulate malicious nodes or having different implementation of the consensus
- @thedoctor: are you using the plugin in your fork?

# Conclusion
Isolate and refactor the code for the consensus before abstract it (we need a stable consensus layer)

ttmc:
Just a reminder: there's a page about consensus plugins in the docs: https://bigchaindb.readthedocs.io/en/latest/appendices/consensus.html

thedoctor:
Sorry guys, I just saw this — I'm not actively using bigchain atm and won't have much time to contribute unless there's something specific I can do to make the code more usable for you guys.

TimDaub:
> @sbellem Is this completed now with #641 merged?

Yes, this ticket has been resolved.

PR #590, opened 2016-08-23
[WIP] Feat/586/integrate tx model

TimDaub:
Hey,

TODO:

- [x] Fix all tests
- [x] [Remove consensus plugin docs](https://github.com/bigchaindb/bigchaindb/pull/590#issuecomment-244771034)
- [x] ~~Change all docs to the new transaction model~~
- [x] Remove all `TODO`s currently in the diff
- [x] Merge PR over [here](https://github.com/bigchaindb/bigchaindb-common/pull/4)
- [ ] Discuss what will happen with the [Cryptocondition tests](https://github.com/bigchaindb/bigchaindb/pull/590/commits/37fb742a7e44da9b1c780860875586ee4c18491a)
- [x] Remove old client and replace with new one
- [x] Replace all util functions with the ones from commons (`timestamp` e.g.)
- [x] Test Block and Transaction validation functions
- [x] Increase test coverage to at least 0.00% diff
- [ ] Rebase against master :see_no_evil:


ttmc:
Note: Removing the consensus plugin is issue #500 so I guess this PR will resolve that issue too. As noted in that issue:

> Just a reminder: there's a page about consensus plugins in the docs: https://bigchaindb.readthedocs.io/en/latest/appendices/consensus.html

I guess that will have to be deleted.

TimDaub:
> I guess that will have to be deleted.

Thanks, I've added it to the TODOs of this PR.

.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
 
 install:
   - sudo apt-get install rethinkdb
-  - pip install -e .[test]
+  - pip install -e .[test] --process-dependency-links
```
TimDaub: Remove this before merging

TimDaub:
> Change all docs to the new transaction model

@sbellem and I talked.
We both agree that fixing https://github.com/bigchaindb/bigchaindb/blob/master/tests/doc/run_doc_python_server_api_examples.py will take a long time (up to a week probably), which is time we do not want to afford (as this PR would get stuck for another week).

Hence, we'll ignore the doc_python_server_api.py and .md file for this PR and properly fix (meaning also refactor) it right after merging this PR.
As most people (hopefully) pull bigchaindb from pip (and since we'll not release a new version before the documentation is not fixed), we shouldn't cause problems to our users that want to replay these examples.

@vrde @r-marques @rhsimplex Thoughts?

diminator:
> Hence, we'll ignore the doc_python_server_api.py and .md file for this PR and properly fix (meaning also refactor) it right after merging this PR.

ok, we can simply keep on serving that file for now as long as it doesnt break?

setup.py
```diff
@@ -105,6 +102,10 @@ def check_setuptools_features():
         'requests==2.9',
         'gunicorn~=19.0',
         'multipipes~=0.1.0',
+        'bigchaindb-common>=0.0.2',
+    ],
+    dependency_links=[
+        'git+https://github.com/bigchaindb/bigchaindb-common.git@integrate-transaction-model#egg=bigchaindb_common-0.0.2',
```
TimDaub: Change this to an actual version of BigchainDB-common.

TimDaub:
Closing this in regard of #641

Issue #125, opened 2016-03-09
Define a json template for digital assets


r-marques:
At this point the digital asset looks like this:
```javascript
{
    "hash": "<sha3 hash of the payload>",
    "payload": {
        ...
    }
}
```
The digital asset is defined by the payload which is a generic json object. The hash is added by bigchaindb to allow users to query bigchaindb for the digital asset hash.

This is a simple way to keep track of digital assets on bigchaindb but it is not enough.
We need a generic template for a digital asset so that we can provide more full feature queries on the digital assets.

r-marques:
# Proposal a JSON template for digital assets.

This proposal may change as needed to address comments or during implementation. Once the implementation is finished we will add the spec to the documentation.

Right now the payload field is being used for 2 different purposes: to define the digital asset; and to add extra information to transactions.

The proposal is to divide them into two separate fields. A digital asset field that is created in a `CREATE` transaction in which its properties are used by the consensus rules. And a payload field which as generic information that users may want to append to their transactions, and are not used by the consensus rules.

A digital asset may have several characteristics:
- Unique ID
- Whether the digital asset is divisible or un-divisible
- Whether the digital asset can or cannot be updated in the future
- The amount of the digital asset
- If divisible it can be possible to refill in the future or not.
- Data field provided by the user

The transaction would have an `asset` field with this structure:
```json
{
    "id": "<uuid>",
    "divisible": "<true | false>",
    "updatable": "<true | false>",
    "amount": "<int>",
    "refillable": "<true | false>",
    "data": "<json document>"
}
```

- __id__: Unique identifier 
- __divisible__: Whether the asset is divisible or not. Defaults to `false`.
- __updatable__: Whether the data in the asset can be updated in the future or not. Defaults to `false`.
- __amount__: The amount of "shares". Only relevant if the asset is marked as divisible. Defaults to `1`.
- __refillable__: Whether the amount of the asset can change after its creation. Defaults to `false`.
- __data__: A user supplied json document with custom information about the asset. Defaults to `null`.

#### Changes to be the current transaction model.
`data` field is renamed to `payload`. Remove the uuid field and nested payload.

Previous `CREATE` transaction model:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "version": "transaction version number",
        "fulfillments": [],
        "conditions": [],
        "operation": "<string>",
        "timestamp": "<timestamp from client>",
        "data": {
[...]

Issue #586, opened 2016-08-23
Integrate new transaction model


TimDaub:
The isolation of the transaction model was done here: https://github.com/bigchaindb/bigchaindb-common/pull/4

This ticket's task is to take the model and use it in BigchainDB.

TimDaub:
As an update: This will probably be deferred till after September 1, as we have more important stuff to do right now.

PR #599, opened 2016-08-25
Feat/564/implement digital asset template

### LINKED ISSUE #564: Implement the digital asset template
This issue is to be done after #125 

New `CREATE` transaction model:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "version": "transaction version number",
        "fulfillments": [],
        "conditions": [{
          "cid": "<condition index>",
          "condition": { },
          "owners_after": ["<new owner public key>"],
          "amount": "<int>"
        }],
        "operation": "<string>",
        "timestamp": "<timestamp from client>",
        "asset": {
            "id": "<uuid>",
            "divisible": "<true | false>",
            "updatable": "<true | false>",
            "refillable": "<true | false>",
            "data": "<json document>"
        },
        "payload": "<json document>"
    },
}
```

Implementation details
We need to add extra arguments to `util.create_tx(..., data=None, divisible=False, updatable=False, amount=1, refillable=False)`
Probably safer to add the uuid server side.

This issue will only implement the digital asset template.
`divisible = True`, `updatable = True`, `refillable = True` won't be implemented in this issue.
Since we don't have divisible assets yet `amount` will always default to 1

r-marques:
resolves #564 #125 

##### Changes introduced

New `CREATE` transaction model:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "version": "transaction version number",
        "fulfillments": [],
        "conditions": [{
          "cid": "<condition index>",
          "condition": { },
          "owners_after": ["<new owner public key>"],
          "amount": "<int>"
        }],
        "operation": "CREATE",
        "timestamp": "<timestamp from client>",
        "asset": {
            "id": "<uuid>",
            "divisible": "<true | false>",
            "updatable": "<true | false>",
            "refillable": "<true | false>",
            "data": "<json document>"
        },
        "metadata": {
             "id": "<uuid>",
             "data": "<json document>"
        },
    },
}
```

New `TRANSFER` transaction model:
```json
{
    "id": "<sha3 hash>",
    "transaction": {
        "version": "transaction version number",
        "fulfillments": [],
        "conditions": [{
          "cid": "<condition index>",
          "condition": { },
          "owners_after": ["<new owner public key>"],
          "amount": "<int>"
        }],
        "operation": "TRANSFER",
        "timestamp": "<timestamp from client>",
        "asset": {
            "id": "<uuid>",
        },
         "metadata": {
             "id": "<uuid>",
             "data": "<json document>"
        },
    },
}
```
- The `CREATE` transaction contains the actual asset.
- The `TRANSFER` transaction only contains the id of the asset
- The amount is specified in the outputs / conditions. This does not implement divisible assets so the amount at this point can only be `1`
- `util.create_tx` and `consensus.validate_transaction` was updated to also create and validate the asset
- most of the code related to digital assets is under `bigchaindb.assets.py`
- Renamed `payload` with `metadata` and `get_tx_by_payload_uuid` with `get_tx_by_metadata_id`
- Created `get_txs_by_asset_id`
- Created a secondary index `asset_id`
- Fixed and wrote tests

With this change every transaction is related to only one asset. For this reason multiple input and outputs now only make sense for divisible assets (I had to skip some tests related to multiple input multiple output, will unskip them once I implement divisible assets). I don't think there is a use case to have a single transaction dealing with multiple assets, but if there is that is a feature we can add after.

Will write the documentation in a separate pull request


ttmc:
Just some minor comments:

* I like `metadata`
* For the `CREATE` transaction model, won't the "operation" always be "CREATE"?
* Similar comment for `TRANSFER`
* You wrote:

> The amount is specified in the outputs / conditions. This does not implement digital assets so the amount at this point can only be `1`

I think you meant "divisible assets".

Regarding updating the docs:  a lot of the examples in the docs will have different outputs now. I wonder if there's a way to auto-generate those outputs using Sphinx + outputs from our actual current code. That would save a lot of hassle. (Of course, there's no guarantee that the text of the docs would stay consistent with the auto-generated outputs. Oh well, nothing's perfect.)

r-marques:
Yes I meant divisible assets.
Fix the `operation` in the models

sohkai:
cc @TimDaub in case you haven't seen this yet.

bigchaindb/assets.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    if not isinstance(txids, list):
+        txids = [txids]
+
+    asset_ids = []
+    for txid in set(txids):
+        tx = bigchain.get_transaction(txid)
+        if tx is None:
+            raise TransactionDoesNotExist('Transaction with txid `{}` does not exist in the bigchain'.format(txid))
```
sohkai: Should probably be documented in the `Raises` above.

bigchaindb/assets.py
```diff
@@ -0,0 +1,78 @@
...
+        asset_ids.append(tx['transaction']['asset']['id'])
+
+    asset_ids = set(asset_ids)
```
sohkai: Could this be a set from the start?

bigchaindb/assets.py
```diff
@@ -0,0 +1,78 @@
...
+    if len(asset_ids) > 1:
+        raise AssetIdMismatch("All inputs of a transaction need to have the same asset id.")
+    return asset_ids.pop()
+
+
+def validate_asset_creation(asset_data, divisible, updatable, refillable, amount):
```
sohkai: Should all these parameters default to `None`? Maybe make everything after `asset_data` keyword only?

bigchaindb/assets.py
```diff
@@ -0,0 +1,78 @@
...
+        asset_data (Optional[dict]): dictionary describing the digital asset (only used on a create transaction)
+        divisible (Optional[boolean): Whether the asset is divisible or not. Defaults to `False`.
+        updatable (Optional[boolean]): Whether the data in the asset can be updated in the future or not.
+                                       Defaults to `False`.
+        refillable (Optional[boolean]): Whether the amount of the asset can change after its creation.
+                                        Defaults to `False`.
+        amount (Optional[int]): The amount of "shares". Only relevant if the asset is marked as divisible.
+                                Defaults to `1`.
+    """
+    if asset_data is not None and not isinstance(asset_data, dict):
```
sohkai: No need to double check `None` and `isinstance`.

bigchaindb/consensus.py
```diff
@@ -162,6 +171,14 @@ def validate_transaction(bigchain, transaction):
                     raise exceptions.DoubleSpend(
                         'input `{}` was already spent'.format(fulfillment['input']))
 
+                txids.append(fulfillment['input']['txid'])
```
sohkai: Seems like a bit of a waste of effort if we've just got the transaction before on https://github.com/bigchaindb/bigchaindb/pull/599/files#diff-d82bc8f43aa0fddab154084c829f426aR160

bigchaindb/consensus.py
```diff
@@ -139,12 +139,21 @@ def validate_transaction(bigchain, transaction):
                 raise exceptions.OperationError(
                     'Only federation nodes can use the operation `CREATE`')
 
+            # validate digital asset
+            # A create transaction will have only one output
+            asset = transaction['transaction']['asset']
+            assets.validate_asset_creation(asset['data'], asset['divisible'], asset['updatable'],
+                                           asset['refillable'], transaction['transaction']['conditions'][0]['amount'])
+            # TODO: Validate uuid?
```
sohkai: This would be really nice, but I guess runs into all sorts of race conditions (and involves a query)?

bigchaindb/assets.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def get_asset_id(txids, bigchain):
+    """Get the asset id from a list of transaction ids.
+
+    This is useful when we want to check if the multiple inputs of a transaction
+    are related to the same asset id.
+
+    Args:
+        txids (list): list of transaction ids.
```
sohkai: Not sure if we ignore `bigchain` arguments across the board in the docstrings, but otherwise, we should add it to the argument list.

bigchaindb/util.py
```diff
@@ -157,14 +159,22 @@ def create_tx(owners_before, owners_after, inputs, operation, payload=None):
...
+        amount (Optional[int]): The amount of "shares". Only relevant if the asset is marked as divisible.
+                                Defaults to `1`.
 
     Returns:
         dict: unsigned transaction.
 
 
     Raises:
-        TypeError: if the optional ``payload`` argument is not a ``dict``.
+        TypeError: if the optional optinal arguments are the wrong type.
```
sohkai: Slight typo.

bigchaindb/util.py
```diff
@@ -216,23 +230,37 @@ def create_tx(owners_before, owners_after, inputs, operation, payload=None):
...
+    asset = None
+
     # transfer
     if inputs:
+        # Since a transaction is specific to a digital asset we need to check if all the inputs are from the same
+        # digital asset (see https://github.com/bigchaindb/bigchaindb/issues/125#issuecomment-241020139)
+        # Here we need to get the transactions pointed by the inputs to check the asset id and the amounts if its a
+        # divisible asset . This is inefficient because we are querying past transactions multiple times
+        # (here and later on in the sign_tx)
+        # This can be improved and probably cached with @TimDaub transaction model
```
sohkai: 👍

bigchaindb/assets.py
r-marques: We need all the parameters to validate a the asset. The documentation is wrong in this case

bigchaindb/consensus.py
r-marques: true

bigchaindb/consensus.py
r-marques: Not sure if we need to validate the uuid. Its set server side. And I don't think we will have collisions with uuid4

bigchaindb/assets.py
r-marques: Yeah I don't know. This is just to make sure we don't just instantiate unnecessary bigchain instances.

tests/assets/test_digital_assets.py
```diff
@@ -0,0 +1,192 @@
...
+    tx_input = b.get_owned_ids(user_vk).pop()
+    tx_create = b.get_transaction(tx_input['txid'])
+    tx_transfer = b.create_transaction(user_vk, user_vk, tx_input, 'TRANSFER')
+    tx_transfer_signed = b.sign_transaction(tx_transfer, user_sk)
+
+    assert b.validate_transaction(tx_transfer_signed) == tx_transfer_signed
+    assert tx_transfer_signed['transaction']['asset']['id'] == tx_create['transaction']['asset']['id']
+
+
+def test_validate_bad_asset_creation(b, user_vk):
```
sohkai: Might be useful to leave a little comment at the top of each of the test blocks here to indicate what they do... didn't catch on that the validation was failing because `'divisible'` should be boolean for quite a while :blush:

tests/assets/test_digital_assets.py
```diff
@@ -0,0 +1,192 @@
...
+    tx = b.create_transaction(b.me, user_vk, None, 'CREATE')
+    tx['transaction']['conditions'][0]['amount'] = 0
+    tx['id'] = get_hash_data(tx['transaction'])
+    tx_signed = b.sign_transaction(tx, b.me_private)
+    with pytest.raises(AmountError):
+        b.validate_transaction(tx_signed)
+
+
+@pytest.mark.usefixtures('inputs')
+def test_validate_bad_asset_transfer(b, user_vk, user_sk):
```
sohkai: A more descriptive test name might be nice... I think the validation is failing because of a bad asset id?

tests/assets/test_digital_assets.py
```diff
@@ -0,0 +1,192 @@
...
+
+@pytest.mark.usefixtures('inputs')
+def test_get_asset_id_transfer_transaction(b, user_vk, user_sk):
+    from bigchaindb.assets import get_asset_id
+
+    tx_input = b.get_owned_ids(user_vk).pop()
+    # create a transfer transaction
+    tx_transfer = b.create_transaction(user_vk, user_vk, tx_input, 'TRANSFER')
+    tx_transfer_signed = b.sign_transaction(tx_transfer, user_sk)
+    # create a block
```
sohkai: I imagine this is used elsewhere to simulate a block creation... maybe put it into a test util?

bigchaindb/consensus.py
sohkai: Not sure if we handle this somewhere else with malicious nodes, but in case one of them changed a transaction's asset id to something else?

sohkai:
> * I like `metadata`

Same :)

I guess the more difficult question now is how we'll integrate this with #586 coming...

bigchaindb/consensus.py
r-marques: This is only important at creation time. To things can happen:
- The asset_id is a duplicate
- The asset_id is not a valid uuid in canonical form
After that all nodes independently check that the asset_id matches the inputs.

I think that this would be solved once we improve the asset creation where a client would contact several nodes instead on only one.

bigchaindb/consensus.py
r-marques: This is related to https://github.com/bigchaindb/bigchaindb/pull/599#discussion-diff-78031204

`get_asset_id` is used in `create_tx` and `validate_transaction`. In `validate_transaction` its true that I already have the input transactions but in `create_tx` I don't. I will need to retrieve the input transactions in `create_tx` before calling `get_asset_id`

bigchaindb/consensus.py
r-marques: The change you proposed will add a bit more code to create_tx but it will simplify validate_transaction which I think is a good tradeoff since we want validate_transaction to be as fast as possible

sohkai: > I think that this would be solved once we improve the asset creation where a client would contact several nodes instead on only one.

In that case though, assuming we keep the backlog algorithm the same and assign each transaction to only one node in the end, wouldn't we still be vulnerable to that one node manipulating the asset? (In fact, that one node could do anything with the transaction it receives; I'm sure you and @ttmc have talked about the fault tolerance of this... it's probably in an issue / project somewhere? For now, are we assuming we can trust each deployed node?)

bigchaindb/consensus.py
r-marques: The change you proposed will add a bit more code to create_tx but it will simplify validate_transaction which I think is a good tradeoff since we want validate_transaction to be as fast as possible

bigchaindb/consensus.py
r-marques: If you have several nodes and/or the client signing the creation then it would be impossible for a node to change the transaction. Worst case scenario he could decide to just drop the transaction and not write it to the backlog but that is another problem

tests/assets/test_digital_assets.py
r-marques: I try leave comments where I think its needed but usually unit tests are quite verbose and self explanatory. This pretty much follows the exceptions thrown by `validate_asset_creation`

tests/assets/test_digital_assets.py
r-marques: This can be proposed in a new issue maybe part of a refactor of the test suite.
Creating a fixture and changing all unit tests to use it are out of the scope of this issue

r-marques:
Should not be different than any other code conflict. The only difference is that #586 is a really big change

r-marques:
deprecated in favor of #680

PR #600, opened 2016-08-25
Ansible playbook now sets up storage for the underlying DB

### LINKED ISSUE #552: Minimal Ansible thing to setup filesystem/storage for RethinkDB
This is a sub-issue of issue #498.

- Set up the file system on the RethinkDB storage device
- Mount that device on `/data`
- Set the I/O scheduler?
- Update `/etc/fstab`

See `prep_rethinkdb_storage()` in https://github.com/bigchaindb/bigchaindb/blob/master/deploy-cluster-aws/fabfile.py

ttmc:
The Ansible playbook now formats the external block storage device at `/dev/xdvp` with the ext4 file system, mounts the `/data` directory on that device, and edits `/etc/fstab` so those changes will persist through a reboot.

This PR also includes additions to ensure that the latest pip/pip3 and setuptools (Python packages) are installed. That turned out to be trickier than expected: doing "sudo apt-get python3-pip" and "pip3 install -U pip" would fail.

Resolves #552


r-marques:
:+1:

PR #609, opened 2016-08-30
Delete transactions after block is written

rhsimplex:
Aims to address #568.  Depends on #359. Ready for review, pending the approval of #359.

Crucial ideas:

- [x] Include stale transaction monitoring code
- [x] Transactions deleted from backlog _after_ being written to a block
- [x] Transactions must be checked for duplicates in valid/undecided blocks and discarded if already in a block

### DIFF
--- a/bigchaindb/pipelines/block.py
+++ b/bigchaindb/pipelines/block.py
@@ -47,7 +47,8 @@ class Block:
     def validate_tx(self, tx):
         """Validate a transaction.
 
-        Also checks if a transaction already exists in the blockchain.
+        Also checks if the transaction already exists in the blockchain. If it
+        does, or it's invalid, it's deleted from the backlog immediately.
 
         Args:
             tx (dict): the transaction to validate.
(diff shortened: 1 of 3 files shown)


ttmc:
Also, if the block-building node determines that a tx is invalid, then that node must delete the tx from the backlog table. (The node did that implicitly before, because it deleted the tx from the backlog *first*, then checked tx validity, and if the tx was invalid, it was not passed on to be included in a block.)

rhsimplex:
@ttmc fixed

bigchaindb/pipelines/block.py
```diff
@@ -44,34 +44,43 @@ def filter_tx(self, tx):
...
+                # to a new block. We can delete and drop it.
+                r.table('backlog').get(tx['id']) \
+                        .delete(durability='hard') \
+                        .run(self.bigchain.conn)
+                return None
+
+        tx_validated = self.bigchain.is_valid_transaction(tx)
+        if tx_validated:
             return tx
+        else:
```
r-marques: I see the need for this and I think its fine for now but I am a bit concerned because this is the most critical method when in comes to speed and this will block doing an io operation, specially with `durability='hard'`.

This is more a note for when we do the benchmarks

r-marques:
:+1:

PR #614, opened 2016-09-01
Add integration testing

rhsimplex:
This PR has morphed somewhat. Main addition is a new test folder for integration tests.

The fixture starts up the `vote`, `block`, etc processes for each test.  I also included some sample tests.

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -1,0 +1,66 @@
+import time
+import pytest
+
+from bigchaindb import Bigchain
+
+
+@pytest.fixture
+def inputs(user_pk):
+    from bigchaindb.models import Transaction
+
+    b = Bigchain()
+
+    # create blocks with transactions for `USER` to spend
+    for block in range(4):
+        transactions = [
+            Transaction.create(
+                [b.me], [([user_pk], 1)],
+                metadata={'i': i})
+            .sign([b.me_private])
+            for i in range(10)
+        ]
+        block = b.create_block(transactions)
+        b.write_block(block, durability='hard')
+
+
+@pytest.mark.usefixtures('processes')
+def test_fast_double_create(b, user_pk):
+    from bigchaindb.models import Transaction
+    tx = Transaction.create([b.me], [([user_pk], 1)],
+                            metadata={'test': 'test'}) \
+            .sign([b.me_private])
+
+    # write everything fast
+    b.write_transaction(tx)
+    b.write_transaction(tx)
+
+    time.sleep(2)
+    tx_returned = b.get_transaction(tx.id)
+
+    # test that the tx can be queried
+    assert tx_returned == tx
+    # test the transaction appears only once
+    last_voted_block = b.get_last_voted_block()
+    assert len(last_voted_block.transactions) == 1
+    assert b.backend.count_blocks() == 2
+
+
+@pytest.mark.usefixtures('processes')
+def test_double_create(b, user_pk):
+    from bigchaindb.models import Transaction
+    tx = Transaction.create([b.me], [([user_pk], 1)],
+                            metadata={'test': 'test'}) \
+            .sign([b.me_private])
+
+    b.write_transaction(tx)
+    time.sleep(2)
+    b.write_transaction(tx)
+    time.sleep(2)
+    tx_returned = b.get_transaction(tx.id)
+
+    # test that the tx can be queried
+    assert tx_returned == tx
+    # test the transaction appears only once
+    last_voted_block = b.get_last_voted_block()
+    assert len(last_voted_block.transactions) == 1
+    assert b.backend.count_blocks() == 2
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ tests_require = [
     'pep8',
     'flake8',
     'pylint',
-    'pytest',
+    'pytest>=3.0.0',
     'pytest-cov>=2.2.1',
     'pytest-xdist',
     'pytest-flask',
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,35 @@
+import pytest
+from bigchaindb.pipelines import block, election, vote, stale
+
+# TODO: fix this import madness
+from ..db import conftest
+
+
+@pytest.fixture(scope='module', autouse=True)
+def restore_config(request, node_config):
+    from bigchaindb import config_utils
+    config_utils.set_config(node_config)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def setup_database(request, node_config):
+    conftest.setup_database(request, node_config)
+
+
+@pytest.fixture(scope='function', autouse=True)
+def cleanup_tables(request, node_config):
+    conftest.cleanup_tables(request, node_config)
+
+
+@pytest.fixture
+def processes(b):
+    b.create_genesis_block()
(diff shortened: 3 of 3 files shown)


rhsimplex:
This is tricky.  When should we check for a duplicate transaction? Logically when voting (block validation).  But what happens when two identical transactions are in `undecided` blocks?  

1. If the voters account only for `tx` in valid blocks, then both undecided blocks can be voted valid simultaneously.  
2. If voters account for `tx` in undecided blocks, then neither can become valid, as they will both see a transaction is in an undecided block.

rhsimplex:
Discussed with @ttmc. If #609 is merged, we can be assured that the `tx` is always located in the backlog or bigchain.  Since the `id` is the primary key of the backlog, we're safe there. This is not a complete solution but does help against clients trying to insert duplicate transactions.

rhsimplex:
So, these tests pass now that #359 and #609 are merged.  I think there are other cases where this is possible though, will investigate further.  This should cover the double-create 'attack,' though.

diminator:
reproduce by:
```
git clone git@github.com:bigchaindb/hyperchaindb.git
git checkout bug/doublespent-on-owned
cd bigchaindb_workshop
# virtualenv
pip install -e . [dev]
# start bigchaindb
bigchaindb start
# start webserver in terminal 2 (cd bigchaindb_workshop)
python -m backend.server
# run example in terminal 3 (cd bigchaindb_workshop)
python -m exercises.bdb.ex1_tx_create_transfer_solution
```

You will now see the following output:

```
[2016-09-28 17:01:00 +0200] [30802] [INFO] Booting worker with pid: 30802
INFO:bigchaindb.pipelines.block:Write new block 1c21b76bcf9ffa522b54f5589c9ae73b871f419ec80be7807dddc40e9b7911fb with 1 transactions
INFO:bigchaindb.pipelines.block:Write new block 9baf086364c79ef20b84c5a8178d4f3f6317b6299df6ca1aa99a6608bb052488 with 1 transactions
INFO:bigchaindb.pipelines.block:Write new block dd25da2c599451f7dd267291004cc01add98dae38f6060c4a76f581ecc05abae with 1 transactions
INFO:bigchaindb.pipelines.election:Rewriting 1 transactions from invalid block dd25da2c599451f7dd267291004cc01add98dae38f6060c4a76f581ecc05abae
INFO:bigchaindb.pipelines.block:Write new block bb8f254f4df521253e75c929fec698bb8fe928682d87fad45612fc9d1f102b2b with 1 transactions
```

Which means that the block was invalid and rewritten...

Running the example again (`python -m exercises.bdb.ex1_tx_create_transfer_solution`) yields Doublespend mania on `get_owned_ids`:

```
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/dimi/code_ascribe/hyperchaindb/src/github.com/hyperchaindb/bigchaindb_workshop/exercises/bdb/ex1_tx_create_transfer_solution.py", line 28, in <module>
    printd(get_tx_owned(dimi['vk']))
  File "/home/dimi/code_ascribe/hyperchaindb/src/github.com/hyperchaindb/bigchaindb_workshop/exercises/bdb/driver.py", line 35, in get_tx_owned
    return b.get_owned_ids(vk)
  File "/home/dimi/code_ascribe/hyperchaindb/src/github.com/hyperchaindb/bigchaindb_workshop/venv/lib/python3.5/site-packages/bigchaindb/core.py", line 368, in get_owned_ids
    if not self.get_spent(tx_input):
  File "/home/dimi/code_ascribe/hyperchaindb/src/github.com/hyperchaindb/bigchaindb_workshop/venv/lib/python3.5/site-packages/bigchaindb/core.py", line 317, in get_spent
    tx_input['txid']))
bigchaindb.exceptions.DoubleSpend: `74bdebbb947bf9c99d7e1510e8b97ad3ba02ea74f97dfe1042a832f64d6f8984` was spent more then once. There is a problem with the chain
```

diminator:
maybe also add a test that does the following: 
- create invalid block with some faulty tx's and good ones
- all tx's go back to backlog
- on next block: 
 - faulty tx gets omitted
 - other tx go into valid block

rhsimplex:
The example has fallen behind the refactor, unfortunately.

diminator:
yeah, you can close it - will reimplement the example with the new version

tests/integration/test_integration.py
```diff
@@ -0,0 +1,96 @@
+import time
+import pytest
+import rethinkdb as r
+
+from bigchaindb import Bigchain
+
+
+@pytest.fixture
+def inputs(user_vk):
```
sohkai: Any reason why this wasn't in the conftest?

tests/integration/test_integration.py
```diff
@@ -0,0 +1,96 @@
...
+            Transaction.create(
+                [b.me], [user_vk], metadata={'i': i}).sign([b.me_private])
+            for i in range(10)
+        ]
+        block = b.create_block(transactions)
+        b.write_block(block, durability='hard')
+
+
+@pytest.mark.usefixtures('processes')
+def test_fast_double_create(b, user_vk):
```
sohkai: I was wondering about a fast double transfer earlier today; something along the lines of:

```python
b.write_transaction(create_tx)

time.sleep(2)
b.write_transaction(transfer_tx_1)
b.write_transaction(transfer_tx_2)
```

where `transfer_tx_1` and `transfer_tx_2` may not be the same but need to be spending the `create_tx`. Not sure what we'd do here (on a single node, would they just keep getting re-added to the same block and fail?)...

sohkai:
This is really awesome! Is this just waiting for Dimi's failing case that needs to be rewritten? Maybe we can merge this and open another issue for his test.

tests/integration/test_integration.py
r-marques: Or import it from `..db.conftest`

sohkai:
I've made a separate issue for @diminator's failing script(https://github.com/bigchaindb/bigchaindb/issues/774) so this can be merged without it.

sbellem:
A [pytest note](https://pytest.org/dev/plugins.html) about the practice of importing "stuff" from a `conftest.py` module:

> **Note**
>
> If you have conftest.py files which do not reside in a python package directory (i.e. one containing an __init__.py) then “import conftest” can be ambiguous because there might be other conftest.py files as well on your PYTHONPATH or sys.path. It is thus good practise for projects to either put conftest.py under a package scope or to never import anything from a conftest.py file.


ref: https://pytest.org/dev/plugins.html

tests/integration/test_integration.py
sohkai: Moved this to https://github.com/bigchaindb/bigchaindb/issues/832 so this can be merged without it.

tests/integration/test_integration.py
```diff
@@ -15,17 +14,18 @@ def inputs(user_vk):
     for block in range(4):
         transactions = [
             Transaction.create(
-                [b.me], [user_vk], metadata={'i': i}).sign([b.me_private])
+                [b.me], [([user_pk], 1)],
+                metadata={'i': i}).sign([b.me_private])
```
sohkai: Maybe put the `sign` on a different line so it's more obvious

PR #619, opened 2016-09-05
Fixed fabfile-monitor.py to use locally-generated SSH key

ttmc:
`fabfile-monitor.py` was still using the old AWS-generated SSH key to communicate with the instance on AWS. I fixed it to use the local user-generated SSH key.


ttmc:
This was just a small fix so I'm going to go ahead and merge it.

PR #620, opened 2016-09-05
Made the security group an AWS deployment parameter

ttmc:
Before, when using the AWS deployment scripts (i.e. the Fabric-based ones), the security group was hard-wired to be "bigchaindb". This just makes it a parameter in the AWS deployment configuration file. This makes it easier to have a more secure security group, if you want one. I tried it and it works.


vrde:
:+1:

Issue #502, opened 2016-07-27
Isolate code to access the Database


vrde:
Access to the database is spread around different layers of the architecture.
Would be nice to encapsulate all the database queries in a different module to keep the code more isolated and easy to test.

Consider also:
- https://github.com/bigchaindb/bigchaindb/pull/623/files#r77772779

ttmc:
See pull requests #754 and #788

sohkai:
Closing as #788 was merged.

PR #623, opened 2016-09-06
Make db connection more robust

vrde:
This PR makes connections to the database more robust.

The new `bigchaindb.db.utils.Connection` class is a proxy to query the database, and it is now used whenever we need to query the database.

`bigchaindb.db.utils.Connection` is:
- lazy, since it creates a connection only when needed
- resilient, because before raising exceptions it tries more times to run the query or to open a connection.

### DIFF
--- a/bigchaindb/commands/utils.py
+++ b/bigchaindb/commands/utils.py
@@ -5,6 +5,7 @@ for ``argparse.ArgumentParser``.
 import argparse
 import multiprocessing as mp
 import subprocess
+import tempfile
 
 import rethinkdb as r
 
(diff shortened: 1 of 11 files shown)


bigchaindb/commands/utils.py
```diff
@@ -14,41 +15,63 @@
 from bigchaindb.version import __version__
 
 
-def start_rethinkdb():
+def start_temp_rethinkdb(port=0, directory=None):
+    directory = directory or tempfile.mkdtemp()
+
+    extra_opts = ['--cluster-port', '0',
```
r-marques: what does it mean setting the cluster port to 0?

bigchaindb/commands/utils.py
```diff
@@ -14,41 +15,63 @@
 from bigchaindb.version import __version__
 
 
-def start_rethinkdb():
+def start_temp_rethinkdb(port=0, directory=None):
```
r-marques: I can't find this being used anywhere.

bigchaindb/core.py
```diff
@@ -219,8 +226,10 @@ def search_block_election_on_index(self, value, index):
             A list of blocks with with only election information
         """
         # First, get information on all blocks which contain this transaction
-        response = r.table('bigchain', read_mode=self.read_mode).get_all(value, index=index)\
-            .pluck('votes', 'id', {'block': ['voters']}).run(self.conn)
+        response = self.connection.run(
+                r.table('bigchain', read_mode=self.read_mode)
+                .get_all(value, index=index)\
```
r-marques: no need for a backslash

bigchaindb/core.py
```diff
@@ -526,12 +536,15 @@ def write_block(self, block, durability='soft'):
         """
 
         block_serialized = rapidjson.dumps(block)
-        r.table('bigchain').insert(r.json(block_serialized), durability=durability).run(self.conn)
+        self.connection.run(
+                r.table('bigchain')
+                .insert(r.json(block_serialized), durability=durability))
 
     # TODO: Decide if we need this method
```
r-marques: I don't think we need this method. Its not being used anywhere and its behavior is now inconsistent with get_transaction.
This is just a note to remember to create an issue for this

r-marques:
Since this makes changes to the way queries are done we should first merge pending pull requests before this one.

bigchaindb/commands/utils.py
vrde: This function is to run a temporary RethinkDB to test things, and the default port can be in use. With port `0` you leave the choice of the port to the OS.

bigchaindb/commands/utils.py
vrde: I used it to spawn and kill RethinkDB to tests the new `Connection` class. Unfortunately Travis wasn't happy about that (I think there is no `rethinkdb` binary when testing, it's just an external service) so I had to rollback and mock up the connection lost behaviour. I can remove this or integrate it with `start_rethinkdb`.

bigchaindb/core.py
vrde: I've added a note to #502

r-marques: Nevermind this comment.
@rhsimplex is using it to check for duplicated transactions when creating new blocks https://github.com/bigchaindb/bigchaindb/pull/609

PR #627, opened 2016-09-08
fix assignment timestamp in block

rhsimplex:
Fixes a bug where the assignment timestamp ended up in the block. Related to #198


r-marques:
:+1:

PR #629, opened 2016-09-12
check if fulfillments are in a valid block

rhsimplex:
A transaction will no longer be validated if any fulfillment is in an invalid, undecided, or non-existent block.

Requires extensive review by @diminator (or @r-marques?) and minor review from @vrde (see line notes for specific questions, please).


tests/db/test_bigchain_api.py
```diff
@@ -542,6 +542,31 @@ def test_valid_non_create_transaction_after_block_creation(self, b, user_vk, use
         assert tx_valid_signed == b.validate_transaction(tx_valid_signed)
         assert tx_valid_signed == b.is_valid_transaction(tx_valid_signed)
```
rhsimplex: @diminator or @r-marques Not sure who wrote these tests, but the big change here is every `tx` has to be put into a block and voted valid.

tests/db/test_bigchain_api.py
```diff
@@ -1760,13 +1789,16 @@ def test_create_and_fulfill_asset_with_hashlock_condition(self, b, user_vk):
...
         block = b.create_block([hashlock_tx_signed])
         b.write_block(block, durability='hard')
 
-        assert len(b.get_owned_ids(b.me)) == 0
+        # vote block valid
+        vote = b.vote(block['id'], b.get_last_voted_block()['id'], True)
+        b.write_vote(vote)
+
+        # this assertion must be wrong...?
+        # assert len(b.get_owned_ids(b.me)) == 0
```
rhsimplex: I'm not sure what this assertion was meant to test.  There are a couple of these.  Can you comment, @diminator?

tests/db/test_bigchain_api.py
```diff
@@ -1822,6 +1856,8 @@ def test_get_subcondition_from_vk(self, b, user_sk, user_vk):
             subcondition = condition.get_subcondition_from_vk(owner_after)[0]
             assert subcondition.public_key.to_ascii().decode() == owner_after
 
+    @pytest.mark.skipif(reason='@Dimi need explanation of what is actually'
+                        'going on here')
     @pytest.mark.usefixtures('inputs')
```
rhsimplex: I just don't know what's going on in these two tests, so it's pretty difficult to adjust them. @diminator let's discuss when you get back?

tests/pipelines/test_vote.py
```diff
@@ -239,13 +239,13 @@ def test_valid_block_voting_with_transfer_transactions(monkeypatch, b):
     vote_pipeline = vote.create_pipeline()
     vote_pipeline.setup(indata=inpipe, outdata=outpipe)
 
+    vote_pipeline.start()
     inpipe.put(block)
+    import time; time.sleep(1)
```
rhsimplex: @vrde We probably don't want this, but I don't have a good alternative.  The voter module is running in different processes, and if one block with the fulfillment `tx` doesn't get written first, this will (correctly) raise an exception.

tests/db/test_bigchain_api.py
diminator: Is there a way to validate a transaction fulfillment regardless of the block it is in?

diminator: then we should only keep that one

tests/db/test_bigchain_api.py
```diff
@@ -792,20 +807,12 @@ def test_multiple_owners_before_single_owner_after_single_input(self, b, user_sk
         assert len(tx_signed['transaction']['fulfillments']) == 1
         assert len(tx_signed['transaction']['conditions']) == 1
 
+    @pytest.mark.usefixtures('inputs')
     def test_multiple_owners_before_single_owner_after_multiple_inputs(self, b, user_sk, user_vk):
         # create a new users
         user2_sk, user2_vk = crypto.generate_key_pair()
         user3_sk, user3_vk = crypto.generate_key_pair()
 
-        # create inputs to spend
```
diminator: are you sure you can delete this?

it's a multi-ownership `[user_vk, user2_vk]` instead of single

tests/db/test_bigchain_api.py
```diff
@@ -844,21 +856,13 @@ def test_multiple_owners_before_multiple_owners_after_single_input(self, b, user
...
 
+    @pytest.mark.usefixtures('inputs')
     def test_multiple_owners_before_multiple_owners_after_multiple_inputs(self, b, user_sk, user_vk):
         # create a new users
         user2_sk, user2_vk = crypto.generate_key_pair()
         user3_sk, user3_vk = crypto.generate_key_pair()
         user4_sk, user4_vk = crypto.generate_key_pair()
 
-        # create inputs to spend
-        transactions = []
```
diminator: same as above

tests/db/test_bigchain_api.py
diminator: yeah, it was to see that we didnt own anything yet before redeeming the hashlock. But a bit of a stupid test

tests/db/test_bigchain_api.py
```diff
@@ -1792,8 +1824,10 @@ def test_create_and_fulfill_asset_with_hashlock_condition(self, b, user_vk):
         block = b.create_block([hashlock_fulfill_tx])
         b.write_block(block, durability='hard')
 
-        assert len(b.get_owned_ids(b.me)) == 0
-        assert len(b.get_owned_ids(user_vk)) == 1
+        # the first one might be wrong, the second one conflicts with the
+        # fixture
+        # assert len(b.get_owned_ids(b.me)) == 0
+        # assert len(b.get_owned_ids(user_vk)) == 1
```
diminator: maybe we can check if the balance of `get_owned_ids(user_vk)` went up?
At least we might wanna check that `user_vk` owns this hashlock

tests/db/test_bigchain_api.py
diminator: These are the escrow tests. Basically they are hold accounts that can either execute or roll back.
The expiry time defines which one of both is valid at a given point in time.

asset flow: `[A] -> [A, B] -> [A] if expired else [B]`

tests/db/test_bigchain_api.py
rhsimplex: Should come from the `inputs` fixture.  The removed lines are identical.

rhsimplex: here they are different, I see

tests/db/test_bigchain_api.py
rhsimplex: got it

tests/db/test_bigchain_api.py
rhsimplex: Is there a reason why this isn't three separate tests? Is one of the `tx` reused? Something is becoming invalid along the way now.

diminator: ok

tests/db/test_bigchain_api.py
rhsimplex: now they're from a new fixture

r-marques:
There may be some logic that we may need to handle.

For the case in which an input is in an undecided block. This does not necessarily mean that the transaction is invalid, it just needs to wait a bit more for block in which the input resides to be validated.

The solutions would be:
1. Leave this as it is and make it the clients responsibility to check that the inputs were correctly validated before using them.
2. Add additional logic in the block creation to cache transactions in which inputs are still in an undecided state and try again later.

If we feel that point 2. is worth it can create an issue for it and do it in a separate pull request as an optimization.

@diminator thoughts?

tests/pipelines/test_vote.py
```diff
@@ -258,6 +258,7 @@ def test_valid_block_voting_with_transfer_transactions(monkeypatch, b):
     from bigchaindb_common import crypto, util
     from bigchaindb.models import Transaction
     from bigchaindb.pipelines import vote
+    import time
```
sbellem: This can be imported at the top of the file.

The idea of importing under a test function is to prevent an entire test module to fail because of some failing import, which is only required by a few tests. Hence, importing under the test function will only make the affected tests fail, meanwhile the other tests will pass. This helps identifying the cause of a failing test when an import statement is causing it. The reasoning is explained in http://docs.pylonsproject.org/en/latest/community/testing.html#rule-never-import-the-module-under-test-at-test-module-scope

> Import failures in the module-under-test (MUT) should cause individual test cases to fail: they should never prevent those tests from being run. Depending on the testrunner, import problems may be much harder to distinguish at a glance than normal test failures.

tests/db/test_bigchain_api.py
```diff
@@ -695,67 +733,52 @@ def test_transfer_single_owner_single_input(self, b, inputs, user_vk,
...
-        tx = Transaction.transfer(inputs, 3 * [[user2_vk]])
+        tx = Transaction.transfer(inputs, len(inputs) * [[user2_vk]])
         tx = tx.sign([user_sk])
 
         # create block with the transaction
         block = b.create_block([tx])
         b.write_block(block, durability='hard')
 
+        # vote block valid
+        vote = b.vote(block.id, b.get_last_voted_block().to_dict()['id'], True)
```
TimDaub: `get_last_voted_block()` returns a `Block`. Just use `.id` attribute.

```
vote = b.vote(block.id, b.get_last_voted_block().id, True)
```

tests/db/test_bigchain_api.py
```diff
@@ -787,40 +810,38 @@ def test_single_owner_before_multiple_owners_after_single_input(self, b,
...
-        tx = Transaction.transfer(inputs, 3 * [[user2_vk, user3_vk]])
+        tx = Transaction.transfer(inputs, len(inputs) * [[user2_vk, user3_vk]])
         tx = tx.sign([user_sk])
 
+        # create block with the transaction
+        block = b.create_block([tx])
+        b.write_block(block, durability='hard')
+
+        # vote block valid
+        vote = b.vote(block.id, b.get_last_voted_block().to_dict()['id'], True)
```
TimDaub: `get_last_voted_block()` returns a `Block`. Just use `.id` attribute.

```
vote = b.vote(block.id, b.get_last_voted_block().id, True)
```

tests/db/test_bigchain_api.py
```diff
@@ -890,6 +908,11 @@ def test_multiple_owners_before_multiple_owners_after_single_input(self, b,
         block = b.create_block([tx])
         b.write_block(block, durability='hard')
 
+        # vote block valid
+        vote = b.vote(block.id, b.get_last_voted_block().to_dict()['id'], True)
```
TimDaub: `get_last_voted_block()` returns a `Block`. Just use `.id` attribute.

```
vote = b.vote(block.id, b.get_last_voted_block().id, True)
```

rhsimplex: I got it

tests/pipelines/test_vote.py
```diff
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 
 import rethinkdb as r
+import time
```
sbellem: as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports), the grouping of imports is as follows:

> Imports should be grouped in the following order:
>
> 1. standard library imports
> 2. related third party imports
> 3. local application/library specific imports
>
> You should put a blank line between each group of imports.

rhsimplex: I didn't know! Thought it was a free-for-all.

sbellem: :smile:

diminator:
:+1:

Issue #603, opened 2016-08-26
rename `get_owned_ids` to `get_unspents`


diminator:
more universal lingo

amrandahvur:
Please Assign me this

ttmc:
Hacktoberfest aside, this issue needs to be revisited so that `get_owned_ids` is renamed something consistent with the new names of `owners_before` and `owners_after`: see https://github.com/bigchaindb/bigchaindb/issues/626

My inclination is to close this issue and bundle it with #626

diminator:
> this issue needs to be revisited so that get_owned_ids is renamed something consistent with the new names of owners_before and owners_after

The `get_owned_ids` function is meant to check what you have in your current wallet, the things you haven't spent yet. This way you can compute your balance. `unspents` is common language in cryptocurrency space, see [here](https://bitcoin.org/en/glossary/unspent-transaction-output)

My take is that we use "known" language for such API functions, and it's agnostic to the implementation under the hood

ttmc:
The problem is, a BigchainDB instance doesn't know the assets _owned_ by a public key. The so-called `owners_after` and `owners_before` lists are not lists of asset owners; they are lists of public keys which have (or had) the ability to enable a transfer. (This is why we want to rename those lists as per issue #626.)

If someone wants to get a list of the assets they own (in a particular BigchainDB database), then they are out of luck. Asset _owners_ aren't stored in BigchainDB, only transfer-enablers / signers / fulfillers / controllers / (whatever we decided to call them).

That function, `get_owned_ids` _will_ return a list of unspent assets associated with a particular key, so `get_unspents` is probably okay. One must be very careful not to interpret those assets as being owned by the key (or keyholder). I worry that this will be a source of confusion and a common source of human error.

diminator:
see the comments in #677

ttmc:
1. `get_unspents` is fine, but
2. the docstring and implementation of that method (in `bigchaindb/bigchaindb/core.py`) must be modified so that users of that method can correctly interpret what it _actually_ returns.

sbellem:
> Change the input parameter from owner to pubkey
> Say that the output is a list of unspent outputs (TransactionLinks, i.e. txid + cid pairs) associated with that pubkey
> Rename the internal variable owned to txlinks or similar

we could turn this into tasks?

ttmc:
@sbellem Okay I changed those to tasks.

sbellem:
[...]

PR #631, opened 2016-09-13
Solves issue #603 rename `get_owned_ids` to `get_unspents`. renamed in docs too.

zack-bitcoin:
issue #603


zack-bitcoin:
It looks like this pull request doesn't actually modify any code.
my mistake.

PR #635, opened 2016-09-20
Allow temporary keypair if no conf file found

### LINKED ISSUE #482: Rename `--experimental-start-rethinkdb` to `--dev-start-rethinkdb`
In a production environment BigchainDB, RethinkDB, and other services (web servers  ) might be run in different machines. To be more explicit that `start-rethinkdb` is more for development purposes, we can change the prefix from `experimental` to `dev`.

### LINKED ISSUE #559: Generate key pair at start-up time if none is found
a subtask of #549

vrde:
I've updated also the `Dockerfile` to use the new `--allow-temp-keypair` on startup.

Closes #482, closes #559

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ ENV BIGCHAINDB_CONFIG_PATH /data/.bigchaindb
 ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 ENV BIGCHAINDB_API_ENDPOINT http://bigchaindb:9984/api/v1
 
-ENTRYPOINT ["bigchaindb", "--experimental-start-rethinkdb"]
+ENTRYPOINT ["bigchaindb", "--dev-start-rethinkdb", "--dev-allow-temp-keypair"]
 
 CMD ["start"]
 
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -157,7 +157,19 @@ def run_drop(args):
 def run_start(args):
     """Start the processes to run the node"""
     logger.info('BigchainDB Version {}'.format(bigchaindb.__version__))
+
     bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
+
+    if args.allow_temp_keypair:
+        if not (bigchaindb.config['keypair']['private'] or
+                                  bigchaindb.config['keypair']['public']):
+
+            private_key, public_key = crypto.generate_key_pair()
+            bigchaindb.config['keypair']['private'] = private_key
+            bigchaindb.config['keypair']['public'] = public_key
+        else:
+            logger.warning('Keypair found, no need to create one on the fly.')
+
 
     if args.start_rethinkdb:
         try:
@@ -174,7 +186,8 @@ def run_start(args):
         sys.exit("Can't start BigchainDB, no keypair found. "
                  'Did you run `bigchaindb configure`?')
 
-    logger.info('Starting BigchainDB main process')
+    logger.info('Starting BigchainDB main process with public key %s',
+                bigchaindb.config['keypair']['public'])
     processes.start()
 
 
@@ -238,10 +251,15 @@ def main():
         description='Control your BigchainDB node.',
         parents=[utils.base_parser])
 
-    parser.add_argument('--experimental-start-rethinkdb',
+    parser.add_argument('--dev-start-rethinkdb',
                         dest='start_rethinkdb',
                         action='store_true',
                         help='Run RethinkDB on start')
+
+    parser.add_argument('--dev-allow-temp-keypair',
+                        dest='allow_temp_keypair',
+                        action='store_true',
+                        help='Generate a random keypair on start')
 
     # all the commands are contained in the subparsers object,
     # the command selected by the user will be stored in `args.command`
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -245,8 +245,7 @@ def run_set_replicas(args):
         except r.ReqlOpFailedError as e:
             logger.warn(e)
 
-
-def main():
+def create_parser():
     parser = argparse.ArgumentParser(
         description='Control your BigchainDB node.',
         parents=[utils.base_parser])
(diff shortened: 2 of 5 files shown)


bigchaindb/commands/bigchain.py
```diff
@@ -157,8 +157,20 @@ def run_drop(args):
...
 
+    if args.allow_temp_keypair:
+        if not (bigchaindb.config['keypair']['private'] or
+                                  bigchaindb.config['keypair']['public']):
+
+            private_key, public_key = crypto.generate_key_pair()
+            bigchaindb.config['keypair']['private'] = private_key
+            bigchaindb.config['keypair']['public'] = public_key
+        else:
+            logger.warning('Keypair found, no need to create one on the fly.')
```
sohkai: Would be nice if we knew where the keypair was found, but not sure if that's easy.

Dockerfile
```diff
@@ -19,7 +19,7 @@ ENV BIGCHAINDB_CONFIG_PATH /data/.bigchaindb
 ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 ENV BIGCHAINDB_API_ENDPOINT http://bigchaindb:9984/api/v1
 
-ENTRYPOINT ["bigchaindb", "--experimental-start-rethinkdb"]
+ENTRYPOINT ["bigchaindb", "--dev-start-rethinkdb", "--dev-allow-temp-keypair"]
```
sohkai: It'd be nice if we added these only to the `Dockerfile-dev` file rather than the `Dockerfile`, which I would assume is less for devs.

sohkai:
Looks good otherwise 👍.

bigchaindb/commands/bigchain.py
vrde: The logs should give to the user enough info about that ("→" mine):

```
INFO:bigchaindb.commands.bigchain:BigchainDB Version 0.6.0
→ INFO:bigchaindb.config_utils:Configuration loaded from `/home/vrde/.bigchaindb`
→ WARNING:bigchaindb.commands.bigchain:Keypair found, no need to create one on the fly.
INFO:bigchaindb.commands.bigchain:RethinkDB started with PID 2191
→ INFO:bigchaindb.commands.bigchain:Starting BigchainDB main process with public key Ajh1ehp4hYQGqgq5w8QhfNxn2pCHpGuNNbhm5p7gnjDn
```

is it enough in your opinion?

sohkai: Ah, nice, didn't notice the configuration was getting logged.

Dockerfile
vrde: Good point, I've moved this to #636 so we can proceed with this PR.

bigchaindb/commands/utils.py
```diff
@@ -59,7 +59,7 @@ def start_rethinkdb():
     raise StartupError(line)
 
 
-def start(parser, scope):
+def start(parser, argv, scope):
```
sbellem: Is there a reason for not adding documentation for the new argument?

tests/test_commands.py
```diff
@@ -62,9 +62,64 @@ def mock_bigchaindb_backup_config(monkeypatch):
...
 
+def test_make_sure_we_dont_remove_any_command():
+    # thanks to: http://stackoverflow.com/a/18161115/597097
+    from bigchaindb.commands.bigchain import utils
+    from bigchaindb.commands.bigchain import create_parser
+
+    parser = create_parser()
+
+    with pytest.raises(SystemExit):
+        utils.start(parser, [], {})
```
sbellem: It seems to me that this:

```python
with pytest.raises(SystemExit):
    utils.start(parser, [], {})
```

could be in its own separate test, e.g., something like `test_utils_start_with_no_args_command()`.

tests/test_commands.py
```diff
@@ -62,9 +62,64 @@ def mock_bigchaindb_backup_config(monkeypatch):
...
+        utils.start(parser, ['configure'], {})
+
+
+@patch('multiprocessing.cpu_count', return_value=42)
+def test_start_sets_multiprocess_var_based_on_cli_args(mock_cpu_count):
+    def run_load(args):
+        return args
+
+    from bigchaindb.commands.bigchain import utils
+    from bigchaindb.commands.bigchain import create_parser
```
sbellem: Simple style related question: "Is there a reason for the import statements to be after the function (`run_load()`) definition?"

bigchaindb/commands/utils.py
vrde: mistake

tests/test_commands.py
vrde: mistake

tests/test_commands.py
vrde: done

tests/test_commands.py
```diff
@@ -229,6 +284,42 @@ def test_start_rethinkdb_exits_when_cannot_start(mock_popen):
...
+
+    # Preconditions for the test
+    assert isinstance(bigchaindb.config['keypair']['private'], str)
+    assert isinstance(bigchaindb.config['keypair']['public'], str)
+
+    args = Namespace(allow_temp_keypair=True, start_rethinkdb=False, config=None, yes=True)
+    run_start(args)
+
+    assert bigchaindb.config['keypair']['private'] != 'private_key'
+    assert bigchaindb.config['keypair']['public'] != 'public_key'
```
sbellem: Would it be possible to check explicitly for the original (unchanged) values? That is:

```python
assert bigchaindb.config['keypair']['private'] != 'original_private_key'
assert bigchaindb.config['keypair']['public'] == 'original_public_key'
```

Why?

Because, say the command sets the keys to `None`, or some random string by error. This test will not catch the problem.

bigchaindb/commands/utils.py
sbellem: :smile:

tests/test_commands.py
sbellem: :+1:

tests/test_commands.py
vrde: done

sbellem:
LRTM! (Looks Ready To Merge!)

PR #624, opened 2016-09-07
[WIP Do Not Merge] Api/601/expose bigchaindb functionality

diminator:
**UPDATE SEE: https://github.com/bigchaindb/bigchaindb/pull/624#issuecomment-259408517**

- [x] `get_owned_ids` -> `GET /transactions/unspents/{owner_id}`
- [ ] `get_block`
- [ ] `get_votes`

Issue #601, opened 2016-08-26
API: expose bigchaindb functionality


diminator:
This issue is to list endpoints that should be exposed by the API.

Useful documents for this discussion, to add context:
- https://github.com/bigchaindb/bigchaindb-examples/blob/master/server/lib/api/views.py
- https://github.com/diminator/rfcs/blob/36e30bd746281a539457eab8b1ba8f2e30e34ae4/0012-ledger-web-api/0012-ledger-web-api.md
### bigchaindb metadata
- [ ] `get_info` or `get_metadata`: info on ledger instance, asset types, federation, etc (to be fleshed out)
### transactions
- [x] `get_transaction`: #554 
- [x] `write_transaction`: #554 
- [x] `get_status`: #555
- [ ] `get_tx_by_metadata_id`
- [x] `get_owned_ids`#624
- [x] <s>`escrow`</s> => move to `driver/templates` cc @TimDaub 
### asset: #599
- [ ] `get_txs_by_asset_id` (all transactions related to the asset)
- [ ] `get_asset` (maybe. get the `CREATE` transaction for the asset)
### blocks
- [ ] `get_block` #624
### votes
- [ ] `get_votes` #624
- ...

r-marques:
The most important for clients would be:
- transactions
- metadata
- asset #599 

Less important to clients (but useful if they want to validate themselves)
- blocks
- votes

r-marques:
Some methods that could go into the api:
- Transactions:
    - `get_transaction`
    - `get_owned_ids`

- Metadata:
    - `get_tx_by_metadata_id`

- Asset:
     - `get_txs_by_asset_id` (all transactions related to the asset)
     - `get_asset` (maybe. get the `CREATE` transaction for the asset)
     - We could have some more digital asset queries to better support divisible assets. The  tradeshare example should give us some ideas

diminator:
I would group `get_tx_by_metadata_id` as part of transactions as it is a query on the transaction and not on the state of the DB

sohkai:
Looks like we'd also like a websocket API for interledger, which is probably worth its own project.

> `get_asset` (maybe. get the `CREATE` transaction for the asset)

[...]

PR #639, opened 2016-09-26
Add semi RPC commands to the API

roderik:
I've been hard at work building stuff based on BigchainDB and since I'm not working in Python, I've come across some difficulties in getting a stable system going. (especially together with docker)

I want to be able to perform all actions the python client can perform (e.g. follow the documentation example) but not in Python, but from my NodeJS code. While there is some REST api calls, there is too much missing to make it usable at the moment. (I understand the difficulty in building and maintaining this REST api)

I started out with an extra Flask api between my code and the bigchaindb node, but that becomes a mess quickly because you need to sync keys, but not hostnames, and an extra http component is not very beneficial for performance.

I took a page from bitcoin and multichain, and by pure googling and stack overflowing I fabricated this addition to the API. It allows any function call or variable call through the api. With no maintenance at all if the client should change.

**warning: I have no idea about the security repercussions of what I did here, please use it as input, not a finished product***

It supports all methods of the bigchaindb client, all variables in the client, and all variables and calls on the crypto component.

Examples:
```
POST http://bigchain.docker:59984/rpc/bigchaindb

{
  "method": "me"
}
```
```
{
  "result": "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse"
}
```
---------
```
POST http://bigchain.docker:59984/rpc/bigchaindb

{
  "method": "create_transaction",
  "params": {
    "owner_before": "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse",
    "owner_after": "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse",
    "tx_input": "None",
    "operation": "CREATE",
    "payload": {
      "naam": "test"
    }
  }
}
```
```
{
  "result": {
    "id": "659171e4921d64cfe7bddbfb3cb00744e6951b0403e52dca4e91c1dc4600e65d",
    "transaction": {
      "conditions": [
        {
          "cid": 0,
          "condition": {
            "details": {
              "bitmask": 32,
              "public_key": "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse",
              "signature": null,
              "type": "fulfillment",
              "type_id": 4
            },
            "uri": "cc:4:20:rOxXPC6VwwilYi1O51Dkv43dFr645s6j2v10o0tm6y0:96"
          },
          "owners_after": [
            "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse"
          ]
        }
      ],
      "data": {
        "payload": {
          "naam": "test"
        },
        "uuid": "525b73f4-ff93-476f-8871-82eef9f0dc31"
      },
      "fulfillments": [
        {
          "fid": 0,
          "fulfillment": null,
          "input": "None",
          "owners_before": [
            "Ce29QivVvJuNgTGbrPBuWEaRwrG2Nhqa8kR55V3QpLse"
          ]
        }
      ],
      "operation": "CREATE",
      "timestamp": "1474880951",
      "version": 1
    }
  }
}
```
---------
```
POST http://bigchain.docker:59984/rpc/crypto

{
  "method": "generate_key_pair"
}
```
```
{
  "result": [
    "4MspmDAsnBGRPVQVDGmGtDhb51zBnbvWe85hy79V6FHD",
    "Fs7omc6FDbTQqdH14ovK6dWjWLXoEBFW1VmF3cHzsz3q"
  ]
}
```
---------
I believe something like this would make it a lot easier for developers in other languages to get started.

Thoughts?

### DIFF
--- a/bigchaindb/web/server.py
+++ b/bigchaindb/web/server.py
@@ -13,6 +13,7 @@ from bigchaindb import util
 from bigchaindb import Bigchain
 from bigchaindb.web.views.info import info_views
 from bigchaindb.web.views.transactions import transaction_views
+from bigchaindb.web.views.rpc import rpc_views
 
 from bigchaindb.monitor import Monitor
 
@@ -66,6 +67,7 @@ def create_app(settings):
 
     app.register_blueprint(info_views, url_prefix='/')
     app.register_blueprint(transaction_views, url_prefix='/api/v1')
+    app.register_blueprint(rpc_views, url_prefix='/rpc')
     return app
 
 
--- a/bigchaindb/web/views/rpc.py
+++ b/bigchaindb/web/views/rpc.py
@@ -1,0 +1,49 @@
+import flask
+from flask import current_app, request, Blueprint
+from bigchaindb import crypto
+
+import traceback
+
+rpc_views = Blueprint('rpc_views', __name__)
+
+@rpc_views.route('/bigchaindb', methods=['POST'])
+def rpc_bigchaindb():
+    try:
+        rpc_request = request.get_json(force=True)
+        method_name = rpc_request["method"]
+        pool = current_app.config['bigchain_pool']
+        with pool() as bigchain:
+            method = getattr(bigchain, method_name)
+            if not callable(method):
+                return flask.jsonify(result=bigchain.__dict__[method_name])
+            else:
+                if "params" in rpc_request:
+                    result = method(**rpc_request["params"])
+                else:
+                    result = method()
+                return flask.jsonify(result=result)
+    except:
+        response = flask.jsonify(exception=traceback.format_exc())
+        response.status_code = 500
+        return response
+
+
+@rpc_views.route('/crypto', methods=['POST'])
+def rpc_crypto():
+    try:
+        rpc_request = request.get_json(force=True)
+        method_name = rpc_request["method"]
+        method = getattr(crypto, method_name)
+        if not callable(method):
+            return flask.jsonify(crypto=bigchain.__dict__[method_name])
+        else:
+            if "params" in rpc_request:
+                result = method(**rpc_request["params"])
+            else:
+                result = method()
+            return flask.jsonify(result=result)
+
+    except:
+        response = flask.jsonify(exception=traceback.format_exc())
+        response.status_code = 500
+        return response


r-marques:
Thanks for this proposal.

We understand the pain right now in using bigchaindb with other languages other then python.
We have discussed internally whether to use REST or rpc calls.

I personally don't feel comfortable opening the nodes so much to rpc calls.
With bitcoin its a bit different, you can just install the bitcoin daemon download a copy of the blockchain and do whatever you want locally with little to no impact with the rest of the network.

With bigchaindb is a bit different assuming that you are a client you will always have to communicate with a remote bigchaindb federation node which needs to have a tight control over its resources.

We are currently working hard in separating the bigchaindb server and client code and at the same time improving the api to bring it into a more useful state.

Another idea that as been floating around would was, the python client driver could not only be a client to the bigchaindb REST API but it could also provide a command line interface similar to `bitcoin-cli` to perform several operations. The reason for this is that a client needs to not only be able to perform REST calls but it also needs to implement all the crypto code to be have to generate and fulfill cryptoconditions, basically the client also needs to implement the cryptoconditions spec.

So I like the idea of possibly doing something like this but in the python client driver .

@diminator thoughts?

roderik:
I'm not following a 100%. Yes, RPC will result in a bit more 'calls' since it's not possible to create methods like in the examples made by @diminator or in the client.py that combine a few bigchaindb python calls. But load wise, these calls do still happen internally and add only some network traffic overhead. At the same time access to these calls can be limited by the IP address of the flask server, and via traditional firewall rules. 

The python client driver is essentially the code that I wrote in NodeJS, but I need a way to get the different commands to bigchaindb, via an API (be it REST or RPC, I couldn't care less 😄 ). A cli component that my code needs to call is not a very clean solution. While I use b.me and b.me_private over the API, an approach like the python client is possible however.

Calls that are needed for sure are:

* crypto.generate_key_pair (I can do this implementation in node as well, but this is duplicating logic)
* create or
    * create_transaction
    * sign_transaction
    * validate_transaction
    * write_transaction
* transfer or
    * create_transaction
    * sign_transaction
    * validate_transaction
    * write_transaction
* get_owned_ids
* everything related to fulfilment and crypto conditions (have not implemented them yet, don't know the full list)

TimDaub:
@roderik First of all: Thanks so much for committing time to this project :tada:!

To give you context:
We're currently in the process of separating client and server logic. Logic that can be used on both server and client can all be found here: https://github.com/bigchaindb/bigchaindb-common

Some methods seem to be quite interesting for you:

- Generate a `CREATE` transaction: https://github.com/bigchaindb/bigchaindb-common/blob/master/bigchaindb_common/transaction.py#L308
- Generate a `TRANSFER` transaction: https://github.com/bigchaindb/bigchaindb-common/blob/master/bigchaindb_common/transaction.py#L360
- Sign any transaction: https://github.com/bigchaindb/bigchaindb-common/blob/master/bigchaindb_common/transaction.py#L435
- Serialize a transaction to a dict (and then json): https://github.com/bigchaindb/bigchaindb-common/blob/master/bigchaindb_common/transaction.py#L555


My idea (and this we do not even have on the roadmap yet) was that we could potentially have a binary (executable on maybe on various Unix systems) that uses these function's parameter signature, but on the command line. Each of those commands would take JSON as an input and would yield JSON as an output.
This way, it would be very easy for other developers to write language-specific bindings for this unix-binary.

I kinda stole this idea from https://github.com/bitpay/bitcoind-rpc/blob/master/lib/index.js

Not sure how much that helps you at this very point, at least it gives you a direction on where we might be going.

diminator:
Thanks @roderik, your effort seems in-line with some of our internal efforts on creating a suitable (python) driver and exposing the API endpoints. Very interesting approach, but as you and @r-marques mentioned, might expose the nodes a bit too much from a security point of view. RPC methods would open up the internals of BDB more, but could likely be restricted to what a standard JSON API would offer. 

Most of the API discussion is grouped in #601 (and PR's like #554, #555, #624 spin out of that)
We'd want to have support for at least a JSON driver, hence I would like to categorize some of the endpoints you proposed and see what's the shortest path to a JS driver.

This that make sense to serve through a/the JSON API are:

- `write_transaction` ie `POST /transactions`
- `get_transaction` ie `GET /transactions/{tx_id}`
- `get_owned_ids` ie `GET /transactions/unspents/{owner_id}` see #624 

Could be part of either JSON API and/or JS/Python driver:

- `validate_transaction` ie `POST /transactions/validate` (not implemented yet, but makes sense). When validation in the client would only the crypto would be validated, not the federation rules (doublespends, ...)
- `generate_key_pair` ie `POST /accounts` (not implemented yet, but could be part of a KPI infrastructure), should be an easy starting point for KPI infrastructure

Probably part of only the (JS/Python) driver:
- `create_transaction`: return a template of a transaction with some prefilled fields. Probably the main driver routine to be foreseen in a JS environment
- `sign_transaction` and everything related to crypto-conditions. This should be relatively easy using the JS version of crypto-conditions.

roderik:
I'm going to have to move forward using my method in a forked docker image, my PoC needs to progress fast 😄 I'll refactor later.

But on the longer term, what can I do to help in this regard. So much is in flux in different PR's that it's hard to get a grasp on the next steps.

TimDaub:
@roderik: The development [here](https://github.com/bigchaindb/bdb-transaction-cli/pull/1) might be of interest to you.

@r-marques @diminator: How to proceed with this PR?

r-marques:
I feel that between the api and the cli everything in here will be covered.
The http api itself should take care of most of https://github.com/bigchaindb/bigchaindb/pull/639#issuecomment-249819462
Except for signing and manipulation of conditions and fulfillments that need to be handle client side and which the cli should provide

TimDaub:
> @r-marques @diminator: How to proceed with this PR?

TimDaub:
> so we should close this in favor of the work being done on the api and cli. Don't think it makes sense to support and maintain another way of making calls to bigchaindb.

OK, so since we're aware of the methods @roderik wants we can close this PR and work through the tickets to solve his problem.
@r-marques @roderik is it OK if I close?

roderik:
Ok!

PR #646, opened 2016-09-28
Remove extra fields from get_transaction

vrde:
Using the `/api/v1/transactions/<txid>` Web API @TimDaub noticed that `assignee` and `assignment_timestamp` are returned for some calls.

After further investigation I noticed this behavior happens only when the transaction is in the `backlog` table. This PR is to fix this misbehavior.

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -237,7 +237,10 @@ class Bigchain(object):
 
         else:
             # Otherwise, check the backlog
-            response = self.connection.run(r.table('backlog').get(txid))
+            response = self.connection.run(r.table('backlog')
+                                           .get(txid)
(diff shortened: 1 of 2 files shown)


bigchaindb/core.py
```diff
@@ -237,7 +237,10 @@ def get_transaction(self, txid, include_status=False):
 
         else:
             # Otherwise, check the backlog
-            response = self.connection.run(r.table('backlog').get(txid))
+            response = self.connection.run(r.table('backlog')
+                                           .get(txid)
+                                           .without('assignee', 'assignment_timestamp')
```
rhsimplex: https://www.youtube.com/watch?v=HPhDcP9y76Y

PR #648, opened 2016-09-29
Relabeled docs in this repository as being for BigchainDB Server

ttmc:
Historically, there was only one set of docs --- the set in this repository --- and it was for the server, the Python driver, and the BigchainDB project as a whole. Now there are separate docs for the Python driver and for bigchaindb-common. This PR changes various labels and descriptions to clarify that the documentation is (or will soon be) for BigchainDB Server only.

(There is still some documentation in this repository that isn't server-specific, but that will be moved elsewhere eventually.)


sbellem:
:+1:

PR #657, opened 2016-10-04
Tidied up the docs for the config settings & the CLI

ttmc:
* Removed the bold notice at the top of the docs about the Configuration Settings and the Command Line Interface, saying that things may change a lot after the separation of client and server. (As it happens, there weren't any changes.)
* Removed the word "BigchainDB" from the title of those two pages. It's now clear from context that the the configuration settings, and the CLI, are for BigchainDB Server.
* Expanded the first sentence on both pages to clarify that we're talking about BigchainDB Server, just in case that wasn't clear (and to help search engines).
* I also clarified that `bigchaindb load` uses the Python Server API (not the HTTP API or a driver that wraps the HTTP API)

### DIFF
--- a/docs/source/server-reference/bigchaindb-cli.md
+++ b/docs/source/server-reference/bigchaindb-cli.md
@@ -1,9 +1,6 @@
-# BigchainDB Command Line Interface (CLI)
+# Command Line Interface (CLI)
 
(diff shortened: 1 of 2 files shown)


ttmc:
@r-marques This is a small PR based on what we discussed yesterday, so I assigned you.

PR #666, opened 2016-10-07
New ability to bind the RethinkDB web interface port to localhost when deploying on AWS

ttmc:
The RethinkDB documentation says it's more secure to bind the RethinkDB web interface (HTTP) port to localhost, so that it can't be accessed directly from external IP addresses (at http://hostname:8080/), see **Binding the web interface port** at:

https://www.rethinkdb.com/docs/security/

This pull request:

* adds a new configuration setting to the AWS configuration file: `BIND_HTTP_TO_LOCALHOST`. It can be set to `True` or `False` (the "default" in the example config file)
* checks to make sure that setting is set and is a boolean
* appends the following two lines to the RethinkDB config file when `BIND_HTTP_TO_LOCALHOST` is true:

```text
## Bind the web interface port to localhost
bind-http=127.0.0.1
```

I tried it out and it seems to work: the RethinkDB web interface ceases to be available at http://hostname:8080/ but I can access it via a SOCKS proxy, e.g. using Firefox as decribed on the RethinkDB security page linked above.


deploy-cluster-aws/awsdeploy.sh
```diff
@@ -116,7 +117,7 @@ fab upgrade_setuptools
 
 if [ "$WHAT_TO_DEPLOY" == "servers" ]; then
     # (Re)create the RethinkDB configuration file conf/rethinkdb.conf
-    python create_rethinkdb_conf.py
+    python create_rethinkdb_conf.py --bind-http-to-localhost $BIND_HTTP_TO_LOCALHOST
```
r-marques: Do we need a cli argument?
Can't we just read the environment variable?

ttmc: Reading the environment variable would also work, but then `create_rethinkdb_conf.py` would have to check if it's set, and complain if it's not. I guess it's just a matter of taste, and I prefer being explicit.

I also used the same "Mr. Explicit" style with the calls to `launch_ec2_nodes.py` and `clusterize_confiles.py`.

deploy-cluster-aws/example_deploy_conf.py
```diff
@@ -74,3 +74,8 @@
 # Setting EBS_OPTIMIZED=True may cost more, but not always.
 # If USING_EBS=False, EBS_OPTIMIZED is irrelevant and not used
 EBS_OPTIMIZED=False
+
+# BIND_HTTP_TO_LOCALHOST is True or False, depending on whether
+# you want the RethinkDB web interface port to be bound to localhost
+# (which is more secure). See https://www.rethinkdb.com/docs/security/
+BIND_HTTP_TO_LOCALHOST=False
```
ttmc: Yeah, you're right. I will change it. I will also update the docs a bit because the default will be for the RethinkDB web interface to not "just work" at http://hostname:8080

r-marques: it just feels more inline with the rest of the deployment since `launch_ec2_nodes.py` will also fail it he key is not set

deploy-cluster-aws/awsdeploy.sh
r-marques: exactly, `launch_ec2_nodes.py` will also fail if the key is not set so you can be confident that the key will always need to exist for the entire deployment to run

ttmc: As I said, it's basically a matter of taste. Moreover, it's not broken, so I'm reluctant to "fix" it.

deploy-cluster-aws/example_deploy_conf.py
ttmc: Okay, I changed the default to `True` and updated the docs accordingly.

r-marques:
Not sure why the tests are failing now. Maybe sync it with master

PR #669, opened 2016-10-09
Docs: Deleted most of the Topic Guides pages

ttmc:
I copied most of the Topic Guides over to the [root BigchainDB project docs](https://docs.bigchaindb.com/en/latest/index.html). This pull request deletes them from the BigchainDB Server docs. It also updates the Topic Guide index page and some links.

The page titled **The Transaction, Block and Vote Models** remains. I plan to break it apart into pieces, but that's going to take elaborate surgery, so I'll do it in a separate pull request.


ttmc:
Before merging this pull request:

- [ ] Merge https://github.com/bigchaindb/docs-root/issues/3 (i.e. bring the Topic Guides live on the root docs)
- [ ] Redo all the ReadTheDocs redirects (on the BigchainDB project) to point to the correct URL starting with https://docs.bighaindb.com/stuff

PR #684, opened 2016-10-14
Fixed the docstring in config_utils.py

ttmc:
I noticed that the docstring in `config_utils.py` was out-of-date. The order of precedence has changed; see the `autoconfigure()` function.

The changed docstring is copied from the current BigchainDB Server docs, e.g. https://docs.bigchaindb.com/projects/server/en/latest/server-reference/configuration.html

@vrde can you take a look and verify that I got the order of precedence right?

### DIFF
--- a/bigchaindb/config_utils.py
+++ b/bigchaindb/config_utils.py
@@ -3,8 +3,8 @@
 The value of each BigchainDB Server configuration setting is
 determined according to the following rules:
 
-* If it’s set by an environment variable, then use that value
-* Otherwise, if it’s set in a local config file, then use that
+* If it's set by an environment variable, then use that value
+* Otherwise, if it's set in a local config file, then use that
   value
 * Otherwise, use the default value (contained in 
   ``bigchaindb.__init__``)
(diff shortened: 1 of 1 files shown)


r-marques:
:+1:

bigchaindb/config_utils.py
```diff
@@ -1,13 +1,13 @@
...
-`$HOME/.bigchaindb`) will be updated with the values contained
-in the configuration file.
+The value of each BigchainDB Server configuration setting is
+determined according to the following rules:
 
-Note that there is a precedence in reading configuration values:
- - local config file;
- - environment vars;
- - default config file (contained in ``bigchaindb.__init__``).
+* If it’s set by an environment variable, then use that value
```
vrde: Might sound OCD-ish, but can we use `'` instead of `’`? I say that because we never used `’` in docstrings.

ttmc: Nice catch. I copied the text from the (rendered) BigchainDB Server docs, which turned the vertical appstrophe into a top-tipping-right apostrophe (all Unicode). I will fix it.

PR #698, opened 2016-10-18
Added installation of libffi-dev (or libffi-devel) to install docs, scripts, Dockerfiles

### LINKED ISSUE #694: New (implicit) PyNaCl dependency depends on ffi.h, so that must be installed first
The new `cryptoconditions` depends on PyNaCl which depends on `ffi.h` (i.e. an OS-level file) but currently we don't install that. According to [this Stack Overflow answer](https://stackoverflow.com/questions/31508612/pip-install-unable-to-find-ffi-h-even-though-it-recognizes-libffi), all we need to do is install:

> `libffi-dev` on Debian/Ubuntu, `libffi-devel` on Redhat/Centos/Fedora.

I already tried that on Ubuntu 14.04 (using `apt-get` in an AWS instance) and it worked.

Things that need updating: Documentation in various places, AWS deployment scripts (`fabfile.py`), some Ansible stuff, various Dockerfiles?, more?

ttmc:
This resolves #694 

I tested it on a bare-bones Ubuntu 14.04 instance on AWS and I didn't get the error message about a missing `ffi.h`

@r-marques Can you test that `sudo dnf install gcc-c++ redhat-rpm-config python3-devel libffi-devel` works on Fedora?

@sbellem Can you take a look to ensure I didn't miss some Docker-related files?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM rethinkdb:2.3
 
 RUN apt-get update
-RUN apt-get -y install python3 python3-pip
+RUN apt-get -y install python3 python3-pip libffi-dev
 RUN pip3 install --upgrade pip
 RUN pip3 install --upgrade setuptools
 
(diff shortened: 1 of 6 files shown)


r-marques:
Works fine on fedora with the provided instructions

PR #701, opened 2016-10-18
Remove conn property from core API

vrde:
I noticed working on #502 that some code is still using the old API. This PR is to fix some issues so I can rebase the other branch and move on with a clean baseline.

### DIFF
--- a/bigchaindb/pipelines/stale.py
+++ b/bigchaindb/pipelines/stale.py
@@ -71,6 +71,5 @@ def start(timeout=5, backlog_reassign_delay=5):
     """Create, start, and return the block pipeline."""
     pipeline = create_pipeline(timeout=timeout,
                                backlog_reassign_delay=backlog_reassign_delay)
-    pipeline.setup()
     pipeline.start()
     return pipeline
--- a/tests/db/test_bigchain_api.py
+++ b/tests/db/test_bigchain_api.py
@@ -263,9 +263,11 @@ class TestBigchainApi(object):
     def test_genesis_block(self, b):
         import rethinkdb as r
         from bigchaindb.util import is_genesis_block
+        from bigchaindb.db.utils import get_conn
+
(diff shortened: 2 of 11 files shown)


bigchaindb/pipelines/stale.py
```diff
@@ -71,6 +71,5 @@ def start(timeout=5, backlog_reassign_delay=5):
     """Create, start, and return the block pipeline."""
     pipeline = create_pipeline(timeout=timeout,
                                backlog_reassign_delay=backlog_reassign_delay)
-    pipeline.setup()
```
r-marques: What did this do?

tests/db/test_bigchain_api.py
```diff
@@ -263,9 +263,11 @@ def test_read_transaction_invalid_block(self, b, user_vk, user_sk):
     def test_genesis_block(self, b):
         import rethinkdb as r
         from bigchaindb.util import is_genesis_block
+        from bigchaindb.db.utils import get_conn
+
         response = list(r.table('bigchain')
                         .filter(is_genesis_block)
-                        .run(b.conn))
+                        .run(get_conn()))
```
r-marques: Shouldn't the tests also use the new api `b.connection.run`?

bigchaindb/pipelines/stale.py
vrde: `setup` prepares the pipeline to be used and it's called in `__init__`. You can call `setup` again if you want to reconfigure the pipeline. Calling `setup` with no args doesn't do anything.

tests/db/test_bigchain_api.py
vrde: This *was* true :smile_cat:

I'm working on #502 to move all the queries to a specific module and the `b.connection.run` might be moved out of `core.py`. In this PR I'm using the generic function to get a connection.

PR #737, opened 2016-10-23
Moved contents of models.md to new pages in root docs

ttmc:
The page of documentation titled "The Transaction, Block and Vote Models" (in the **Topic Guides** section) was getting too big, so I broke that page into six new pages, all of which are now in [the root BigchainDB docs](https://docs.bigchaindb.com/en/latest/index.html).

* Transaction Concepts
* The Transaction Model
* The Digital Asset Model
* Crypto-Conditions and Fulfillments
* The Block Model
* The Vote Model

I didn't completely delete the "The Transaction, Block and Vote Models" page because people may be in the habit of looking there. I just changed that page to explain that it's going away, with links to where the new pages can be found.

You may wonder why I moved the docs about the Block Model and the Vote Model to the root docs. The reason is that some day, even BigchainDB clients will want to request blocks and votes; they aren't server-specific.


docs/server/source/topic-guides/models.md
```diff
@@ -1,313 +1,6 @@
...
-        "is_block_valid": "<true|false>",
-        "invalid_reason": "<None|DOUBLE_SPEND|TRANSACTIONS_HASH_MISMATCH|NODES_PUBKEYS_MISMATCH",
-        "timestamp": "<Unix time when the vote was generated, provided by the voting node>"
-    },
-    "signature": "<signature of vote>",
-}
-```
-
-Note: The `invalid_reason` was not being used as of v0.1.3 and may be dropped in a future version of BigchainDB.
+* [Transaction concepts](https://docs.bigchaindb.com/en/latest/transactions.html)
```
ttmc: Oh, yes, I changed that but forgot to update the link here. I'll fix it.

sohkai:
LGTM, should anyone else take a look?

ttmc:
In some sense, it's a small PR because it replaces a long page with a short one, and you checked the new short page. The contents of the old long page (edited) are already live on the root BigchainDB docs, so there will be no time interval when the documentation isn't available anywhere. 

I will merge this PR now.

PR #748, opened 2016-10-26
Update outdated or incorrect docstrings

sohkai:
Small quality of life updates to docstrings that are outdated from the code, or had slight formatting errors.

*complaint*: We don't really seem to have guidelines on usage of sphinx's scopes (e.g. `:class:`, etc) so it's unclear at times how to represent some things.

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -198,10 +198,10 @@ class Bigchain(object):
                                    the return value is then a tuple: (tx, status)
 
         Returns:
-            A dict with the transaction details if the transaction was found.
-            Will add the transaction status to payload ('valid', 'undecided',
-            or 'backlog'). If no transaction with that `txid` was found it
-            returns `None`
+            A :class:`~.Transaction` instance if the transaction was
+            found, otherwise ``None``.
+            If :attr:`include_status` is ``True``, also returns the
+            transaction's status if the transaction was found.
         """
 
         response, tx_status = None, None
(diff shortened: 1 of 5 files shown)


r-marques:
:+1:

bigchaindb/core.py
```diff
@@ -198,10 +198,10 @@ def get_transaction(self, txid, include_status=False):
                                    the return value is then a tuple: (tx, status)
 
         Returns:
-            A dict with the transaction details if the transaction was found.
-            Will add the transaction status to payload ('valid', 'undecided',
-            or 'backlog'). If no transaction with that `txid` was found it
-            returns `None`
+            A :class:`~.Transaction` instance if the transaction was
```
sbellem: what's the `.` doing, in :class:`~.Transaction`?

sohkai: From the sphinx domain docs:

> Normally, names in these roles are searched first without any further qualification, then with the current module name prepended, then with the current module and class name (if any) prepended. **If you prefix the name with a dot, this order is reversed.** For example, in the documentation of Python’s codecs module, :py:func:`open` always refers to the built-in function, while :py:func:`.open` refers to codecs.open().

Reading this again, it seems to suggest that I might need to use :class:`~.models.Transaction` but I haven't been forced to explicitly state the module in pycoalaip's docstrings since sphinx usually does a pretty good job of autofinding these.

sohkai: Yeah, that's a good point. I'll change it to be more explicit.

sbellem:
:+1:  we can merge this now, right?

PR #750, opened 2016-10-26
Explicitly pass settings for flask into flask app factory

sohkai:
Slight changes to how we instantiate the flask server for clarity. AFAIK, the `settings` dict that's passed around is actually for gunicorn and not flask so this makes it more clear which flask settings we're configuring.

### DIFF
--- a/bigchaindb/web/server.py
+++ b/bigchaindb/web/server.py
@@ -49,19 +49,22 @@ class StandaloneApplication(gunicorn.app.base.BaseApplication):
         return self.application
 
 
-def create_app(settings):
+def create_app(*, debug=False, threads=4):
     """Return an instance of the Flask application.
 
     Args:
         debug (bool): a flag to activate the debug mode for the app
             (default: False).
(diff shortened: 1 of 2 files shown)


sbellem:
:+1:

bigchaindb/web/server.py
```diff
@@ -88,6 +91,7 @@ def create_server(settings):
     if not settings.get('threads'):
```
diminator: not sure why this is still using settings and not explicit args

sohkai: Mmm true, `settings` could be passed in as arguments by destructing them from the bdb config, but then we'd have to reconstruct them into a dict when we create the gunicorn app (see https://github.com/bigchaindb/bigchaindb/pull/750/files#diff-9943b8877e792a45dc46b4e483e1c02eR35).

diminator: ok, true - you can go for the general settings then

PR #753, opened 2016-10-27
Moved docs dir to docs/server

ttmc:
I want to move the root BigchainDB docs into this repository (bigchaindb/bigchaindb). To make that possible, this pull request moves the contents of the `docs` directory to `docs/server`, so that the root docs can be moved into `docs/root` in the future.

I didn't move `docs/README.md` into `docs/server`. Instead I edited it to be instructions for _all_ docs subdirectories.

I also updated the path to `version.py` in `docs/server/source/conf.py`.

Note: After merging this PR, some settings on ReadTheDocs will have to be updated, mainly the path to the BigchainDB Server docs.


sbellem:
:+1:

PR #754, opened 2016-10-27
Isolate database calls

vrde:
This is a first step towards database isolation.

This PR introduces the first backend called `RethinkDBBackend` to encapsulate all the queries that CRUD data. I'll follow up with a more precise description of the next steps.


bigchaindb/core.py
```diff
@@ -60,6 +60,7 @@ def __init__(self, host=None, port=None, dbname=None,
         self.host = host or bigchaindb.config['database']['host']
         self.port = port or bigchaindb.config['database']['port']
         self.dbname = dbname or bigchaindb.config['database']['name']
+        self.backend = backend or get_backend()
```
r-marques: we should update the docstrings

r-marques: I don't think we need `self.connection` anymore

vrde: > I don't think we need self.connection anymore

`self.connection` is still used in some tests. Before updating the tests I want to understand better how to abstract some other database calls. I'll change this part in my next PR, but before I'll put all the details about the refactoring in an issue.

r-marques: Ok, also note that if you need to access the `connection` from a `bigchain` instance you can always do `self.backend.connection`

bigchaindb/core.py
```diff
@@ -131,23 +129,30 @@ def reassign_transaction(self, transaction, durability='hard'):
             # There is no other node to assign to
             new_assignee = self.me
 
-        response = self.connection.run(
-                r.table('backlog')
-                .get(transaction['id'])
-                .update({'assignee': new_assignee, 'assignment_timestamp': time()},
-                        durability=durability))
-        return response
+        return self.backend.update_transaction(
```
r-marques: `durability` ?

vrde: good :eye:

`durability` is `hard` by default, I'll remove the option

vrde: This might change, I have to check if MongoDB has a similar option.

bigchaindb/core.py
```diff
@@ -382,13 +352,7 @@ def get_spent(self, txid, cid):
         """
         # checks if an input was already spent
         # checks if the bigchain has any transaction with input {'txid': ..., 'cid': ...}
-        response = self.connection.run(
-                r.table('bigchain', read_mode=self.read_mode)
-                .concat_map(lambda doc: doc['block']['transactions'])
```
r-marques: At some point we should consider a secondary index here

bigchaindb/db/backends/rethinkdb.py
```diff
@@ -0,0 +1,379 @@
...
+        dict that contains extra information that can be appended to the transaction.
+
+        To make it easy to query the bigchain for that particular metadata we create a UUID for the metadata and
+        store it with the transaction.
+
+        Args:
+            metadata_id (str): the id for this particular metadata.
+
+        Returns:
+            A list of transactions containing that metadata. If no transaction exists with that metadata it
```
vrde: If we consider a cursor an `iterable` we are quite safe. But yes, this might be subject to change, I'll wait for some more inputs for MongoDB to decide what's the best approach.

bigchaindb/db/utils.py
```diff
@@ -67,6 +67,18 @@ def _connect(self):
...
 
 
+def get_backend():
+    '''Get a backend instance.'''
+
+    from bigchaindb.db.backends import rethinkdb
+
+    # NOTE: this function will be re-implemented when we have real
+    # multiple backends to support. Right now it returns the RethinkDB one.
+    return rethinkdb.RethinkDBBackend(host=bigchaindb.config['database']['host'],
```
r-marques: These parameters should be passed to `get_backend` and not read from the config or else if someone instantiates Bigchain with custom parameters they will not be passed to `get_backend`.

bigchaindb/db/backends/rethinkdb.py
```diff
@@ -0,0 +1,381 @@
...
+    def has_transaction(self, transaction_id):
+        """Check if a transaction exists in the bigchain table.
+
+        Args:
+            transaction_id (str): the id of the transaction to check.
+
+        Returns:
+            ``True`` if the transaction exists, ``False`` otherwise.
+        """
+        return bool(self.connection.run(
```
r-marques: This is correct but I just want to make a note because `bool` always makes me nervous. There are a lot of values that evaluate to true and this can lead to undefined behavior, especially when the input of `bool` comes from an external call

bigchaindb/core.py
rhsimplex: MongoDB has read and [write concerns](https://docs.mongodb.com/manual/reference/write-concern/) where one can specify whether to return immediately, wait until written to disk, or wait until written to a majority of nodes.  This can definitely be set globally (collection or db-wide...or both?) but I'm not sure if it can be set on a per query basis.

sohkai:
I vote for purging db calls from tests.

r-marques:
:+1:

PR #757, opened 2016-10-27
Updated the docs page about the HTTP API

ttmc:
- Updated the example transactions (JSON documents) because they've changed (e.g. with the merge of the digital asset model).
- Removed the text with tips about how to build a valid transaction using the Python Server API. That's beyond the scope of this page. I pointed to some other docs pages about that.
- Added headers to make it easier to navigate to specific API endpoints.


sbellem:
:+1:

PR #759, opened 2016-10-28
Removed docs page "The Python Server API by Example"

ttmc:
The examples on the doce page "The Python Server API by Example" were mostly broken, and Sylvain has been translating them all to Python Driver examples, so I: 

- deleted the page "The Python Server API by Example"
- removed that page from the index
- changed the last line of the "Quickstart" page (for BigchainDB Server) to point the reader to the BigchainDB Python Driver docs


sbellem:
:+1:

PR #771, opened 2016-10-31
Security: updated the Ubuntu AMI in example_deploy_config.py

ttmc:
Recently, a critical security bug was reported in the Linux kernel: CVE-2016-5195 a.k.a. "[Dirty COW](https://dirtycow.ninja/)". This pull request just upgrades the AMI given in the example AWS deployment configuration file to one with a version of Ubuntu that isn't vulnerable.

Specifically, it uses the AMI with image ID `ami-9c09f0f3`. I launched an instance with that AMI to check the kernel version:

```text
$ uname -rv
3.13.0-100-generic #147-Ubuntu SMP Tue Oct 18 16:48:51 UTC 2016
```

I looked around and it seems that that version is right on the borderline of being patched, but _is_ patched.


r-marques:
:+1:

PR #783, opened 2016-11-02
Isolate db testcalls

rhsimplex:
Ready to merge:

I've made some progress getting db-specific calls out of the tests:

**Files which contain db calls not supported by `core.py`**

1. `tests/pipelines/test_vote.py` **done**
2. `tests/pipelines/test_block_creation` **done**
3. `tests/pipelines/test_election.py` **done**
4. `tests/db/test_bigchain_api.py` **done**

These are mostly refactorable, but some testing conditions are created using direct calls to RethinkDB that  I can't mimic with the existing API.

**I suggest adding the following functions to `core.py` to help** since they might be helpful in other contexts too.

1. `get_block` get a block by id #799 **merged**
2. `backlog_count` get the size of the backlog.  You do this now with `b.backend.get_stale_transactions(0)` but it's hacky #806 **merged**
3. ~~`block_count` all blocks, as filtering by validity is prohibitively expensive~~ this already exists
4.  `get_genesis_block` #809 **merged**
 
`test_vote` also uses db calls to return the entire chain/vote table, which will probably need to be rewritten.

The following will be handled in another PR:
**Files which contain db-specific tests:**  I don't think these ones can be abstracted out, or the test setup is too complicated for me to see an easy abstraction.  We'll need parallel tests for every backend:

1. `tests/test_command.py`
2. `tests/test_core.py`
3. `tests/test_run_query_util.py`
4. `tests/db/test_utils.py`

See also #784

### DIFF
--- a/tests/db/test_bigchain_api.py
+++ b/tests/db/test_bigchain_api.py
@@ -303,44 +303,24 @@ class TestBigchainApi(object):
 
     @pytest.mark.usefixtures('inputs')
     def test_genesis_block(self, b):
-        import rethinkdb as r
-        from bigchaindb.util import is_genesis_block
-        from bigchaindb.db.utils import get_conn
+        block = b.backend.get_genesis_block()
 
-        response = list(r.table('bigchain')
-                        .filter(is_genesis_block)
-                        .run(get_conn()))
-
-        assert len(response) == 1
-        block = response[0]
         assert len(block['block']['transactions']) == 1
         assert block['block']['transactions'][0]['transaction']['operation'] == 'GENESIS'
         assert block['block']['transactions'][0]['transaction']['fulfillments'][0]['input'] is None
 
     def test_create_genesis_block_fails_if_table_not_empty(self, b):
-        import rethinkdb as r
         from bigchaindb.common.exceptions import GenesisBlockAlreadyExistsError
-        from bigchaindb.util import is_genesis_block
-        from bigchaindb.db.utils import get_conn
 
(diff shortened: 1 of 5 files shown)


vrde:
> I suggest adding the following functions to core.py to help since they might be helpful in other contexts too.

:+1: 

There are two calls also in `pipelines/election.py` and `pipelines/block.py`. Can you move those two as well? If you think it's out of the scope of this PR we can open another one.

sohkai:
Can't comment too much on the queries except that they look fine to my untrained eye and `b.backend.get_stale_transactions(0)` is definitely a hack.

In the future, with more experience in implementing backends, we'll have to make a decision about how far we go with the test abstraction: maybe we'll still need isolated backend tests? Maybe we could have a suite of tests (e.g. testing core.py and the pipelines) that are generic enough to be run on any backend to check basic implementation sanity?

This is certainly a good step towards that :).

rhsimplex:
Yes definitely.  A lot of the tests now require pulling an entire table.  I'm sure you agree there shouldn't be a direct backend call that allows this.  I rewrote some of the tests around this, but grabbing all the `tx` with `get_stale_transactions` sure is the easy way out.

sohkai:
> I'm sure you agree there shouldn't be a direct backend call that allows this

Yeah... Maybe for now we can wrap all the `b.backend.get_stale_transactions(0)` in a testing util that we can factor out later. This also makes it easier to document why we're using this.

PR #788, opened 2016-11-03
Abstract database interface

vrde:
This is a proposal to abstract the database and offer a unified interface to developers.

# Rationale
BigchainDB uses the database for ~6 different types of operations:
 - schema operations, like creating the database, tables, and the indexes
 - queries, like create/read/update/delete operations
 - changefeeds, to get changes in real time
 - connections, to connect to the actual concrete database
 - testing, because when we test we need to check in the DB if the value is there
 - firewall, to block some specific potentially malicious activity on the database (like drop/delete)

We need to support all those operations for different databases.
The scope of this PR is to:
 - abstract the database layer
 - move RethinkDB specific code in different modules.

@rhsimplex is currently working to add support for MongoDB. This PR should help him find a home for his code :house_with_garden:.

# Architecture
~The architecture is based on the pattern *Abstract Factory*.~ [no more].
After some talking with @sbellem and @sohkai, we realized that we can try a different strategy and use [`@singledispatch`](https://github.com/bigchaindb/bigchaindb/pull/788#issuecomment-261210667).

We agree on:
- queries have no internal state (so they can be just functions)
- connections have internal state (so they can be classes)

This fits quite well our problem since we can model queries just as simple functions, where the first argument of the function is the connection object. `@singledispatch` will use that first argument to decide which registered function to call.

As an overview, we will have interfaces in those modules. Almost all the functions in those modules will be wrapped using `@singledispatch`:
 - `bigchaindb.backend.query` to define the interface for querying the database.
 - `bigchaindb.backend.connection` to define the interface for connecting the database and a function to create the correct `Connection` instance based on the configuration. The function `bigchaindb.backend.connect(backend, host, port, dbname)` will use the name of the `backend` to instantiate the correct `Connection` object.
 - `bigchaindb.backend.schema` to define the interface for creating and dropping tables and indexes.
 - `bigchaindb.backend.changefeed` to define the interface to get events from the database.

Those other modules will implement all the functionalities for RethinkDB. The functions in those modules will be wrapped with something like `@<function_name>.register(RethinkDBConnection)`:
 - `bigchaindb.backend.rethinkdb.query` 
 - `bigchaindb.backend.rethinkdb.connection`
 - `bigchaindb.backend.rethinkdb.schema`
 - `bigchaindb.backend.rethinkdb.changefeed`

After those changes the workflow to communicate with the database will something like:
 - use `bigchaindb.backend.connect` to get a connection object
 - use the generic *interface* contained in `bigchaindb.backend.[query|schema|changefeed]` to interact with the database, e.g. `bigchaindb.backend.query.write_vote(conn, vote)`. `@singledispatch` will look up the type of the connection instance and will call the correct function.

# Todo:
- [x] ~~review fixtures/conftest @sbellem~~ (tests will be fully abstracted later)
- [x] ~~investigate how to connect to MongoDB @rhsimplex~~ (mongo will be in a separate PR)
- [x] add function to select the correct backend when creating a connection @vrde #886 
- [x] port queries to `@singledispatch` @sohkai 
- [x] update configuration to add support to multiple backends

# Before merging:
- [x] Document new backend setting in configuration
- [x] Document how to use the backend module, as docstrings
- [x] Document new pytest option
- [x] ~~Rebase on master~~ Merge on master
- Fix commit history to better represent moves of code
    - [x] backend/rethinkdb/connection.py was moved from db/utils.py
    - [x] backend/rethinkdb/schema.py was moved from db/utils.py (after Connection was moved out)
    - [x] backend/rethinkdb/query.py was moved from db/backends/rethinkdb.py
    - [x] tests/db/rethinkdb/test_schema.py was moved from test_utils.py
    - [ ] Rebase clean up branch with correct attribution and force push to this branch


sohkai:
From talks with @vrde.

Definitive actions we've more or less agreed on:
* `Query` should have its functions split into model-specific classes or modules
* `get_backend_factory()` should cache its output
* Some parts of `RethinkDBConnection` could be generalized so other backend's Connection classes don't have to reimplement them
* The `core.Bigchain` class should be a user-friendly facade to the backend and not be coupled with a `Connection`. They may also wrap around the backend's raw queries with model objects (e.g. light ORM).

---------

A big question is whether we need classes for the `Backend`, and if we need two layers of factories (to create the *abstract factory* pattern):

1. How much state should the `Backend` implementations hold? They could hold onto a `Connection` and some settings, but those could also be passed in.
1. If the `Backend` implementations don't hold onto state, do we need classes for them or can they be functions in a module?
   * How would we expose and use these modules? How would we enforce their interfaces?
1. If we use classes, could we use just a *factory* pattern instead of an *abstract factory*, so that each `Backend` class holds an instance of `Query`, `Schema`, and `Changefeed` as state rather than using a `BackendFactory` to generate these separately? Perhaps the *abstract factory* pattern is necessary for the `Changefeed` and `Firewall` implementations, in case they may need to be instantiated multiple times?
1. Should the `Backend`s be implemented as singletons? (Note: singletons that don't hold state should just become modules).
1. If we use classes, should we use `AbstractBaseClass`s for the interfaces?

On factories: I view factories as necessary only when we need to make multiple instances of something during a run. I can see us doing that for `Connection`s, but I'm not convinced this is the case for the other groups of logic (`Query`, `Schema`, `Changefeed`, and `Firewall`).

On classes: I prefer to decouple logic from state when the logic is generic enough to not be dependent on the state.

sbellem:
As suggested offline, we could benefit from acknowledging the architecture of a few (say 3) mature systems that have similar needs of abstraction. This could help us validating design decisions, and/or be inspired by ideas we had not thought about. Also, given the amount of details, refinements, patches, etc that a mature system needs to go through, looking at such systems may help us foresee what could be needed later on. Foreseeing these needs may help us make design decisions today that will ease the evolution of our current design and implementation.

sbellem:
Following up on previous comment https://github.com/bigchaindb/bigchaindb/pull/788#issuecomment-259139741, some suggestions, and references for systems to look at, that may or may not be helpful, but just to get started (please feel free to add to the list):

* [The Architecture of Open Source Applications](http://aosabook.org/en/index.html)
* Django
* Celery

vrde:
> Definitive actions we've more or less agreed on:
> - Query should have its functions split into model-specific classes or modules
No need for now. We can keep everything in the same "query" class. We can always split it up later if we need.

> get_backend_factory() should cache its output
No need for now. We can add it later.

> Some parts of RethinkDBConnection could be generalized so other backend's Connection classes don't have to reimplement them

Yep!

> The core.Bigchain class should be a user-friendly facade to the backend and not be coupled with a Connection. They may also wrap around the backend's raw queries with model objects (e.g. light ORM).

Yep!

The following questions are related to the **state** so I'll bundle them together.

> How much state should the Backend implementations hold? They could hold onto a Connection and some settings, but those could also be passed in.

and

> Should the Backends be implemented as singletons? (Note: singletons that don't hold state should just become modules).

and

> If the Backend implementations don't hold onto state, do we need classes for them or can they be functions in a module?

Even if the don't have state **now** doesn't mean we won't have it **later**. I'm scared that it's a big assumption that we *do not have state* there. The abstract factory pattern is easy and known by devs, so I think it's actually better than something tailored for a specific use case (that we still didn't flesh out completely).

> If we use classes, could we use just a factory pattern instead of an abstract factory, so that each Backend class holds an instance of Query, Schema, and Changefeed as state rather than using a BackendFactory to generate these separately? Perhaps the abstract factory pattern is necessary for the Changefeed and Firewall implementations, in case they may need to be instantiated multiple times?

Django uses a similar approach to the one described in my PR, check out the base class and the implementation of some backends operations:
 - https://github.com/django/django/blob/master/django/db/backends/base/operations.py
 - https://github.com/django/django/blob/master/django/db/backends/mysql/operations.py

On the other side, instantiating a backend happens in a different way. What is pretty interesting is the function `load_backend` in [`django.db.utils`](https://github.com/django/django/blob/master/django/db/utils.py#L105):

```python
# ...
from importlib import import_module
# ...

def load_backend(backend_name):
    """
    Return a database backend's "base" module given a fully qualified database
    backend name, or raise an error if it doesn't exist.
    """

    try:
        return import_module('%s.base' % backend_name)
    except ImportError as e_user:
        # ... manage the error
```

So in Django they return the whole module instead of an instance of a class to generate other instances. I truly believe that this is the pythonic way to do an *abstract factory pattern*. It's much lighter, it uses the dynamic properties of the language, and we get rid of extra classes.

> How would we expose and use these modules? How would we enforce their interfaces?

This is not something you'd do in Python. Duck typing FTW imo, and we can check this with testing. 

> If we use classes, should we use AbstractBaseClasss for the interfaces?

I don't have arguments :+1: or :-1: it. @sbellem do you have any experience with that?

sbellem:
Please forgive me to add to this already long issue. Here's a note about `singledispatch` as it may be useful somewhere in this undertaking.

* docs: https://docs.python.org/3/library/functools.html#functools.singledispatch
* Interesting blog about it: http://lukasz.langa.pl/8/single-dispatch-generic-functions/

`singledispatch` allows to dispatch based on the type of an object.

No action required by anyone here, just wish to have this note somewhere to avoid forgetting about it.

vrde:
> singledispatch allows to dispatch based on the type of an object. In other words, one can write one generic function, such that depending on the type of its first argument, it may dispatch to another function that is meant to handle the type of the first argument.

Interesting, it's basically overloading a function. I think we cannot apply it here, since it applies to types (and enums).

sohkai:
>> How would we expose and use these modules? How would we enforce their interfaces?

> This is not something you'd do in Python. Duck typing FTW imo, and we can check this with testing.

>> If we use classes, should we use AbstractBaseClasss for the interfaces?

> I don't have arguments 👍 or 👎 it. @sbellem do you have any experience with that?

If we're not somehow enforcing the interfaces directly, there's no reason to use ABCs.

The question then becomes: how does someone write a test that will check that their implementation is correct? Do we heavily abstract our testing to only use core API functionality? Somehow let the database implementations cheat and provide database-specific queries?

sohkai:
**TL;DR:**
- For loading database interfaces, **SQLAlchemy essentially uses the same strategy as Django**: dynamically loading an external or internal module by parsing a configuration.
- After the right module's been loaded (from above), we have to figure out how to create the right classes / use the right interfaces. I'm not familiar with how Django expects its backend implementations to be structured, but SQLAlchemy presents some ideas:
   - Use a single class as the entry point into a database implementation, using classmethods that return other classes (not instances!) as a point of customization.
    - Each module declares its entrypoint class through a global variable

**Ideas for us:**
- We could enforce implementations to implement functions that return their version of a particular interface (e.g. `Connection` or `Schema`) from the module entry point
  - These could be instances of a class, or the class itself
- We could add entrypoints through `setup_tools` to allow external packages to be automatically found and used as a database implementation
 
=================

Expanded notes on how SQLAlchemy does things:

- Databases are implemented through "dialects", which have a [standard interface](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/engine/interfaces.py?at=master&fileviewer=file-view-default#interfaces.py-16) and a [default implementation](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/engine/default.py?at=master&fileviewer=file-view-default#default.py-36) that does things like set up the default pool classes, etc, that aren't actually database-implementation specific.
- Loading dialects:
   1. The [docs on configuration](http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.create_engine) have good diagrams and examples, but it's interesting to note they use the standard URL for configuring which database and driver; something like `create_engine('postgresql+psycopg2://scott:tiger@localhost/mydatabase')` where the notation is `dialect+driver://username:password@host:port/database`.
   1. When `create_engine()` starts creating a new `Engine` instance (essentially all of SQLAlchemy), it [parses the given URL](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/engine/strategies.py?at=master&fileviewer=file-view-default#strategies.py-50) and tries to load the dialect module and find the actual dialect class:

    ```python
    u = url.make_url(name_or_url)
    ...
    entrypoint = u._get_entrypoint() # loads the dialect module
    dialect_cls = entrypoint.get_dialect_cls(u)
    ```

  1.  It's interesting to note that they use a [cached registry](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/util/langhelpers.py?at=master&fileviewer=file-view-default#langhelpers.py-183) for all plugins and dialects, and that one can [register new dialects via the standard `setup_tools` plugin infrastructure or by doing it manually](http://sqlalchemy.readthedocs.io/en/latest/core/connections.html#registering-new-dialects)
  1. To load the dialects, the loader either uses [`pkg_resources.entry_points` (from `setup_tools`)](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/util/langhelpers.py?at=master&fileviewer=file-view-default#langhelpers.py-183) or a [direct import of the dialect module through `__import__('sqlalchemy.dialects.%s' % (dialect)`](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/dialects/__init__.py?at=master&fileviewer=file-view-default#__init__.py-43)
- Dialect structure:
    - The dialect class has classmethods like [`get_dialect_class`](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/engine/interfaces.py?at=master&fileviewer=file-view-default#interfaces.py-771) and [`get_pool_class`](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/engine/default.py?at=master&fileviewer=file-view-default#default.py-238) that return the class SQLAlchemy should instantiate for certain things, e.g. the dialect or pool.
    - A dialect module is [expected to contain certain globals](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/dialects/__init__.py?at=master&fileviewer=file-view-default#__init__.py-47), e.g. [`dialect` (points to the dialect class to use](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/dialects/__init__.py?at=master&fileviewer=file-view-default#__init__.py-50): see [psycopg2's module](https://bitbucket.org/zzzeek/sqlalchemy/src/30abab55726b0a156180c7c3da0eeb8ae61303d5/lib/sqlalchemy/dialects/postgresql/psycopg2.py?at=master&fileviewer=file-view-default#psycopg2.py-702) 
    - It's not really a factory, but I guess they [call it one](http://docs.sqlalchemy.org/en/latest/core/internals.html#sqlalchemy.engine.interfaces.Dialect)  ¯\_(ツ)_/¯

sohkai:
Remaining tests that are still coupled with RethinkDB (and the changefeed) will be done in another PR following this one to make it easier for the MongoDB implementation.

sohkai:
The current travis error should be fixed when https://github.com/bigchaindb/bigchaindb/pull/918 is merged.

PR #793, opened 2016-11-05
get_transaction will return a tx from backlog, even if there are some in invalid blocks

ttmc:
I made a slight change to how `Bigchain.get_transaction()` works. Before, if there was a transaction with the specified `txid` in the backlog _and_ in some invalid blocks (but nowhere else), `Bigchain.get_transaction()` would return `None`. I changed it so that in that case, it returns the transaction from the backlog (and if the user asked for the transaction's status, it will be `'backlog'`).

I updated and expanded the docstring for `Bigchain.get_transaction()`.

I also made a minor change to a test: `	test_read_transaction_invalid_block` didn't need to write the transaction to the backlog, so I commented-out that line.

### DIFF
--- a/tests/db/test_bigchain_api.py
+++ b/tests/db/test_bigchain_api.py
@@ -245,7 +245,7 @@ class TestBigchainApi(object):
         inputs = input_tx.to_inputs()
         tx = Transaction.transfer(inputs, [user_vk], input_tx.asset)
         tx = tx.sign([user_sk])
-        b.write_transaction(tx)
+        # There's no need to b.write_transaction(tx) to the backlog
 
         # create block
         block = b.create_block([tx])
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -189,19 +189,25 @@ class Bigchain(object):
             return False
 
     def get_transaction(self, txid, include_status=False):
-        """Retrieve a transaction with `txid` from bigchain.
+        """Get the transaction with the specified `txid` (and optionally its status)
 
-        Queries the bigchain for a transaction, if it's in a valid or invalid
-        block.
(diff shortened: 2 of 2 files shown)


ttmc:
Note: I originally didn't add the second test and code coverage went down slightly. Once I added the second test, it went up by 0.06%.

tests/db/test_bigchain_api.py
```diff
@@ -257,9 +257,38 @@ def test_read_transaction_invalid_block(self, b, user_vk, user_sk):
...
 
     @pytest.mark.usefixtures('inputs')
+    def test_read_transaction_invalid_block_and_backlog(self, b, user_vk, user_sk):
+        from bigchaindb.models import Transaction
+
+        input_tx = b.get_owned_ids(user_vk).pop()
+        input_tx = b.get_transaction(input_tx.txid)
+        inputs = input_tx.to_inputs()
+        tx = Transaction.transfer(inputs, [user_vk], input_tx.asset)
+        tx = tx.sign([user_sk])
```
sbellem: Just a note, not a blocker to merge this PR.

I wonder whether it is really needed to write a "real" tx for this test, as it seems all we would need is a "dummy" tx, just to make sure the block in-validation can take place.

sbellem:
:+1:

tests/db/test_bigchain_api.py
ttmc: Yeah, but I'm not sure how to go about that. I based the two "new" tests on the one original test, so at least this is no worse than before.

sbellem: Ah, yeah, ok, no problem!

Issue #488, opened 2016-07-26
Implementation and testing of divisible assets


r-marques:
Specs defined in #487

ttmc:
Resolved by pull request #794

Issue #327, opened 2016-05-23
Add assurances that a node can't modify the payload in a CREATE transaction


ttmc:
Right now, when a client sends a payload to a node to create a new asset, the node can change the payload and sign _that_ to create the asset/transaction that actually gets stored.

Sophisticated clients can do things to make such shenanigans detectable by other clients (e.g. by including their public key and a signature in the payload itself, and by sending two transactions, with the first one having the hash of the second payload).

We could make it easier for clients to rest assured that a node can't corrupt their payload before the node creates the asset:
- Require the payload to include the public key of the client.
- Add a client-provided payload signature to the transaction model.
- Add two new steps to transaction validation: 1) Check that the public key of the original owner equals the public key in the payload, 2) Verify the client-provided signature of the payload.

The initial node can't change the client's public key (in both places): the client-provided signature would change.

The initial node can't change the payload only: the client-provided signature would change.

The initial node could change the client's public key _and_ payload, but then it's basically a totally different transaction! The client could see that their transaction never ended up in the bigchain, report a problem, and send their transaction to a different node.

**To Consider Next**: What about TRANSFER transactions? They can also have a payload (but might not).

ttmc:
Issue #347 is in the same vein as this one. I wouldn't say that one is a duplicate of the other, but whoever tackles one should look at both.

TimDaub:
@ttmc Removing `transform_create` was handled in #634 and #590.

Since a `CREATE` transaction is now signed with the client's private key and since the bdb-driver is evaluating the correctness of the transaction's id that was sent to a node, a scenario like this is resolved.

If correct, please close this ticket.

r-marques:
### Proposal

In the initial design of BigchainDB it was decided that only federation nodes could issue assets on behalf of the users. This seemed like a good idea at the time but we never added any additional logic or found a use case where this was helpful. Moreover we introduced transaction malleability and other problems #347 .

With the introduction of bigchaindb common one of the solutions was to have the clients themselves issue the assets. This way a client creates a signs a `CREATE` transaction and sends it to the network. This removes any transaction malleability problems and ensures the asset was created by the actual owner of a particular public key.

The problem now is that the when creating an asset the client does not have any condition to fulfill however it still needs to provide a signature in the form of a fulfillment.

Since `CREATE` transactions are a special case here is how I propose we should handle the conditions and fulfillments:

- The fulfillment will be an implicit fulfillment:
[...]

PR #794, opened 2016-11-06
Divisible assets

### LINKED ISSUE #487: Write specs for divisible assets
Define data model specs and core consesus protocol to to support divisible assets

r-marques:
resolves #487 #488 #327 

Adds support for divisible assets.
The main change is in how we define `owners_after` since now we need to pass the public keys and amounts.

`owners_after` is now a list of tuples of `([public_keys], amount)`

Example:
```python

# Non-divisible assets
# non-divisible assets will always have one output with amount=1
asset = Asset(divisible=False)

# single owner
Transaction.create([b.me], [([user_vk], 1)], asset=asset)

# multiple owners
Transaction.create([b.me], [([user_vk, user2_vk], 1)], asset=asset)

# Divisible Assets
asset = Asset(divisible=True)

# single output single owner
Transaction.create([b.me], [([user_vk], 100)], asset=asset)

# multiple output single owner per output
Transaction.create([b.me], [([user_vk], 50), ([user_vk), 50])], asset=asset)

# multiple outputs multiple owners per output
Transaction.create([b.me], [([user_vk, user2_vk], 50), ([user_vk, user2_vk], 50)], asset=asset)

# multiple outputs: one output with a single owner, one output with multiple owners
Transaction.create([b.me], [([user_vk], 50), ([user_vk, user2_vk], 50)], asset=asset)
```

We can perform any combination we want. Each tuple creates a condition.
If there is only one key in the list we create a simple signature condition.
If there are multiple keys in the list we create a threshold signature condition

Most of the changes in the tests is due to the change in the api, except for `tests/assets/test_divisible_assets.py` which contains the new tests for divisible assets.

Some changes in `common/transaction.py`:
- Removed boilerplate for timelock conditions. It should be implemented on the client side
- Removed support for hashlocks. The implementation only worked for some specific use cases. Its a custom use case that does not work well with the `create, transfer, sign` api. Also to be implemented on the client side

#### TODO:
- [x] Fix failing tests
- [x] Add support for multiple `owners_before` in `CREATE` transactions
- [x] Cleanup code


sbellem:
Just a general comment and opinion regarding the names of variables, for fulfilments and conditions. I think using the explicit names `fulfilments`, and `conditions` rather than the shorten forms `ffils` and `conds` makes the code clearer and easier to understand.

Short forms may perhaps be used in loops, or in `list`/`dict` comprehensions where it's obvious what one deals with. But when the variable appears in multiple places and is being passed around, the full name helps making the code clearer, and faster to understand.

TimDaub:
> from PEP 8:

> in many cases, abbreviations and technical terms are used which aren't English

Quite funny how they literally start to import `os` and `sys` one paragraph below.

sbellem:
@TimDaub 

> > from PEP 8:
> >
>  >   in many cases, abbreviations and technical terms are used which aren't English
> 
> Quite funny how they literally start to import os, mypkg and sys one paragraph below.

Well, they do clearly state "For Python 3.0 and beyond, ...".  `os` and `sys` pre-date Python 3.

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    assert 'subfulfillments' in ffill
+    assert len(ffill['subfulfillments']) == 2
+
+
+# TRANSFER divisible asset
+# Single input
+# Single owners_before
+# Single output
+# Single owners_after
+# TODO: I don't really need inputs. But I need the database to be setup or
```
TimDaub: Create conftest setup for this?

I think if we don't do it now, we'll probably never do it.

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    b.write_block(block, durability='hard')
+    # vote
+    vote = b.vote(block.id, b.get_last_voted_block().id, True)
+    b.write_vote(vote)
+
+    # TRANSFER divisible asset
+    # `b` transfers its 50 shares to `user_vk`
+    # after this transaction `user_vk` will have a total of 100 shares
+    # split across two different transactions
+    tx_transfer1 = Transaction.transfer([tx_create.to_inputs()[1]],
```
TimDaub: `to_inputs` actually accepts a list of condition indexes `cid`s.

`tx_create.to_inputs([0])`

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    assert block.validate(b) == block
+    b.write_block(block, durability='hard')
+    # vote
+    vote = b.vote(block.id, b.get_last_voted_block().id, True)
+    b.write_vote(vote)
+
+    # TRANSFER
+    # `user_vk` combines two different transaction with 50 shares each and
+    # transfers a total of 100 shares back to `b`
+    tx_transfer2 = Transaction.transfer([tx_create.to_inputs()[0],
```
TimDaub: `.to_inputs([0])`

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+                                         tx_transfer1.to_inputs()[0]],
```
TimDaub: `.to_inputs([0])`

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    tx_transfer_signed = tx_transfer.sign([user_sk])
+    with pytest.raises(AmountError):
+        tx_transfer_signed.validate(b)
+
+
+@pytest.mark.skip(reason='Figure out how to handle this case')
+@pytest.mark.usefixtures('inputs')
+def test_threshold_same_public_key(b, user_vk, user_sk):
+    # If we try to fulfill a threshold condition where each subcondition has
+    # the same key get_subcondition_from_vk will always return the first
```
TimDaub: As far as I know, it returns a list: https://github.com/bigchaindb/cryptoconditions/blob/master/cryptoconditions/types/threshold_sha256.py#L153

Maybe the error is in `bigchaindb.common.transaction` then?

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    block = b.create_block([tx_create_signed])
+    assert block.validate(b) == block
+    b.write_block(block, durability='hard')
+    # vote
+    vote = b.vote(block.id, b.get_last_voted_block().id, True)
+    b.write_vote(vote)
+
+    # create a transfer transaction with one output and check if the amount
+    # is 3
+    tx_transfer = Transaction.transfer(tx_create.to_inputs(), [([b.me], 3)],
```
TimDaub: That's pretty cool :)

tests/common/test_transaction.py
```diff
@@ -614,16 +598,16 @@ def test_validate_multiple_fulfillments(user_ffill, user_cond, user_priv):
...
+    tx = Transaction(Transaction.CREATE, Asset(divisible=True),
                      [user_ffill, deepcopy(user_ffill)],
-                     [user_ffill, deepcopy(user_cond)])
+                     [user_cond, deepcopy(user_cond)])
 
     expected_first = deepcopy(tx)
     expected_second = deepcopy(tx)
     expected_first.fulfillments = [expected_first.fulfillments[0]]
-    expected_first.conditions = [expected_first.conditions[0]]
+    expected_first.conditions = expected_first.conditions
```
TimDaub: I'd remove that line.

tests/common/test_transaction.py
```diff
@@ -614,16 +598,16 @@ def test_validate_multiple_fulfillments(user_ffill, user_cond, user_priv):
...
     expected_second.fulfillments = [expected_second.fulfillments[1]]
-    expected_second.conditions = [expected_second.conditions[1]]
+    expected_second.conditions = expected_second.conditions
```
TimDaub: I'd remove that line.

tests/common/test_transaction.py
```diff
@@ -715,10 +699,6 @@ def test_validate_fulfillments_of_transfer_tx_with_invalid_params(transfer_tx,
     with raises(TypeError):
         transfer_tx.operation = "Operation that doesn't exist"
         transfer_tx.fulfillments_valid([utx.conditions[0]])
-    with raises(ValueError):
```
TimDaub: Why remove this check?

tests/common/test_transaction.py
```diff
@@ -876,11 +848,13 @@ def test_validate_threshold_create_transaction(user_pub, user_priv, user2_pub,
...
     from bigchaindb.common.transaction import Transaction, Asset
 
-    tx = Transaction.create([user_pub], [user_pub, user2_pub], data, Asset())
+    tx = Transaction.create([user_pub], [([user_pub, user2_pub], 1)],
+                            data, Asset())
     tx = tx.sign([user_priv])
     assert tx.fulfillments_valid() is True
 
 
+@mark.skip(reason='Hashlocks are not implemented')
```
TimDaub: Lets delete those tests for now.

tests/common/test_transaction.py
```diff
@@ -927,6 +901,7 @@ def test_create_create_transaction_hashlock(user_pub, data, data_id):
     assert tx == expected
 
 
+@mark.skip(reson='Hashlocks are not implemented')
```
TimDaub: Lets delete those tests for now.

tests/common/test_transaction.py
```diff
@@ -1014,16 +989,15 @@ def test_create_transfer_transaction_single_io(tx, user_pub, user2_pub,
     assert transfer_tx.fulfillments_valid([tx.conditions[0]]) is True
 
 
-@mark.skip(reason='FIXME: When divisible assets land')
 def test_create_transfer_transaction_multiple_io(user_pub, user_priv,
```
TimDaub: This test was testing to create a `TRANSFER` with multiple inputs and outputs.
Now the `TRANSFER` method just spends the outputs of a single transaction.

Either:

- rename this test's method
- implement multiple IO

Thoughts @r-marques?

tests/db/test_bigchain_api.py
```diff
@@ -797,77 +802,14 @@ def test_transfer_single_owner_single_input(self, b, inputs, user_vk,
...
-        tx = Transaction.transfer(inputs, [user2_vk], input_tx.asset)
+        tx = Transaction.transfer(inputs, [([user2_vk], 1)], input_tx.asset)
         tx = tx.sign([user_sk])
 
         # validate transaction
         assert b.is_valid_transaction(tx) == tx
         assert len(tx.fulfillments) == 1
         assert len(tx.conditions) == 1
 
-    @pytest.mark.skipif(reason=('Multiple inputs are only allowed for the '
```
TimDaub: I guess these have been replaced with tests from `test_divisible_assets.py`?

tests/db/test_bigchain_api.py
```diff
@@ -881,47 +823,14 @@ def test_single_owner_before_multiple_owners_after_single_input(self, b,
...
-        tx = Transaction.transfer(input_tx.to_inputs(), [[user2_vk, user3_vk]], input_tx.asset)
+        tx = Transaction.transfer(input_tx.to_inputs(),
+                                  [([user2_vk, user3_vk], 1)], input_tx.asset)
         tx = tx.sign([user_sk])
 
         assert b.is_valid_transaction(tx) == tx
         assert len(tx.fulfillments) == 1
         assert len(tx.conditions) == 1
 
-    @pytest.mark.skipif(reason=('Multiple inputs are only allowed for the '
```
TimDaub: I guess these have been replaced with tests from `test_divisible_assets.py`?

tests/db/test_bigchain_api.py
```diff
@@ -945,37 +854,15 @@ def test_multiple_owners_before_single_owner_after_single_input(self, b,
...
+        transfer_tx = Transaction.transfer(inputs, [([user3_vk], 1)],
+                                           input_tx.asset)
         transfer_tx = transfer_tx.sign([user_sk, user2_sk])
 
         # validate transaction
         assert b.is_valid_transaction(transfer_tx) == transfer_tx
         assert len(transfer_tx.fulfillments) == 1
         assert len(transfer_tx.conditions) == 1
 
-    @pytest.mark.skipif(reason=('Multiple inputs are only allowed for the '
```
TimDaub: I guess these have been replaced with tests from `test_divisible_assets.py`?

tests/db/test_bigchain_api.py
```diff
@@ -1000,46 +887,22 @@ def test_multiple_owners_before_multiple_owners_after_single_input(self, b,
...
-        tx = Transaction.transfer(tx_input.to_inputs(), [[user3_vk, user4_vk]], tx_input.asset)
+        tx = Transaction.transfer(tx_input.to_inputs(),
+                                  [([user3_vk, user4_vk], 1)], tx_input.asset)
         tx = tx.sign([user_sk, user2_sk])
 
         assert b.is_valid_transaction(tx) == tx
         assert len(tx.fulfillments) == 1
         assert len(tx.conditions) == 1
 
-    @pytest.mark.skipif(reason=('Multiple inputs are only allowed for the '
```
TimDaub: I guess these have been replaced with tests from `test_divisible_assets.py`?

tests/db/test_bigchain_api.py
```diff
@@ -1269,22 +1130,22 @@ def test_get_spent_single_tx_multiple_outputs(self, b, user_sk, user_vk):
...
 
-        # select inputs to use
-        inputs = sum([tx.to_inputs() for tx in transactions[:2]], [])
-
-        # create a transaction and block
-        tx = Transaction.transfer(inputs, len(inputs) * [[user2_vk]])
-        tx = tx.sign([user_sk])
-        block = b.create_block([tx])
+        # transfer the first 2 inputs
+        tx_transfer = Transaction.transfer(tx_create.to_inputs()[:2],
```
TimDaub: Formerly this test spent inputs from different transactions. Any reason why we got rid of this?

bigchaindb/common/transaction.py
```diff
@@ -649,54 +707,27 @@ def create(cls, owners_before, owners_after, metadata=None, asset=None,
             raise TypeError('`owners_before` must be a list instance')
         if not isinstance(owners_after, list):
             raise TypeError('`owners_after` must be a list instance')
+        if len(owners_before) == 0:
```
r-marques: That issue was closed. There is no use case where owners_before can be an empty list

bigchaindb/common/transaction.py
```diff
@@ -638,9 +709,23 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
...
+        else:
+            # In transactions other then `CREATE` we don't know if its a divisible asset
+            # or not, so we cannot validate the amount here
+            self.asset._validate_asset()
+
+
     @classmethod
     def create(cls, owners_before, owners_after, metadata=None, asset=None,
                secret=None, time_expire=None):
+        # TODO: Update docstring
```
libscott: Will this kind of TODO not be resolved in the PR?

bigchaindb/common/transaction.py
```diff
@@ -1075,14 +1188,19 @@ def gen_tx(fulfillment, condition, input_condition_uri=None):
...
+        #     raise ValueError('Fulfillments, conditions and '
+        #                      'input_condition_uris must have the same count')
+        # else:
+        if not fulfillments_count == input_condition_uris_count:
+            raise ValueError('Fulfillments and '
                              'input_condition_uris must have the same count')
         else:
             partial_transactions = map(gen_tx, self.fulfillments,
                                        self.conditions, input_condition_uris)
-            return all(partial_transactions)
```
r-marques: I removed the indent. Looking at it again there is actually no reason for the `else`
It either raises an exception or returns `all(partial_transactions)`

bigchaindb/common/transaction.py
```diff
@@ -677,28 +762,43 @@ def create(cls, owners_before, owners_after, metadata=None, asset=None,
             raise TypeError('`owners_after` must be a list instance')
 
         metadata = Metadata(metadata)
+
+        ffils = []
+        conds = []
+
+        # generate_conditions
+        for owner_after in owners_after:
+            # TODO: Check types so this doesn't fail unpacking
```
libscott: I think you can remove this, we don't need type checking everywhere for every value.

bigchaindb/common/transaction.py
```diff
@@ -770,20 +870,27 @@ def transfer(cls, inputs, owners_after, asset, metadata=None):
...
+        # if len(inputs) == len(owners_after):
+        #     if len(owners_after) == 1:
+        #         conditions = [Condition.generate(owners_after)]
+        #     elif len(owners_after) > 1:
+        #         conditions = [Condition.generate(owners) for owners
+        #                       in owners_after]
+        # else:
+        #     # TODO: Why??
+        #     raise ValueError("`inputs` and `owners_after`'s count must be the "
+        #                      "same")
```
libscott: Why is this commented code left here?

bigchaindb/common/transaction.py
```diff
@@ -770,20 +870,27 @@ def transfer(cls, inputs, owners_after, asset, metadata=None):
...
+
+        conds = []
+        for owner_after in owners_after:
+            pub_keys, amount = owner_after
```
libscott: Could also do `for pub_keys, amount in owners_after`... no biggie tho

bigchaindb/common/transaction.py
```diff
@@ -770,20 +870,27 @@ def transfer(cls, inputs, owners_after, asset, metadata=None):
...
+            conds.append(Condition.generate(pub_keys, amount))
+
 
         metadata = Metadata(metadata)
         inputs = deepcopy(inputs)
-        return cls(cls.TRANSFER, asset, inputs, conditions, metadata)
+        return cls(cls.TRANSFER, asset, inputs, conds, metadata)
```
libscott: be nicer to keep it as `conditions`

bigchaindb/core.py
```diff
@@ -343,6 +344,21 @@ def get_txs_by_asset_id(self, asset_id):
...
+                asset_id (str): The asset id
+
+            Returns:
+                :class:`~bigchaindb.common.transaction.Asset` if the asset
+                exists else None
+        """
+        cursor = self.backend.get_asset_by_id(asset_id)
+        cursor = list(cursor)
+        if cursor:
+            return Asset.from_dict(cursor[0]['transaction']['asset'])
```
r-marques: `self.backend.get_asset_by_id` actually returns a rethinkdb cursor. Just following what the rest of `backends.rethinkdb` methods do. `backends` calls they all return raw db documents and then the methods in `core` transform those documents in model instances

bigchaindb/models.py
```diff
@@ -101,11 +73,28 @@ def validate(self, bigchain):
...
+
+            # get the asset creation to see if its divisible or not
+            asset = bigchain.get_asset_by_id(asset_id)
+            # validate the asset
+            asset._validate_asset(amount=input_amount)
+            # validate the amounts
+            output_amount = sum([condition.amount for
+                                 condition in self.conditions])
+            if output_amount != input_amount:
+                raise AmountError(('The amout used in the inputs `{}`'
```
ttmc: There's a minor typo here: `The amout used` should be `The amount used`.

Maybe a more helpful message would be something like 'The sum of the input amounts (14) differs from the sum of the output amounts (23)'

ttmc: The variables `input_amount` and `output_amount` are sums but their names don't convey that. On first inspection, one might think `input_amount` refers to one of the input amounts, for example. Maybe use `input_amounts_sum` and `output_amounts_sum` or similar?

tests/assets/test_divisible_assets.py
r-marques: nice :+1: didn't noticed that

tests/assets/test_divisible_assets.py
r-marques: I think the problem comes from https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L969

This only happens when you create a threshold condition with the same public key which makes no sense in the first place. Example if owners_after = `[user_pub, user_pub, user_pub]` you end up with 3 subconditions for the same public key but when you do `subffill = ccffill.get_subcondition_from_vk(owner_before)[0]` you end up always fulfilling the first subcondition

tests/assets/test_divisible_assets.py
r-marques: And the best part is that the only check you need when validating is `output_amount == input_amount`

tests/common/test_transaction.py
r-marques: This was testing the case where you enforced the number of fulfillments to be the same as the number of conditions on a transaction. That is no longer true.

The only case where this raises a `TypeError` is when the number of fulfillments is different then the number of `input_condition_uris` which is tested with `transfer_tx.fulfillments_valid([])`

tests/common/test_transaction.py
r-marques: The transfer transaction still has 2 inputs and two outputs (the only difference is that the two inputs come from 2 conditions of the same transaction)

tests/db/test_bigchain_api.py
r-marques: Yes, I handle all those cases in `test_divisible_assets.py`.
The only cases I didn't handle was the ones with `get_owned_ids` and `get_spent`

tests/db/test_bigchain_api.py
r-marques: yes

tests/db/test_bigchain_api.py
r-marques: This seemed to be testing partial ownership of outputs. I don't think that the simplification changed that.
Let me know if you think otherwise

bigchaindb/core.py
vrde: Exactly, #788 will address this. The idea is that the DB layer handles raw documents. Once the document is loaded in the `core.py::Bigchain` class it will be imported in a model.

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    block = b.create_block([tx_create_signed])
+    assert block.validate(b) == block
+    b.write_block(block, durability='hard')
+    # vote
+    vote = b.vote(block.id, b.get_last_voted_block().id, True)
+    b.write_vote(vote)
+
+    # TRANSFER
+    tx_transfer = Transaction.transfer(tx_create.to_inputs(),
+                                       [([b.me], 50), ([b.me], 50)],
```
vrde: I don't understand this transfer. You are taking the first CREATE `tx_create` and you are transferring 50 shares to the node and other 50 shares to the same node? Is it a real use case or is just for testing?

tests/assets/test_divisible_assets.py
```diff
@@ -0,0 +1,686 @@
...
+    block = b.create_block([tx_create_signed])
+    assert block.validate(b) == block
+    b.write_block(block, durability='hard')
+    # vote
+    vote = b.vote(block.id, b.get_last_voted_block().id, True)
+    b.write_vote(vote)
+
+    # TRANSFER
+    tx_transfer = Transaction.transfer(tx_create.to_inputs(),
+                                       [([b.me, b.me], 100)],
```
vrde: Does this line transfer all the shares to a threshold condition that is satisfied only when signed by `b.me` and `b.me`?

bigchaindb/common/transaction.py
```diff
@@ -612,17 +665,27 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
...
-               secret=None, time_expire=None):
+    def create(cls, owners_before, owners_after, metadata=None, asset=None):
         """A simple way to generate a `CREATE` transaction.
 
             Note:
                 This method currently supports the following Cryptoconditions
                 use cases:
                     - Ed25519
                     - ThresholdSha256
-                    - PreimageSha256.
```
r-marques: The problem with hashlocks is that its difficult to provide a generic method to create and validate them. They don't really go well with the `create, sign, validate` api.

This type of custom cryptoconditions should be added later on the driver side

tests/assets/test_divisible_assets.py
r-marques: Its just a test for multiple outputs. You are taking a token with amount 100 and dividing it into two tokens with amount 50 (one input, two outputs). The actual public keys don't really matter. It would work the same for any public key

tests/assets/test_divisible_assets.py
r-marques: That is the objective of the test.
On more practical terms I know it does not make sense to create a threshold condition with the same public key

ttmc:
Is there a size limit on the `amount`? Should there be?

I don't think integer overflow is a problem if all the arithmetic is done in Python (with no calls to C stuff). Is that right?

The only other problem I can think of is making a transaction so big that it causes problems, but that's possible by stuffing other fields too (e.g. metadata data). Should we create a new issue to put an upper bound on overall transaction size?

r-marques:
I guess at some point that can be a problem but I don't know what the limit should be

ttmc:
I'll make a new issue for that so we don't forget.

:+1:

TimDaub:
> I guess at some point that can be a problem but I don't know what the limit should be

http://stackoverflow.com/a/3212065/1263876

r-marques:
Thanks @TimDaub 
So the answer is that we should limit the number of style sheets to 31 per page

bigchaindb/common/transaction.py
```diff
@@ -275,34 +283,22 @@ def generate(cls, owners_after):
...
-        # TODO: We probably want to remove the tuple logic for weights here
-        # again:
-        # github.com/bigchaindb/bigchaindb/issues/730#issuecomment-255144756
-        if isinstance(owners_after, tuple):
-            owners_after, threshold = owners_after
-        else:
-            threshold = len(owners_after)
-
+        threshold = len(owners_after)
+        if not isinstance(amount, int):
```
vrde: Should we add a check like:

```python
if amount < 0:
    raise ValueError('`amount` must be a natural number')
```

r-marques: Good catch

bigchaindb/models.py
```diff
@@ -71,14 +41,16 @@ def validate(self, bigchain):
             if inputs_defined:
                 raise ValueError('A CREATE operation has no inputs')
             # validate asset
-            self.asset._validate_asset()
+            amount = sum([condition.amount for condition in self.conditions])
```
sbellem: repeating earlier https://github.com/bigchaindb/bigchaindb/pull/794#discussion_r87822113:

@r-marques pointed out that although it may be more memory efficient, it would be slower. Small tests show that to be true:

```python
>>> timeit.timeit('sum(n for n in range(10000))', number=10000)
6.362047109985724

>>> timeit.timeit('sum([n for n in range(10000)])', number=10000)
4.701008878997527

>>> timeit.timeit('sum((n for n in range(10000)))', number=10000)
6.287681774992961
```

PR #798, opened 2016-11-09
Schema definition

### LINKED ISSUE #693: Node should check on validity of UUID's schema
See: [How to check if a UUIDv4 is compliant to it's schema](https://gist.github.com/ShawnMilo/7777304).

I think this can be done in `Transaction.validate` or `Asset.validate`.

//cc @r-marques

### LINKED ISSUE #720: Add schema validation and sanity checks to models
_From @sohkai on September 30, 2016 11:38_

As of now, we do very little to ensure that the given inputs are what we expect them to be when constructing the models. We **really** should be doing at least some checks with json schema or any other validation approach; not doing so results in the **worst** and most **nefarious** types of bugs for users.

Earlier, I created a transfer transaction and accidentally passed through a keypair dict as part of the `owners_after` argument in the [driver](https://github.com/sbellem/bigchaindb-driver/) (rather than just a string verifying key). The driver was able to create, sign, and send this wrongly-formatted condition all the way to BigchainDB, who accepted it and wrote it onto RethinkDB (see below). Now my public/private keypair is forever visible to the world, and this condition is probably impossible to fulfill.

If I was a naive user and assumed everything had worked correctly, since nothing had told me otherwise, I might now start creating `n` transactions this way without ever realizing that I've completely shot myself in the foot. Oops.

What really should've happened is me getting an error immediately upon the `owners_after` signature being wrong. Crisis averted.

The persisted transaction, available in a block in RethinkDB:

```
{'id': '960a684bb92de50384ebf568b7e1e78d88f8afbb674872b3d25356eae77fba51',
 'transaction': {'conditions': [{'cid': 0,
                                 'condition': {'uri': {'signing_key': '4S1dzx3PSdMAfs59aBkQefPASizTs728HnhLNpYZWCad',
                                                       'verifying_key': '2dBVUoATxEzEqRdsi64AFsJnn2ywLCwnbNwW7K9BuVuS'}},
                                 'owners_after': [{'signing_key': '4S1dzx3PSdMAfs59aBkQefPASizTs728HnhLNpYZWCad',
                                                   'verifying_key': '2dBVUoATxEzEqRdsi64AFsJnn2ywLCwnbNwW7K9BuVuS'}]}],
                 'data': None,
                 'fulfillments': [{'fid': 0,
                                   'fulfillment': 'cf:4:4HwjqBgNkDK0fD1ajmFn0OZ75N3Jk-xIV2zlhgPxP2a7KeSZy56l5uU_Wqj6FmQz9O8xUYS3eIdR0RGDAobM4grfbtmeWbv4FXwtd5FyYbh6Azs4sle_eYFOKwsGREcL',
                                   'input': {'cid': 0,
                                             'txid': 'fad5ec32672b75cccf92992cc977f456bd587a7d49956fed67a227fb26d13250'},
                                   'owners_before': ['G7J7bXF8cqSrjrxUKwcF8tCriEKC5CgyPHmtGwUi4BK3']}],
                 'operation': 'TRANSFER',
                 'timestamp': '1475228620'},
 'version': 1}
```

**Notice the condition, which lists the entire given dict**:

```
{
    ...
    'conditions': [{'cid': 0,
                    'condition': {'uri': {'signing_key': '4S1dzx3PSdMAfs59aBkQefPASizTs728HnhLNpYZWCad',
                                          'verifying_key': '2dBVUoATxEzEqRdsi64AFsJnn2ywLCwnbNwW7K9BuVuS'}},
                    'owners_after': [{'signing_key': '4S1dzx3PSdMAfs59aBkQefPASizTs728HnhLNpYZWCad',
                                      'verifying_key': '2dBVUoATxEzEqRdsi64AFsJnn2ywLCwnbNwW7K9BuVuS'}]}],
    ...
```

(In fact, I probably could've put anything there; it just so happened that it was the keypair.)

---

At the very least, as suggested by @sbellem, it should be made very clear (with big warning signs and all that) that whatever you give that is turned into the `owners_after` is forever written in the transaction's condition. We could also add some sort of identity verification system, but that's out of scope.

cc @TimDaub @diminator 

_Copied from original issue: bigchaindb/bigchaindb-common#31_

libscott:
Fixes #693, fixes #720.

**Note**: transactions are not yet actually validated, but the functionality to do so is available.

- [x] Canonical schema defined in `bigchaindb.common` module
- [x] Sphinx documentation generated from schema
- [x] Reflects current data structure completely
- [x] All fields documented and explained
- [x] Is used to actually validate Transactions on deserialization
- [x] Has reasonable (non show stopping / sub ~100ms) performance
- [x] Validates that objects do not have extra keys where appropriate
- [x] Add a step in our `Release_Process.md` file: generate documentation from json schema
- [x] Address all concerns in https://github.com/bigchaindb/bigchaindb/issues/720 (all relevant concerns addressed).
- [x] Check what happens when passing invalid transactions via HTTP API
- [x] Refactor validation exceptions into ValidationError, InvalidHash, and InvalidTransactionStructure

screenshot from 2016-11-16 14-18-15


bigchaindb/common/schema.py
```diff
@@ -0,0 +1,166 @@
...
+readability and to make documenting easier.
+"""
+
+
+def nullable(data):
+    """ Describe optional data """
+    return {"anyOf": [data, {"type": "null"}]}
+
+
+class SchemaObject(object):
```
vrde: You can drop `(object)` since in Py3 all classes are "new style" classes by default. The examples in the [official documentation](https://docs.python.org/3/tutorial/classes.html#class-definition-syntax) don't inherit from `object`.

libscott:
Currently leaning towards YaML for the schema definition:

* More readable than JSON
* Can write multiline strings, neccesary for documentation
* Can decode to object suitable for use with JSON Schema

bigchaindb/common/transaction.py
```diff
@@ -1194,6 +1197,13 @@ def from_dict(cls, tx_body):
             Returns:
                 :class:`~bigchaindb.common.transaction.Transaction`
         """
+
+        # Attempt full validation of tx body, fail quickly if possible
+        try:
+            jsonschema.validate(tx_body, TX_JSON_SCHEMA)
+        except jsonschema.ValidationError as e:
+            raise ValidationError(str(e))
```
libscott: Yep, agreed. Question though, should verification of transaction ID remain inside `from_dict`?

libscott:
It appears to take about 4ms to validate a transaction using `jsonschema`. For comparison, the `Transaction.from_dict` method takes 0.3ms, without schema validation.

sohkai:
> It appears to take about 4ms to validate a transaction using jsonschema. For comparison, the Transaction.from_dict method takes 0.3ms, without schema validation.

Ouch :(. At most ~250 tx can be validated in a second now, on our puny machines. Might have to find a faster solution for the voting pipelines when we're pressed for performance.

Release_Process.md
```diff
@@ -2,18 +2,19 @@
 
 This is a summary of the steps we go through to release a new version of BigchainDB Server.
 
-1. Update the `CHANGELOG.md` file
-2. Update the version numbers in `bigchaindb/version.py`. Note that we try to use [semantic versioning](http://semver.org/) (i.e. MAJOR.MINOR.PATCH)
-3. Go to the [bigchaindb/bigchaindb Releases page on GitHub](https://github.com/bigchaindb/bigchaindb/releases)
+1. Run `python docs/server/generate_schema_documentation.py` and commit the changes to docs/server/sources/schema.rst, if any.
+2. Update the `CHANGELOG.md` file
```
sohkai: We can use autonumbering with markdown, so each of these items can just be `1.`. Makes it much easier to diff.

bigchaindb/common/exceptions.py
```diff
@@ -22,7 +22,11 @@ class DoubleSpend(Exception):
     """Raised if a double spend is found"""
 
 
-class InvalidHash(Exception):
+class ValidationError(Exception):
+    """Raised if any error with object schema"""
```
sohkai: Would rephrase to `Raised if there was any error validating an object's schema`

bigchaindb/common/schema/transaction.json
```diff
@@ -0,0 +1 @@
..., "conditions": {"items": {"additionalProperties": false, "required": ["owners_after", "condition", "amount"], "type": "object", "properties": {"cid": {"type": "integer"}, "amount": {"type": "integer"}, "owners_after": {"$ref": "#/definitions/owners_list"}, "condition": {"additionalProperties": false, "required": ["uri"], "type": "object", "properties": {"uri": {"type": "string"}, "details": {"additionalProperties": true, "type": "object"}}}}}, "type": "array"}, "timestamp": {"type": "string"}, "metadata": {"anyOf": [{"additionalProperties": false, "required": ["id", "data"], "type": "object", "properties": {"data": {"additionalProperties": true, "type": "object"}, "id": {"$ref": "#/definitions/uuid4"}}}, {"type": "null"}]}, "asset": {"additionalProperties": false, "required": ["id"], "type": "object", "properties": {"updatable": {"type": "boolean"}, "id": {"$ref": "#/definitions/uuid4"}, "refillable": {"type": "boolean"}, "data": {"anyOf": [{"additionalProperties": true, "type": "object"}, {"type": "null"}]}, "divisible": {"type": "boolean"}}}, "operation": {"enum": ["CREATE", "TRANSFER", "GENESIS"], "type": "string"}}}, "version": {"type": "integer", "maximum": 1, "minimum": 1}}}
```
sohkai: I don't think this file is necessary?

bigchaindb/common/schema/__init__.py
```diff
@@ -0,0 +1,23 @@
+from collections import OrderedDict
+import os.path
+
+import yaml
+
+
+def ordered_load(stream):
```
sohkai: Is there a reason we need to do this rather than a direct `yaml.safe_load()`?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,245 @@
+---
```
sohkai: Needs the `id` and `$schema` properties from https://github.com/bigchaindb/bigchaindb/pull/311/files#diff-1f1becfa0fbd3b2647c0b02c4fe99ebeR2.

bigchaindb/common/exceptions.py
ttmc: I would also rename this exception to something more precise, e.g. `TxSchemaValidationError`. (There are many ways an error might arise when validating a transaction; an error in validating its JSON structure against the JSON schema is only a subset.)

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,245 @@
...
+type: object
+additionalProperties: false
+title: Transaction Schema
+description: |
+    This is the outer transaction wrapper. It contains the ID, version and the body of the transaction, which is also called `transaction`.
+properties:
+  id:
+    pattern: "[0-9a-f]{64}"
```
sohkai: I think this can be replaced by the `sha3_hexdigest` definition below, along with moving the description to down there?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,245 @@
+---
+type: object
+additionalProperties: false
```
sohkai: Should we put a `required` list in the wrapper too?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,245 @@
...
+        "$ref": "#/definitions/base58"
+    - type: 'null'
+  sha3_hexdigest:
+    pattern: "[0-9a-f]{64}"
+    type: string
+  uuid4:
+    pattern: "[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}"
+    type: string
+    description: |
+        A `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_
```
sohkai: Might be better to reference the RFC: https://tools.ietf.org/html/rfc4122.html.

bigchaindb/common/exceptions.py
libscott: Ok, good ideas both.

bigchaindb/common/schema/transaction.json
libscott: Right! Thanks...

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,245 @@
...
+definitions:
+  base58:
+    pattern: "[1-9a-zA-Z^OIl]{43,44}"
+    type: string
+  owners_list:
+    anyOf:
+    - type: array
+      items:
+        "$ref": "#/definitions/base58"
+    - type: 'null'
```
libscott: It does, `null` translates to None in Python, wheras the JSON Schema spec uses `'null'`.

libscott:
> Ouch :(. At most ~250 tx can be validated in a second now, on our puny machines. Might have to find a faster solution for the voting pipelines when we're pressed for performance.

Yep. There are some libraries that do python code generation. Either way we will probably have to get our hands dirty.

libscott:
By the way, I have a nifty pre-commit hook that automatically generates the schema when the relavent inputs change. It goes in `.git/hooks/pre-commit`. I can't commit it because git doesn't allow versioning files inside the `.git` folder (though it could be achieved with symlinks).

```shell
#!/bin/bash

# Function that updates the schema documentation if any of the dependencies
# of the schema documentation change.
if [[ $(git diff --cached --name-status bigchaindb/common/schema/transaction.yaml docs/server/generate_schema_documentation.py) ]]; then
    echo -n "Updating schema documentation... "
    python docs/server/generate_schema_documentation.py
    git add docs/server/source/schema.rst
    echo "All done."
fi
```

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+    type: object
+    description: |
+        An output of a transaction. A condition describes a quantity of an asset
+        and what conditions must be met in order for it to be fulfilled. See also:
+        fulfillment_.
+    additionalProperties: false
+    required:
+    - owners_after
+    - condition
+    - amount
```
sohkai: I think `cid` is also always required?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+      condition:
+        description: |
+            Body of the condition. Has the properties:
+
+            - **details**: Details of the condition.
+            - **uri**: Condition encoded as an ASCII string.
+        type: object
+        additionalProperties: false
+        required:
+        - uri
```
sohkai: I think the `details` is also always required (at least for now).

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+        properties:
+          details:
+            type: object
+            additionalProperties: true
+          uri:
+            type: string
```
sohkai: We should include the pattern: https://github.com/interledgerjs/five-bells-condition/blob/master/schemas/Condition.json#L6

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+    type: "object"
+    description:
+      A fulfillment is an input to a transaction, named as such because it
+      fulfills a condition of a previous transaction. In the case of a
+      **CREATE** transaction, a fulfillment may provide no `input`.
+    additionalProperties: false
+    required:
+    - owners_before
+    - input
+    - fulfillment
```
sohkai: I think `fid` is always required as well.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+        type: integer
+        description: |
+            The offset of the fulfillment within the fulfillents array.
+      owners_before:
+        "$ref": "#/definitions/owners_list"
+        description: |
+            List of public keys of the previous owners of the asset.
+      fulfillment:
+        anyOf:
+        - type: object
```
sohkai: Mmm, we can keep the object version of `fulfillment` here, but, depending on when we expect to do validation, this might not be necessary. The object version should only be used when fulfillments haven't been signed yet, so this is strictly something only the driver would want.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+            type:
+              type: string
+            signature:
+              anyOf:
+              - type: string
+              - type: 'null'
+            type_id:
+              type: integer
+          description: |
+              Fulfillment of a ondition_, or put a different way, this is a
```
sohkai: **condition_**

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+        anyOf:
+        - type: 'object'
+          description: |
+            Reference to a condition of a previous transaction
+          additionalProperties: false
+          properties:
+            cid:
+              type: 'integer'
+            txid:
+              type: 'string'
```
sohkai: I think this can be validated with the `sha3_hexdigest` pattern?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+        and what conditions must be met in order for it to be fulfilled. See also:
+        fulfillment_.
+    additionalProperties: false
+    required:
+    - owners_after
+    - condition
+    - amount
+    properties:
+      cid:
+        type: integer
```
sohkai: All `cid`s and `fid`s should also be validated to be minimum 0.

bigchaindb/common/schema/transaction.yaml
libscott: It's actually not: https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L272

bigchaindb/common/schema/transaction.yaml
libscott: It's not: https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L257

bigchaindb/common/transaction.py
```diff
@@ -1248,19 +1251,15 @@ def __str__(self):
...
-        """Transforms a Python dictionary to a Transaction object.
+    @staticmethod
+    def validate_structure(tx_body_orig):
+        """Validate the structure of a transaction body.
 
             Args:
                 tx_body (dict): The Transaction to be transformed.
-
-            Returns:
-                :class:`~bigchaindb.common.transaction.Transaction`
```
sohkai: Should document the raises types

bigchaindb/common/transaction.py
```diff
@@ -1272,17 +1271,34 @@ def from_dict(cls, tx_body):
 
         if proposed_tx_id != valid_tx_id:
             raise InvalidHash()
+
+        try:
```
sohkai: Might be worth putting this in the common.schema model, maybe as `validate_transaction_schema`?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,248 @@
...
+              anyOf:
+              - type: string
+              - type: 'null'
+            type_id:
+              type: integer
+          description: |
+              Fulfillment of a ondition_, or put a different way, this is a
+              payload that satisfies a condition in order to spend the associated
+              asset.
+        - type: string
```
libscott: I assume you mean for the `signature` field?

bigchaindb/common/schema/transaction.yaml
libscott: :+1:

bigchaindb/common/schema/transaction.yaml
libscott: :+1:

docs/server/generate_schema_documentation.py
```diff
@@ -0,0 +1,156 @@
...
+        return ' or '.join(property_type(p) for p in prop['anyOf'])
+    if '$ref' in prop:
+        return property_type(resolve_ref(prop['$ref']))
+    raise ValueError("Could not resolve property type")
+
+
+def resolve_ref(ref):
+    """ Resolve reference """
+    assert ref.startswith('#/definitions/')
+    return TX['definitions'][ref[14:]]
```
sohkai: Probably better to use `len('#/definitions/')` than `14`

bigchaindb/common/transaction.py
libscott: :+1:

docs/server/generate_schema_documentation.py
```diff
@@ -0,0 +1,156 @@
...
+%(title)s
+%(underline)s
+
+**type:** %(type)s
+
+%(description)s
+"""
+
+
+TPL_DOC = """\
```
sohkai: Should probably put a rst comment at the start of this document to let others know that it's deliberately auto generated from the schemas.

tests/common/test_schema.py
```diff
@@ -0,0 +1,34 @@
...
+
+def test_validate_transaction_signed_create(signed_create_tx):
+    jsonschema.validate(signed_create_tx.to_dict(), TX_SCHEMA)
+
+
+def test_validate_transaction_signed_transfer(signed_transfer_tx):
+    jsonschema.validate(signed_transfer_tx.to_dict(), TX_SCHEMA)
+
+
+def test_addition_properties_false():
```
sohkai: addition**al**

tests/common/test_transaction.py
```diff
@@ -1062,3 +1091,9 @@ def test_cant_add_empty_fulfillment():
         tx = Transaction(Transaction.CREATE, None)
     with raises(TypeError):
         tx.add_fulfillment(None)
+
+
+def validate(tx):
```
sohkai: This should probably be in a common.tests.util module, as `validate_transaction`

libscott:
> MANIFEST.in (I think so, for it to be included in the packaging? @vrde might know better here)
> setup.py (again, @vrde might know better here for include_package_data vs package_data)

I added the yaml file to `package_data` in setup.py, tested install and it worked.

libscott:
> - schema/README.md: might as well keep it

Feel free :)

bigchaindb/common/schema/transaction.yaml
libscott: Right now it's neccesary, tests fail without it.

bigchaindb/common/transaction.py
libscott: That should go for the whole file then, indeed the whole codebase. I'd put it in a separate issue.

docs/server/generate_schema_documentation.py
libscott: > schema/README.md: might as well keep it

Explain?

docs/server/generate_schema_documentation.py
libscott: :+1:

bigchaindb/common/transaction.py
sohkai: Actually that's a fair point

brother-cary-elwes-gnJgBlPgHtcnS

libscott:
@ttmc @sohkai it's possible to serialize a `Conditon` without a `cid`, do we agree that a transaction body containing fulfillments is invalid? This case comes up sometimes in the tests, I agree with @sohkai it should be considered invalid. Same with `Fulfillment.fid`. But I'm not so sure about `Condition.details`.

ttmc:
@libscott I'm currently uncertain about the contents and meanings of all the things inside conditions and fulfillments. Tim and Sylvain are good people to ask about those.

TimDaub:
Not 100% sure but, `fid` and `cid` are technically unnecessary as all functions in consensus take the actual index of the lists `fulfillments` and `conditions` as indexes and do not search in the lists for an object with a specific `fid` or `cid`. (Again, not 100% sure)

bigchaindb/common/schema/__init__.py
sohkai: Given that the `OrderedDict` is only used for rendering the .rst, I'd move it to `generate_schema_documentation.py`.

bigchaindb/common/schema/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        loader.flatten_mapping(node)
+        return OrderedDict(loader.construct_pairs(node))
+    OrderedLoader.add_constructor(
+        yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
+        construct_mapping)
+    return yaml.load(stream, OrderedLoader)
+
+
+TX_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), 'transaction.yaml')
+TX_SCHEMA_DATA = open(TX_SCHEMA_PATH).read()
```
sohkai: This should be using `with` in case of exceptions or etc.

bigchaindb/common/schema/transaction.yaml
TimDaub: I have no memory why providing `details` is optional in `Condition`, but it shouldn't.
I'll create a ticket. Make sure that it's required for now in the schema.

TimDaub: I remember: For hash lock serialization, this was included: https://github.com/bigchaindb/bigchaindb/commit/66f7ab6ea4f398a4e05dfcf3932359eec02c6926

It's now legacy, meaning that `details` are mandatory.

TimDaub: Ticket created: #836

ttmc:
Some comments about the new auto-generated docs page (`schema.rst`):

* In `docs/server/source/index.rst`, could you move `schema` right after `topic-guides/index`? This is because **Topic Guides** used to contain the data models docs and **Appendices** should go last. (The **Topic Guides** section will be going away; I kept it there for a while to help people find where what was in there went.)

* We also have a block schema and a vote schema (although we might not be so formal about those), so could you rename `schema.rst` to `transaction-schema.rst`? Maybe do a `grep` to find all mentions of `schema.rst` (e.g. in `Release_Process.md`).

* When I did the `make html` to get Sphinx to generate the docs, I got a new warning message:
```text
/home/troy/repos/bigchaindb/docs/server/source/schema.rst:165: WARNING: Inline interpreted text or phrase reference start-string without end-string.
```
The offending line (number 165) is:
```text
Index of this transaction's appearance in the `transaction.conditions`_
```
I guess the single quote in `transaction's` is causing the problem? Sphinx is weird sometimes.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,256 @@
...
+    description: |
+        A sha3 digest of the transaction. The ID is calculated by removing all
+        derived hashes and signatures from the transaction, serializing it to
+        JSON with keys in sorted order and then hashing the resulting string
+        with sha3.
+  transaction:
+    type: object
+    title: transaction
+    description: |
+        Body of the transaction.
```
ttmc: Below, you call this thing the "Transaction Body" so for consistency it would be nice to use the same terminology here. Is it possible to hyperlink from here to that section below?

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,256 @@
...
+      metadata:
+        "$ref": "#/definitions/metadata"
+      timestamp:
+        "$ref": "#/definitions/timestamp"
+  version:
+    type: integer
+    minimum: 1
+    maximum: 1
+    description: |
+        Transaction version identifier.
```
ttmc: This is the version number of the BigchainDB transaction _schema_. (One might think, incorrectly, that a particular transaction might be updated by a subsequent transaction with a higher version number.)

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,256 @@
...
+    type: string
+    description: |
+        A `UUID <https://tools.ietf.org/html/rfc4122.html>`_
+        of type 4 (random).
+  operation:
+    type: string
+    description: |
+        Type of the transaction:
+
+        A `CREATE` transaction creates an asset in BigchainDB. This
```
ttmc: Sphinx is rendering these strings inside single backticks as italicized strings. For literal strings like "CREATE", maybe use double quotes around the string instead? An alternative is double backticks.

bigchaindb/common/schema/__init__.py
sohkai: Also, maybe we should use an initialization function for initializing the schemas before we start all the processes.

This lets us preload the schema into every process, as well as throw an error and exit BigchainDB if the schemas aren't found.

bigchaindb/common/schema/transaction.yaml
libscott: Very nice.

screenshot from 2016-11-17 18-30-08

libscott:
> I'd suggest:
> 
> Make it mandatory for now
> Create a ticket to remove them from fulfillments and conditions (should be quite easy)

@TimDaub: Seems to be some different opinions on this one, anyhow I'll move it to another issue.

bigchaindb/common/schema/transaction.yaml
libscott: :+1:

bigchaindb/common/schema/transaction.yaml
libscott: :+1:

bigchaindb/common/schema/transaction.yaml
libscott: But I agree it's not ideal. Another one for the TX milestone?

bigchaindb/common/schema/transaction.yaml
libscott: Done

bigchaindb/common/schema/transaction.yaml
```diff
@@ -0,0 +1,257 @@
...
+          contain any valid JSON payload provided by the user.
+      additionalProperties: false
+      required:
+      - id
+      - data
+      properties:
+        data:
+          type: object
+          description: |
+            Freeform object containing user provided asset metadata.
```
ttmc: The metadata data is for this specific transaction. It might not say anything about the asset being transferred. For example, it might say, "There was a thunderstorm in Sydney when Jake Hoopman constructed this transaction."

Maybe just borrow language from the description of the metadata object above.

libscott: Hm, how did "asset" get in there. Right you are!

libscott:
@sbellem @sohkai can you review the changes and dismiss reviews if appropriate?

sohkai:
@libscott The one last thing for me, I think, is to use `with` to make sure the file is closed here: https://github.com/bigchaindb/bigchaindb/pull/798/files#diff-90055ffda3fc3dc83838121300897616R11.

libscott:
> @libscott The one last thing for me, I think, is to use with to make sure the file is closed here: https://github.com/bigchaindb/bigchaindb/pull/798/files#diff-90055ffda3fc3dc83838121300897616R11.

Done.

> 
> (And maybe address this: #834 (comment))

I made it a separate issue so we could discuss and implement it separately, is there something else that needs to be changed here?

tests/common/test_transaction.py
```diff
@@ -356,20 +369,22 @@ def test_transaction_deserialization(user_ffill, user_cond, data, data_id):
...
 
-def test_tx_serialization_with_incorrect_hash(utx):
+def test_tx_serialization_with_incorrect_hash(utx, ):
     from bigchaindb.common.transaction import Transaction
-    from bigchaindb.common.exceptions import InvalidHash
+    from bigchaindb.common.exceptions import InvalidHash, SchemaValidationError
 
     utx_dict = utx.to_dict()
-    utx_dict['id'] = 'abc'
+    utx_dict['id'] = ('abc' * 30)[:64]
```
libscott: You got it.

bigchaindb/web/views/transactions.py
```diff
@@ -98,7 +98,7 @@ def post(self):
 
         try:
             tx_obj = Transaction.from_dict(tx)
-        except (InvalidHash, InvalidSignature):
+        except (ValidationError, InvalidSignature):
             return make_error(400, 'Invalid transaction')
```
libscott: Let's make this a separate issue, I'm not so sure yet.

bigchaindb/common/transaction.py
```diff
@@ -1248,17 +1248,14 @@ def __str__(self):
         tx = Transaction._remove_signatures(self.to_dict())
         return Transaction._to_str(tx)
 
-    @classmethod
-    # TODO: Make this method more pretty
-    def from_dict(cls, tx_body):
-        """Transforms a Python dictionary to a Transaction object.
+    @staticmethod
+    def validate_structure(tx_body):
```
libscott: `validate_structure` as it is... It is validating against the schema and also checking the ID.

bigchaindb/web/views/transactions.py
libscott: To expand on this, the exception data is huge and messy and im not sure it's appropriate to pass it to a client as-is, maybe it's not the highest priority to solve it right now.

libscott:
> Let's keep the auto-validate on from_dict scoped to only the server's models for now, until we know for sure that the driver would also like this behaviour

https://github.com/bigchaindb/bigchaindb/pull/798/commits/c1db722793a85315a9ee5b108a1fbd7f415f98c3

Discussion of this issue is moved to https://github.com/bigchaindb/bigchaindb/issues/834 and is no longer the scope of this PR.

> Let's setup an init function to load these schemas (and perhaps gracefully exit the program if loading fails) before any other processes are spun off in processes.py so that each process doesn't have to load them themselves.

This seems a bit premature and I would prioritise other things but if you'd like to add it I won't have objections

bigchaindb/common/schema/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+import os.path
+
+import jsonschema
+import yaml
+
+from bigchaindb.common.exceptions import SchemaValidationError
+
+
+TX_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), 'transaction.yaml')
+TX_SCHEMA_YAML = open(TX_SCHEMA_PATH).read()
```
sbellem: Is there a reason for not using `with`? E.g.:

```python
with open(TX_SCHEMA_PATH) as f:
    TX_SCHEMA_YAML = f.read()
```

sbellem:
@libscott could you please clean up the history, and use meaningful commit messages?

PR #802, opened 2016-11-10
Format http api docs to 79 chars

TimDaub:
@ttmc I hope this change is OK.

If not, I guess I can also work with more chars per line, but in vim that's kinda tricky.

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -3,21 +3,35 @@ The HTTP Client-Server API
 
 .. note::
 
-   The HTTP client-server API is currently quite rudimentary. For example, there is no ability to do complex queries using the HTTP API. We plan to add querying capabilities in the future.
+   The HTTP client-server API is currently quite rudimentary. For example,
+   there is no ability to do complex queries using the HTTP API. We plan to add
+   querying capabilities in the future.
 
-When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at the address stored in the BigchainDB node configuration settings. The default is:
+When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at
+the address stored in the BigchainDB node configuration settings. The default
+is:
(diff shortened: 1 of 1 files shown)


docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -3,21 +3,35 @@ The HTTP Client-Server API
...
+   The HTTP client-server API is currently quite rudimentary. For example,
+   there is no ability to do complex queries using the HTTP API. We plan to add
+   querying capabilities in the future.
 
-When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at the address stored in the BigchainDB node configuration settings. The default is:
+When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at
+the address stored in the BigchainDB node configuration settings. The default
+is:
 
-`http://localhost:9984/api/v1/ <http://localhost:9984/api/v1/>`_
```
ttmc: The underscore `_` at the end is a RST thing to make a hyperlink.

TimDaub: :laughing: What a bunch of :trollface:! I'll readd it.

ttmc:
I didn't realize that RST files with lines over 79 characters were problematic for vim. I don't have any objections so long as the final rendered docs still look okay.

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: Readded it.

ttmc: :confused: When I check the latest commit or the latest version of this branch, it seems to still have the underscore removed

TimDaub: Yeah, I added a whitespace instead of an underscore. Not sure what I was thinking :laughing:.

Should be fine now.

ttmc:
:+1:

PR #765, opened 2016-10-29
Rename FulfillmentNotInValidBlock to TransactionNotInValidBlock

ttmc:
A minor pull request that renames the exception FulfillmentNotInValidBlock to TransactionNotInValidBlock.

The new name is in line with the name of the exception TransactionDoesNotExist (which serves a similar purpose).

I also added TransactionNotInValidBlock to the list of exceptions that `Transaction.validate()` can raise (i.e. in that method's docstring).

PR #803, opened 2016-11-10
Docs/601/get owned ids

TimDaub:
This is ready for review.

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -11,7 +11,7 @@ When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at
 the address stored in the BigchainDB node configuration settings. The default
 is:
 
-`http://localhost:9984/api/v1/ <http://localhost:9984/api/v1/>` 
+`http://localhost:9984/api/v1/ <http://localhost:9984/api/v1/>`_
 
 but that address can be changed by changing the "API endpoint" configuration
 setting (e.g. in a local config file). There's more information about setting
@@ -294,8 +294,8 @@ GET /unspents/
    a previous transaction and could hence be called unspent conditions/outputs
    (or simply: unspents).
 
-   This endpoint doesn't return anything if the querystring ``owner_after``
-   happens to not be defined in the request.
+   This endpoint will return a ``HTTP 400 Bad Request`` if the querystring
+   ``owner_after`` happens to not be defined in the request.
 
    Note that if unspents for a certain ``owner_after`` have not been found by
    the server, this will result in the server returning a 200 OK HTTP status
@@ -303,7 +303,7 @@ GET /unspents/
 
    :param owner_after: A public key, able to validly spend an output of a
    transaction, assuming the user also has the corresponding private key.
-   :type owner_after: hex string
+   :type owner_after: base58 encoded string
 
    **Example request**:
 
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -320,14 +320,9 @@ GET /unspents/
       Content-Type: application/json
 
       [
-        {
-          "txid": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
-          "cid": 0
-        },
-        {
-          "txid": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
-          "cid": 1
-        },
+
+        '../transactions/2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e/conditions/0',
+        '../transactions/2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e/conditions/1'
       ]
 
    :statuscode 200: A list of outputs were found and returned in the body of
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
 
    :statuscode 200: A transaction with that ID was found.
    :statuscode 404: A transaction with that ID was not found.
+
+
+GET /unspents/
+-------------------------
+
+.. http:get:: /unspents?owner_after={owner_after}
+
+   Get a list of links to transactions' conditions that have not been used in
+   a previous transaction and could hence be called unspent conditions/outputs
+   (or simply: unspents).
+
+   This endpoint doesn't return anything if the querystring ``owner_after``
+   happens to not be defined in the request.
+
+   Note that if unspents for a certain ``owner_after`` have not been found by
+   the server, this will result in the server returning a 200 OK HTTP status
+   code and an empty list in the response's body.
+
+   :param owner_after: A public key, able to validly spend an output of a
+   transaction, assuming the user also has the corresponding private key.
+   :type owner_after: hex string
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      GET /unspents?owner_after=1AAAbbb...ccc HTTP/1.1
+      Host: example.com
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      [
+        {
+          "txid": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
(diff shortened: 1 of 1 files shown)


docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
...
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      [
+        {
```
TimDaub: The decision to return a list of JSON `TransactionLink`s here was rationalized with the following argument:

When creating a `TRANSFER`, the creator is interested both in the transaction's id and the position of the condition (`cid`), but also in the actual condition to be fulfilled/spent.

`txid` and `cid` need to be included as `input` in `fulfillments[x].fulfillment`.
To create `fulfillments[x].fulfillment.fulfillment`, the original condition solves as the basis of the fulfillment.

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
 
    :statuscode 200: A transaction with that ID was found.
    :statuscode 404: A transaction with that ID was not found.
+
+
+GET /unspents/
```
TimDaub: Any objections to call the endpoint `unspents`?

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
...
+-------------------------
+
+.. http:get:: /unspents?owner_after={owner_after}
+
+   Get a list of links to transactions' conditions that have not been used in
+   a previous transaction and could hence be called unspent conditions/outputs
+   (or simply: unspents).
+
+   This endpoint doesn't return anything if the querystring ``owner_after``
+   happens to not be defined in the request.
```
TimDaub: Any objections to this?

docs/server/source/drivers-clients/http-client-server-api.rst
ttmc: It's consistent with out intention to rename the method `get_owned_ids()` to `get_unspents()`, as per issue #603

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -11,7 +11,7 @@ When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at
 the address stored in the BigchainDB node configuration settings. The default
 is:
 
-`http://localhost:9984/api/v1/ <http://localhost:9984/api/v1/>`_
+`http://localhost:9984/api/v1/ <http://localhost:9984/api/v1/>`
```
ttmc: Oops, need to put back the underscore (to make this a valid RST hyperlink).

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
...
+   Get a list of links to transactions' conditions that have not been used in
+   a previous transaction and could hence be called unspent conditions/outputs
+   (or simply: unspents).
+
+   This endpoint doesn't return anything if the querystring ``owner_after``
+   happens to not be defined in the request.
+
+   Note that if unspents for a certain ``owner_after`` have not been found by
+   the server, this will result in the server returning a 200 OK HTTP status
+   code and an empty list in the response's body.
```
ttmc: An alternative would be to return the HTTP 404 status code, i.e. 404 Not Found. From [the HTTP/1.1 spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), this means:

> The server has not found anything matching the Request-URI. No indication is given of whether the condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address. This status code is commonly used when the server does not wish to reveal exactly why the request has been refused, or when no other response is applicable.

docs/server/source/drivers-clients/http-client-server-api.rst
ttmc: There should be some response to tell the requester (user agent) that they did something wrong, i.e. their request was badly formed. The most general catch-all would be `HTTP 400 Bad Request`, but a more specific HTTP 40? might work too. Some discussion:

https://stackoverflow.com/questions/9454811/which-http-status-code-to-use-for-required-parameters-not-provided

https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
 
    :statuscode 200: A transaction with that ID was found.
    :statuscode 404: A transaction with that ID was not found.
+
+
+GET /unspents/
+-------------------------
+
+.. http:get:: /unspents?owner_after={owner_after}
```
r-marques: Since the only parameter here is `owners_after`, why use a query parameter at all.
Why not just do like the transactions endpoint `.. http:get:: /unspents/{owners_after}`

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -283,3 +283,54 @@ GET /transactions/{tx_id}
...
+   This endpoint doesn't return anything if the querystring ``owner_after``
+   happens to not be defined in the request.
+
+   Note that if unspents for a certain ``owner_after`` have not been found by
+   the server, this will result in the server returning a 200 OK HTTP status
+   code and an empty list in the response's body.
+
+   :param owner_after: A public key, able to validly spend an output of a
+   transaction, assuming the user also has the corresponding private key.
+   :type owner_after: hex string
```
r-marques: its actually a base58 encoded string

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: Since `owner_after` is just applying a filter on `/unspents` I think it would be correct to just return an empty list instead of a 404.

r-marques:
:+1:

ttmc:
:+1:

docs/server/source/drivers-clients/http-client-server-api.rst
sbellem: **tl;dr** We'll stick to `/unspents`

A small summary of an offline conversation between @TimDaub and @sbellem:

As @TimDaub pointed out the endpoint does not strictly return `conditions`, but `TransactionLinks`, which from the point of view of relational data modelling, could be seen as a kind of "intermediate table". That is, one could think of conditions, fulfilments and transactions living in separate tables, along with intermediate tables to keep track of their relationships.

In such scenarios, querying can be done in various ways: on transactions, on conditions, or on the intermediate table. 

In this case, `unspents` is the intermediate path. That is we could be querying on either `/transactions`, `/conditions` or the intermediate table resource, which in this case is most commonly known as `unspents`.

So, we'll stick to `/unspents`.

TimDaub:
From talking to @sbellem, we came up with a improvement.

### Don't return the `dict` version of `TransactionLink`

This is a HTTP API. An HTTP client hasn't any notion of a `TransactionLink`, however a HTTP client understands good old URIs. So instead of returning:

```
[{
  "txid": "abc",
  "cid": 0
}, ...]
```

We figured that it would make sense to return URIs like shown below:

```
[
  'http://ipdb.foundation/transactions/txid?conditions=cid', '...'
]
```

sbellem:
> We figured that it would make sense to return URIs like shown below:
> 
> ```bash
> [
>   'http://ipdb.foundation/transactions/txid?conditions=cid', '...'
> ]
> ```
>
> or
>
> ```bash 
> [
>   'http://ipdb.foundation/transactions/txid/conditions/cid', '...'
> ]
> ```
>
> Where in both cases only a single condition from the transaction with txid is being returned.
> I prefer the query string form.

A few thoughts:

1. Perhaps it is better not to return the node's url path, e.g.: `/transactions/txid/conditions/cid` This way, we do not make assumptions with respect to which node will be queried next.
2. I prefer the path `/transactions/txid/conditions/cid` because it seems to me more explicit with regards to what the resource is; in this case `conditions`. The query parameter form I find somewhat ambiguous for this use case as it seems to me that it could also return a full blown transaction payload, (having the specified `txid` and condition id (`cid`)).

ttmc:
This is a nice idea (returning a list of URIs rather than a list of dicts).

ttmc:
@TimDaub The Travis CI test is failing because of two minor flake8 errors that I fixed in PR #765 (now merged), so maybe if you merge origin/master those will go away.

PR #804, opened 2016-11-10
Moved data models docs back to server

ttmc:
As discussed this morning with @TimDaub @libscott and @sbellem , I moved the docs about the data models from the "root" BigchainDB docs back to the Server docs.

I also made a few edits to those docs.

Those docs will probably be revised or even replaced in the coming days, because this is something @libscott is working on. They'll also have to be updated to account for divisible assets. And more...


ttmc:
This PR will conflict quite substantially with PR #807 because it edited some of the same docs pages---after moving them.

Independent of PR #807, this PR is also a bit unfocused in that it both moved _and_ edited some docs. I think it might be better to break it up into two smaller and more-focused PRs, one that moves the docs, and one that does some edits.

Therefore I'm closing this PR and will create two new PRs in its place.

PR #806, opened 2016-11-11
add backlog count

rhsimplex:
Adds a backend call to count the number of transactions in the backlog


r-marques:
Should we also add a method on `core.py` since for the rest we don't call `rethinkdb.py` methods directly

r-marques:
Nevermind, this does not seem to be the case

PR #807, opened 2016-11-11
Feat/746/new naming convention for keys

### LINKED ISSUE #746: Naming convention: public/private keys versus verifying/signing keys?
In issue #603, @sbellem [wrote](https://github.com/bigchaindb/bigchaindb/issues/603#issuecomment-256032321):

> ### Naming convention: public/private keys versus verifying/signing keys
> 
> This question could perhaps have its own issue, but I'll ask here, and we can create an issue if needed.
> 
> I simply wonder whether we should stick to one rather than using both interchangeably. It may help make the code simpler, especially for new comers.

@r-marques [responded](https://github.com/bigchaindb/bigchaindb/issues/603#issuecomment-256038312):

> Looking at the literature they seem to be used interchangeably.
> 
> In asymmetric cryptography algorithms that provide both signing and encryption they use the terms public/private key.
> For asymmetric cryptography algorithms that only provide signing like ed25519 they sometimes refer to the keys as signing/verifying key.
> 
> It seems that public/private keys is a more universal term. But at the same time is always advisable to use different keys for signing and encryption and in that case a distinction between public/private and signing/verifying would be desirable.
> 
> At this point we only use the keys for signing and verifying maybe we could switch to using public/private everywhere since its the most used term

r-marques:
resolves #746 

Replaces all occurrences of verifying key/signing key with public key/private key.
Replaces `VerifyingKey` with `PublicKey`
Replaces `SigningKey` with `PrivateKey`
Replaces the short name for public key `vk` with `pk`
Short name for private key `sk` remains the same. It seems to a common short name for private key in both the literature and NaCL code

### DIFF
--- a/deploy-cluster-aws/write_keypairs_file.py
+++ b/deploy-cluster-aws/write_keypairs_file.py
@@ -10,8 +10,8 @@ Using the list in other Python scripts:
     # in a Python 2 script:
     from keypairs import keypairs_list
     # keypairs_list is a list of (sk, pk) tuples
-    # sk = signing key (private key)
-    # pk = verifying key (public key)
+    # sk = private key
+    # pk = public key
 """
 
 import argparse
--- a/docs/root/source/data-models/block-model.rst
+++ b/docs/root/source/data-models/block-model.rst
@@ -10,8 +10,8 @@ A block has the following structure:
       "block": {
         "timestamp": "<block-creation timestamp>",
         "transactions": ["<list of transactions>"],
-        "node_pubkey": "<public/verifying key of the node creating the block>",
-        "voters": ["<list of federation nodes verifying keys>"]
+        "node_pubkey": "<public key of the node creating the block>",
+        "voters": ["<list of federation nodes public keys>"]
       },
       "signature": "<signature of block>"
     }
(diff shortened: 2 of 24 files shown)


deploy-cluster-aws/write_keypairs_file.py
```diff
@@ -10,8 +10,8 @@
     # in a Python 2 script:
     from keypairs import keypairs_list
     # keypairs_list is a list of (sk, pk) tuples
-    # sk = signing key (private key)
-    # pk = verifying key (public key)
+    # sk = private key
```
ttmc: For clarity, maybe change this comment to:

```python
    # sk = private key (also called a signing key)
```

ttmc:
There will be merge conflicts between this PR and the PR where I moved & edited the Data Models docs (PR #804 ). I can take care of those if you merge this PR first.

docs/root/source/data-models/block-model.rst
```diff
@@ -10,8 +10,8 @@ A block has the following structure:
       "block": {
         "timestamp": "<block-creation timestamp>",
         "transactions": ["<list of transactions>"],
-        "node_pubkey": "<public/verifying key of the node creating the block>",
-        "voters": ["<list of federation nodes verifying keys>"]
+        "node_pubkey": "<public key of the node creating the block>",
+        "voters": ["<list of federation nodes public keys>"]
```
sohkai: Could be more descriptive, e.g. `list of public keys representing the federation nodes voting on this block`.

sohkai:
shane-compilation-amazin-AewTtLsNJBE9G

docs/root/source/data-models/block-model.rst
ttmc: @sohkai There is a longer description of what "voters" means on the same page, below the outline of the JSON structure at the top.

ttmc:
@r-marques Never mind my little quibble above, let's get this PR merged soon before resolving the merge conflicts becomes an absurd undertaking.

PR #813, opened 2016-11-13
Updated example AWS deployment config file to use Ubuntu 16.04 LTS

### LINKED ISSUE #773: Update aws deployment scripts to use ubuntu 16.04 LTS
Since there is a new LTS version of ubuntu we should consider changing the aws deploy scripts to use the new version.
BigchainDB would probably benefit from improvements and optimizations introduced in the file system and kernel io

ttmc:
* Updated the example AWS deployment config file to use an Ubuntu 16.04 AMI. Updated the documentation accordingly.
* Also added a paragraph to those docs about how to get an up-to-date Ubuntu AMI image ID for your AWS region
* In `fabfile.py`, updated the first command to install RethinkDB, based on [the instructions on the RethinkDB website](https://www.rethinkdb.com/docs/install/ubuntu/) (also updated). The old command was failing with Ubuntu 16.04

I tested it and it works.

This PR resolves #773

### DIFF
--- a/deploy-cluster-aws/example_deploy_conf.py
+++ b/deploy-cluster-aws/example_deploy_conf.py
@@ -43,9 +43,9 @@ USE_KEYPAIRS_FILE=False
 # Canonical (the company behind Ubuntu) generates many AMIs
 # and you can search for one that meets your needs at:
 # https://cloud-images.ubuntu.com/locator/ec2/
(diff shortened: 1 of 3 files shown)


sohkai:
> I tested it and it works.

Sounds good to me ;).

PR #814, opened 2016-11-14
web/test_basic_views --> web/test_transactions

TimDaub:
Gonna add a new endpoint and view soon with #601. Currently all tests in `test_basic_views.py` test `web/views/transactions.py`, so I thought it would make sense to just rename it, so that we can have the tests in separate files.


ttmc:
`web/test_basic_views.py` includes a test for the "basic info endpoint":

```python
def test_api_endpoint_shows_basic_info(client):
    from ...
``` 

Maybe that and any similar tests should be moved into a new file?

TimDaub:
> Maybe that and any similar tests should be moved into a new file?

Solved in 5c0616e

tests/web/test_info.py
```diff
@@ -0,0 +1,5 @@
+def test_api_endpoint_shows_basic_info(client):
+    from bigchaindb import version
```
libscott: Why is this in the function and not at the top?

ttmc:
screenshot from 2016-11-15 12 01 35

Shouldn't that test (`test_api_endpoint_shows_basic_info`) also be removed from the renamed tests file?

TimDaub:
> Shouldn't that test (test_api_endpoint_shows_basic_info) also be removed from the renamed tests file?

I messed that up. Sorry.

TimDaub:
> Shouldn't that test (test_api_endpoint_shows_basic_info) also be removed from the renamed tests file?

Should be fixed now. @ttmc

tests/web/test_info.py
TimDaub: Shouldn't it be OK to import it in the test in this case though, as you're actually using it to do an assertion?

libscott: @diminator, that still doesn't explain the reasoning, in `test_transaction.py` the imports are done over and over again, but the modules that are imported are not re-initialized each time (I just tested this), so I'm not sure what the reasoning behind it is.

sbellem: @libscott the practice of putting some imports under function scope is based on http://docs.pylonsproject.org/en/latest/community/testing.html

sbellem:
:+1:

tests/web/test_info.py
libscott: Ok, wasn't aware of this practice.

PR #815, opened 2016-11-14
Add a button to the docs linking to the bdb cli

libscott:
Screenie below. Any ideas where else to link these docs?

Links to https://docs.bigchaindb.com/projects/cli/en/latest/.

screenshot from 2016-11-14 15-50-08

### DIFF
--- a/docs/server/source/drivers-clients/index.rst
+++ b/docs/server/source/drivers-clients/index.rst
@@ -5,7 +5,7 @@ Currently, the only language-native driver is written in the Python language.
 
 We also provide the Transaction CLI to be able to script the building of
 transactions. You may be able to wrap this tool inside the language of
-your choice, and then use the server api directly to post transactions.
+your choice, and then use the HTTP API directly to post transactions.
 
 
 .. toctree::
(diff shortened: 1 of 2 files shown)


ttmc:
Maybe also add a link in the BigchainDB Server docs (/docs/server/) in the section titled **Drivers & Clients**?

https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/index.html

Note: In a Sphinx `index.rst` file, you can make a direct link using syntax like the following in the table of contents (toctree) section:
```
    This is a Link Title <http://url.goes/here.html>
```

docs/server/source/drivers-clients/index.rst
```diff
@@ -1,10 +1,18 @@
...
+We also provide the Transaction CLI to be able to script the building of
+transactions. You may be able to wrap this tool inside the language of
+your choice, and then use the server api directly to post transactions.
+
+
 .. toctree::
    :maxdepth: 1
 
    http-client-server-api
    python-driver
```
sohkai: @ttmc With these changes, I wonder if we should just link to the python driver's docs here instead of having a different page that does that?

We can handle this separately.

ttmc:
@libscott You'll probably recognize the two Flake8 errors that Travis CI is tripping over. Once those are handled, you should be able to merge this PR.

Issue #808, opened 2016-11-11
Proposal: remove user-provided Transaction timestamp


libscott:
Following on from #132.

Short version is, I propose to remove the Transaction timestamp, for the reasons given below. Currently nothing depends on the user provided timestamp, so it's a good time to do this.

This issue is separate from the issue of providing a timestamping service to the user, for which the server federation will provide some guarantees.

> As part of #798 I am wondering what kind of validation to add to the transaction timestamps, and having looked through the code and talked to people it seems like there is every reason to remove them:
> * Not used in the code, correct order depends on transaction dependency graph and not client provided timestamp.
> * Should not be used in the code, since client provided timestamps cannot be trusted, and we create our own when we do need them (tx backlog priority). 
> * Ambiguous meaning, see page: https://docs.bigchaindb.com/en/latest/timestamps.html
> * Maintenance overhead, see page: https://docs.bigchaindb.com/en/latest/timestamps.html
> * Ambiguity on what the data field *should* be, currently unix epoch in seconds, but milliseconds or a trusted timestamp service provided value are also legitimate candidates
> * Amounts to hand-holding of the user, when the user could very well put the timestamp in the `metadata` field if they so wished.

> The only possible value I can see to them being there is for operational analysis, ie, extracting a set of possible client latencies, but it's not clear _who_ this will benefit either.

ttmc:
This makes sense to me.

ttmc:
Resolved by pull request #817

PR #816, opened 2016-11-14
Fix database timestamp indexes

libscott:
2 issues found with timestamp indexes in DB, this PR fixes. The rationale on removing the transaction_timestamp index is that it's not used. The other fix is due to an index using the wrong timestamp.

* remove database index on transaction.timestamp
* fix database index on assignee__transaction_timestamp to use correct timestamp

See also: https://github.com/bigchaindb/bigchaindb/issues/808


ttmc:
This is code that @rhsimplex is most familiar with, as he added the `assignment_timestamp` as a way to determine when transactions in the backlog become stale.

libscott:
Going to push this through then - if there are any other requirements for timestamps or indexes we can address them separately.

Issue #697, opened 2016-10-18
Use transaction ID as asset ID


libscott:
When an asset is created, it needs to be identifiable with a unique ID, which can identify the set of transactions of this asset.

This ID can't be a hash of the asset, as the asset may have no distinguishing properties to give it a unique hash. The client may be trusted to generate a random asset ID, but this is also a potential attack vector and makes it more complicated to validate a block.

Using the ID of the transaction that created the asset has some nice properties:
- Client can still generate valid transactions
- No extra validation step is required
- No potential to create a duplicate (without privkey compromise), because the transaction ID depends on ~~a signature of the person creating  the asset~~ the public keys of the people creating the asset, verified by the signatures.

Thoughts?

ttmc:
This is a nice idea.

One concern I had:  A client could make two CREATE transactions with the same hash/id at substantially different times (e.g. by lying about the timestamp in the second one, setting it equal to the timestamp of the first one). I just checked with Ryan about what would happen in that situation. Answer: The second transaction would be checked to see if there's already a transaction with the same ID in an undecided or valid block, and if there is, it would be deleted from the backlog. Good!

sbellem:
I thought that the ID of the transaction was the hash of the transaction content. Has this changed?

ttmc:
@sbellem That is also my understanding (where "the transaction content" doesn't include the signatures [fulfillments]). Why do you ask?

libscott:
@sbellem I think you may be referring to the chicken / egg problem, how to record the asset ID inside the transaction, which then changes the transaction ID... The answer is that the asset definition inside the CREATE tx doesn't need to include the transaction ID, only subsequent transactions do.

sbellem:
@libscott ok then

ttmc:
Actually, it might not even be possible for a client to generate two transactions with the same hash/id, even if they try (e.g. by lying about the timestamp), because of the UUID in the `transaction.metadata.id`. I guess it depends on who generates _that_ UUID.

It doesn't matter though. Even if a tricksy client somehow manages to generate two CREATE transactions with the same id/hash, BigchainDB server will nuke the second one.

libscott:
@ttmc: to be clear, the dummy fulfillment in the signed CREATE tx is critically important because it allows the creator to control their address space (only the controller of that privkey can generate a tx with that hash).

[...]

PR #817, opened 2016-11-14
Remove user-provided Transaction timestamp

libscott:
See #808. This includes https://github.com/bigchaindb/bigchaindb/pull/816 also. If this gets merged I'll do the same for bdb cli and the driver.


ttmc:
The Travis CI tests are failing because of two small flake8 errors --- errors which should now be fixed in the master branch (since PR #765 got merged).

libscott:
Thanks, did a rebase.

docs/root/source/timestamps.md
```diff
@@ -1,95 +0,0 @@
-# Timestamps in BigchainDB
```
sohkai: I think we still need these since we still add timestamps for blocks and voters (as well as when we validate a fulfillment, although I'm not sure if that's actually shown)?

libscott: True.

It was a bit rushed to drop the entire file, not to mention it's nice documentation, but it will need a fairly comprehensive rewrite, as a large portion of it was related to the user provided timestamps.

I can rewrite it, but since it seems like the timestamps help file was born out of the ambiguity of the user provided transaction timestamp, what should be the new goals of the file?

docs/root/source/data-models/block-model.rst
```diff
@@ -20,7 +20,7 @@ A block has the following structure:
 - ``id``: The hash of the serialized ``block`` (i.e. the ``timestamp``, ``transactions``, ``node_pubkey``, and ``voters``). This is also a database primary key; that's how we ensure that all blocks are unique.
 
 - ``block``:
-    - ``timestamp``: The Unix time when the block was created. It's provided by the node that created the block. See `the page about timestamps <https://docs.bigchaindb.com/en/latest/timestamps.html>`_.
+    - ``timestamp``: The Unix time when the block was created. It's provided by the node that created the block.
```
libscott: Cool, in agreement with this, I expected there would be some discussion about this page, but as in my other comment it would be helpful to me to re-state the objectives of the timestamps page, as it seemed to deal primarily with the ambiguity arising from the user provided timestamp.

sohkai: I'd say the objective is to make clear that the node generated timestamps (and, really, any timestamps not verifiable) should not be trusted:

> The server-provided timestamps are still not 100% reliable or trustworthy. (A Byzantine node could put whatever timestamp it wants in its blocks or votes.)

diminator:
do we have a way to figure out which block a transaction belongs to?
Otherwise it is very tedious to know at which time a transaction occurred (or at which block_number)
In that case it becomes really hard to use BDB as a timestamping blockchain

libscott:
@diminator, it seems though that the issue of using BDB as a timestamping blockchain and the utility of the user provided `timestamp` field are separate issues. Perhaps we should have another ticket for the former?

diminator:
> Perhaps we should have another ticket for the former?

Yes, making this change really needs the other ticket of figuring out which block the transaction belongs to (and maybe additionally providing a block sequence number)

ttmc:
@diminator @libscott Yes, getting all the (node-generated) timestamps associated with a transaction will take a bit of effort:

* Get the block the transaction is in
* Get the timestamp from that block, somehow (mabye by getting the whole block and parsing it)
* Get the votes on that block
* Get the timestamps in those votes, somehow (maybe by getting all the contents of those votes and parsing them)

Anyway, those are separate things to do outside the scope of this PR.

sohkai:
> do we have a way to figure out which block a transaction belongs to?
Otherwise it is very tedious to know at which time a transaction occurred (or at which block_number)
In that case it becomes really hard to use BDB as a timestamping blockchain

In these cases, would adding a timestamp yourself (perhaps verifiable, e.g. OpenTimestamps) in the metadata not be sufficient? This might also be alleviated if a time oracle were present.

docs/root/source/timestamps.md
```diff
@@ -36,35 +34,26 @@ In all likelihood, you will never have to worry about leap seconds because they
...
 
-Those timestamps come from many sources, so you can look at all of them to get some statistical sense of when the transaction "actually happened." The timestamp of the block should always be after the timestamp of the transaction, and the timestamp of the votes should always be after the timestamp of the block.
-
 
 ## How BigchainDB Uses Timestamps
 
 BigchainDB _doesn't_ use timestamps to determine the order of transactions or blocks. In particular, the order of blocks is determined by RethinkDB's changefeed on the bigchain table.
 
-BigchainDB does use timestamps for some things. It uses them to determine if a transaction has been waiting in the backlog for too long (i.e. because the node assigned to it hasn't handled it yet). It also uses timestamps to determine the status of timeout conditions (used by escrow).
+BigchainDB does use timestamps for some things. It uses them to determine if a transaction has been waiting in the backlog for too long (i.e. because the node assigned to it hasn't handled it yet).
```
sohkai: Don't remember exactly how we do this, but is this still possible if we remove the timestamp from a transaction? Does the added `assignee` information contain a timestamp?

ttmc: There's a special `assignment_timestamp` added to a transaction when it gets written to the backlog table. It's what's used to decide if a transaction is stale.

@libscott Maybe this paragraph should say something about that (i.e. to distinguish the two kinds of timestamps)? We're not getting rid of the `assignment_timestamp`.

Note: The `assignment_timestamp` is removed before the transaction gets written to a block. It's an ephemeral backlog-table-specific appendage.

ttmc:
@sbellem It looks like we'll be using the transaction id to identify assets in the future (i.e. the id of the CREATE transaction where the asset was created). The asset.id will be going away. See the discussion on issue #697 

This raises an interesting question: if there are two transactions with the same id (hash), then is it _okay_ to delete one (because it's presumed identical)?

docs/root/source/timestamps.md
libscott: ok, I'll mention `assignment_timestamp` explicitly.

libscott:
> This raises an interesting question: if there are two transactions with the same id (hash), then is it okay to delete one (because it's presumed identical)?

Two transactions with the same ID are the same transaction. Currently I'm not sure if we do check for this when we write a transaction to the backlog or to a block.

> NOTE: The time stamp (and asset id) play a critical role in the uniqueness of the transaction id. If both are removed two transactions can easily hash to the same value, and consequently have identical ids.

Correct, and #697 relies on proper incentivisation ie of the user who `CREATE`s an asset to not collide with his/her own transactions. In this case, a signature must be part of the transaction hash.

ttmc:
@libscott The signatures (fulfillment URIs) _aren't_ currently included in what gets hashed to calculate the transaction hash (transaction id). I suppose they could be. Maybe make a new issue for that, including an explanation of why it's necessary (in the new issue's opening comment).

r-marques:
There is one use case where we did trust the user provided timestamp and that was with timelocks.
In this case the user had an incentive to provide a correct timestamp.

When re-adding the timelock functionality to the server the only other way I see is to use the timestamp of the block in which the transaction was included but this approach has two problems:
- The client will need to trust the server provided timestamp:
- The timestamp will be of when the transaction was validated and not when the transaction was created.

@diminator do you have any input for this since you have more knowledge about timelocks

ttmc:
Today, each node must validate each transaction, usually. Why not leave the determination of the "current time" up to each node: the node would just get the current time as-needed, from its own clock.

If it's definitely before the timeout time, then most nodes will agree that now is before the timeout time. If it's definitely after the timeout time, then most nodes will agree that now is after the timeout time. They'll only disagree, maybe, when now is near the timeout time, and then voting will decide.

I think that approach would be the most decentralized.

For people who care about super-duper time precision, then one option would be to rely on some trusted timestamp from a trusted timestamp provider (i.e. an oracle, a third party). The trusted timestamp could be put in the transaction's metadata. Support for that feature could be added later if there's enough demand.

r-marques:
I understand what you are saying and there is some confusion about how the timelock works at the moment.
When creating a timelock a user did not actually give a time, just an offset in seconds. Something like this timelock expires 30 min after the creation of this transaction.

With what you are proposing we will need to change the actual timelock condition to carry a timestamp instead of an offset

diminator:
The timelocks work currently by having each node executing `now()` upon validation of the transaction, and comparing the local node time with a UNIX timestamp (the `expiry_time`).

This poses a problem when validating a timelocked transaction in the future - after the expiry time.
The user would claim that the transaction time needs to be frozen (timestamped) upon submitting it to BDB. Or at least, the time at which each node in the system sees the transaction.

FYI: In the future-validation can happen in a few scenario's, eg:
- when a block was `INVALID` due to another transaction and the transaction bounces back and forth between the `backlog` and `INVALID` blocks.
- when you want to validate an entire asset history

One suggestion, made by @libscott is to use a `locktime` instead of `expiry_time`. This would mean that a transaction needs to wait a while to become valid - but it would remain valid in all future validations. This becomes more complicated when using `execute_condition` and `abort_condition` in hold/escrow transactions.

diminator:
> With what you are proposing we will need to change the actual timelock condition to carry a timestamp instead of an offset

agreed

r-marques:
OK

So since the timelock functionality has not yet been implemented by common now it would be a good time to redesign this feature.

Regarding this pr it is safe to remove the transaction timestamp if we change the timelock condition to carry a timestamp instead of a time offset. This way we can add server side validation code in which each node compares this client provided timestamp with the current server time

r-marques:
@diminator pointed out I was wrong. The timelock already contains a unix timestamp so this pr does not interfere with the timelock conditions

libscott:
Cool. @sbellem, are we good to go?

PR #818, opened 2016-11-15
Add /unspents/ endpoint

TimDaub:
As already discussed in #803, this is the follow up PR to implement the functionality that was proposed.


TimDaub:
See #830 for updates.

We'll implement this functionality, but probably using another path scheme.
I'll reopen this once #830 is merged.

sbellem:
Has been replaced by #955

PR #819, opened 2016-11-15
Corrections to render sphinx correctly

TimDaub:
> I took a look to see how they render via ReadTheDocs. It seems Sphinx got confused with how to render the Parameters and Status Codes.

Thx @ttmc, I should have checked the rendered output :/ (I did this time!)

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -301,8 +301,7 @@ GET /unspents/
    the server, this will result in the server returning a 200 OK HTTP status
    code and an empty list in the response's body.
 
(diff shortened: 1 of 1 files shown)


ttmc:
This time I checked out the branch and rendered it locally just to see what happens. The Parameters and Status codes render fine now.

I also noticed it fails to syntax-highlight the final example response (HTTP). Sphinx gave the warning:
```text
/home/troy/repos/bigchaindb/docs/server/source/drivers-clients/http-client-server-api.rst:316: WARNING: Could not lex literal_block as "http". Highlighting skipped.
```
I honestly don't see what's wrong or different from previous example HTTP response blocks.  I'm not sure it matters that much though. I'd be okay with merging as is.

PR #821, opened 2016-11-15
Removed the "api_endpoint" setting from everywhere (in this repo)

### LINKED ISSUE #780: Remove "api_endpoint" from the configuration settings?
The `api_endpoint` is [a configuration setting](https://docs.bigchaindb.com/projects/server/en/latest/server-reference/configuration.html#api-endpoint).

I think it was originally going to be used by the `bigchaindb load` command (for load testing), but I don't think it's being used for that. It appears that command uses the `database.host`, `database.port` and `database.name` config settings to know where to write the transactions.

As far as I can tell, the `api_endpoint` setting isn't being used anywhere in the functional part of the code. For example, it's not used by the HTTP API web server. The URL prefix "/api/v1" for the transaction endpoints [is actually hard-coded](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/web/server.py#L68) (and that's okay).

To do:

- [x] Remove the `api_endpoint` setting, and all the code associated with it. Don't forget the AWS deployment scripts.
- [x] Update the docs about the BigchainDB Server Configuration Settings: remove the section about `api_endpoint`. Write something about how the API URL is actually determined.

ttmc:
As described in issue #780, the `api_endpoint` setting isn't actually used anywhere (as far as I can tell), so I removed it from everywhere.

I added some docs to the page about the HTTP API to explain how the API Root URL is _actually_ determined.

* Ran all unit tests. They passed.
* Ran flake8. It passed.
* Tested an AWS deployment. It worked.
* Played around with Docker based on edited Dockerfile. It seemed to work (but the `bigchaindb load` command seems to be broken now; that's another issue)

Resolves #780


ttmc:
@sbellem Could you try out the Dockerfile and Docker Compose file to see if that stuff is still okay? I did some testing but I'm not a Docker pro.

I created a related issue to update the BigchainDB Python Driver (docs etc.): https://github.com/bigchaindb/bigchaindb-driver/issues/142

sohkai:
Oh interesting, we have `BIGCHAINDB_SERVER_BIND` that actually sets the server host and port. There's a lot of dockerfiles to go through and check in all the different repos, but since this env variable actually isn't used, nothing shouldn't be affected... right?

ttmc:
`BIGCHAINDB_SERVER_BIND` isn't the whole story of how the public HTTP API URLs are determined. I added some documentation about how those are _actually_ determined (in this PR, in `docs/server/source/drivers-clients/http-client-server-api.rst`).

sohkai:
Just grepped through my local repos (which should include most of our docker-related files) and only found two instances of `BIGCHAINDB_API_ENDPOINT` (both depend on the python driver). Once we confirm that the driver doesn't need it, I'll remove those instances as well.

ttmc:
This pull request can probably be merged around the same time as https://github.com/bigchaindb/bigchaindb-driver/pull/171

ttmc:
https://github.com/bigchaindb/bigchaindb-driver/pull/171 was merged earlier today so I'm merging this pull request now. Farewell, "api_endpoint"

Issue #823, opened 2016-11-16
Fix "bigchaindb load"


ttmc:
The `bigchaindb load` command is broken. It exists to do load testing by dumping transactions straight into the backlog table. Here's an example of what happens when I tried to run it today (on my local machine with default configs):
```bash
$ bigchaindb load -c 100
INFO:bigchaindb.config_utils:Configuration loaded from `/home/troy/.bigchaindb`
INFO:bigchaindb.commands.bigchain:Starting 1 processes
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib/python3.4/multiprocessing/process.py", line 254, in _bootstrap
    self.run()
  File "/usr/lib/python3.4/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/troy/repos/bigchaindb/bigchaindb/commands/bigchain.py", line 197, in _run_load
    tx = Transaction.create([b.me], [b.me])
  File "/home/troy/repos/bigchaindb/bigchaindb/common/transaction.py", line 764, in create
    raise ValueError(('Each `owner_after` in the list must be a'
ValueError: Each `owner_after` in the list must be a tuple of `([<list of public keys>], <amount>)`
```

The culprit seems to be the call to `Transaction.create()`. Its interface changed with the introduction of divisible assets.

I think the fix would be easy, but maybe there's some subtlety I'm not seeing.

sohkai:
Fixed by #824.

PR #824, opened 2016-11-16
Fixed bug in `bigchaindb load`

ttmc:
Resolves issue #823 

I also updated the top docstring in `commands/bigchain.py`: there is more than one "BigchainDB CLI" these days, so I clarified that that file has the implementation of the BigchainDB Server CLI.


ttmc:
@sohkai I created a new issue to write more unit tests for `bigchaindb load`: #829 

I will merge this PR.

PR #830, opened 2016-11-16
es[WIP/Feedback] Design BigchainDB HTTP API v0.9

TimDaub:
This morning @diminator and @sbellem had a discussion about the BigchainDB HTTP API.
From the discussion, the conclusion was drawn that in order to properly design the API, we should - instead of iteratively - a single document outlining the individual calls using HTTP as this methodology:

- provides a better overview of actual endpoints to be implemented; and
- allows us to plan and split tasks more efficiently.

Hence, for the rest of the day I hacked together a first draft of what the BigchainDB HTTP API could look like.

## How to review

When your reviewing, see this rather as a proposal. It's here to be discussed with everyone. Things can and will change.
A few things upfront though:

- All paths, responses, status codes, headers were assigned according to widely-accepted specifications and heuristics. Suggestions like: "I prefer X to be like schema Y" is therefore only acceptable as feedback if founded on a credible source (specification > blog posts  and so on - obviously!)
- Reasoning for changes is provided in the message section of the commits. Read them.
- The diff of the `.rst` file will very likely not give you a good overview of the changes. Pull the changes from this branch and render them in sphinx for reviewing/reading.

When merging this PR, we should have an HTTP API everyone is happy with :tada:.

#### Countering "endpoint madness"

A very simple rule:
> Any entity that is retrievable with a unique key has its own top-level endpoint.

This only makes sense as REST mainly implements pathing after the following scheme:

- `/resource`: List endpoint
- `/resource?manipulator=X`: Filtered/manipulated list endpoint
- `/resource/id`: Single endpoint
- `/resource/id?manipulator=X` Filtered/manipulated single endpoint

Concretely, this means the following resources will have a top-level endpoint:

- blocks
- votes
- transactions
- statuses
- assets
- metadata

Keep in mind: Best practices advise strongly for a flat path hierarchy.

##### "But, Fulfillments and Conditions should have their own top-level endpoints!"

Theoretically they could. In fact, we could pretend that a condition's or fulfillment's unique key of retrieval is their URI. This would enable the following endpoints:

- `GET /conditions`
- `GET /conditions/uri`
- `GET /fulfillments`
- `GET /fulfillments/uri`

However, there has yet to be a use case found, where a HTTP API user would benefit from this. Hence in this proposal I thought about it, but didn't include it in the design.

#### ~Retrieval of specific fields in a resource~

~I followed [this very popular guide](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api):~

> ~Limiting which fields are returned by the API~
> ~The API consumer doesn't always need the full representation of a resource. The ability select and chose returned fields goes a long way in letting the API consumer minimize network traffic and speed up their own usage of the API.~
> ~Use a fields query parameter that takes a comma separated list of fields to include. For example, the following request would retrieve just enough information to display a sorted listing of open tickets:~
> ~GET /tickets?fields=id,subject,customer_name,updated_at~

#### `201 Created` for `POST /transactions` becomes `202 Accepted`

https://www.ietf.org/rfc/rfc2616.txt `10.2.2 201 Created` reads:

> The origin server MUST create the resource before returning the 201 status code.

Since a transaction is not created before the `POST` request returns, we cannot use status code 201 to respond to the request. However (same document) `10.2.3 202 Accepted` reads:

> The request has been accepted for processing, but the processing has not been completed.
> The entity returned with this response SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled.

Using the HTTP `Location` header, we redirect the client to the status endpoint to ask for the status of the creation.

Once the resource has been created, we use HTTP code `303 See Other`, to notify the client about the updated resource and, using the  `Location` header, link back to the actual transaction `/transactions/txid`.

#### `/transaction/{txid}/status` becomes `/statuses`

1. [The plural of status is `statuses`](http://english.stackexchange.com/questions/877/what-is-the-plural-form-of-status)
2. `/transaction/{txid}/status` is not a RESTful path
3. Statuses do not only occur on transactions. In fact, a status is derived from a block, a transaction just inherits it from the block.

Therefore, an endpoint `/status` was introduced that allows to query any entity that is retrievable by a unique ID (block, transaction, asset, metadata, votes(not sure)).

## Things to be done

- [x] Get feedback
- [x] [Find solution to the `/unspents` endpoint-problem](https://github.com/bigchaindb/bigchaindb/pull/830#issuecomment-261502197)
- [x] Block retrieval by ID
- [x] Block retrieval by transaction
- [x] Asset retrieval by ID
- [x] Asset list retrieval
- [x] Asset history retrieval
- [x] ~Metadata retrieval by ID~
- [x] Include related context in statuses response (type, related resources)
- [x] ~Votes retrieval by ID **(use case missing)**~
- [x] Votes retrieval by Block ID
- [x] Outline how to retrieve the timestamp of the block containing the transaction
- [ ] Outline how to retrieve the timestamps of the votes for a certain block
- [x] Consider resolvement of #835 (remove `?fields` query strings)
- [x] Consider changes made in #856 (remove `metadata.id`)
- [x] ~Consider implementing `get_spent` functionality~
- [x] Consider timestamp removal made in #817
- [x] Consider resolving #906
- [x] Consider changes made in #925 (Inputs and Outputs)
- [x] Consider changes made in #945 (remove `owners_before`)
- [x] Consider changes made in #963 (rename `owners_after`)

## Things to better be discussed later

- Sorting (timestamps are influx right now)
- Pagination (easy to add later on)
- How this proposal relates to a [Interledger Common Web API](https://github.com/mDuo13/rfcs/blob/7a6a4b0723c347759c0366836c48c546efc9f268/0012-common-ledger-api/0012-common-ledger-api.md#common-ledger-api) (this is the BigchainDB™ API, lets separate concerns for now)
- Retrieval of full blocks vs hyperlinked versions [see comment](https://github.com/bigchaindb/bigchaindb/pull/830#issuecomment-269996902)
- Websockets: see issue #862 
- Client-defined queries and indices [see comment](https://github.com/bigchaindb/bigchaindb/pull/830#issuecomment-268258036)

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -128,15 +128,13 @@ Transactions
    queried correctly. Some of them include retrieving a list of transactions
    that include:
 
-   * `Unfulfilled conditions <#get--transactions?fields=id,conditions&fulfilled=false&owners_after=owners_after>`_
-   * `A specific asset <#get--transactions?fields=id,asset,operation&operation=CREATE|TRANSFER&asset_id=asset_id>`_
-   * `Specific metadata <#get--transactions?fields=id,metadata&metadata_id=metadata_id>`_
+   * `Unfulfilled conditions <#get--transactions?fulfilled=false&owners_after=owners_after>`_
+   * `A specific asset <#get--transactions?operation=CREATE|TRANSFER&asset_id=asset_id>`_
+   * `Specific metadata <#get--transactions?&metadata_id=metadata_id>`_
 
    In this section, we've listed those particular requests, as they will likely
    to be very handy when implementing your application on top of BigchainDB.
    A generalization of those parameters can follows:
-
-   :query string fields: Comma separated list, allowed values are: ``asset``, ``conditions``, ``fulfillments``, ``id``, ``metadata``, ``operation``, ``owners_after``, ``version``.
 
    :query boolean fulfilled: A flag to indicate if transaction's with fulfilled conditions should be returned.
 
@@ -151,7 +149,7 @@ Transactions
    :statuscode 404: BigchainDB does not expose this endpoint.
 
 
-.. http:get:: /transactions?fields=id,conditions&fulfilled=false&owners_after={owners_after}
+.. http:get:: /transactions?fulfilled=false&owners_after={owners_after}
 
    Get a list of transactions with unfulfilled conditions.
 
@@ -162,8 +160,6 @@ Transactions
    This endpoint returns conditions only if the transaction they're in are
    included in a ``VALID`` or ``UNDECIDED`` block on ``bigchain``.
 
-   :query string fields: A comma separated string to expand properties on the transaction object to be returned.
-
    :query boolean fulfilled: A flag to indicate if transaction's with fulfilled conditions should be returned.
 
    :query string owners_after: Public key able to validly spend an output of a transaction, assuming the user also has the corresponding private key.
@@ -172,7 +168,7 @@ Transactions
 
    .. sourcecode:: http
 
-      GET /transactions?fields=id,conditions&fulfilled=false&owners_after=1AAAbbb...ccc HTTP/1.1
+      GET /transactions?fulfilled=false&owners_after=1AAAbbb...ccc HTTP/1.1
       Host: example.com
 
    **Example response**:
@@ -203,7 +199,29 @@ Transactions
               ]
             }
           ],
+          "operation": "CREATE",
+          "asset": {
+            "divisible": false,
+            "updatable": false,
+            "data": null,
+            "id": "aebeab22-e672-4d3b-a187-bde5fda6533d",
+            "refillable": false
+          },
+          "metadata": null,
+          "timestamp": "1477578978",
+          "fulfillments": [
+            {
+              "fid": 0,
+              "input": null,
+              "fulfillment": "cf:4:GG-pi3CeIlySZhQoJVBh9O23PzrOuhnYI7OHqIbHjkn2VnQaEWvecO1x82Qr2Va_JjFywLKIOEV1Ob9Ofkeln2K89ny2mB-s7RLNvYAVzWNiQnp18_nQEUsvwACEXTYJ",
+              "owners_before": [
+                "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+              ]
+            }
+          ]
+        },
         "id": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
+        "version": 1
       }]
 
    :resheader Content-Type: ``application/json``
@@ -211,14 +229,12 @@ Transactions
    :statuscode 200: A list of transaction's containing unfulfilled conditions was found and returned.
    :statuscode 400: The request wasn't understood by the server, e.g. the ``owners_after`` querystring was not included in the request.
 
-.. http:get:: /transactions?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}&asset_id={asset_id}
+.. http:get:: /transactions?operation={GENESIS|CREATE|TRANSFER}&asset_id={asset_id}
 
    Get a list of transactions that use an asset with the ID ``asset_id``.
 
    This endpoint returns assets only if the transaction they're in are
    included in a ``VALID`` or ``UNDECIDED`` block on ``bigchain``.
-
-   :query string fields: A comma separated string to expand properties on the transaction object to be returned.
 
    :query string operation: One of the three supported operations of a transaction: ``GENESIS``, ``CREATE``, ``TRANSFER``.
 
@@ -228,7 +244,7 @@ Transactions
 
    .. sourcecode:: http
 
-      GET /transactions?fields=id,asset,operation&operation=CREATE&asset_id=1AAAbbb...ccc HTTP/1.1
+      GET /transactions?operation=CREATE&asset_id=1AAAbbb...ccc HTTP/1.1
       Host: example.com
 
    **Example response**:
@@ -240,6 +256,26 @@ Transactions
 
       [{
         "transaction": {
+          "conditions": [
+            {
+              "cid": 0,
+              "condition": {
+                "uri": "cc:4:20:GG-pi3CeIlySZhQoJVBh9O23PzrOuhnYI7OHqIbHjkk:96",
+                "details": {
+                  "signature": null,
+                  "type": "fulfillment",
+                  "type_id": 4,
+                  "bitmask": 32,
+                  "public_key": "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+                }
+              },
+              "amount": 1,
+              "owners_after": [
+                "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+              ]
+            }
+          ],
+          "operation": "CREATE",
           "asset": {
             "divisible": false,
             "updatable": false,
@@ -247,8 +283,21 @@ Transactions
             "id": "1AAAbbb...ccc",
             "refillable": false
           },
-        "operation": "CREATE",
+          "metadata": null,
+          "timestamp": "1477578978",
+          "fulfillments": [
+            {
+              "fid": 0,
+              "input": null,
+              "fulfillment": "cf:4:GG-pi3CeIlySZhQoJVBh9O23PzrOuhnYI7OHqIbHjkn2VnQaEWvecO1x82Qr2Va_JjFywLKIOEV1Ob9Ofkeln2K89ny2mB-s7RLNvYAVzWNiQnp18_nQEUsvwACEXTYJ",
+              "owners_before": [
+                "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+              ]
+            }
+          ]
+        },
         "id": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
+        "version": 1
       }]
 
    :resheader Content-Type: ``application/json``
@@ -256,14 +305,12 @@ Transactions
    :statuscode 200: A list of transaction's containing an asset with ID ``asset_id`` was found and returned.
    :statuscode 400: The request wasn't understood by the server, e.g. the ``asset_id`` querystring was not included in the request.
 
-.. http:get:: /transactions?fields=id,metadata&metadata_id={metadata_id}
+.. http:get:: /transactions?metadata_id={metadata_id}
 
    Get a list of transactions that use metadata with the ID ``metadata_id``.
 
    This endpoint returns assets only if the transaction they're in are
    included in a ``VALID`` or ``UNDECIDED`` block on ``bigchain``.
-
-   :query string fields: A comma separated string to expand properties on the transaction object to be returned.
 
    :query string metadata_id: metadata ID.
 
@@ -271,7 +318,7 @@ Transactions
 
    .. sourcecode:: http
 
-      GET /transactions?fields=id,metadata&metadata_id=1AAAbbb...ccc HTTP/1.1
+      GET /transactions?metadata_id=1AAAbbb...ccc HTTP/1.1
       Host: example.com
 
    **Example response**:
@@ -283,12 +330,53 @@ Transactions
 
       [{
         "transaction": {
-        "metadata": {
-          "id": "1AAAbbb...ccc",
-          "data": {
-            "hello": "world"
+          "conditions": [
+            {
+              "cid": 0,
+              "condition": {
+                "uri": "cc:4:20:GG-pi3CeIlySZhQoJVBh9O23PzrOuhnYI7OHqIbHjkk:96",
+                "details": {
+                  "signature": null,
+                  "type": "fulfillment",
+                  "type_id": 4,
+                  "bitmask": 32,
+                  "public_key": "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+                }
+              },
+              "amount": 1,
+              "owners_after": [
+                "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+              ]
+            }
+          ],
+          "operation": "CREATE",
+          "asset": {
+            "divisible": false,
+            "updatable": false,
+            "data": null,
+            "id": "aebeab22-e672-4d3b-a187-bde5fda6533d",
+            "refillable": false
           },
+          "metadata": {
+            "id": "1AAAbbb...ccc",
+            "data": {
+              "hello": "world"
+            },
+          },
+          "timestamp": "1477578978",
+          "fulfillments": [
+            {
+              "fid": 0,
+              "input": null,
+              "fulfillment": "cf:4:GG-pi3CeIlySZhQoJVBh9O23PzrOuhnYI7OHqIbHjkn2VnQaEWvecO1x82Qr2Va_JjFywLKIOEV1Ob9Ofkeln2K89ny2mB-s7RLNvYAVzWNiQnp18_nQEUsvwACEXTYJ",
+              "owners_before": [
+                "2ePYHfV3yS3xTxF9EE3Xjo8zPwq2RmLPFAJGQqQKc3j6"
+              ]
+            }
+          ]
+        },
         "id": "2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e",
+        "version": 1
       }]
 
    :resheader Content-Type: ``application/json``
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -26,6 +26,7 @@ with something like the following in the body:
     {
       "_links": {
         "docs": "https://docs.bigchaindb.com/projects/server/en/v0.9.0/"
+        "api_v1": "http://example.com:9984/api/v1/"
       },
       "keyring": [
         "6qHyZew94NMmUTYyHnkZsB8cxJYuRNEiEpXHe1ih9QX3",
@@ -50,12 +51,10 @@ that allows you to discover the BigchainDB API endpoints:
 
     {
       "_links": {
-        "blocks": "https://example.com:9984/api/v1/blocks",
         "docs": "https://docs.bigchaindb.com/projects/server/en/v0.9.0/drivers-clients/http-client-server-api.html",
         "self": "https://example.com:9984/api/v1",
         "statuses": "https://example.com:9984/api/v1/statuses",
         "transactions": "https://example.com:9984/api/v1/transactions",
-        "votes": "https://example.com:9984/api/v1/votes"
       },
       "version" : "0.9.0"
     }
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -382,7 +382,48 @@ Blocks
 
 .. http:get:: /blocks/{block_id}
 
-   Descriptions: TODO
+   Get the block with the ID ``block_id``.
+
+   A block is only returned if it was labeled ``VALID`` or ``UNDECIDED`` and
+   exists in the table ``bigchain``.
+
+   :param block_id: block ID
+   :type block_id: hex string
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      GET /blocks/6152fbc7e0f7686512ed6b92c01e8c73ea1e3f51a7b037ac5cc8c860215e7202 HTTP/1.1
+      Host: example.com
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "block":{
+          "node_pubkey":"ErEeVZt8AfLbMJub25tjNxbpzzTNp3mGidL3GxGdd9bt",
+            "timestamp":"1479389911",
+            "transactions":[
+              '<transaction1>',
+              '<transaction2>'
+            ],
+            "voters":[
+              "ErEeVZt8AfLbMJub25tjNxbpzzTNp3mGidL3GxGdd9bt"
+            ]
+        },
+        "id":"6152fbc7e0f7686512ed6b92c01e8c73ea1e3f51a7b037ac5cc8c860215e7202",
+        "signature":"53wxrEQDYk1dXzmvNSytbCfmNVnPqPkDQaTnAe8Jf43s6ssejPxezkCvUnGTnduNUmaLjhaan1iRLi3peu6s5DzA"
+      }
+
+   :resheader Content-Type: ``application/json``
+
+   :statuscode 200: A block with that ID was found. Its status is either ``VALID`` or ``UNDECIDED``.
+   :statuscode 404: A block with that ID was not found.
 
 .. http:get:: /blocks?tx_id={tx_id}
 
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -12,8 +12,8 @@ and you're not sure what the API Root URL is,
 then see the last section of this page for help.
 
 
-API Root URL
-------------
+BigchainDB Root URL
+-------------------
 
 If you send an HTTP GET request to the API Root URL
 e.g. ``http://localhost:9984`` 
@@ -31,7 +31,28 @@ with something like the following in the body:
       ],
       "public_key": "AiygKSRhZWTxxYT4AfgKoTG4TZAoPsWoEt6C6bLq4jJR",
       "software": "BigchainDB",
-      "version": "0.6.0"
+      "version": "0.9.0"
+    }
+
+
+API Root URL
+-------------------
+
+If you send an HTTP GET request to the API Root URL
+e.g. ``http://localhost:9984/api/v1/``
+or ``http://apihosting4u.net:9984/api/v1/``,
+then you should get an HTTP response
+that allows you to discover the BigchainDB endpoints:
+
+.. code-block:: json
+
+    {
+      "_links": {
+        "self": { "href": "/" },
+        "transactions": { "href": "/transactions" },
+        "statuses": { "href": "/statuses" },
+        "blocks": { "href": "/blocks" },
+        "votes": { "href": "/votes" }
     }
 
 Transactions
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -247,8 +247,11 @@ Transactions
    .. literalinclude:: samples/post-tx-response.http
       :language: http
 
-   :statuscode 202: The pushed transaction was accepted, but the processing has not been completed.
-   :statuscode 400: The transaction was invalid and not created.
+   :resheader Content-Type: ``application/json``
+   :resheader Location: As the transaction will be persisted asynchronously, an endpoint to monitor its status is provided in this header.
+
+   :statuscode 202: The pushed transaction was accepted in the ``BACKLOG``, but the processing has not been completed.
+   :statuscode 400: The transaction was malformed and not accepted in the ``BACKLOG``.
 
 
 Statuses
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -636,14 +636,8 @@ Blocks
 Votes
 --------------------------------
 
-.. http:get:: /votes/{vote_id}
-
-   Descriptions: TODO
-
 .. http:get:: /votes?block_id={block_id}
 
    Descriptions: TODO
 
 .. http:get:: /votes?block_id={block_id}&voter={voter}
-
-   Descriptions: TODO
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -257,12 +257,13 @@ Transactions
 Statuses
 --------------------------------
 
-.. http:get:: /statuses/{tx_id|block_id}
+.. http:get:: /statuses?tx_id={tx_id}|block_id={block_id}
 
    Get the status of an asynchronously written resource by their id.
 
    Supports the retrieval of a status for a transaction using ``tx_id`` or the
-   retrieval of a status for a block using ``block_id``.
+   retrieval of a status for a block using ``block_id``. Only use exactly one of both
+   queries, as they are required but mutually exclusive.
 
    The possible status values are ``backlog``, ``undecided``, ``valid`` or
    ``invalid``.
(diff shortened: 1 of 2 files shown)


docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -115,82 +119,133 @@ POST /transactions/
...
+   :type operation: string
+
+   :query asset_id: asset ID.
+   :type asset_id: uuidv4
+
+   :query metadata_id: metadata ID.
+   :type metadata_id: uuidv4
+
+
+.. http:get:: /transactions?fields=id,conditions&fulfilled=false&owner_afters={owners_after}
```
diminator: It might be more pleasant for the user to use 
``GET /transactions/unfulfilled?fields=id,conditions&owners_after={owners_after}``
opposed to/or in conjunction with
``GET /transactions/?fields=id,conditions&fulfilled=false&owners_after={owners_after}``

this way it's more clear that ``GET /transactions`` is a different endpoint - and this one has actually better mapping to what people in the blockchain space call unspent outputs (UTXO)

from http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api
## Aliases for common queries
To make the API experience more pleasant for the average consumer, consider packaging up sets of conditions into easily accessible RESTful paths. For example, the recently closed tickets query above could be packaged up as GET /tickets/recently_closed

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -115,82 +119,133 @@ POST /transactions/
...
+   A generalization of those parameters can follows:
+
+   :query fields: A comma separated string to expand properties on the transaction object to be returned.
+   :type fields: string
+
+   :query fulfilled: A flag to indicate if transaction's with fulfilled conditions should be returned.
+   :type fulfilled: boolean
+
+   :query owners_after: Public keys able to validly spend an output of a transaction, assuming the user also has the corresponding private key.
+   :type owners_after: base58 encoded string
```
diminator: Is it a list of keys or a single key?

I assume that it's a list, disregarding of whether we'll be able to make the `owners_after` consistent with the `condition_uri`

docs/server/source/drivers-clients/http-client-server-api.rst
diminator: `owner_afters={owners_after}` > `owners_after={owners_after}`

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -115,82 +119,133 @@ POST /transactions/
...
+   :query asset_id: asset ID.
+   :type asset_id: uuidv4
+
+   :query metadata_id: metadata ID.
+   :type metadata_id: uuidv4
+
+
+.. http:get:: /transactions?fields=id,conditions&fulfilled=false&owner_afters={owners_after}
+
+   Get a list of transactions with unfulfilled conditions.
```
diminator: this section requires to set the `fulfilled` always to false, hence shortening the name to `/transactions/unfulfilled` resolves this - ie always implying `fulfilled=false`

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -115,82 +119,133 @@ POST /transactions/
...
-   :statuscode 201: A new transaction was created.
-   :statuscode 400: The transaction was invalid and not created.
+   :resheader Content-Type: ``application/json``
 
+   :statuscode 200: A list of transaction's containing unfulfilled conditions was found and returned.
+   :statuscode 400: The request wasn't understood by the server, e.g. the ``owners_after`` querystring was not included in the request.
 
-GET /transactions/{tx_id}/status
---------------------------------
+.. http:get:: /transactions?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}&asset_id={asset_id}
```
diminator: (proposal) we could shorten this query to 
``GET /transactions/assets/{asset_id}?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}``
or even
``GET /assets/{asset_id}?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}``
or (under the assumption `operation=CREATE`)
``GET /assets/{asset_id}?fields=id,asset,operation``

or maybe I'm being silly ;-)

anyhow, I know shortening should only be additional - and not core, but it might be useful for the end user to distinguish between the various flavors of transactions...

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -318,10 +433,67 @@ GET /unspents/
...
+
+   :statuscode 200: A transaction or block with that ID was found. The status is either ``backlog``, ``invalid``.
+   :statuscode 303: A transaction or block with that ID was found and persisted to the chain. A location header to the resource is provided.
+   :statuscode 404: A transaction or block with that ID was not found.
+
+
+Assets
+--------------------------------
+
+.. http:get:: /assets/{asset_id}
```
diminator: maybe is that related to the comment above?

ttmc:
Assuming the client-provided timestamp gets removed from the Transaction data model (via pull request #817), clients will want some way to get the other timestamps associated with a transaction:

* the timestamp of the block containing the transaction
* the timestamps of the votes on that block

I'm not sure how to design the endpoints to get those timestamps.

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: `get_owned_ids` just takes a single owner, but for the API we could allow multiple keys.
Right now, we don't have a use case to support multiple keys as far as I'm aware of.

My suggestion:

Keep it a single key for now. Allow multiple keys as soon as someone asks for it, as this functionality can be added without breaking the input and output of the endpoint.
OK?

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: > Aliases for common queries

Yes, I saw that as well and I think it's a really good idea. I'll think about it and make proposals later in this PR.

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: I acknowledge that the request is kinda complex.

I have my problems with the suggested URL though, as I think `/transactions/unfulfilled` doesn't really follow the RESTful path scheme.
`unfulfilled` is an adjective, rather than a resource.
It describes the resource, but doesn't name it. For adjectives one would usually use query parameters.

I can't think of something better, but I'll try to.

docs/server/source/drivers-clients/http-client-server-api.rst
TimDaub: Consider that the paths you proposed yield a different semantic meaning:

- `/transactions`: I expect the outer most object to be a transaction
- `/assets`: I expect the outer most object to be an asset
- ...

- `GET /transactions/assets/{asset_id}?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}` has only two chars less than what was proposed
- `GET /assets/{asset_id}?fields=id,asset,operation&operation={GENESIS|CREATE|TRANSFER}` is significantly shorter than the proposal (15 chars), but doesn't carry the same semantic meaning. From interpreting only the path, I'd expect to: Return me an asset by its id and include only the fields `id`, `asset` and `operation` (while the asset doesn't have most of those fields)
- `GET /assets/{asset_id}?fields=id,asset,operation` same as above.

TimDaub:
> Assuming the client-provided timestamp gets removed from the Transaction data model (via pull request #817), clients will want some way to get the other timestamps associated with a transaction:

I've included these two requirements into the TODOs of this PR (original post).
Thanks.

docs/server/source/drivers-clients/http-client-server-api.rst
diminator: > Making it a noun could work:
> 
> - `/transactions/unspents`
> - `/transactions/unfulfilled_conditions`

maybe even back to 
- `GET /unspents`
- `GET /utxo`

> `/transactions/unfulfilled_conditions`
> `/conditions?fulfilled=false&owners_after={owners_after}`

I kinda like these (especially the first one), it speaks a more correct tech language

Not sure if it's a real problem when the aliases share the namespace of the items

TimDaub:
> Retrieval of specific fields in a resource
> 
> I followed this very popular guide:
> 
> Limiting which fields are returned by the API
> The API consumer doesn't always need the full representation of a resource. The ability select and chose returned fields goes a long way in letting the API consumer minimize network traffic and speed up their own usage of the API.
> Use a fields query parameter that takes a comma separated list of fields to include. For example, the following request would retrieve just enough information to display a sorted listing of open tickets:
> GET /tickets?fields=id,subject,customer_name,updated_at

See ticket #835. We cannot return partial payloads as long as we don't provide parts of a merkle tree for validation.
I'd propose, we remove the `?fields` query string for now and return only full transactions.
When we're more advanced, we can add partial payloads by including merkle trees or eventually just use IPLD out of the box.

OK: @sbellem @diminator?

vrde:
> I'd propose, we remove the ?fields query string for now and return only full transactions.

:+1:

sbellem:
> I'd propose, we remove the ?fields query string for now and return only full transactions.
> When we're more advanced, we can add partial payloads by including merkle trees in a response or eventually just use IPLD out of the box.

Yes, agreed, and for the time being we can suggest to those really interested in more powerful querying capabilities, to pipe the data into a data store they control and/or trust which can support the advanced querying needed.

In other words, we can suggest that BigchainDB can be part of an [ETL architecture](https://en.wikipedia.org/wiki/Extract,_transform,_load), as the initial source of trusted data. The rest of the pipeline is the hands of the user.

TimDaub:
Here is a fun little problem:

The whole purpose to have this query:

`/transactions?fulfilled=false&owners_after={owners_after}`

is to allow a client to figure out what unfulfilled conditions are left to fulfill for a certain public (`owners_after`) and respective private key.
Before, the combination of `?fields=conditions,txid` and `fulfilled=false`, would have taken care of filtering the list `transactions.conditions` to a subset of conditions that were actually not fulfilled yet.
In #830, we realized that partial payloads are harder to verify though, meaning an expansion of `conditions` as well as its filtering for unfulfilled conditions cannot happen.
As a result, we're now returning a full transaction payload.

Now the problem with this is that a transaction can have for example (or any other combination):

- 1 fulfilled condition; and
- N unfulfilled condition.

by requesting: `/transactions?fulfilled=false&owners_after={owners_after}`, we're hence not sending information about unfulfilled conditions anymore, meaning the request doesn't serve the use case anymore.

In #803 that exact problem was solved in a different way.
An endpoint `/unspents` was introduced that essentially returns the URL equivalent of a `TransactionLink`. From the standpoint of cryptographic-verification and RESTfulness, that endpoint was alright.

Now, it was however criticized for:

- introducing a new top-level endpoint (an intermediate table) (`/unspents`)
- using a language, we'd like to get rid of: `unspents`
- introducing overhead in that it would take two requests to get the same information (txid, cid, and the condition itself)

The fact is though, in order to fulfill a condition in a `TRANSFER` transaction, we need exactly three pieces of information that one or multiple endpoints will have to deliver:

- txid
- cid
- the condition itself

Therefore, I'd like to address the initial criticisms to reintroduce the endpoint again.
A few things up front:

- Essentially we're returning a `TransactionLink` resource (it's pretty hard to run away from that fact tbh)
- An adjective cannot be part of a resource name, but must always be a query parameter (e.g. `/unfulfilled_conditions` is embarrassingly unRESTful)
- A `TransactionLink` is not a condition, in no way does it fit under `/conditions`
- `/conditions?fulfilled=false&owners_after={owners_after}` isn't fulfilling the use case (there is no way of telling to which transaction the condition belongs + no cryptographic-verification can be provided)
- Essentially `TransactionLink` links two transactions, but de-facto it really points from a fulfillment to a condition (we don't use it in any other way)

A few proposals:

- Keep using `/unspents`
  - It's a known concept in the Blockchain world
  - It describes what you're actually getting
  - It's a noun
  - It cannot be retrieved by a unique key

- `/links`
  - it's a noun
  - It's non-"intuitive", meaning a concept that is specific to BigchainDB
  - It describes what you're getting

-  `/transactionlinks`
  - It's two words mashed together (all cases should be avoided in the path, no `_` or `L`s
  - It's specific to BigchainDB

- `/transactions/links`
  - It's not a real RESTful path
  - It has the same problems as `/transactionlinks`
  - It shares a namespace with `/transactions`, you're not getting what you're expecting

- `/outputs?fulfilled=false`
  - It's not specific to BigchainDB
  - It kinda describes what you're getting (even though the actual "output" it the condition itself - one could argue though that `output != condition` as an output is always the combination of: transaction + lock one needs to provide a key for)
  - it's not in line with the Interledger lingo


If you have better ideas, let me know. I'm currently leaning towards `/unspents` or `/outputs`.

ttmc:
Thoughts on how to name the endpoint to get links to all the unfulfilled conditions associated with a public key:

What we'd get back would be a set of TransactionLinks, i.e. a set of (txid, cid) pairs (maybe as URIs). Those really should be called ConditionLinks because they're links to specific conditions, not transactions. (I realize we have an internal class named TransactionLink but it could be renamed.)

The endpoint could be, for example:

`/conditionlinks?fulfilled=false&owners_after={owners_after}`

It could be seen as a "convenience endpoint" that exists to make life easier for users.

diminator:
> The endpoint could be, for example:
> 
> `/conditionlinks?fulfilled=false&owners_after={owners_after}`

Agreed, from a user point of view, you would be asking:
"which conditions have I yet to fulfill"
In bitcoin lingo it's the same as:
"which outputs can I spent - ie create an input for"
I would probably want to hide the "links" part as you could treat is as a Hyperlinked response

Hence I am pro:
`/conditions?fulfilled=false&owners_after={owners_after}`

ttmc:
An endpoint like:

`/conditions?fulfilled=false&owners_after={owners_after}`

suggests that you'll get back a list of condition objects, but those, by themselves, wouldn't say which transaction they're in (`txid`), or which condition in that transaction (`cid`). I think you need the `txid` and `cid` to build a fulfillment for a condition.

Also, as discussed in issue #835, it seems like we're leaning towards returning whole transactions, rather than just transaction parts (such as conditions). Returning a (`txid`, `cid`) pair isn't exactly a transaction "part" so much as a pointer to a full transaction and a hint where to look in that transaction for an unfulfilled condition.

diminator:
> suggests that you'll get back a list of condition objects, but those, by themselves, wouldn't say which transaction they're in (txid), or which condition in that transaction (cid). I think you need the txid and cid to build a fulfillment for a condition.

It actually gives you the locations of the conditions - as specified by `(txid, cid)`, which are the absolute paths for that condition. So I guess that's allright

TimDaub:
> Asset retrieval by ID
>  Metadata retrieval by ID

I just thought about this.
Initially, I thought there would be the two endpoints:

- `/assets/{asset_id}`
- `/metadata/{metadata_id}`

Now, both IDs in usage here are just uuidv4. When retrieving an asset or meta data, the user cannot check the payload, meaning the node could return something completely different then what was persisted. It's basically the same problem that was outlined in #835.

In noticed though, that with a little bit of query strings magic, these two endpoint's functionality can be used already:

- `/transactions?operation=CREATE&asset_id=1AAAbbb...ccc`: Returns the original in the `CREATE` transaction (+ it's verifiable). Consensus of BigchainDB ensures us that it will always only return ONE asset.
- `/transactions?metadata_id={metadata_id}` returns us a transaction including metadata of a certain id. Consensus ensures us, that there is always one of a kind persisted to the data.

As long as assets and metadata are not retrievable by their own hash therefore, we deliver them in the transaction itself. Implicitly, this means that for now there will be no `/assets` or `/metadata` endpoints returning ONLY the objects without the transaction.

TimDaub:
Here is another question:

Is there any value in allowing the user to retrieve blocks labeled as `INVALID`?

For transactions, we say that only valid (or undecided) transaction's can be retrieved via the API.
This is due to the fact that a:

- invalid transaction will never by included into a valid block
- that a valid transaction will eventually end up in a valid block.

For blocks though, this obviously doesn't happen.
Blocks can be invalid, still they're part of the chain. They'll never get discarded.

Therefore, I'd like to argue that, in contrast to the retrieval of transactions, invalid blocks can be retrieved using the HTTP API.

Any thoughts on that?

diminator:
> Is everybody OK with that? Feedback? Please tell me if you have a better solution :).

I like this

TimDaub:
> the server through websockets

Agree.

> the server through a /validate endpoint. The result is temporary - as the actual write occurs on a different time

:crying_cat_face: 
> REST stands for Representational State Transfer. (It is sometimes spelled "ReST".) It relies on a **stateless**, client-server, cacheable communications protocol.

In fact, I was able to address the `INVALID` block question quite nicely. I'll update it to the commits, once I push them.

diminator:
> REST stands for Representational State Transfer. (It is sometimes spelled "ReST".) It relies on a stateless, client-server, cacheable communications protocol.
> In fact, I was able to address the INVALID block question quite nicely. I'll update it to the commits, once I push them

Afraid that doesn't solve the problem of users wanting to:
- check if the transaction is valid before pushing to the network
- check why the transaction was invalid after pushing to the network

TimDaub:
> check if the transaction is valid before pushing to the network

It's out of the scope of this PR (so I'd prefer if we discuss this somewhere else if necessary to discuss this further), but checking the validity of a transaction before pushing it to the network is part of the client's task.

The only thing the server can add to the validation of a transaction is to check for DoubleSpents (assuming we'll get rid of the uuid's in metadata and assets - which I think will happen).

Now, there is an argument to be made about if the server should expose functionality to check for "spents", as we right now only expose the reverse ("unspents" (lol btw., since we don't like to use the "spents" metaphor I think "Double**Spents**" is not appropriate anymore :D).
I think it should have a `/spents`-esque endpoint. If you want I can add it to the TODOs.



> check why the transaction was invalid after pushing to the network

Websockets.

diminator:
> Now, there is an argument to be made about if the server should expose functionality to check for "spents",

What do you mean exactly here?
Getting every transaction that I ever spent?
Or getting the full spent history of an asset chain - starting at a certain transaction and going back in time?

TimDaub:
> the timestamp of the block containing the transaction

I was thinking about this a little bit.
I see two options here.

**Option 1:**

Outline in text how this is done doing a combination of calls on the API:

1. Call `/transactions/tx_id` to see if a transaction was included in a valid block
2. Call `/blocks?tx_id=tx_id` to see in which block the transaction was included
3. If a block was returned, use `block.timestamp` as the timestamp of the transaction.


This is already possible with the API that's being designed here, so it's just the question if we want to put this information-retrieval strategy into writting.


**Option 2:**

I looked into [Custom HTTP Entity Headers](https://www.safaribooksonline.com/library/view/restful-web-services/9780596809140/ch01s13.html).  Here is a definition:

> Request and Response messages MAY transfer an entity if not otherwise restricted by the request method or response status code. An entity consists of entity-header fields and an entity-body, although some responses will only include the entity-headers.

- https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html

This means, when doing a call to `/transactions/tx_id`, we could include a `X-BigchainDB-Timestamp` header into the response that would have been taken from the block.

The RESTful Cookbook gives useful advice:

> Use custom headers for informational purposes.

In fact, the timestamp for a transaction is only informal. Per default the `/transactions/tx_id` endpoint only returns valid transactions meaning a timestamp should always retrievable.


> Avoid using custom HTTP headers to change the behavior of HTTP methods

It doesn't change `GET /transactions/tx_id`'s behavior.

> If the information you are conveying through a custom HTTP header is important for the correct interpretation of the request or response, include that information in the body of the request or response or the URI used for the request. Avoid custom headers for such usages.

It's not. Any transaction returned is valid so any transaction has a timestamp. However a timestamp is not a vital part of a transaction. It can be completely ignored. In fact it can be ignored and Option 1. can be favored.


I'm leaning towards option 2. Still option 1 is and available and can be used if preferred.

diminator:
There are some suggestions from @roderik and possibly others:
- retrieve all asset (id's)
- list the entire history of an asset
- select or navigate in an asset: useful for selecting a specific event in a time series
- order the asset history: since we do not rely on timestamps, a counter or timestamp could be provided by the client

roderik:
Bonus points for 
- getting a list of assets with a json property matching x

diminator:
> I just realized that we should be thinking of owners_after more as like a parameter for a certain locking script that was persisted in a transaction.

Yes, this is part of the proposal to split up the attributes in an output script from the topology, the attributes are then good for indexing

diminator:
> select or navigate in an asset: useful for selecting a specific event in a time series

listener on asset, see #862

> select or navigate in an asset: useful for selecting a specific event in a time series

can be done by simple client recipe (once ordering and pagination is provided)

> getting a list of assets with a json property matching x

Custom indexing would be prohibitively complex, quoting @r-marques 

"queryability and indexing are two different things. You can query data without it being indexed by a particular key. For instance mongodb has a hard limit of 64 secondary indexes and if you are using more then 20 you are probably doing something wrong. The overhead would just be to much"

diminator:
> Consider changes made in #856 (remove metadata.id)

`metadata.id` is removed, hence there is no specific ID to look for `metadata` besides the `tx.id`

Proposal to remove this endpoint? @TimDaub @libscott  @ttmc

TimDaub:
> Proposal to remove this endpoint? @TimDaub @libscott @ttmc

:+1:

ttmc:
I just got around to looking at the latest version of this proposal. Some comments:

* For the `GET /transactions` endpoint, there's a proposal for a query parameter named `fulfilled`. Should that be changed to `spent`? I'm a bit confused about how our vocabulary has settled for version 0.9. Do "outputs get spent" or "conditions get fulfilled" or something else? @libscott can you comment?

* There are a few places where the plural of transactions is written as "transaction's" (with an apostrophe) rather than "transactions", e.g. "A flag to indicate if transaction’s with fulfilled conditions should be returned."

* Using the `HTTP 303 See Other` response for the status of a valid transaction/block seemed a bit odd at first, but on closer inspection, seems to be a good use of that standard.

I noticed some minor formatting and stylistic things that I would change, but it is not yet time to be bothering with such things.

ttmc:
The note at the top of the HTTP API docs page ("The HTTP client-server API is currently quite rudimentary...") can maybe be removed now.

diminator:
> getting a list of assets with a json property matching x

some steps for moving this forward:
- provide an endpoint to get all assets: ie `GET /transactions?operation=CREATE&is_asset=true`, added the `is_asset` flag and one could narrow it further down by specifying `public_keys`
- provide pagination (and maybe ordering, but that can be done client side)

suggestions @TimDaub ?

diminator:
When returning a list of objects, do we return the full object or a url path to limit size? Imagine returning a list of blocks, blocks can be really big datastructures (with 1000+ transactions)... so that wouldnt be feasible to send in one go over the wire

diminator:
> Imagine returning a list of blocks, blocks can be really big datastructures (with 1000+ transactions)... so that wouldnt be feasible to send in one go over the wire

This relates to #835 
For lists of URIs it would be possible to verify each object behind the URI separately, it becomes more tedious to verify that each URI belongs to the list...

comments? @TimDaub @ttmc @libscott @sbellem

diminator:
Some notes:
- Use `Location` in header or body? or both?
- status code `303` might automatically redirect on the client, never showing the original response
- does it make sense to have a single `statuses` for multiple objects?
- include more info in the `status`: eg object type and related reference (for a tx, give the `block_id` if `VALID`
- `/votes/block_id` isnt restful (using queryparams then yes). Alternative: `/votes/vote_id` is and `/blocks/block_id/votes` is restful.
- asset list retrieval `GET /transactions?operation=create&is_asset=true`
- Put everything in the response body and not only in the header. We'd like this to be a JSON API

libscott:
> Put everything in the response body and not only in the header. We'd like this to be a JSON API

API fully expressed with JSON :D but with HTTP protocol on top.

diminator:
> I wonder why it won't return the transaction if it is in the backlog table (but not in a VALID or UNDECIDED block). That's what the get_transaction() method in the Bigchain class does. See https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/core.py

This is interesting and I like to revive it. the status returned could be simply `BACKLOG`. It's good to know that your transaction is in the pool of unconfirmed transactions.

diminator:
> This means, when doing a call to /transactions/tx_id, we could include a X-BigchainDB-Timestamp header into the response that would have been taken from the block.

After review, the `X-BigchainDB-Timestamp` seemed a bit fuzzy. What does it mean?
For timestamps on transactions the user would care about a few things:
- the first time the system sees the transaction (in the `backlog` or upon the first inclusion in an `undecided` block)
- when the containing block was voted `valid`
- maybe when included in every subsequent (possibly multiple `invalid` but only one `valid`) blocks

One proposal is to move this information to the `/statuses/{tx_id}` endpoint:

```
{
  "status": "<valid|invalid|undecided>",
  "_links": [
    {
      "block": "/block/<block_id_invalid>",
       "status": "/statuses/<block_id_invalid>"
    },
    {
      "block": "/block/<block_id_valid>",
       "status": "/statuses/<block_id_valid>"
    },
    {
      "vote": "/votes/<vote_1>"
    },
    {
      "vote": "/votes/<vote_2>",
    },
  ]
}
```

diminator:
We agreed to go with the language of "unspent outputs" rather than "unfulfilled outputs"

A transaction can have many outputs, and it's possible that some are spent but not others. (They must all be for the same asset --- they can have different amounts and conditions.) That is, it's possible for a transaction to be partially "spent". How shall we handle the intermediate cases between all outputs unspent, and all outputs spent?

diminator:
Remove all mention of GENESIS blocks from the HTTP API.

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -1,58 +1,226 @@
...
 
-If you send an HTTP GET request to the API Root URL
-e.g. ``http://localhost:9984`` 
-or ``http://apihosting4u.net:9984``
-(with no ``/api/v1/`` on the end), 
-then you should get an HTTP response 
+If you send an HTTP GET request to the BigchainDB Root URL
+e.g. ``http://localhost:9984``
+or ``https://example.com:9984``
+(with no ``/api/v0.9/`` on the end),
```
sohkai: Should be `/api/v1/`?

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -1,58 +1,226 @@
...
+        "api_v1": "http://example.com:9984/api/v1/"
+      },
       "keyring": [
         "6qHyZew94NMmUTYyHnkZsB8cxJYuRNEiEpXHe1ih9QX3",
         "AdDuyrTyjrDt935YnFu4VBCVDhHtY2Y6rcy7x2TFeiRi"
       ],
       "public_key": "AiygKSRhZWTxxYT4AfgKoTG4TZAoPsWoEt6C6bLq4jJR",
       "software": "BigchainDB",
-      "version": "0.6.0"
+      "version": "0.9.0",
```
sohkai: Trailing commas aren't supported in JSON

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -64,110 +232,187 @@ POST /transactions/
...
+   :statuscode 202: The pushed transaction was accepted in the ``BACKLOG``, but the processing has not been completed.
+   :statuscode 400: The transaction was malformed and not accepted in the ``BACKLOG``.
 
 
-GET /transactions/{tx_id}/status
+Statuses
 --------------------------------
 
-.. http:get:: /transactions/{tx_id}/status
+.. http:get:: /statuses?tx_id={tx_id}|block_id={block_id}
```
sohkai: I think the pipe (`|`) is confusing here; maybe it'd be nicer just to have two examples?

E.g.:

```
http:get:: /statuses?tx_id={tx_id}
http:get:: /statuses?block_id={block_id}
```

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -64,110 +232,187 @@ POST /transactions/
...
+
+
+
+Blocks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. http:get:: /blocks/{block_id}
+
+   Get the block with the ID ``block_id``. Any blocks, be they ``VALID``, ``UNDECIDED`` or ``INVALID`` will be
+   returned. To check a block's status independently, use the `Statuses endpoint <#get--statuses?tx_id=tx_id|block_id=block_id>`_.
```
sohkai: (If changed the status endpoint above, also change this)

diminator:
> I can make the changes too.

+1

sohkai:
Noticed in our implementation that whenever we return a `404`, we also return a json payload with

```json
{
  "message": "Not found",
  "status": 404
}
```

AFAIK the spec doesn't really define this, and seems somewhat redundant. Did we have a discussion about this at some point?

cc @sbellem.

diminator:
> AFAIK the spec doesn't really define this, and seems somewhat redundant. Did we have a discussion about this at some point?

Good point

See more granular messages in the `GET /votes` tests
https://github.com/bigchaindb/bigchaindb/pull/1031/files#diff-e8747065feb7bc8df82da2650da680f4R132

We should indeed adapt something uniform and document it

docs/server/source/drivers-clients/http-client-server-api.rst
sohkai: On further examination, this is kind of messy right now so we'll review it later.

sohkai: Done in https://github.com/bigchaindb/bigchaindb/pull/830/commits/fc06b6348dca1f69246dd7f2445a5cc94e48e54f

sohkai:
I replaced all usages of the `:param` domain with `:query` in https://github.com/bigchaindb/bigchaindb/pull/830/commits/1007bf9b582dfd8e3c863b7867b4e05f48328234 (was there a reason `:param` was used instead @diminator?).

diminator:
> I replaced all usages of the :param domain with :query in 1007bf9 (was there a reason :param was used instead @diminator?).

not sure, I think this was introduced by @TimDaub

sohkai:
Turns out I didn't realize they were actually url parameters (and not query strings). I've removed 1007bf9.

libscott:
> For the GET /transactions endpoint, there's a proposal for a query parameter named fulfilled. Should that be changed to spent? I'm a bit confused about how our vocabulary has settled for version 0.9. Do "outputs get spent" or "conditions get fulfilled" or something else? @libscott can you comment?

Yes, this endpoint doesn't make any sense so it's going to be `/outputs?unspent=...&public_key=...`.

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -1,58 +1,226 @@
...
+
+   **Example response**:
+
+   .. literalinclude:: samples/get-tx-id-response.http
+      :language: http
+
+   :resheader Content-Type: ``application/json``
+
+   :statuscode 200: A transaction with that ID was found.
+   :statuscode 404: A transaction with that ID was not found.
```
sbellem: Following #1038, we need to add a query string `'status'`, such that by default only VALID transactions are queried. If a consumer wishes to query transactions that may be in the BACKLOG, or in an UNDECIDED block, a `'status'` query parameter must be provided, e.g.:

```text
GET /transaction/{id}?status=backlog
GET /transaction/{id}?status=undecided
GET /transaction/{id}?status=valid
```

the last case being equivalent to the default, when no status is provided.

@diminator @libscott @TimDaub could you please confirm?

libscott: Yep, thats great.

TimDaub:
Yes, great!

docs/server/source/drivers-clients/http-client-server-api.rst
sbellem: thanks, oh, yes, just made the change, thanks

libscott:
After discussions with @TimDaub and @sbellem we are merging the spec as is, currently the feature delta is to return http 202 on POST of a transaction, I am working on this now. Great job everyone!

Issue #836, opened 2016-11-17
Remove hashlock support from models


TimDaub:
As was pointed out in https://github.com/bigchaindb/bigchaindb/pull/798/files#r88443521 `Condition` currently doesn't enforce `details` in its de(serialization) methods.

[This was introduced when wanting to support hashlock conditions](https://github.com/bigchaindb/bigchaindb/commit/66f7ab6ea4f398a4e05dfcf3932359eec02c6926) (which we do not anymore for the time being)

The goal of this ticket is to remove all left over parts from the hashlock support.

PR #837, opened 2016-11-17
Added notes about support for hashlock & timeout conditions

ttmc:
A minor update to the docs to keep them current.

### DIFF
--- a/docs/root/source/data-models/crypto-conditions.md
+++ b/docs/root/source/data-models/crypto-conditions.md
@@ -2,9 +2,11 @@
 
 To create a transaction that transfers an asset to new owners, one must fulfill the asset’s current conditions (crypto-conditions). The most basic kinds of conditions are:
 
-* **A hashlock condition:** One can fulfill a hashlock condition by providing the correct “preimage” (similar to a password or secret phrase)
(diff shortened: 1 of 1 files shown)


TimDaub:
What's being updated here is not correct.

All conditions and fulfillments provided by bigchaindb/cryptoconditions are fully supported by both the core as well as the Transaction model.

How?

Create a `Transaction()`.
Use `Transaction.add_condition` and `Transaction.add_fulfillment`.
Sign. Serialize, Push it to the server.

The only thing that's currently missing is documentation.

ttmc:
Okay, I'll close this PR. I'm pretty sure Sylvain will be documenting some of how to use the Python driver to work with hashlock and timeout conditions, because some of the Advanced Usage examples involve those.

sohkai:
@TimDaub but https://github.com/bigchaindb/bigchaindb/issues/836 was made to remove support for hashlocks? If we've already gone down this road, I think we should be providing some warning signs.

Also, I'm not sure anybody can actually use `add_condition` or `add_fulfillment` since the driver (AFAIK) doesn't expose the transaction model in a way that someone could use it like that.

TimDaub:
That's the problem of the driver (or the driver's docs) not the node's documentation.

PR #851, opened 2016-11-22
[Discuss] another asset id strategy

libscott:
This is a demo of another strategy for asset IDs. Below I will outline the pros and cons of different approaches.

### Transaction ID

**+** No need for extra database call in validation of CREATE tx
**+** No need for index of asset IDs
**+** Client can create valid transactions
**+** Client cannot falsify asset ID
**-** Confusing nomenclature & API (asset_id isn't quite right.... asset_txid?)
**-** Not clear how to do refills (guess we could add an edge case to override the asset ID on creation?)
**-** Can create duplicate assets (per pubkey)

### hash(fulfillment_pubkeys + asset_data)

**+** Client can create valid transactions
**+** Client cannot falsify asset ID
**+** Clear nomenclature & API (tx.asset.id)
**+** Clear how to do refills (Create another TX with same pubkeys and asset data)
**+** Can't create duplicate assets (per pubkey)
**-** Need extra DB index and DB validation step

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -409,8 +409,8 @@ class Asset(object):
     def __init__(self, data=None, data_id=None, divisible=False,
                  updatable=False, refillable=False):
         """An Asset is not required to contain any extra data from outside."""
+        self.data_id = data_id
         self.data = data
-        self.data_id = data_id if data_id is not None else self.to_hash()
         self.divisible = divisible
         self.updatable = updatable
         self.refillable = refillable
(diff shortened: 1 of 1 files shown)


ttmc:
This is how I understand what you mean when you write, "Can't create duplicate assets (per pubkey)":

If the transaction ID is used as the asset ID, then a given pubkey can create many copies of the same asset, in different transactions, simply by changing the transaction metadata (leaving the asset data identical).

If the pubkeys + asset data are used to generate the asset id, then the same pubkey can't generate a new asset with the same asset data, because that asset would have the exact same asset id as an already-existing asset.

Maybe that is okay? If the owner of the pubkey desperately wants to create a new identical asset, then he or she can just generate a new keypair and use the new public key. The new public key should be announced, lest others think some random Joe has started minting fake tickets to the Giant Orange Treehouse.

libscott:
> If the transaction ID is used as the asset ID, then a given pubkey can create many copies of the same asset, in different transactions, simply by changing the transaction metadata (leaving the asset data identical).

Well, in that case, the asset body would be the same, and the creators would be the same, but the asset ID would be different, since the server won't accept multiple transactions with the same ID.

> If the pubkeys + asset data are used to generate the asset id, then the same pubkey can't generate a new asset with the same asset data, because that asset would have the exact same asset id as an already-existing asset.

Correct. Except in the case that that asset is `refillable` in which case the server could allow it.

> Giant Orange Treehouse.

:heart_eyes:

libscott:
Closed in favor of using asset id. Maybe we get to this, but for now Asset ID is good enough and we're not sure what the requirements of Asset are going to be in the future.

PR #855, opened 2016-11-23
Flat transaction

### LINKED ISSUE #828: Revisit "inner" transaction in transaction model
Having an "inner" transaction inside the transaction model seems to be confusing and redundant:

```
tx = {
    'id': '…',
    'transaction': {
        …
    }
    …
}
```

From talks with @ttmc, @sbellem, @libscott.

libscott:
Implementation of #828. ~~This hasn't been generally approved yet.~~ This has been generally approved.

Still to do, if/when given green light:

- [x] Docs changes

And separately: 

- Corresponding PR in bigchaindb-driver
- Ditto bdb-transaction-cli

Fixes #828

### DIFF
--- a/bigchaindb/db/backends/rethinkdb.py
+++ b/bigchaindb/db/backends/rethinkdb.py
@@ -178,7 +178,8 @@ class RethinkDBBackend:
              .filter(lambda transaction:
                      transaction['asset']['id'] == asset_id)
              .filter(lambda transaction:
-                     transaction['operation'] == 'CREATE'))
+                     transaction['operation'] == 'CREATE')
+             .pluck('asset'))
 
     def get_spent(self, transaction_id, condition_id):
         """Check if a `txid` was already used as an input.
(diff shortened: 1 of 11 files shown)


bigchaindb/db/backends/rethinkdb.py
```diff
@@ -202,10 +202,9 @@ def get_asset_by_id(self, asset_id):
              .get_all(asset_id, index='asset_id')
              .concat_map(lambda block: block['block']['transactions'])
              .filter(lambda transaction:
-                     transaction['transaction']['asset']['id'] == asset_id)
+                     transaction['asset']['id'] == asset_id)
              .filter(lambda transaction:
-                     transaction['transaction']['operation'] == 'CREATE')
-             .pluck({'transaction': 'asset'}))
```
r-marques: Why was this removed?

libscott: Hmm, looks odd indeed! Will investigate...

r-marques: everything will work the same with or without pluck because the structure of the document is maintained. The only difference is that all other keys are empty

PR #856, opened 2016-11-23
Remove metadata uuid

### LINKED ISSUE #852: Is Metadata.id neccesary?
Do we need transaction metadata id? Is there a use case for it that can't be served by using `[txid].metadata`?

libscott:
This is the implementation of https://github.com/bigchaindb/bigchaindb/issues/852 and has the following changes:

* No more `tx.metadata.data`, now the user dict is `tx.metadata`.
* Remove `bigchaindb.common.transaction.Metadata` class.
* Remove metadata DB index.
* Queries to get metadata go: `bigchaindb.get_transaction(tx.id).metadata` rather than `bigchaindb.get_metadata(tx.metadata.id)`.

Question:

Should `metadata` be a nullable field? Arguably, it should default to an empty dict, and not be nullable, to remove the (somewhat meaningless) distinction between `null` and `{}`.

Fixes #852.


TimDaub:
Since metadata to transaction is 1:1, I agree that:

- `bigchaindb.get_transaction(tx.id).metadata` is better than
- `bigchaindb.get_metadata(tx.metadata.id)`

sohkai:
> Should metadata be a nullable field? Arguably, it should default to an empty dict, and not be nullable, to remove the (somewhat meaningless) distinction between `null` and `{}`.

The `asset.data` may be nullable right now, so perhaps these two should be considered together. There's a lot of really small trade-offs on going either way, but the small distinction between the two:

- `null`: No metadata provided at all
- `{}`: Metadata provided, but it's empty.

Makes allowing for both plausible to me.

libscott:
> null: No metadata provided at all
> {}: Metadata provided, but it's empty.

My question was if this is useful at all. Removing the ambiguity now may save someone a headache in the future.

r-marques:
:+1: assuming that we go with `null` for the empty field. This seems to be the behavior for fields that can be empty `asset.data`, `input`, ...

If we feel that we should change this behavior to empty dicts or empty lists then this behavior should be done to all data models in a different pull request

ttmc:
@libscott Please hold off on merging this pull request until after we've released version 0.8.0 (which should be soon), because after merging this, changes will have to be made to the Python driver docs, at least, if nothing else.

PR #858, opened 2016-11-23
Bug/787/secondary index on outputs

### LINKED ISSUE #105: Consider using secondary indexes in some queries

### LINKED ISSUE #787: Performance of get_owned_ids
Extract from Gitter:

> **Roderik van der Veer @roderik Nov 02 20:29**:
> Is anyone experiencing very slow “get_owned_ids”? I've got 5 addresses with combined 80 assets in them, and when i do get_owned_ids on each of them (concurrently) i still end up with about 4.5 seconds to get all the results back. This would totally suck if i put in 100k assets 
I’m still on 0.6 btw

> **Rodolphe Marques @r-marques 09:22**:
> @roderik can you open a github issue. We should consider creating a secondary index for this query

r-marques:
resolves #105 
resolves #787 (partially?)

#### Changes
Added secondary index on `owners_after`: used by `get_unspents`.
Added secondary index on `input`: used by `get_spent`.
Renamed `get_owned_ids` to `get_unspents` to follow the language agreed on the api. `get_unspents` now accepts a list of keys instead of a single key. This way we can check if a subset of keys exists in `owners_after` instead of only one key. The logic was also simplified
Example:
```python
{
    'id': 'txid1',
    'conditions': [
        {
            'cid': 0,
            'owners_after': [user1]
        },
        {
            'cid': 1,
            'owners_after': [user1, user2]
        },
        {
            'cid': 2,
            'owners_after': [user1, user2, user3]
        }
    ]
}

b.get_unspents([user1]) # this returns cid 0, 1 and 2
b.get_unspents([user2]) # this returns cid 1 and 2
b.get_unspents([user3]) # this returns cid 2
b.get_unspents([user1, user2]) # this returns cid 1 and 2
b.get_unspents([user2, user3]) # this returns cid 2
```

#### Benchmarks
This measures the performance of `get_owned_ids` and `get_unspents` on a set of 25k, 50k, 75k and 100k assets randomly assigned to a public key in a 500 key pool.
Also each particular asset may be assigned to between 1 and 3 users also chosen randomly.

This tests were done on a single node `m3.2xlarge` ec2 instance running this script https://gist.github.com/r-marques/123e4f788b946f720262e0cc793b1bc1

| # assets | v0.7.0 (db) | v0.7.0 (bdb) | this brach (db) | this branch (bdb) |
| --- | --- | --- | --- | --- |
| 25k | 0.177 | 18.029 | 0.032 | 15.453 |
| 50k | 0.206 | 50.192 | 0.026 | 26.844 |
| 75k | 0.231 | 92.609 | 0.039 | 28.462 |
| 100k | 0.239 | 150.835 | 0.023 | 42.488 |

The values are the average call time in seconds for the database query (**db**) and the bigchaindb method (**bdb**)

The secondary indexes speed up the database queries about 10x but most of the time is being spent on the bigchaindb side to make sure that all returned unspents are not in invalid blocks.
Although the changes make the call about 2-3x faster we need to revisit the election code to speed this up.

cc @roderik


roderik:
Would creating these indexes by hand also speed up a 0.6 bigchaindb? Need some speed but don't want to rewrite a lot of the integrations at this time.

sohkai:
> Renamed `get_owned_ids` to `get_unspents` to follow the language agreed on the api.

This really should've been a different PR.

r-marques:
@roderik hard to say. As you can see the database query are still pretty fast. The biggest problem is with the rest of the logic to make sure the returned unspents are not in invalid blocks.

You can try. They should work as is. I don't think there was any change in the data models since version 0.6.0 that would break the secondary indexes

diminator:
@r-marques LGTM (besides the conflicts)

r-marques:
Closing this in favour of #1125

PR #860, opened 2016-11-23
refactor database calls out of block

rhsimplex:
Removes an explicit database call from the `block` module.

Actually, this code probably can be absorbed into the stale transaction monitor code.  It's kind of useless that a node can come back and pick up its misplaced transactions when `stale_transaction_monitor` would reassign them shortly anyway.  Also costs us an index (maybe that's needed elsewhere anyway).

But that's another PR!

### DIFF
--- a/bigchaindb/db/backends/rethinkdb.py
+++ b/bigchaindb/db/backends/rethinkdb.py
@@ -90,7 +90,7 @@ class RethinkDBBackend:
                 r.table('backlog')
                 .filter(lambda tx: time() - tx['assignment_timestamp'] > reassign_delay))
 
-    def get_old_transactions(self, node):
+    def get_old_transactions(self, node_pubkey):
         """Returns the oldest transactions.
 
         Old transactions are not necessarily stale transactions.  If a node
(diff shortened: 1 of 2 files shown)


libscott:
Code looks good, but I wonder if there's an opportunity to make it clearer what's going on. The transactions are described as "old", but the database query appears to select all transactions assigned to the node. So what is the qualification of "old" in this case?

vrde:
> Code looks good, but I wonder if there's an opportunity to make it clearer what's going on. The transactions are described as "old", but the database query appears to select all transactions assigned to the node. So what is the qualification of "old" in this case? (this may be more for my edification than the PR!)

When a transaction is pushed to a federation it is randomly assigned to a node. If the node is down it won't get the transaction from its changefeed. A solution to this is to query the federation on startup, asking for all the transactions that are in the backlog and that are assigned to that specific node. But @rhsimplex remark is interesting →

> Actually, this code probably can be absorbed into the stale transaction monitor code. It's kind of useless that a node can come back and pick up its misplaced transactions when stale_transaction_monitor would reassign them shortly anyway. Also costs us an index (maybe that's needed elsewhere anyway).

This is actually true, and it would simplify the overall architecture. I'm wondering if we can just close this PR and create a new one to remove the *prefeed*, so we have less back and forth.

r-marques:
>  It's kind of useless that a node can come back and pick up its misplaced transactions when stale_transaction_monitor would reassign them shortly anyway

I agree.
In that case this pr is not needed and we need one to actually remove this functionality.
We also need to make sure that we test this scenario

rhsimplex:
I'll close and open a new issue

PR #861, opened 2016-11-23
chore: fix printing out transactions on block creation

diminator:
instead of printing the list `block.transactions` print `len(block.transactions)`

FYI: this edit was done in the browser ;-)


r-marques:
:+1:

sohkai:
> FYI: this edit was done in the browser ;-)

@diminator You're 1/1 for proposing awesome changes with your new IDE. Maybe you should write all your code in the browser ;)

Issue #217, opened 2016-04-23
Remove "invalid_reason" from the vote model?


ttmc:
The `vote` model (documented at http://bigchaindb.readthedocs.org/en/master/models.html#the-vote-model ) for the vote-on-a-block includes an "invalid_reason":

``` javascript
"invalid_reason": "<None|DOUBLE_SPEND|TRANSACTIONS_HASH_MISMATCH|NODES_PUBKEYS_MISMATCH",
```

but a block may contain several invalid transactions, each one invalid for a different reason, therefore it doesn't make sense to give only one reason for voting that a block is invalid.

A node stops looking at the transactions in a block once it encounters the first invalid transaction, so it doesn't even know if there are other invalid transactions in the block, or why they are invalid.

Alternatives to removing "invalid_reason":
- rename it "first_invalid_reason"
- replace it with "first_invalid_transaction_id" (but don't store the reason; anyone else can check the transaction to figure out the reason)

I prefer the latter option: replacing "invalid_reason" with "first_invalid_transaction".

r-marques:
I also like the option of of `first_invalid_transaction`.
Another solution would be to just removed that field for now since we are not using it.

diminator:
or one could provide a `set` or `dict` of reasons is there are multiple?

could be:
```
"invalid_reason": {
    "DOUBLE_SPEND": <count>/[<txid>], 
    "TRANSACTIONS_HASH_MISMATCH": <count>/[<txid>],
    ...
}
```

r-marques:
Right now the transaction validation raises an exception at the first invalid condition so we don't have multiple reasons (this also increases performance). I guess that the txid would be enough for other nodes to just run he validator and find out what was wrong with the transaction.

r-marques:
Since @libscott has been working on simplifying the models and adding schema validation now would be a good time to just remove it.
We can re-add it later if we have a use for it

ttmc:
BigchainDB 2.0 works differently from BigchainDB 1.3 and earlier.

Invalid blocks are no longer a thing in BigchainDB 2.0. There is no need for a reason to be given for a block being invalid, since there is no block being invalid.

isaacwalkercox:
> or one could provide a `set` or `dict` of reasons is there are multiple?
> 
> could be:
> 
> ```
> "invalid_reason": {
>     "DOUBLE_SPEND": <count>/[<txid>], 
[...]

PR #865, opened 2016-11-24
Vote schema

### LINKED ISSUE #447: Vote schema validation
as #275 but for votes. Although, votes can only be written by other nodes, they should be rejected if the structure is incorrect.

libscott:
Fixes #447.

- [x] Complete vote schema .yaml file
- [x] Validate votes in during `BigchainDB.block_election_status`
- [x] Logs validation failures during block election
- [x] Generate schema reference documentation
- [x] Fix all broken tests
- [x] New tests for valid and invalid vote
- [x] Test .yaml sanity (via `additionalProperties`)
- [x] Remove descriptions from yaml schemas loaded for validation to shorten error messages

screenshot from 2016-11-24 16-58-08


bigchaindb/common/schema/vote.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - invalid_reason
+    - is_block_valid
+    - previous_block
+    - voting_for_block
+    - timestamp
+    properties:
+      previous_block:
+        "$ref": "#/definitions/sha3_hexdigest"
+        description: |
+          Sha3 identifier of the block that preceeds the block being voted on.
```
ttmc: SHA3 is an acronym for Secure Hash Algoritm 3. It can be written in many ways, e.g. SHA3, SHA-3, maybe sha3 (e.g. in code), but Sha3 seems wrong. There are other instances but I'll just note this one and let you grep the others.

bigchaindb/common/schema/vote.yaml
```diff
@@ -0,0 +1,69 @@
...
+  node_pubkey:
+    type: "string"
+    pattern: "[1-9a-zA-Z^OIl]{43,44}"
+    description: |
+        Ed25519 public key identifying the voting node.
+  signature:
+    type: "string"
+    pattern: "[1-9a-zA-Z^OIl]{86,88}"
+    description:
+        Ed25519 signature of the ``vote`` object.
```
ttmc: This sentence doesn't make it clear which object is the "vote object." Is it the outer one or the inner one? I think it's the inner one, the one you also call the "Vote Details".

bigchaindb/common/schema/vote.yaml
ttmc: In this case, I think it would be okay to say it's "The id of the block that the node considers to be the block right before the block being voted on. In principle, another node might think a different block is the previous block."

bigchaindb/common/schema/vote.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - is_block_valid
+    - previous_block
+    - voting_for_block
+    - timestamp
+    properties:
+      previous_block:
+        "$ref": "#/definitions/sha3_hexdigest"
+        description: |
+          Sha3 identifier of the block that preceeds the block being voted on.
+          The notion of a "previous" block is subject to vote.
```
ttmc: This isn't relevant. The value of this field is _not_ determined by voting. It's determined solely by one node.

ttmc:
This pull request calls into question the need for a separate page like https://docs.bigchaindb.com/projects/server/en/latest/data-models/vote-model.html

The only extra information on that page is the line:

> Note: The `invalid_reason` was not being used and may be dropped in a future version of BigchainDB. See [Issue #217](https://github.com/bigchaindb/bigchaindb/issues/217) on GitHub.

That line could be moved inside the new **Vote Schema** page and **The Vote Model** page could be deleted.

bigchaindb/common/schema/vote.yaml
```diff
@@ -0,0 +1,69 @@
...
+      voting_for_block:
+        "$ref": "#/definitions/sha3_hexdigest"
+        description: |
+          Sha3 identifier of the block being voted on.
```
libscott: I kinda of light keeping "SHA3" here, it doesn't occupy much space and it's informative.

bigchaindb/common/schema/__init__.py
```diff
@@ -7,18 +7,47 @@
...
+        for val in node.values():
+            drop_schema_descriptions(val)
+
+
+def _load_schema(name):
+    """ Load a schema from disk """
+    path = os.path.join(os.path.dirname(__file__), name + '.yaml')
+    with open(path) as handle:
+        schema = yaml.safe_load(handle)
+        drop_schema_descriptions(schema)
```
libscott: Deindent

docs/server/source/schema/vote.rst
```diff
@@ -0,0 +1,131 @@
...
+
+`Vote Details`_ to be signed.
+
+
+Vote.previous_block
+^^^^^^^^^^^^^^^^^^^
+
+**type:** string
+
+SHA3 identifier of the block that preceeds the block being voted on.
```
ttmc: If you want to keep "SHA3" in the description, then that's okay, but you should make it clear that the "SHA3 identifier of the block" is the same thing as the `id` of the block. Also, I think it's important to note that this is the opinion of the voting node, not some absolute truth. How about:

`id` (SHA3 hash) of the block that precedes the block being voted on (as determined by the voting node).

(I also fixed the spelling of "precedes".)

I have similar comments for the `Vote.voting_for_block` below, i.e.

`id` (SHA3 hash) of the block being voted on.

rhsimplex:
Just an architectural question.  In [consensus.py](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/consensus.py) we use class methods of the `Block` and `Transaction` classes for validation.  

Is there a reason `vote` doesn't follow this pattern?  Specifically why the vote model isn't defined like transaction and block?

rhsimplex:
Other than that, :+1: from me

libscott:
> Is there a reason vote doesn't follow this pattern? Specifically why the vote model isn't defined like transaction and block?

Yea, I think that the design isn't completely clean and could be improved. `consensus.py` in particular illustrates this, it does next to nothing.

bigchaindb/common/schema/vote.yaml
```diff
@@ -0,0 +1,73 @@
...
+          This field is ``true`` if the block was deemed valid by the node.
+      invalid_reason:
+        anyOf:
+        - type: "string"
+          description: |
+            Reason the block is voted invalid, or ``null``.
+
+            .. container:: notice
+
+              **Note**: The invalid_reason was not being used and may be dropped in a future version of BigchainDB. See Issue `#217 <https://github.com/bigchaindb/bigchaindb/issues/217>`_ on GitHub.
```
r-marques: I am ok with this

docs/server/source/data-models/index.rst
```diff
@@ -16,4 +16,4 @@ This section unpacks each one in turn.
    asset-model
    crypto-conditions
    block-model
-   vote-model
+   ../schema/vote
```
r-marques: This is the same as https://github.com/bigchaindb/bigchaindb/pull/865/files#diff-4098e6de490b39b8141bf1abd2b474b6R19 so you end up with the same documentation inside the `data-models` and the root of the server docs.

I prefer what you did with the transaction schema. The description page still exists as before but now the key names are links to the schema.

tests/test_consensus.py
```diff
@@ -1,2 +1,37 @@
-class TestBaseConsensusRules(object):
-    pass
+from bigchaindb.consensus import BaseConsensusRules
```
r-marques: I don't know if this was agreed by the team but we have been following the pattern in which we import test dependencies inside the tests instead of at the the top of the file.

http://docs.pylonsproject.org/en/latest/community/testing.html#rule-never-import-the-module-under-test-at-test-module-scope

docs/server/source/data-models/index.rst
libscott: Troy asked me to replace the `vote-model` page with that link, let's offline it

bigchaindb/common/schema/vote.yaml
libscott: Cool maybe add your voice to the issue #217 then

bigchaindb/common/schema/__init__.py
```diff
@@ -7,18 +7,46 @@
...
-__all__ = ['TX_SCHEMA', 'TX_SCHEMA_YAML', 'validate_transaction_schema']
+TX_SCHEMA_PATH, TX_SCHEMA = _load_schema('transaction')
+
+
+def validate_transaction_schema(tx):
+    """ Validate a transaction dict """
+    _validate_schema(TX_SCHEMA, tx)
+
+
+VOTE_SCHEMA_PATH, VOTE_SCHEMA = _load_schema('vote')
```
sohkai: Might be nicer to group all the constants together

bigchaindb/common/schema/__init__.py
```diff
@@ -7,18 +7,46 @@
...
+
+
+def validate_vote_schema(vote):
+    """ Validate a vote dict """
+    _validate_schema(VOTE_SCHEMA, vote)
```
sohkai: This might file be big enough now that it's worth taking out of the `__init__.py`, but `¯\_(ツ)_/¯`

bigchaindb/consensus.py
```diff
@@ -1,4 +1,11 @@
+import logging
+
 from bigchaindb.util import verify_vote_signature
+from bigchaindb.common.schema import SchemaValidationError, \
+    validate_vote_schema
```
sohkai: Usually we prefer:

```python
from ... import (
    a,
    b,
    ...
)
```

(See [our style guide](https://github.com/bigchaindb/bigchaindb/blob/master/PYTHON_STYLE_GUIDE.md#how-to-format-long-import-statements))

docs/server/generate_schema_documentation.py
```diff
@@ -77,10 +90,60 @@
 --------
 
 %(metadata)s
-"""
+""" + TPL_STYLES
```
sohkai: I think this might be more obvious if it were `TPL_TRANSACTION = TPL_STYLES + """...` rather than the other way around.

I was really confused what `""" + TPL_STYLE` was doing for a while 😌 .

docs/server/generate_schema_documentation.py
```diff
@@ -91,13 +154,26 @@ def construct_mapping(loader, node):
...
         yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
         construct_mapping)
-    return yaml.load(stream, OrderedLoader)
+    with open(path) as handle:
+        return yaml.load(handle, OrderedLoader)
 
 
-TX_SCHEMA = ordered_load_yaml(TX_SCHEMA_YAML)
+def load_schema(path):
+    global DEFS
```
sohkai: I'm k with this here ;).

tests/test_consensus.py
```diff
@@ -1,2 +1,40 @@
...
+    from bigchaindb.consensus import BaseConsensusRules
+    from bigchaindb.common import crypto
+    from bigchaindb.common.util import serialize
+    vote_body = {}
+    vote_data = serialize(vote_body)
+    signature = crypto.PrivateKey(b.me_private).sign(vote_data.encode())
+    vote_signed = {
+        'node_pubkey': b.me,
+        'signature': signature.decode(),
+        'vote': vote_body
```
sohkai: Might be nice to have a comment like "Missing required vote properties" here to explain what's causing the schema validation to fail.

tests/common/schema/test_schema.py
```diff
@@ -0,0 +1,69 @@
+from bigchaindb.common.schema import TX_SCHEMA, VOTE_SCHEMA, \
```
sohkai: See above for the bit on multi-line imports.

tests/common/schema/test_schema.py
```diff
@@ -0,0 +1,69 @@
...
+                ]
+            }
+        },
+        'definitions': {
+            'wat': {
+                'description': "go"
+            }
+        }
+    }
+    drop_schema_descriptions(node)
```
sohkai: Maybe move this to be just above the `assert` for clarity.

bigchaindb/consensus.py
```diff
@@ -19,10 +26,19 @@ def validate_block(bigchain, block):
...
+        if verify_vote_signature(voters, signed_vote):
+            try:
+                validate_vote_schema(signed_vote)
+                return True
+            except SchemaValidationError as exc:
+                logger.warning(exc)
+        else:
+            logger.warning("Vote failed signature verification: "
+                           "%s with voters: %s", signed_vote, voters)
+        return False
```
sohkai: It's kind of weird that this returns a bool status whereas the transaction validation throws... but that's a problem for another day.

bigchaindb/common/schema/__init__.py
```diff
@@ -7,18 +7,46 @@
...
+    return path, schema
+
+
+def _validate_schema(schema, body):
+    """ Validate data against a schema """
     try:
-        jsonschema.validate(tx_body, TX_SCHEMA)
+        jsonschema.validate(body, schema)
     except jsonschema.ValidationError as exc:
         raise SchemaValidationError(str(exc))
```
sohkai: Oh yeah, we can also reraise the original to keep the stacktrace: `raise SchemaValidationError(str(exc)) from exc`

tests/test_consensus.py
libscott: It gives you the output:

```
tests/test_consensus.py 'invalid_reason' is a required property

Failed validating 'required' in schema['properties']['vote']:
    {'additionalProperties': False,
     'properties': {'invalid_reason': {'anyOf': [{'type': 'string'},
                                                 {'type': 'null'}]},
                    'is_block_valid': {'type': 'boolean'},
                    'previous_block': {'$ref': '#/definitions/sha3_hexdigest'},
                    'timestamp': {'pattern': '[0-9]{10}', 'type': 'string'},
                    'voting_for_block': {'$ref': '#/definitions/sha3_hexdigest'}},
     'required': ['invalid_reason',
                  'is_block_valid',
                  'previous_block',
                  'voting_for_block',
                  'timestamp'],
     'type': 'object'}

On instance['vote']:
    {}
```

PR #870, opened 2016-11-25
Fixed bug in create_rethinkdb_conf.py

ttmc:
Before, `bind_http_to_localhost` was a string, either "True" or "False", and so the line:
```python
    if bind_http_to_localhost:
```
saw a nonempty string and always evaluated it as `True` (even if the string was "False").

I fixed it.


ttmc:
@r-marques This is why you weren't able to see the RethinkDB web admin on port 8080, even when you set `BIND_HTTP_TO_LOCALHOST=False` in the AWS deployment configuration file.

ttmc:
Also, thanks to @libscott for helping me figure out this bug.

Issue #841, opened 2016-11-18
Remove `divisible` keyword


libscott:
It seems that a `divisible`, non `updatable`, non `refillable` asset with an amount `1` would be equal to a non-divisible asset, so perhaps we can remove the `divisible` keyword. I think it would be worthwhile as it would remove a bit of complexity from tests and make our product less conceptually complicated.

libscott:
Have a branch for this: https://github.com/bigchaindb/bigchaindb/tree/remove-divisible-keyword.

ttmc:
Oh interesting. So you're saying that the amount already determines whether an asset is divisible or not?

What about a `divisible` asset that is `refillable` and starts out with an amount `1`? Initially, it will be effectively indivisible, but later, if it gets more amount added to it, its divisibility would become meaningful. It's true that `refillable` assets aren't implemented today, but they may be in the future. Were you thinking that the `divisible` keyword/flag could be added back at that time?

P.S. I took a look at the `remove-divisible-keyword` branch and see that you spotted some places where the docs became out-of-date when divisible assets became supported. Those docs _should_ be updated, so if this idea gets shelved, let's try to remember to fix those docs at least.

libscott:
> What about a divisible asset that is refillable and starts out with an amount 1? Initially, it will be effectively indivisible, but later, if it gets more amount added to it, its divisibility would become meaningful. It's true that refillable assets aren't implemented today, but they may be in the future. Were you thinking that the divisible keyword/flag could be added back at that time?

If it starts out with amount 1, no-one will be able to divide it, and when it has amount >= 1, it will become `divisible` automatically. In the case of `refillable` assets, they don't appear to have any bearing or whether or not we need the `divisible` keyword. The only case that we would need a `divisible` keyword is if we want to support non-divisible assets with an amount > 1, but from what I can see this is currently not the intention:

```python
    # If the amount is supplied we can perform extra validations to 
    # the asset                                                     
    if amount is not None:                                          
        if not isinstance(amount, int):                             
            raise TypeError('`amount` must be an int')              
                                                                    
        if self.divisible is False and amount != 1:                 
            raise AmountError('non divisible assets always have'    
                              ' amount equal to one')               
                                                                    
        # Since refillable assets are not yet implemented this shoul
        # raise and exception                                       
        if self.divisible is True and amount < 2:                   
            raise AmountError('divisible assets must have an amount'
                              ' greater than one')                  
```

TimDaub:
[...]

PR #872, opened 2016-11-26
Update some docs re divisible assets

ttmc:
I updated a few docs pages that were out-of-date now that BigchainDB Server supports divisible assets.

Thanks to @libscott for bringing some of these to my attention via issue #841 and the associated branch at https://github.com/bigchaindb/bigchaindb/tree/remove-divisible-keyword

### DIFF
--- a/docs/root/source/transaction-concepts.md
+++ b/docs/root/source/transaction-concepts.md
@@ -15,7 +15,7 @@ one might register an identity or a creative work. The things are often called
 
 BigchainDB supports divisible assets as of BigchainDB Server v0.8.0.
 That means you can create/register an asset with an initial quantity,
-e.g. 700 thumbtacks. Divisible assets can be split apart or recombined
+e.g. 700 oak trees. Divisible assets can be split apart or recombined
 by transfer transactions (described more below).
 
 A CREATE transaction also establishes the conditions that must be met to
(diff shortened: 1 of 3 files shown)


ttmc:
Note: This PR is a blocker to releasing v0.8.0 of `bigchaindb` (server).

docs/root/source/transaction-concepts.md
```diff
@@ -4,31 +4,47 @@ In BigchainDB, _Transactions_ are used to register, issue, create or transfer
...
 given public key. More sophisticated conditions are possible.
 BigchainDB's conditions are based on the crypto-conditions of the [Interledger
 Protocol (ILP)](https://interledger.org/).
 
-A _transfer transaction_ can transfer an asset by fulfilling the current
-conditions on the asset. It can also specify new transfer conditions.
+## TRANSFER Transactions
+
+A TRANSFER transaction can transfer an asset
+(or set of assets of the same type)
```
r-marques: I find this sentence confusing because at this point a transaction is related to a particular asset.
I understand that what you are trying to say is that I can spend multiple outputs related to the same divisible asset but I wonder if there is a better language for this

r-marques:
what's a thumbtack?

docs/root/source/transaction-concepts.md
ttmc: Okay I'll stick to saying just "asset" here, even though it's not totally accurate, because sometimes clarity is more important than strict accuracy. This page is meant to be a gentle introduction, not a detailed exposition of all the possibilities.

I'll add a simple divisible assets example below to illustrate that possibility.

ttmc:
Here's a photo of a thumbtack. They come in many colors, including clear (as shown).

pin-517647_640

PR #881, opened 2016-11-29
Remove Condition.cid and Fulfillment.fid

### LINKED ISSUE #838: Condition.cid & Fulfillment.fid: Either should be always neccesary or never neccesary.
Currently our (still pending in #798) schema definition has `cid` and `fid` as optional fields, in their respective objects. These fields should either be mandatory or non-present.

They are necessary if we cannot depend on array order from our underlying database. Otherwise, they are not neccesary, and the `cid` and `fid` will be implicit given the array of conditions and fulfillments, respectively.

Personally I'm on the side of relying on databases to maintain sequence, as we are using arrays and not sets.

Given the result of this we should either remove `[c|f]id` or make them mandatory.

cc @sohkai tx milestone?

libscott:
Remove `Condition.cid` and `Fulfillment.fid` in favour of dereferencing conditions and fulfillments by their location in their respective arrays.

Fixes #838


tests/common/test_transaction.py
```diff
@@ -1023,7 +1018,6 @@ def test_create_transfer_transaction_multiple_io(user_pub, user_priv,
                     'owners_before': [
                         user2_pub
                     ],
-                    'fid': 1,
                     'fulfillment': None,
                     'input': {
                         'txid': tx.id,
```
diminator: Can you also design a test with multiple inputs/outputs and see if you can screw the sorting of either conditions or fulfillments?

libscott: Sure, I'm not sure what approach to take exactly but I'll have another look at the code and see if I can find any vectors... :pick:

sohkai: This really depends on what we want to test for.

If rethink and mongo provide guarantees (e.g. states it somewhere) for the order of arrays to be kept, then I don't think we need to consider anything more than just model-level tests.

Would just taking a transaction dict with multiple conditions / fulfillments, modifying this dict to have different orders in each, re-serializing each of the dicts to `Transaction` instances, and then checking that their hashes and condition / fulfillment orders are still different be enough?

PR #882, opened 2016-11-29
Fix HTTP API docs for upcoming release

### LINKED ISSUE #871: Needed for v0.8.0 release: updated JSON examples in HTTP API docs
This is a blocker for releasing v0.8.0 of `bigchaindb` (server).

The HTTP API docs at:

https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/http-client-server-api.html

must be updated to have examples of valid transaction JSON --- valid for version v0.8.0 of BigchainDB Server.

I'm worried that if I attempt this, I might mess it up. Using the BigchainDB Python Driver is tricky, because it depends on the `bigchaindb` package, but it should be using the latest code from the master branch of BigchainDB Server.

@sbellem @sohkai Maybe one of you could do this?

sohkai:
Fixes #871.

Not much has changed, but this transaction now passes validation for me. Might be good to double check on your side :).

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -54,12 +54,12 @@ POST /transactions/
 
    Push a new transaction.
 
-   Note: The posted transaction should be valid `transaction
+   Note: The posted transaction should be a valid and signed `transaction
    <https://bigchaindb.readthedocs.io/en/latest/data-models/transaction-model.html>`_.
    The steps to build a valid transaction are beyond the scope of this page.
    One would normally use a driver such as the `BigchainDB Python Driver
    <https://docs.bigchaindb.com/projects/py-driver/en/latest/index.html>`_ to
-   build a valid transaction.
+   build a valid transaction for a public/private keypair.
 
    **Example request**:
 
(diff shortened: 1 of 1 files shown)


docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -54,12 +54,12 @@ POST /transactions/
...
    Push a new transaction.
 
-   Note: The posted transaction should be valid `transaction
+   Note: The posted transaction should be a valid and signed `transaction
    <https://bigchaindb.readthedocs.io/en/latest/data-models/transaction-model.html>`_.
    The steps to build a valid transaction are beyond the scope of this page.
    One would normally use a driver such as the `BigchainDB Python Driver
    <https://docs.bigchaindb.com/projects/py-driver/en/latest/index.html>`_ to
-   build a valid transaction.
+   build a valid transaction for a public/private keypair.
```
ttmc: Not every transaction will have just one associated public/private keypair. For example, a 2-of-2 threshold condition would have two associated keypairs. Maybe just leave this line the way it was.

ttmc:
I tested the examples and they all worked.

PR #886, opened 2016-11-29
Abstract DataBase connection

vrde:
Load a specific connection class based on the configuration of the node.

I've changed also all the calls in the main `core.py::Bigchain` class. I've simplified the signature of the `__init__` by removing all the config params for the DB, such as `host`, `port`, `dbname` and replaced them with `connection`.

### DIFF
--- a/bigchaindb/backend/connection.py
+++ b/bigchaindb/backend/connection.py
@@ -1,3 +1,4 @@
+import bigchaindb
 from bigchaindb.common.exceptions import ConfigurationError
 from importlib import import_module
 
@@ -7,7 +8,11 @@ BACKENDS = {
 }
 
 
-def connect(backend, host, port, dbname):
+def connect(backend, host=None, port=None, name=None):
+    host = host or bigchaindb.config['database']['host']
+    port = port or bigchaindb.config['database']['port']
+    name = name or bigchaindb.config['database']['name']
+
     try:
         module_name, _, class_name = BACKENDS[backend].rpartition('.')
         Class = getattr(import_module(module_name), class_name)
@@ -17,7 +22,7 @@ def connect(backend, host, port, dbname):
     except (ImportError, AttributeError) as exc:
         raise ConfigurationError('Error loading backend `{}`'.format(backend)) from exc
 
-    return Class(host, port, dbname)
+    return Class(host, port, name)
 
 
 class Connection:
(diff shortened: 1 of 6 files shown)


bigchaindb/backend/rethinkdb/connection.py
```diff
@@ -0,0 +1,64 @@
...
+        Args:
+            host (str, optional): the host to connect to.
+            port (int, optional): the port to connect to.
+            db (str, optional): the database to use.
+            max_tries (int, optional): how many tries before giving up.
+        """
+
+        self.host = host or bigchaindb.config['database']['host']
+        self.port = port or bigchaindb.config['database']['port']
+        self.db = db or bigchaindb.config['database']['name']
```
sohkai: We can do this later, but I would avoid making a Connection class depend on the BDB config.

bigchaindb/core.py
```diff
@@ -10,7 +10,7 @@
 
 import bigchaindb
 
-from bigchaindb.db.utils import Connection, get_backend
+from bigchaindb import backend
```
sohkai: Can go into the next line?

bigchaindb/backend/connection.py
```diff
@@ -0,0 +1,15 @@
+from importlib import import_module
+
+BACKENDS = {
+    'rethinkdb': 'bigchaindb.backend.rethinkdb.connection.RethinkDBConnection'
+}
+
+
+def connect(backend, host, port, dbname):
```
vrde: > What do you think about making this a single dispatched call (that just returns a Connection class) and moving the import of the backend module to be an explicit part of the initialization phase?

we discussed this in person, but as a quick answer: it's not feasible, we need to instantiate the concrete class somewhere.



> Erroring out immediately if we can't find the backend module protects us from bad configs and also makes it easier to accept backend plugins in the future (I stress that I think this would be very easy, but we can save it for when we feel we're ready ). I'm sure it'll still error out like this right now, but it's not very obvious.

That's a good point, as discussed I can wrap it in a `try ... except` and raise a `sys.exit` for now, then we can iterate again and make it better.

bigchaindb/backend/rethinkdb/connection.py
vrde: Oh yes, totally yes :tada:

bigchaindb/core.py
vrde: sure

bigchaindb/backend/rethinkdb/connection.py
sohkai: Should we remove the config default here, since `connect()` now does this?

sbellem:
:+1:

PR #890, opened 2016-11-30
[DO NOT MERGE] MongoDB calls under new architecture

rhsimplex:
See #788, #886, and the [branch containing MongoDB work to date](https://github.com/bigchaindb/bigchaindb/tree/isolate-mongodb-calls)

WIP, assignees beware =)


sbellem:
@rhsimplex have a look at https://github.com/bigchaindb/bigchaindb/compare/master...sbellem:docker-mongo?expand=1 

all your commits have been cherry picked, and I started to try to run the tests with the backend set to mongodb

bigchaindb/backend/mongodb/connection.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+
+        self.host = host or bigchaindb.config['database']['host']
+        self.port = port or bigchaindb.config['database']['port']
+        self.db = db or bigchaindb.config['database']['name']
+        self.max_tries = max_tries
+        self.conn = None
+
+    @property
+    def db(self):
```
r-marques: You are defining a property that has the same name as the `self.db` variable in `__init__`.
Doesn't this create problems?

rhsimplex:
Ok final update:

The setup actually works.  If you build the docker images and run 
`docker-compose run --rm bdb-mdb py.test --maxfail=1 tests/db/test_bigchain_api.py`

This will set up the tests in MongoDB mode, start the db API tests and exit on the first error/failure.  You can actually start debugging the queries. A few notes:

1. The mongodb connection class is defined really awkwardly, because it's trying to mimic the rethinkdb connection style.
2.  There's no changefeed function defined yet for mongo, so in `backends/mongodb/__init__.py` the changefeed import is commented out so the other tests can run.  Be sure to uncomment this when you add the changefeed!

I got stuck at `write_block`.  It apparently takes a string representation as an argument.  It uses a rethink-specific deserializer in the rethink case, so I guess you need the python or mongodb equivalent here.

r-marques:
For in rethinkdb `write_block` was using `block.to_str` which was serializing the json using rapidjson. We did this because the block can get big and `rapidjson` was much faster then rethinkdb at serializing.

For mongodb this does not work because it uses a different serialization method.

tests/conftest.py
```diff
@@ -64,6 +64,9 @@ def restore_config(request, node_config):
 def node_config(request):
     config = copy.deepcopy(CONFIG)
     config['database']['backend'] = request.config.getoption('--database-backend')
+    if config['database']['backend'] == 'mongodb':
+        # not a great way to do this
+        config['database']['port'] = 27017
```
sohkai: We could probably include some default settings for each backend when we create the configuration to avoid this (and some of the docker-compose duplication).

r-marques:
closed in favour of #942

Issue #607, opened 2016-08-30
Need an easier way to update example Python code output in the docs


ttmc:
Some of our docs have examples of Python code and the output of those examples:

https://bigchaindb.readthedocs.io/en/latest/drivers-clients/python-server-api-examples.html

https://bigchaindb.readthedocs.io/en/latest/drivers-clients/python-driver-api-examples.html

I'd rather not have to be updating those manually all the time. In principle, it should be possible to get the outputs from a live, running instance of BigchainDB + RethinkDB. I don't think ReadTheDocs can run RethinkDB on their side though. One solution:

Step 1, on our side: run RethinkDB, BigchainDB, and some scripts to auto-generate those outputs and insert them as static text in the relevant Markdown files (in /docs/source/...). This wouldn't have to be done all the time, just when we expect the outputs in examples to be changed.

Step 2, use Sphinx to generate the docs from the contents of the GitHub repo, the same as always. This works locally and on ReadTheDocs.

Maybe Step 1 can be accomplished using an IPython Notebook?
- Start RethinkDB
- Start BigchainDB
- Start IPython
- Run an IPython notebook version of the documentation page (contains the text, the Python commands, and the output of those commands)
- Convert the "finished" IPython notebook to something Sphinx can ingest (.rst or .md or whatever)

A quick Google search reveals that there are several options for that last step.

ttmc:
Sorry I bumped something on my keyboard and created this issue before I was done writing it up. I edited the description and it's better now.

ttmc:
Issue #637 is about the same example code but it's concerned with _testing_ that code.

ttmc:
I discussed the Python Server API examples page at

https://bigchaindb.readthedocs.io/en/latest/drivers-clients/python-server-api-examples.html

with Dimi (who wrote that page). He said that all the Python Server API functionality should (eventually) be exposed via the various client-side APIs, so that page will no longer be necessary. Maybe it can be moved to the Appendices, with some notes added to the top.

As for the Python Driver API examples page at 

https://bigchaindb.readthedocs.io/en/latest/drivers-clients/python-driver-api-examples.html

... the example code is now gone from that page (as of PR #625). Some new example code will be going in the new Python driver documentation at https://bigchaindb-driver.readthedocs.io/en/latest/index.html so this issue isn't going away, it's just moving over there.

Maybe we can use the Sphinx extension named doctest? See http://www.sphinx-doc.org/en/stable/ext/doctest.html

[...]

PR #891, opened 2016-11-30
Auto generate examples in http server API docs

libscott:
Recently we've had the issue that every time the schema is altered, the example transaction payloads in the [http server docs](https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/http-client-server-api.html#post--transactions-) need re-hashing and there is no easy way of doing it.

Given this, it was easier to automate the process such that the payloads are generated when the docs are built with Sphinx.

There were several options for performing this build step, one was using a GIT hook such that generated `.rst` assets are built and committed to the repository. But for simplicity and ease of use, we opted for the building to take place as a build step in the Sphinx documentation build.

The next problem was that Sphinx doesn't have a notion of pre-build steps, and ReadTheDocs doesn't allow you to override the entry point for building. To solve this we are somewhat abusing the Sphinx `extensions` in `conf.py`; documentation generators are imported pre-build and perform their work out of the `setup()` method.

Additionally, a test has been added to check that the documentation builds are working as expected.

Related: #82, #607 

screenshot from 2016-11-30 16-21-15


docs/server/generate_http_server_api_documentation.py
```diff
@@ -0,0 +1,87 @@
...
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+%(tx)s
+"""
+
+
+def main():
+    """ Main function """
+    pub = '9RaWxppkP9UyYWA7NJb5FcgkzfJNPfvPX3FCNw2T5Pwb'
```
ttmc: Maybe use a more descriptive name for this variable (currently `pub`). Elsewhere in BigchainDB, I often see `pubkey` or `public_key`.

sohkai:
justin-wow-jaw-drop-so-3osxY7xVVSgMfAPZ16

PR #893, opened 2016-12-02
Implement singledispatch'd database abstraction

sohkai:
Implements the singledispatch pattern for abstracting database backends. Includes a convenience decorator to make the registration of dispatches *slightly* easier.

~~Cherrypicks (and slightly modifications, for convenience) a few things from #886 as a stopgap before #886 is merged.~~ Originally, I thought I could make this branch pass all the tests, but this is proving unlikely without larger changes (mostly database setup and commands) with more of #886. Making the tests pass, with just RethinkDB, will be done in a another pull request.

Should be merged after #886; I'll solve the merge conflicts that might happen as a result of the cherrypicking.


bigchaindb/backend/util.py
```diff
@@ -0,0 +1,20 @@
...
+                dispatch_registrar = getattr(module, func_name)
+                return dispatch_registrar.register(obj_type)(func)
+            except AttributeError as ex:
+                raise BackendModuleDispatchRegisterError(
+                    ("'{module}' does not contain a single-dispatchable "
+                     "function named '{func}'. The backend being used has not "
+                     "been implemented correctly!").format(
+                         func=func_name, module=module.__name__)) from ex
+        return wrapper
+    return dispatch_wrapper
```
sbellem: Could we put the `return` line outside? E.g.: 

```python
def make_module_dispatch_registrar(module):
    def dispatch_wrapper(obj_type):
        def wrapper(func):
            func_name = func.__name__
            try:
                dispatch_registrar = getattr(module, func_name)
            except AttributeError as ex:
                raise BackendModuleDispatchRegisterError(
                    ("'{module}' does not contain a single-dispatchable "
                     "function named '{func}'. The backend being used has not "
                     "been implemented correctly!").format(
                         func=func_name, module=module.__name__)) from ex
            return dispatch_registrar.register(obj_type)(func)
        return wrapper
    return dispatch_wrapper
```

sohkai: Didn't want to do this because `dispatch_registrar.register` might not work (`dispatch_registrar` might not actually be a single dispatchable function)

bigchaindb/backend/utils.py
```diff
@@ -0,0 +1,20 @@
...
+                dispatch_registrar = getattr(module, func_name)
+                return dispatch_registrar.register(obj_type)(func)
+            except AttributeError as ex:
+                raise BackendModuleDispatchRegisterError(
+                    ("'{module}' does not contain a single-dispatchable "
+                     "function named '{func}'. The backend being used has not "
+                     "been implemented correctly!").format(
+                         func=func_name, module=module.__name__)) from ex
+        return wrapper
+    return dispatch_wrapper
```
sbellem: How about renaming `module` to `generic_backend_module`?

sbellem: How about renaming `obj_type` to `backend_type`, or `backend_class`?

sbellem: How about chameleon ? :)

bigchaindb/core.py
```diff
@@ -46,33 +46,28 @@ def __init__(self, host=None, port=None, dbname=None, backend=None,
...
             private_key (str): the base58 encoded private key for the ED25519 curve.
             keyring (list[str]): list of base58 encoded public keys of the federation nodes.
         """
 
         config_utils.autoconfigure()
+
+        self.me = public_key or bigchaindb.config['keypair']['public']
+        self.me_private = private_key or bigchaindb.config['keypair']['private']
+        if not self.me or not self.me_private:
+            raise exceptions.KeypairNotFoundException()
```
sbellem:
```python
raise exceptions.KeypairNotFoundException
```

sbellem:
:+1:

PR #895, opened 2016-12-05
Add generic init_database() function to set up database schema

sohkai:
Adds back the ability to initialize a database's schema, based on their dispatched functions.

@rhsimplex I think this is quite general, but to double check, can we set up mongo like this too?


rhsimplex:
Yeah, should be ok

sohkai:
Oops, merged into wrong branch. Opened https://github.com/bigchaindb/bigchaindb/pull/901 instead.

PR #896, opened 2016-12-05
Fix commands and related tests broken from the database abstraction

sohkai:
Attempts to fix most of the bigchaindb commands and associated tests that were broken by the database abstraction, mostly from the Connection changing interfaces.

Tests, obviously, won't be all passing yet.

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -22,7 +22,8 @@ import bigchaindb
 import bigchaindb.config_utils
 from bigchaindb.models import Transaction
 from bigchaindb.util import ProcessGroup
-from bigchaindb import db
+from bigchaindb import backend
+from bigchaindb.backend import schema
 from bigchaindb.commands import utils
 from bigchaindb import processes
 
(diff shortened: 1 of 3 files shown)


bigchaindb/commands/bigchain.py
```diff
@@ -133,14 +134,25 @@ def run_export_my_pubkey(args):
         # exits with exit code 1 (signals tha an error happened)
 
 
+def _run_init():
+    # Try to access the keypair, throws an exception if it does not exist
+    b = bigchaindb.Bigchain()
+
+    schema.init_database(b.connection)
```
vrde: :heart_eyes: this looks so cool

bigchaindb/commands/bigchain.py
```diff
@@ -222,23 +243,25 @@ def run_load(args):
 
 
 def run_set_shards(args):
+    conn = backend.connect()
     for table in ['bigchain', 'backlog', 'votes']:
         # See https://www.rethinkdb.com/api/python/config/
-        table_config = r.table(table).config().run(db.get_conn())
+        table_config = conn.run(r.table(table).config())
```
vrde: we can move this to a different module, but for now it's fine IMO (it's more important to have others working on the MongoDB queries).

bigchaindb/commands/utils.py
```diff
@@ -39,11 +39,11 @@ def start_rethinkdb():
             #        of the database. This code assumes the tables are ready
             #        when the database is ready. This seems a valid assumption.
             try:
-                conn = db.get_conn()
+                conn = backend.connect()
                 # Before checking if the db is ready, we need to query
                 # the server to check if it contains that db
-                if r.db_list().contains(dbname).run(conn):
-                    r.db(dbname).wait().run(conn)
+                if conn.run(r.db_list().contains(dbname)):
```
vrde: This one as well, we can do it later. We can eventually add a new module for "database operations".

PR #898, opened 2016-12-06
Fix tests broken from abstracting the database

sohkai:
Based on merging #895 and #896 into #788.

All tests are a go with Rethinkdb 🚨  (some tests are still coupled to Rethinkdb, but this will be done in the future).

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,4 +1,3 @@
-import builtins
 import json
 from unittest.mock import Mock, patch
 from argparse import Namespace
@@ -216,7 +215,7 @@ def test_bigchain_run_init_when_db_exists(mock_db_init_with_existing_db):
     run_init(args)
 
 
-@patch('bigchaindb.schema.drop_database')
+@patch('bigchaindb.backend.schema.drop_database')
 def test_drop_db_when_assumed_yes(mock_db_drop):
     from bigchaindb.commands.bigchain import run_drop
     args = Namespace(config=None, yes=True)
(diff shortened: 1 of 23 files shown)


tests/test_core.py
```diff
@@ -31,9 +32,7 @@ def test_bigchain_class_default_initialization(config):
     from bigchaindb.core import Bigchain
     from bigchaindb.consensus import BaseConsensusRules
     bigchain = Bigchain()
-    assert bigchain.host == config['database']['host']
-    assert bigchain.port == config['database']['port']
-    assert bigchain.dbname == config['database']['name']
+    assert bigchain.connection
```
sohkai: Not really sure what to assert anymore; is just checking that `bigchain.connection` enough?

tests/test_commands.py
```diff
@@ -225,10 +215,33 @@ def test_bigchain_run_init_when_db_exists(mock_db_init_with_existing_db):
     run_init(args)
 
 
-def test_drop_existing_db(mock_rethink_db_drop):
+@patch('bigchaindb.backend.schema.drop_database')
+def test_drop_db_when_assumed_yes(mock_db_drop):
```
sohkai: These drop db interactive tests were (more or less) previously done in the `db/test_utils.py` file that was moved to `db/rethinkdb/test_schema.py`.

tests/test_core.py
```diff
@@ -30,10 +31,12 @@ def config(request, monkeypatch):
 def test_bigchain_class_default_initialization(config):
     from bigchaindb.core import Bigchain
     from bigchaindb.consensus import BaseConsensusRules
+    from bigchaindb.backend.connection import Connection
     bigchain = Bigchain()
-    assert bigchain.host == config['database']['host']
-    assert bigchain.port == config['database']['port']
-    assert bigchain.dbname == config['database']['name']
+    assert isinstance(bigchain.connection, Connection)
```
sohkai: Wasn't sure if we necessarily wanted to do this check, but I'm k with it.

vrde: I wasn't 100% sure as well, but I think it's OK for now. From my experience `asset something` is too loose. I prefer to have a stricter tests that will eventually fail, so we can fix it again in the proper way afterwards.

PR #899, opened 2016-12-06
Add option to switch backend

vrde:
I've added a new `--database-backend` option to py.test, so we can run the whole suite of tests on different backends. This is just a proposal and it **should** work. IMO we can integrate it soon because we need the `backend` key anyway.


vrde:
> Are we able to access the pytest options elsewhere in the tests (e.g. for skipping test modules)?

Yes! We can skip tests based on a condition. We can work on this when we start integrating MongoDB.

PR #902, opened 2016-12-06
Fix metadata description in transaction-model.rst

libscott:
Metadata description in docs was wrong after `remove-metadata-uuid`, this pr fixes, and also reflects that `metadata` may be NULL even in the case of CREATE transactions.


ttmc:
I rendered it locally and it looks good!

:+1:

libscott:
I forgot to include my usual screenshot this time, will try to remember next time :stuck_out_tongue:

PR #903, opened 2016-12-07
remove old tx recovery on block process start

### LINKED ISSUE #876: remove get old transactions
Currently, when a `block` process starts, it queries the backlog for all transactions assigned to it.  The idea is that if the node goes down, there should be a way for it to get assigned transactions back without relying on the changefeed.

This behavior is redundant, however, since we have a stale transaction monitor which will reassign old transactions anyway.  I propose we therefore remove the initialization step from `block` which fetches old transactions.

Found this while writing PR #860

rhsimplex:
Resolves #876.


rhsimplex:
After discussion with @vrde, I suggest we merge this as is.  The test suggested by @r-marques in #860 is necessary, but tricky to implement.  I will do it in a second PR, so I can focus on MongoDB integration for the time being.

PR #908, opened 2016-12-08
Fix 781 and 124

### LINKED ISSUE #124: Queues are unbounded
The voter and block processes use queues to communicate internally. Right now these queues are unbounded. We need to benchmark the current implementation, understand its memory requirements and put a limit to the size of the queues.

### LINKED ISSUE #781: How can we release the higher memory bigchaindb used or leak?
With the bigchaindb older **0.6.0**, run  `bigchiandb load -m 10` may produce the timeout exception.
Now, the **0.8.0.dev** solves it. But we find the memory is run higher when run `bigchaindb load -m 50` .
When stop the load,it will go higher before the backlog is empty or the memory is running out.And it will keep that until kill the bigchaindb processes.

**MayBe it`s trouble by memory leak or else ?**

Now we can release it by stop the bigchaindb. How can we solve that ？

**Thks for replying.**
The partial `top` results as follow:

[bigchaindb 0.8.0.dev]
screenshot from 2016-11-01 20 23 57

krish7919:
The queue/pipeline used in the block creation process is unbounded.
When too many transactions occur, the backlog table fills up quite easily and
bigchaindb just reads all the transaction in the backlog to create a block.
This causes memory usage to grow indefinitely.

Limiting the queue size to 1000 transactions for now as the block creation and
voting happens in batches of 1000. Can be increased later in case block size
is increased.

Resolves #781 
Resolves #124

### DIFF
--- a/bigchaindb/pipelines/block.py
+++ b/bigchaindb/pipelines/block.py
@@ -8,7 +8,7 @@ function.
 import logging
 
 import rethinkdb as r
-from multipipes import Pipeline, Node
+from multipipes import Pipeline, Node, Pipe
 
 from bigchaindb.models import Transaction
 from bigchaindb.pipelines.utils import ChangeFeed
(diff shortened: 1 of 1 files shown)


ttmc:
This looks good to me (LGTM) but @vrde should review it. You can assign reviewers now (over in the right sidebar on GitHub).

I added two "Resolves #____" type comments to your opening comment, so that when this pull request gets merged, GitHub will automatically close those issues.

PR #909, opened 2016-12-08
Add logging statement

vrde:
As easy as that


bigchaindb/backend/rethinkdb/connection.py
```diff
@@ -68,7 +68,11 @@ def _connect(self):
         for i in range(self.max_tries):
             try:
                 self.conn = r.connect(host=self.host, port=self.port, db=self.dbname)
+                logging.debug('Database connection established')
```
sohkai: Maybe be explicit about the host, port, and dbname here?

vrde: Lol I did it and then I removed it. Agree :)

Issue #633, opened 2016-09-16
`get_owned_ids` breaks after DoubleSpent


diminator:
How to reproduce:
- do a `DoubleSpent`, you'll see the node putting the transactions back
- try `get_owned_ids`: it will throw an error on `DoubleSpent`

rhsimplex:
Can you include the stacktrace? `DoubleSpend` is raised in both core and consensus

rhsimplex:
I can't reproduce this. I have a PR open for integration tests ( #614).  See `tests/integration/test_double_creation.py::test_get_owned_ids_works_after_double_spend`.

PR #910, opened 2016-12-08
Broken integration test

rhsimplex:
See issue #633 and PR #614.  Basically, a doublespend test that worked before doesn't work anymore.

As far as I can tell, no doublespend occurs (i.e. no duplicate tx in bigchain)

To repeat, though, **a doublespend exception should be raised in this test but isn't**

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -44,3 +44,35 @@ def test_double_create(b, user_pk):
     last_voted_block = b.get_last_voted_block()
     assert len(last_voted_block.transactions) == 1
     assert count_blocks(b.connection) == 2
+
+
+@pytest.mark.usefixtures('processes', 'inputs')
+def test_get_owned_ids_works_after_double_spend(b, user_pk, user_sk):
+    """See issue 633."""
+    from bigchaindb.models import Transaction
+    input_valid = b.get_owned_ids(user_pk).pop()
+    input_valid = b.get_transaction(input_valid.txid)
+    tx_valid = Transaction.transfer(input_valid.to_inputs(),
+                                    [([user_pk], 1)],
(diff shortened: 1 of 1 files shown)


libscott:
closes #633

libscott:
This is ready to review now, the test shows that the bug is not reproducible.

libscott:
> Ah... I guess the thing that broke the test earlier was because of colliding tx ids from changes to the model (e.g. removal of timestamp)?

The test wasn't broken, it was trying to reproduce #633. See https://github.com/bigchaindb/bigchaindb/pull/910/commits/9d47fc75acfa5fe6ccbff214bf85c93c42e66d7f#diff-4c7022537015fa4876e90501d3136011R63

I fixed the test to work with the new asset changes so we can close the bug report.

> (Although if that made the test return the double spend as a valid transaction, maybe we should add tests for duplicate transactions too--if we don't already have some?)

That wasn't a double spend, it was the same transaction. `tx_valid` and tx_double_spend` were the same because no timestamp.

libscott:
I don't know what's going on with codecov, the change includes one test but coverage has gone down?

tests/integration/test_integration.py
```diff
@@ -44,3 +44,36 @@ def test_double_create(b, user_pk):
...
+                                           input_valid.id,
+                                           {'2': 2}).sign([user_sk])
+
+    # write the double spend tx
+    b.write_transaction(tx_double_spend)
+    time.sleep(2)
+
+    # still doesn't throw an exception
+    b.get_owned_ids(user_pk)
+    assert b.is_valid_transaction(tx_double_spend) is False
```
r-marques: Can we change this to run `validate_transaction` and check that it is actually failing with a `DoubleSpend`

PR #918, opened 2016-12-08
Improve dev documentation for backend abstraction

sohkai:
Adds documentation on the architecture of the backend, how singledispatch works, how to use the module in other parts of BigchainDB, and some general rst fixes.

(Should be merged after #788 is cleaned up).


bigchaindb/backend/README.md
```diff
@@ -0,0 +1,43 @@
+# Backend Interfaces
+
+## Structure
+
+- [`changefeed.py`](./changefeed.py): Changefeed-related interfaces
+- [`connection.py`](./connection.py): Database connection-related interfaces
+- [`query.py`](./query.py): Database query-related interfaces, dispatched through single-dispatch
+- [`schema.py`](./schema.py): Database setup and schema-related interfaces, dispatched through
+  single-dispatch
```
sohkai: Couldn't really find more to say about these files... any ideas?

bigchaindb/backend/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Built-in implementations (e.g. [RethinkDB's](./rethinkdb)) are provided in sub-directories and
+have their connection type's location exposed as `BACKENDS` in [`connection.py`](./connection.py).
+
+## Single-Dispatch
```
ttmc: I'm glad you wrote this section, but it seems to be written in a strange order. It starts out by talking about single-dispatch (which may be a new idea) and only later gets around to showing what it enables.

I think it might be more readable if you started out with the examples of how to e.g. create a connection object, and then when they wonder how it knows which backend implementation to use, you can say, "Aha, that's the magic of Python's new single-dispatch thingy." with links to the relevant PEP etc.

sohkai: I've reworded the section: https://github.com/bigchaindb/bigchaindb/pull/918/commits/8d903bc459e5225b128a9a4323fb761350c92d6b.

sohkai:
Closing for #932.

PR #922, opened 2016-12-09
Upgrade python-rapidjson to 0.0.8

sohkai:
Fixes build issues when travis thinks that 0.0.7 is the latest.


sbellem:
@sohkai @r-marques we can merge this right?

PR #923, opened 2016-12-09
[Proposal]: Suggest starting multi-line imports on the next line to pass flake8

sohkai:
@libscott found out that our [current recommendation for multi-line imports in the style guide](https://github.com/bigchaindb/bigchaindb/blob/master/PYTHON_STYLE_GUIDE.md#how-to-format-long-import-statements) will cause a flake8 error. If we do want to keep the import on the same line, it should be intended to where the brace starts:

 ```python
 from Tkinter import (Tk, Frame, Button, Entry, Canvas, Text,
                      LEFT, DISABLED, NORMAL, RIDGE, END)
```

However, I think it usually looks nicer if the imports just start on the next line (either grouped or each import in each line):

```python
 from Tkinter import (
     Tk, Frame, Button, Entry, Canvas, Text,
     LEFT, DISABLED, NORMAL, RIDGE, END,
)
 ```

I've left both as options.

### DIFF
--- a/PYTHON_STYLE_GUIDE.md
+++ b/PYTHON_STYLE_GUIDE.md
@@ -47,8 +47,15 @@ It seems the preference is for slashes, but using parentheses is okay too. (Ther
 
 If you need to `import` lots of names from a module or package, and they won't all fit in one line (without making the line too long), then use parentheses to spread the names across multiple lines, like so:
 ```python
+from Tkinter import (
+    Tk, Frame, Button, Entry, Canvas, Text,
+    LEFT, DISABLED, NORMAL, RIDGE, END,
+)
+
+# Or
+
 from Tkinter import (Tk, Frame, Button, Entry, Canvas, Text,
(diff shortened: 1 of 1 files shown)


ttmc:
This suggestion is fine with me, but I'm curious if there are any objections, and why.

libscott:
It seems that when it comes to breaking strings across lines we prefer
slashes, but for long import statements it's parenthesis. My preferences
happen to be the other way around, I usually prefer parenthesis for strings
and on occasion I'd use slashes for imports. I don't have a problem reading
either style.

I wouldn't try to argue in favour of my particular preferences. I do
however have my doubts that these discussions really add alot of value. I'd
be in favour of relaxing the rules and not expecting code to be perfect
(flake8 is plenty). But if we are going to choose one over the other, it
makes sense to standardize it for breaking strings over lines, and multi
line import statements, since both the \ and the () are achieving the same
end in this case.

Issue #756, opened 2016-10-27
Project: Renaming things for clarity, consistency, and correctness


ttmc:
There is a small cluster of issues around renaming things for clarity, consistency, and correctness. I thought it might be useful to bundle them all together for reference:
- [x] #746 - Naming convention: public/private keys versus verifying/signing keys?
- [x] #743 - Replace the notions of inputs and outputs with fulfillments and conditions
- [x] #717 - Rename `data_id` in the Metadata and Asset models
- [x] #715 - Revisit `to_hash()` in Assets and Metadata
- [x] #709 - Sanitize naming/positioning of fulfillments, conditions, inputs, outputs
- [x] #708 - Why use "id" and "data_id" for the same thing?
- [x] #679 - Rename variable "validity" to more descriptive name, such as "block_statuses"
- [x] #626 - Rename 'owners_after' to 'fulfillers'
- [ ] #603 - rename `get_owned_ids` to `get_unspents`

Others?

ttmc:
There's only one open issue on this list and it can live on its own, so I'm closing this issue (which is just a list of other issues).

Issue #677, opened 2016-10-12
Add checks to ensure conditions & fullfillments are internally consistent


ttmc:
When validating transactions, we should --- if possible --- check fulfillments and conditions to ensure that:

* The list of keys in `owners_before` (to be renamed: see issue #626) agrees with the other information in the fulfillment.
* The list of keys in `owners_after` (to be renamed: see issue #626) agrees with the other information in the condition.

This may not be possible today, so we should consider changing the structure of conditions and/or fulfillments to make it possible. Here is some of what @diminator wrote about this in issue #626:

> ...there is potentially a discrepancy to the public keys in the list and the actual actors in the condition: In the case of only a condition URI provided, there is no way for the system to cross-check the condition actors with the listed public keys.

> I am currently leaning towards providing the full condition details when public keys are involved in the system (Threshold, Ed25519, ...) This can be detected from the type or feature bitmask.

and:

> ...the `fulfillment_uri` can be deserialized into a full object with all internals (`verifying_keys`) known

diminator:
> so we should consider changing the structure of conditions and/or fulfillments to make it possible

At the lowest level we want the system to work only with `condition_uri` and `fulfillment_uri`. So no additional details revealed

At a higher level we want a user to be able to query the transactions where he is a single owner. This should be easy to check since the condition would always be an `ED25519Condition`, where the `verifying_key` can be extracted from the `condition_uri`. This would mean we can do the consistency check of the `owners_after` (or whatever we call it in the future) without having to expose the internal details of the condition. You can easily do it from only the URI in this case.

At the highest level we might want to give the user the option to query all transactions where he is partially involved in. Now we are talking about complex `ThresholdConditions`, when they hashed into a  URI all the internals are masked and entangled. A consistency check would then need more info than only the URI. Notice that bitcoin and others only allow to query `pure` ownership (only a single key) and not every involvement in a transaction (which could be a branch you didnt need to sign in a 1-of-2 multisig). Question here is: do we need to provide this functionality in the short/long run? Or should a service on top provide this? Or is it enough to provide the internal details of a complex condition?

ttmc:
I'm keeping this issue _open_ for now because it belongs to a set of "tx model" issues that **do** need to get addressed sometime soon, hopefully in transaction spec 3.0 or 4.0. I guess we'll make a proper BigchainDB Enhancement Proposal, but lets let this annoying open issue sit here to remind us to do that for now.

ttmc:
[...]

PR #925, opened 2016-12-09
Inputs & Outputs

### LINKED ISSUE #626: Rename 'owners_after' to 'fulfillers'
@ttmc, @TimDaub, and I had a small discussion on the misleading wording of 'owners_after' (they're not _exactly_ the owners after).

Thoughts @diminator @r-marques?

This should probably be done after https://github.com/bigchaindb/bigchaindb-common/pull/4 and #590 are merged, to make it easier to do.

### LINKED ISSUE #743: Replace the notions of inputs and outputs with fulfillments and conditions
In a comment on issue https://github.com/bigchaindb/bigchaindb/issues/626#issuecomment-256020754 , @r-marques wrote:

> This is a good step to probably replace the notion of inputs and outputs with fulfillments and conditions

The mission, therefore, is to grep through all the code and documentation, and to:
- replace all mentions of (Bitcoin-style) "output(s)" with "condition(s)"
- replace all mentions of (Bitcoin-style) "input(s)" with "fulfillment(s)"

Note: This is not as easy as it might seem. The words "output" and "input" might be used in other contexts where we don't mean Bitcoin-style outputs or inputs (i.e. conditions or fulfillments in BigchainDB). For example:

> Susan phoned and asked me for my input.

Don't change those ones.

Related phrases are "spending an output" or "unspent outputs." That's a cryptocurrency-specific phrase. In BigchainDB, one doesn't "spend an output," one "fulfills a condition" (a txid+cid pair). Therefore, a third thing to do is:
- replace all phrases such as "spends an output" with "fulfills a condition". Some methods and exceptions should be renamed accordingly. If you like, you can create a new issue for this chore but it makes sense to tackle it at the same time as the other two, because occurrences will be in the same neighborhood as "inputs" and "outputs."

### LINKED ISSUE #822: Rename non-cryptoconditions classes (Condition, Fulfillment) to different names
This is confusing, if someone somehow is using both libraries. 

Something like `BdbCondition`, and `BdbFulfillment` would prevent the confusion.

### LINKED ISSUE #875: [TBD] Generalize conditions/fulfillments according to proposal
This is a placeholder issue so that we can start closing and referring old tickets. TBD.

libscott:
This branch does the following:

* Change `conditions` to `outputs`.
* Change `fulfillments` to `inputs`.
* Change `input` in the Input to `fulfills`.
* Change `cid` in the TransactionLink to `output`.
* Change `owners_after` in the Output to `public_keys`.

It was originally planned to remove `owners_before` from the Input in 0.9, however, there are some remaining questions. The issue will be addressed in #945.

Still to do:

- [x] Fix test names 

Before merging:

- [x] ~~Alert mongodb integration about change to query~~ Talked to @r-marques about this; we won't worry about mongodb until it's in a more complete state.

Fixes #875, fixes #626, fixes #822, fixes #743 

Related #756, #677


bigchaindb/common/schema/transaction.yaml
```diff
@@ -88,11 +88,10 @@ definitions:
         Type of the transaction:
 
         A ``CREATE`` transaction creates an asset in BigchainDB. This
-        transaction has outputs (conditions) but no inputs (fulfillments),
-        so a dummy fulfillment is used.
+        transaction has outputs but no inputs, so a dummy input is created.
 
         A ``TRANSFER`` transaction transfers ownership of an asset, by providing
-        fulfillments to conditions of earlier transactions.
+        fulfillments to outputs conditions of earlier transactions.
```
ttmc: The wording here is a bit odd. How about:

fulfillments to conditions in outputs of earlier transactions.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -131,21 +130,27 @@ definitions:
...
           additionalProperties: true
         - type: 'null'
-  condition:
+  output:
     type: object
     description: |
-        An output of a transaction. A condition describes a quantity of an asset
-        and what conditions must be met in order for it to be fulfilled. See also:
-        fulfillment_.
+        A transaction output. Describes a quantity of an asset and a locking
```
ttmc: The "locking script" terminology isn't used anywhere else in our docs, as far as I know. We might use it in the future, but for now I think we can do without. How about:

A transaction output. Describes the quantity of an asset and the requirements that must be met to spend the output.

docs/root/source/transaction-concepts.md
```diff
@@ -28,19 +28,19 @@ Protocol (ILP)](https://interledger.org/).
 ## TRANSFER Transactions
 
 A TRANSFER transaction can transfer an asset
-by fulfilling the current conditions on the asset.
+by provding inputs which fulfill the current output conditions on the asset.
```
ttmc: Minor typo: provding --> providing

sohkai:
Curious: does this also close https://github.com/bigchaindb/bigchaindb/issues/743 (seems like #875 supersedes #743?), based on the discussions that have happened?

sohkai:
> Change `cid` in the `Input.fulfills` to `idx`.

Curious: does `idx` stand for anything? In some ways `cid` still makes sense, since we're still talking about conditions inside an output, but I agree it might not be the best.

bigchaindb/common/schema/transaction.yaml
sohkai: @ttmc I've fixed this in: https://github.com/bigchaindb/bigchaindb/pull/952.

libscott:
> Curious: does this also close #743 (seems like #875 supersedes #743?), based on the discussions that have happened?

Yea. I'll update #875 with the other changes that Dimi proposed. And yes I think this should count as resolving #743.

libscott:
> Curious: does idx stand for anything?

`idx` stands for index. 

> In some ways cid still makes sense, since we're still talking about conditions inside an output

Not quite, because it's: `tx.outputs[idx].condition`. I also considered:

* `oid`: sounds too much like 'object id'
* `output_idx`: Seems unnecessarily long. I got byte shaving ocd :grimacing: 
* `output_id`: Feels wrong to call it an ID, usually when we talk about an ID it's relative to the entire set of the thing in question, ie, the combination of transaction ID and output array offset is an output ID, which we are now calling `fulfills`...
* `offset`: Could also work...

sohkai:
> idx stands for index (fairly common acronym)

Hmm, I guess if it passed the eyeball tests of a few others, it's good.

My favourites below then, in decreasing descriptive-ness:

```json
{
    "transaction_id": "",
    "output_index": ""
}

{
    "transaction": "",
    "output": ""
}

{
    "txid": "",
    "output": ""
}

{
    "txid": "",
    "idx": ""
}
```

But let's not get hung up over this, I'll take what you like best 😄.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -161,30 +166,22 @@ definitions:
...
         description: |
-            List of public keys associated with asset ownership at the time
-            of the transaction.
-      amount:
-        type: integer
-        description: |
-            Integral amount of the asset represented by this condition.
-            In the case of a non divisible asset, this will always be 1.
-  fulfillment:
+            List of public keys associated with asset control.
```
ttmc: Maybe "List of public keys associated with the conditions on an output."

Or "List of public keys associated with the transfer conditions on an output."

bigchaindb/common/schema/transaction.yaml
```diff
@@ -161,30 +166,22 @@ definitions:
...
     description:
-      A fulfillment is an input to a transaction, named as such because it
-      fulfills a condition of a previous transaction. In the case of a
-      ``CREATE`` transaction, a fulfillment may provide no ``input``.
+      An input spends a previous output, by providing a fulfillment to the output's
+      condition.
     additionalProperties: false
     required:
     - owners_before
-    - input
```
ttmc: @libscott will know.

docs/server/source/data-models/transaction-model.rst
```diff
@@ -33,12 +33,12 @@ Here's some explanation of the contents of a :ref:`transaction <transaction>`:
 
 - id: The :ref:`id <transaction.id>` of the transaction, and also the database primary key.
 - version: :ref:`Version <transaction.version>` number of the transaction model, so that software can support different transaction models.
-- **fulfillments**: List of fulfillments. Each :ref:`fulfillment <Fulfillment>` contains a pointer to an unspent asset
+- **inputs**: List of inputs. Each :ref:`input <Input>` contains a pointer to an unspent asset
```
ttmc: An asset can have many shares (e.g. an amount of 300 in the initial CREATE transaction), and those shares can, get spread out across many outputs (up to 300 in my example). If you sum the amounts of an asset across all the unspent outputs, it should always equal the initial amount created.

Therefore it doesn't make sense to talk about an "unspent asset" --- it only makes sense to talk about an unspent output.

docs/server/source/data-models/transaction-model.rst
```diff
@@ -33,12 +33,12 @@ Here's some explanation of the contents of a :ref:`transaction <transaction>`:
...
   and a *crypto fulfillment* that satisfies a spending condition set on the unspent asset. A *fulfillment*
   is usually a signature proving the ownership of the asset.
   See :doc:`./crypto-conditions`.
 
-- **conditions**: List of conditions. Each :ref:`condition <Condition>` is a *crypto-condition* that needs to be fulfilled by a transfer transaction in order to transfer ownership to new owners.
+- **outputs**: List of outputs. Each :ref:`output <Output>` is a *crypto-condition* that needs to be fulfilled by a transfer transaction in order to transfer ownership to new owners.
```
ttmc: Yes @sohkai I think you're right.

bigchaindb/common/schema/transaction.yaml
libscott: See also: #945

bigchaindb/common/transaction.py
```diff
@@ -613,33 +612,33 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
...
         # for transactions other then CREATE we only have an id so there is
         # nothing we can validate
         if self.operation == self.CREATE:
-            amount = sum([condition.amount for condition in self.conditions])
+            amount = sum([condition.amount for condition in self.outputs])
             self.asset.validate_asset(amount=amount)
 
     @classmethod
-    def create(cls, owners_before, owners_after, metadata=None, asset=None):
+    def create(cls, creators, recipients, metadata=None, asset=None):
```
libscott: This is just for the purposes of the method name.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -205,19 +202,22 @@ definitions:
             type_id:
               type: integer
           description: |
-              Fulfillment of a condition_, or put a different way, this is a
-              payload that satisfies a condition in order to spend the associated
-              asset.
+              Fulfillment of an `Output.condition`_, or put a different way, this is a
+              payload that satisfies an output condition in order to spend
```
libscott: We're using "spend" in quite a few places*, if it's wrong we should examine it separately imo.

* Example: https://github.com/bigchaindb/bdb-transaction-cli/blob/master/docs/examples.rst#spend

bigchaindb/common/transaction.py
libscott: I think the docs are sufficient in this case:

```
creators (:obj:`list` of :obj:`str`): A list of keys that  
    represent the creators of the asset created by this    
    Transaction.                                           
recipients (:obj:`list` of :obj:`str`): A list of keys that
    represent the recipients of the asset created by this  
    Transaction.                                           
```

ttmc: In a CREATE transaction, if the "owners_before" list is just one public key, then the associated private key must be used to produce a signature and that signature must be in the transaction. That public key may be unrelated to the "creator" of the asset. It might be the public key of some agent that submits transactions on behalf of the creator, a "transaction builder & submitter agent."

If the "owners_before" list contains m public keys, then there must be m corresponding signatures. Again though, those might not be "creators" of the asset in question.

Idea: Instead of renaming "owners_before" to "creators" here, how about "tx_signers"? That is, the list of signers of the transaction in question.

As for "recipients," that's probably okay, so long as one correctly interprets what it means to "receive," i.e. one receives not ownership, but transfer permissions.

bigchaindb/common/schema/transaction.yaml
```diff
@@ -88,11 +88,10 @@ definitions:
         Type of the transaction:
 
         A ``CREATE`` transaction creates an asset in BigchainDB. This
-        transaction has outputs (conditions) but no inputs (fulfillments),
-        so a dummy fulfillment is used.
+        transaction has outputs but no inputs, so a dummy input is created.
 
         A ``TRANSFER`` transaction transfers ownership of an asset, by providing
-        fulfillments to conditions of earlier transactions.
+        fulfillments to conditions in outputs of earlier transactions.
```
sohkai: Fixed in https://github.com/bigchaindb/bigchaindb/pull/925/commits/36d0ff519efce04b5703c1d9edd040345e25f021

bigchaindb/common/schema/transaction.yaml
sohkai: Fixed in https://github.com/bigchaindb/bigchaindb/pull/925/commits/36d0ff519efce04b5703c1d9edd040345e25f021, related to https://github.com/bigchaindb/bigchaindb/pull/925#discussion_r92399996.

libscott:
So I'm not exactly sure why the tests are failing but I found another issue which needs to be fixed before merging:

https://github.com/bigchaindb/bigchaindb/blob/inputs-outputs/docs/server/source/data-models/crypto-conditions.md

libscott:
@ttmc https://github.com/bigchaindb/bigchaindb/pull/925/commits/7c12ee251dea8426b9653bd3bf63f8b4b1e6ca24

bigchaindb/common/transaction.py
```diff
@@ -661,8 +661,8 @@ def create(cls, creators, recipients, metadata=None, asset=None):
                     - Multiple inputs and outputs.
 
             Args:
-                creators (:obj:`list` of :obj:`str`): A list of keys that
-                    represent the creators of the asset created by this
+                tx_signers (:obj:`list` of :obj:`str`): A list of keys that
+                    represent the signers of the asset created by this
```
ttmc: The `tx_signers` are the signers of this CREATE transaction, not the "signers of the asset" :smile:

sohkai:
@ttmc I've fixed your suggestions in b64026441ba83ca6bbeb806b90e2581a42c2501d and the backticks in 8470854769a93a7a7fcbd1b3282c136559d42460.

tests/db/test_bigchain_api.py
```diff
@@ -1016,8 +1016,8 @@ def test_get_spent_single_tx_single_output(self, b, user_sk, user_pk):
 
         # check spents
         input_txid = owned_inputs_user1.txid
-        input_cid = owned_inputs_user1.cid
-        spent_inputs_user1 = b.get_spent(input_txid, input_cid)
+        input_idx = owned_inputs_user1.output
```
libscott: @sohkai what should we change it to

ttmc:
I approved because I'm leaving on vacation soon and don't want my approval to be a blocker. I see you've resolved most everything I commented on anyway.

PR #926, opened 2016-12-09
Tx ID as Asset ID

### LINKED ISSUE #697: Use transaction ID as asset ID

### LINKED ISSUE #708: Why use "id" and "data_id" for the same thing?
_From @sbellem on October 18, 2016 16:29_

``` python
class Asset(object):
    def __init__(self, data=None, data_id=None, divisible=False,
                 updatable=False, refillable=False):
        self.data = data
        self.data_id = data_id if data_id is not None else self.to_hash()
        self.divisible = divisible
        self.updatable = updatable
        self.refillable = refillable

    def to_dict(self):
        return {
            'id': self.data_id,
            'divisible': self.divisible,
            'updatable': self.updatable,
            'refillable': self.refillable,
            'data': self.data,
        }
```

doing so prevents one from passing directly the results of `Asset.to_dict()` to the Python built-in class initialization

_Copied from original issue: bigchaindb/bigchaindb-common#58_

### LINKED ISSUE #715: Revist `to_hash()` in Assets and Metadata
_From @sohkai on October 13, 2016 8:53_

Assets and metadata don't have hashes associated with them (yet?), so the naming of `to_hash()` to get a uuid is somewhat misleading.

_Copied from original issue: bigchaindb/bigchaindb-common#43_

### LINKED ISSUE #717: Rename `data_id` in the Metadata and Asset models
_From @r-marques on October 10, 2016 15:25_

I find the name `data_id` confusing in both the `Metadata` and `Asset` model.
Since they are already inside a namespace I propose to rename `data_id` to `id`.
This way we have:
- `Transaction.id`
- `Transaction.metadata.id`
- `Transaction.asset.id`

_Copied from original issue: bigchaindb/bigchaindb-common#39_

libscott:
Use the ID of the transaction that CREATEd the asset to refer to the asset.

**Note**: with these changes, you'll now need to provide either a unique asset or metadata payload (along with the list of signers) to create a unique TX ID. If you're going to be creating the same asset without a unique metadata payload, you should probably be thinking about using a divisible asset (and a refillable one, later).

- [x] Make code changes to use TX ID to refer to Asset ID
- [x] Fix tests
- [x] Docs changes (@ttmc I wasn't able to find too much that had to be changed with this change in the server docs, maybe you know of more places?)

Fixes #697, fixes #717. fixes #715, fixes #708.

### DIFF
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -105,7 +105,8 @@ definitions:
     description: |
         Description of the asset being transacted. In the case of a ``TRANSFER``
         transaction, this field contains only the ID of asset. In the case
-        of a ``CREATE`` transaction, this field may contain properties:
+        of a ``CREATE`` transaction, this field contains only a user payload
+        associated with the asset.
     additionalProperties: false
     properties:
       id:
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -105,8 +105,8 @@ definitions:
     description: |
         Description of the asset being transacted. In the case of a ``TRANSFER``
         transaction, this field contains only the ID of asset. In the case
-        of a ``CREATE`` transaction, this field contains only a user payload
-        associated with the asset.
+        of a ``CREATE`` transaction, this field contains only the user-defined
+        payload and asset-related keywords.
     additionalProperties: false
     properties:
       id:
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -106,7 +106,7 @@ definitions:
         Description of the asset being transacted. In the case of a ``TRANSFER``
         transaction, this field contains only the ID of asset. In the case
         of a ``CREATE`` transaction, this field contains only the user-defined
-        payload and asset-related keywords.
+        payload.
     additionalProperties: false
     properties:
       id:
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -107,11 +107,11 @@ definitions:
         transaction, this field contains only the ID of asset. In the case
         of a ``CREATE`` transaction, this field may contain properties:
     additionalProperties: false
-    required:
-    - id
     properties:
       id:
-        "$ref": "#/definitions/uuid4"
+        "$ref": "#/definitions/sha3_hexdigest"
+        description: |
+          ID of the transaction that created the asset.
       data:
         description: |
           User provided metadata associated with the asset. May also be ``null``.
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -511,6 +511,22 @@ class AssetLink(object):
         except TypeError:
             return cls()
 
+    @classmethod
+    def from_inputs(cls, *inputs):
+        """Creates an AssetLink based on given input :class:`~.Transactions`.
+
+        Args:
+            inputs (:class:`~.Transaction`): One or more input transactions
+
+        Returns:
+            An :class:`~.AssetLink` linking to the asset linked by the inputs.
+
+        Raises:
+            :exc:`AssetIdMismatch`: If the :attr:`inputs` do not link to the
+                same asset.
+        """
+        return cls(Asset.get_asset_id(list(inputs)))
+
     def to_dict(self):
         """Transforms the object to a Python dictionary.
 
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -581,8 +581,8 @@ class Transaction(object):
 
             Args:
                 operation (str): Defines the operation of the Transaction.
-                asset (:class:`~bigchaindb.common.transaction.Asset`): An Asset
-                    to be transferred or created in a Transaction.
+                asset (:class:`~.Asset`|:class:`~.AssetLink`): An Asset to be
+                    created or an AssetLink linking an asset to be transferred.
                 fulfillments (:obj:`list` of :class:`~bigchaindb.common.
                     transaction.Fulfillment`, optional): Define the assets to
                     spend.
@@ -599,10 +599,17 @@ class Transaction(object):
             raise ValueError('`operation` must be one of {}'
                              .format(allowed_ops))
 
-        # Only assets for 'CREATE' operations can be un-defined.
-        if (asset and not isinstance(asset, Asset) or
-                not asset and operation != Transaction.CREATE):
-            raise TypeError('`asset` must be an Asset instance')
+        # Assets for 'CREATE' and 'GENESIS' operations must be None or of Asset
+        # type and Assets for 'TRANSFER' operations must be of AssetLink type.
+        if (operation in [Transaction.CREATE, Transaction.GENESIS] and
+                asset is not None and
+                not isinstance(asset, Asset)):
+            raise TypeError(("`asset` must be an Asset instance for "
+                             "'{}' Transactions".format(operation)))
+        elif (operation == Transaction.TRANSFER and
+                not (asset and isinstance(asset, AssetLink))):
+            raise TypeError(("`asset` must be an valid AssetLink instance for "
+                             "'TRANSFER' Transactions".format(operation)))
 
         if conditions and not isinstance(conditions, list):
             raise TypeError('`conditions` must be a list instance or None')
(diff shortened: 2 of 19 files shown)


bigchaindb/common/schema/transaction.yaml
```diff
@@ -105,13 +105,14 @@ definitions:
     description: |
         Description of the asset being transacted. In the case of a ``TRANSFER``
         transaction, this field contains only the ID of asset. In the case
-        of a ``CREATE`` transaction, this field may contain properties:
+        of a ``CREATE`` transaction, this field contains only a user payload
+        associated with the asset.
     additionalProperties: false
-    required:
-    - id
```
sohkai: Looked around to see if we could change the asset model based on `CREATE` and `TRANSFER` transaction but [it'd look pretty ugly](https://github.com/json-schema/json-schema/issues/158#issuecomment-86124035) and might be slower.

Thoughts @libscott? Right now the asset schema is a bit of a lie ;).

libscott: I'm actually in favour of using the oneOf approach. That way it can match both `{operation:CREATE, asset:{...}}` and `{operation:TRANSFER, asset_id:...}}`. But the more complicated these constraints become, the worse the schema will perform...

sohkai: Turns out there's a few other places where we're generalizing properties to fit different use cases (e.g. `CREATE`'s inputs).

This is a fairly different problem at hand, so we'll leave it for another issue: https://github.com/bigchaindb/bigchaindb/issues/946.

tests/common/test_transaction.py
```diff
@@ -481,6 +475,15 @@ def test_asset_link_deserialization_with_empty_payload():
     assert asset_link == expected
 
 
+def test_asset_link_from_inputs(utx, transfer_utx):
+    from bigchaindb.common.transaction import AssetLink
+
+    expected = AssetLink(utx.id)
+    asset_link = AssetLink.from_inputs(utx, transfer_utx)
```
libscott: This appears to be the only time that `from_inputs` is used with multiple arguments. Is it neccesary?

bigchaindb/common/transaction.py
```diff
@@ -508,20 +503,20 @@ def validate_asset(self, amount=None):
                                   ' greater than one')
 
 
-class AssetLink(Asset):
+class AssetLink(object):
```
libscott: Is this guy still needed? What about just having a dict rather than a whole object to store an ID?

libscott:
Two possibilities:

1: polymorphic 'asset' key
```
{operation: CREATE, asset: {data: blah}}
{operation: TRANSFER, asset: {id: abc}}
```

2: separate keys
```
{operation: CREATE, asset: {blah}}
{operation: TRANSFER, asset_id: abc}
```

Interested to hear reasoning one way or the other.

But, from a logical point of view they are the same and it's not clear that we need the `Asset` or `AssetLink` classes anymore. Or at least, we won't once we remove the keywords from the `Asset`.

diminator:
are we also considering squatting of the `tx_id`? ie: what happens if someone makes 2 separate `CREATE` txs with the same content? Will they yield the same `tx_id` or do we need to force the user to provide a `UUID`?

bigchaindb/common/transaction.py
sohkai: Agreed, but let's move this to another PR after #962 is merged so that we can clean up both.

tests/common/test_transaction.py
sohkai: Not really... but then again, none of the common stuff is actually necessary in the server-side code (ala @sbellem's realization that changing the common's API is actually more driver-related work rather than server-related).

sohkai:
@diminator: See note on https://github.com/bigchaindb/bigchaindb/issues/697#issuecomment-267335527.

We basically decided the unique factor must either come in the metadata or asset, otherwise you should just be making divisible assets.

sohkai:
@libscott For the `asset` key, I think it's more elegant if we make it polymorphic. That way the root structure of a transaction stays the same between the two types.

bigchaindb/common/transaction.py
libscott: I'm thinking it could be a good idea to bust out some `namedtuple`s :smiling_imp:

tests/common/test_transaction.py
libscott: The API here looks a bit odd to me. It's not clear that you're supposed to pass in all your transactions so that it can validate that they have the same asset id.

sohkai: Fair enough, I guess this is actually doing more validation than building anything 😉 . I'll take a look into the driver to see how it uses `Transaction.transfer()` and try to remove this.

libscott:
LGTM

bigchaindb/common/schema/transaction.yaml
```diff
@@ -105,13 +105,14 @@ definitions:
     description: |
         Description of the asset being transacted. In the case of a ``TRANSFER``
         transaction, this field contains only the ID of asset. In the case
-        of a ``CREATE`` transaction, this field may contain properties:
+        of a ``CREATE`` transaction, this field contains only the user-defined
+        payload and asset-related keywords.
     additionalProperties: false
-    required:
-    - id
```
r-marques: The id is still required for `TRANSFER` transactions, right?

sohkai: It is, but we don't currently have a good way to split the requirements based on the operation (see #946).

bigchaindb/common/transaction.py
```diff
@@ -575,10 +575,17 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
...
-            raise TypeError('`asset` must be an Asset instance')
+        # Assets for 'CREATE' and 'GENESIS' operations must be None or of Asset
+        # type and Assets for 'TRANSFER' operations must be of AssetLink type.
+        if (operation in [Transaction.CREATE, Transaction.GENESIS] and
+                asset is not None and
+                not isinstance(asset, Asset)):
+            raise TypeError(("`asset` must be an Asset instance for "
+                             "'{}' Transactions".format(operation)))
+        elif (operation == Transaction.TRANSFER and
+                not (asset and isinstance(asset, AssetLink))):
```
r-marques: This can be simplified to `asset not isinstance(asset, AssetLink)`. It already returns `True` if asset is `None`

bigchaindb/common/transaction.py
```diff
@@ -575,10 +575,17 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
...
-                not asset and operation != Transaction.CREATE):
-            raise TypeError('`asset` must be an Asset instance')
+        # Assets for 'CREATE' and 'GENESIS' operations must be None or of Asset
+        # type and Assets for 'TRANSFER' operations must be of AssetLink type.
+        if (operation in [Transaction.CREATE, Transaction.GENESIS] and
+                asset is not None and
+                not isinstance(asset, Asset)):
+            raise TypeError(("`asset` must be an Asset instance for "
+                             "'{}' Transactions".format(operation)))
+        elif (operation == Transaction.TRANSFER and
```
r-marques: This is a big breaking change. `AssetLink` was something that only add meaning in the server code.
Clients would always pass an asset regardless and we would serialized it according to the operation.

Its probably worth it to open an issue in the bigchaindb-driver

sohkai: @sbellem and I had a quick conversation about this change, since it'll affect the driver most. There's [this issue](https://github.com/bigchaindb/bigchaindb-driver/issues/202) for fixing the driver once this is merged.

bigchaindb/common/transaction.py
sohkai: Did you mean:

```py
elif (operation == Transaction.TRANSFER and not isinstance(asset, AssetLink)):
    ...
```

?

The `not asset` part is still important to keep because we want to check that the AssetLink actually contains a link (with its bool operator), although this is not entirely obvious.

r-marques: Ok, I guess that is on way of doing it.

PR #939, opened 2016-12-13
Feat/905/abstract db changefeed

### LINKED ISSUE #900: Remove remaining RethinkDB query in block pipeline
I think we more or less know about this one, but I'm not sure if we have a good abstraction for this.

**Required for database abstraction.**

See https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/pipelines/block.py#L143.

### LINKED ISSUE #905: Create abstraction for changefeed
The changefeed is still coupled with RethinkDB for now. We need to design an abstraction so that multiple databases can implement and expose their own versions of the changefeed for use with BigchainDB.

r-marques:
resolves #905, resolves #900 

- `backend.changefeed` implements a wrapper around Node and the `__init__` method. Database specific changefeeds subclass this method and need to implement the `run_changefeed` and `run_forever` methods. Also the `ChangeFeed` now requires only a connection instead of a `Bigchain` instance. This file also implements the generic function for `get_changefeed` which will return the correct changefeed depending on the connection.
- `backend.rethinkdb.changefeed` implements the actual rethinkdb changefeed and get_changefeed for rethinkdb
- removed `pipelines.utils` since was no longer needed
- moved `tests/pipelines/test_utils.py` to `tests/backend/test_changefeed.py` (some tests are still rethinkdb specific)
- removed some harcoded rethinkdb queries in the block pipeline
- updated the pipeline code to use the new abstraction


sohkai:
> - removed some harcoded rethinkdb queries in the block pipeline

As a note, I think @rhsimplex had done some work on this in https://github.com/bigchaindb/bigchaindb/pull/903/ and is related to #860. I think similar things were done, but the only thing was to add some tests to double check the behaviour.

bigchaindb/backend/changefeed.py
```diff
@@ -1 +1,88 @@
 """Changefeed interfaces for backends."""
+
+from functools import singledispatch
+from multipipes import Node
```
sbellem: The pep 8 police says:

> Imports should be grouped in the following order:
>
> * standard library imports
> * related third party imports
> * local application/library specific imports
>
> You should put a blank line between each group of imports.

bigchaindb/backend/changefeed.py
```diff
@@ -1 +1,88 @@
...
+    is volatile. This class is a helper to create changefeeds. Moreover,
+    it provides a way to specify a ``prefeed`` of iterable data to output
+    before the actual changefeed.
+    """
+
+    INSERT = 1
+    DELETE = 2
+    UPDATE = 4
+
+    def __init__(self, table, operation, *, prefeed=None, connection=None):
```
sbellem: Nice! [PEP 3102](https://www.python.org/dev/peps/pep-3102/) -- Keyword-Only Arguments

bigchaindb/backend/changefeed.py
```diff
@@ -1 +1,88 @@
...
+        """Create a new ChangeFeed.
+
+        Args:
+            table (str): name of the table to listen to for changes.
+            operation (int): can be ChangeFeed.INSERT, ChangeFeed.DELETE, or
+                ChangeFeed.UPDATE. Combining multiple operation is possible
+                with the bitwise ``|`` operator
+                (e.g. ``ChangeFeed.INSERT | ChangeFeed.UPDATE``)
+            prefeed (iterable): whatever set of data you want to be published
```
sbellem: not a blocker at all, but just as a tip: 

```rst
prefeed (:class:`~collections.abc.Iterable`):
```

r-marques:
@sohkai Yes there is an issue to remove that query. @rhsimplex should merge his pr and then I can remove it from here also

bigchaindb/pipelines/election.py
```diff
@@ -50,10 +52,6 @@ def requeue_transactions(self, invalid_block):
         return invalid_block
 
 
-def get_changefeed():
-    return ChangeFeed(table='votes', operation=ChangeFeed.INSERT)
```
sohkai: For the sake of testing (in the tests below), I could see the usefulness of keeping these as separate functions (in each pipeline) to avoid duplicating how we get a pipeline's changefeed later.

bigchaindb/backend/changefeed.py
```diff
@@ -1 +1,89 @@
...
+        Args:
+            table (str): name of the table to listen to for changes.
+            operation (int): can be ChangeFeed.INSERT, ChangeFeed.DELETE, or
+                ChangeFeed.UPDATE. Combining multiple operation is possible
+                with the bitwise ``|`` operator
+                (e.g. ``ChangeFeed.INSERT | ChangeFeed.UPDATE``)
+            prefeed (:class:`~collections.abc.Iterable`): whatever set of data
+                you want to be published first.
+            connection (:class:`~bigchaindb.backend.connection.Connection`):
+                A connection to the database (can be None).
```
sohkai: It's not something we follow consistently, but I think a nice way of noting an argument as optional is to tag it in the object description (as seen in the [napoleon docs](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google)):

```python
"""
...
connection (:class:`~bigchaindb.backend.connection.Connection`, optional):
    A connection to the database.
"""
```

We could also add that the `connection` defaults to a new `Connection` instance.

bigchaindb/backend/changefeed.py
```diff
@@ -1 +1,90 @@
...
+        This method is responsible for first feeding the prefeed to the
+        outqueue and after that starting the changefeed and recovering from any
+        errors that may occur in the backend.
+        """
+        raise NotImplementedError
+
+    def run_changefeed(self):
+        """Backend specific method to run the changefeed.
+
+        The changefeed is is usually a backend cursor that is not closed when
```
vrde: remove one `is`

PR #942, opened 2016-12-13
mongodb integration

sbellem:
to replace #890


sohkai:
Just something small I noticed: should we also be testing mongodb on 3.4?

PR #943, opened 2016-12-13
include all .yaml files in package_data

libscott:
Use a wildcard to include all .yaml files in `package_data`.

http://setuptools.readthedocs.io/en/latest/setuptools.html


ttmc:
Yay wildcards FTW.

:+1:

PR #944, opened 2016-12-13
Treat warnings as errors in documentation builds

libscott:
This will cause the build to fail if the documentation contains any warnings, like an invalid section reference. The tests will now fail if the docs dont build without warnings: https://github.com/bigchaindb/bigchaindb/blob/master/tests/test_docs.py. This is both during travis build and locally.

### DIFF
--- a/docs/root/Makefile
+++ b/docs/root/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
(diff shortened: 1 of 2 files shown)


ttmc:
I looked up the `-W` option in [the Spinx docs](http://www.sphinx-doc.org/en/1.5.1/invocation.html). For convenience, those docs (for Sphinx version 1.5.1) say:

**-W**
Turn warnings into errors. This means that the build stops at the first warning and sphinx-build exits with exit status 1.

If that's what you want, then :+1:

Issue #728, opened 2016-10-20
`owners_before` and `owners_after` should be derived from Cryptoconditions


r-marques:
_From @TimDaub on September 21, 2016 9:19_

@diminator gave a hint here: https://github.com/bigchaindb/bigchaindb-common/pull/4#r76391859

_Copied from original issue: bigchaindb/bigchaindb-common#15_

r-marques:
_From @diminator on September 29, 2016 15:40_

this will be troublesome/impossible for `Thresholdconditions` as the internals are masked, can only be done for `Fulfillments`

note that `owners_before` and `owners_after` are required mainly for:
- indexing
- `get_owned_ids`

libscott:
This SHOULD be part of 1.0 guarantees IMO.

PR #945, opened 2016-12-13
[Spec] [WIP] Remove Input.owners_before

libscott:
"owners_before" is inaccurate and perhaps also unneccesary, we should consider removing it. However, in the case of a threshold condition with `m` public keys it is useful to indicate which `n` keys were involved in fulfillment.

Questions:

* Does there exist a function: `did_sign(signature, pubkey)` which would indicate if a key was involved in a signature?
* From a product perspective (@TimDaub), do we need to keep the list of people that signed, perhaps for the purposes of auditability?
* If we do, how can we be sure that the list of public keys given in `owners_after` is actually the real list? The signatorys could enter the wrong list, so perhaps it's meaningless. And, if it's verifiable via a `did_sign` method as described above, we don't need it because we have the list of `m` keys in the `output`, alongside the condition.
* Speaking of which, I suppose that list is not verified either, what are the implications?

### DIFF
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -179,14 +179,9 @@ definitions:
       ``CREATE`` transaction, a fulfillment may provide no ``input``.
     additionalProperties: false
     required:
-    - owners_before
     - input
     - fulfillment
     properties:
-      owners_before:
-        "$ref": "#/definitions/owners_list"
-        description: |
-            List of public keys of the previous owners of the asset.
       fulfillment:
         anyOf:
         - type: object


diminator:
> Does there exist a function: did_sign(signature, pubkey) which would indicate if a key was involved in a signature?

All info of the entire cryptocondition is encoded in the fulfillment URI. No need for duplicating this in `owners_before`



> If we do, how can we be sure that the list of public keys given in owners_after is actually the real list?

This is the entire discussion of splitting up the attributes (to be indexed) from the topology of a condition... see especially #677 and there are more related issues in the transaction model update slides

Important to note is that `owners_before` is currently only used by the `CREATE` to combine it with a `TRANSFER` in one shot. I believe we should consider not combining them - only have a pure `CREATE`. This way we can get rid of `owners_before`

ttmc:
A pure `CREATE` transaction is a nice idea, but we need to be careful to ensure that the transaction gets signed on the client side. If "owners_before" goes away, we will still need a list of the public keys who signed the transaction.

libscott:
> This is the entire discussion of splitting up the attributes (to be indexed) from the topology of a condition... see especially #677 and there are more related issues in the transaction model update slides

Ah, so I didn't understand this. we had this conversation, about `owners_after` not being verified, and @vrde suggested that we don't need to include the condition URI in the TX since it can be generated deterministically from the public keys. So that solves that problem.

I just need to understand the `owners_before` create & transfer thing then.

diminator:
> @vrde suggested that we don't need to include the condition URI in the TX since it can be generated deterministically from the public keys. So that solves that problem.

that's actually not true. you provide either the compressed condition by URI or the extended by details.

the compressed version/URI/hash: **always needed** for fast verification that the fulfillment matches the input condition

the extended version/details/topology: optional to provide DB-level indices on pubkeys, hashlocks, etc. Needs to match the URI always

sohkai:
Related: #728.

r-marques:
> Speaking of which, I suppose that list is not verified either, what are the implications?

When fulfilling a transaction we make sure that the private keys provided belong to the the list of `owners_before` https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L930 https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L968

sohkai:
Also relevant: #677.

TimDaub:
I'll create a Cryptoconditions DSL ticket and link from this PR to it later.

Closed for now.

PR #950, opened 2016-12-13
More test abstractions for multiple databases

sohkai:
Removes some leftover coupling in the tests with RethinkDB.

**[Proposal]**: I've moved anything RethinkDB-specific into its own file or folder under the higher-level logical grouping of tests, rather than moving them all into their own folder. I'm not in love with how this is organized, but because of the issues surrounding the `tests/db/` folder, I think this is the best way to structure the tests for now.

**Todo**:
- [x] Skip tests unrelated to the currently selected database (ended up going with a pytest hook)
- [x] Move `tests/db/rethinkdb/test_schema.py` to `tests/backends/rethinkdb`


tests/commands/rethinkdb/test_commands.py
```diff
@@ -0,0 +1,123 @@
+import pytest
```
sohkai: RethinkDB-specific tests moved from the old `tests/test_commands.py`.

tests/commands/conftest.py
```diff
@@ -0,0 +1,46 @@
+import pytest
```
sohkai: Fixtures moved from the old `tests/test_commands.py`.

r-marques:
Regarding your todo item http://doc.pytest.org/en/latest/example/pythoncollection.html#customizing-test-collection-to-find-all-py-files

Assuming the tests are well organized, it should be easy to just add to `collect_ignore` the folders we want to skip

r-marques:
There is also `tests/db/rethinkdb/test_schema.py` that should be moved to `tests/backends/rethinkdb`

PR #952, opened 2016-12-14
Extra renames and small fixes for inputs-outputs

### LINKED ISSUE #822: Rename non-cryptoconditions classes (Condition, Fulfillment) to different names

sohkai:
@libscott this ended up being larger than expected (after finding some usages of fulfillments / conditions weren't changed in `common/transaction.py`), so I've pulled this out into its own PR. Feel free to rollback any commit that you don't want merged :).

**[Proposal]**: I've used `input_` to avoid hiding the built-in for now, but I'd agree if you think this looks ugly (since we're using it as argument names, etc). I guess we could change it to `in` / `out`?

Fixes #822.

### DIFF
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -173,8 +173,8 @@ definitions:
   input:
     type: "object"
     description:
-      An input spends a previous output, by providing a fulfillment to the output's
-      condition.
+      An input spends a previous output, by providing one or more fulfillments
+      that fulfill the conditions of the previous output.
     additionalProperties: false
     required:
     - owners_before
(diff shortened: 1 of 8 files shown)


libscott:
Stellar :star2:

sohkai:
Lol @ttmc, @libscott I hope you two didn't miss the `common/transaction.py` changes!

libscott:
> Lol @ttmc, @libscott I hope you two didn't miss the common/transaction.py changes!

You mean because Github is lazy and doesn't bother showing you the large diff unless you press a button? I saw that...

TimDaub:
Also resolves: #822

PR #953, opened 2016-12-14
remove rethinkdb calls

rhsimplex:
Remove explicit rethinkdb calls.  A little slower, because we can't just tear down tables each time; the whole db has to be regenerated

### DIFF
--- a/tests/db/conftest.py
+++ b/tests/db/conftest.py
@@ -7,12 +7,12 @@ Tasks:
 """
 
 import pytest
-import rethinkdb as r
 
 from bigchaindb import Bigchain
 from bigchaindb.backend import connect, schema
 from bigchaindb.common import crypto
-from bigchaindb.common.exceptions import DatabaseAlreadyExists
+from bigchaindb.common.exceptions import (DatabaseAlreadyExists,
+                                          DatabaseDoesNotExist)
 
 
 USER2_SK, USER2_PK = crypto.generate_key_pair()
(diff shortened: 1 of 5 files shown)


tests/db/conftest.py
```diff
@@ -30,23 +30,22 @@ def setup_database(request, node_config):
...
-    if conn.run(r.db_list().contains(db_name)):
-        conn.run(r.db_drop(db_name))
-
     try:
-        schema.init_database()
+        schema.init_database(conn)
     except DatabaseAlreadyExists:
         print('Database already exists.')
+        schema.drop_database(conn, db_name)
+        schema.init_database(conn)
```
sohkai: Might be nice to restructure this to avoid duplicating the `init_database` call:

```python
try:
    schema.drop_database(conn, db_name)
except DatabaseDoesNotExist:
    pass
schema.init_database(conn)
```

tests/db/conftest.py
```diff
@@ -30,23 +30,22 @@ def setup_database(request, node_config):
...
         conn = connect()
+        print('Deleting `{}` database'.format(db_name))
         try:
-            conn.run(r.db_drop(db_name))
-        except r.ReqlOpFailedError as e:
-            if e.message != 'Database `{}` does not exist.'.format(db_name):
+            schema.drop_database(conn, db_name)
+        except DatabaseDoesNotExist as e:
+            if str(e) != 'Database `{}` does not exist'.format(db_name):
                 raise
```
sohkai: With the more granular exception, maybe we could just ignore this exception:

```python
except DatabaseDoesNotExist:
    pass
```

tests/db/conftest.py
```diff
@@ -60,11 +59,12 @@ def cleanup_tables(request, node_config):
...
         try:
-            conn.run(r.db(db_name).table('bigchain').delete())
-            conn.run(r.db(db_name).table('backlog').delete())
-            conn.run(r.db(db_name).table('votes').delete())
-        except r.ReqlOpFailedError as e:
-            if e.message != 'Database `{}` does not exist.'.format(db_name):
+            schema.drop_database(conn, db_name)
+            schema.create_database(conn, db_name)
+            schema.create_tables(conn, db_name)
+            schema.create_indexes(conn, db_name)
```
sohkai: Not sure if we need to recreate the db, tables, and indexes again here; wouldn't `setup_database` handle this with `init_database()`?

rhsimplex: No, that's the annoying thing -- in the RDB specific verison, you could do that, but since the API doesn't expose dropping tables (just the whole db) you have to set it up from scratch each time.

rhsimplex:
Hey @sohkai fixed the redundant cleanup method, and made it consistent accross all tests.  An unrelated test is broken now, and I have no idea why.  No time to work on this anymore =/

sohkai:
> An unrelated test is broken now, and I have no idea why.

Apparently this is something to do with `commands.bigchain.run_init()`, but looks like it should be unrelated to the changes and may be caused by a bad local environment.

PR #961, opened 2016-12-16
Fixed HTTP API docs re/ API Endpoint

ttmc:
I noticed that the documentation page about the HTTP API needed some fixes:

- It still mentioned the now-gone `api_endpoint` setting (indirectly, so a grep for "api_endpoint" didn't find it). Oops.
- It went into way too much detail, up front, about how the API Root URL is determined. Most people visiting that page don't care about that: they already know the API Root URL and just want to know what the endpoints are. I moved the stuff about how the API Root URL is determined to the end of the page.
- I also added some clarification about how the default is to bind to localhost, and what to do if they want to make the API publicly accessible.

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -5,31 +5,23 @@ The HTTP Client-Server API
 
    The HTTP client-server API is currently quite rudimentary. For example,
    there is no ability to do complex queries using the HTTP API. We plan to add
-   querying capabilities in the future.
+   more querying capabilities in the future.
 
-When you start Bigchaindb using `bigchaindb start`, an HTTP API is exposed at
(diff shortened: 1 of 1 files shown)


docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -5,31 +5,23 @@ The HTTP Client-Server API
...
 
 API Root URL
 ------------
 
-If you send an HTTP GET request to e.g. ``http://localhost:9984`` 
+If you send an HTTP GET request to the API Root URL
+e.g. ``http://localhost:9984`` 
 or ``http://apihosting4u.net:9984``
 (with no ``/api/v1/`` on the end), 
 then you should get an HTTP response
```
sohkai: The line length here is super funky haha, it's almost haiku level short.

PR #962, opened 2016-12-16
Remove asset keywords

libscott:
This change removes keywords from the `Asset`.

* `divisible`: This is indicated by the amount of the asset. 
* `refillable`: This logic will be implemented in an asset Policy
* `updatable`: This logic will be implemented in an asset Policy


r-marques:
resolves #841

libscott:
Huzzah!

PR #963, opened 2016-12-16
Rename owner_after to public_key

sbellem:


sbellem:
> I'm ok with this, assuming this is what's been decided in #830?

perhaps @ttmc can confirm?

ttmc:
@sohkai wrote:

> I'm ok with this, assuming this is what's been decided in #830?

@sbellem wrote:

> perhaps @ttmc can confirm?

I've lost track. @libscott will probably know. I'll add him as a reviewer.

sohkai:
@ttmc We've changed "owners_after" to "public_key" in the code in #925, but I'm not sure if the API should reflect that given the conversations surrounding `get_unspents` in #830 and elsewhere.

libscott:
This reflects the approach we are taking in #925, but IMO "owners" is still kind of an open question so I'd recommend merging this and if at some point in the future we decide that "owners" is the appropriate terminology we can change it.

ttmc:
@sbellem I see that you already merged this pull request. It occurred to me that the HTTP API docs also need to be updated accordingly. See 

https://docs.bigchaindb.com/projects/server/en/master/drivers-clients/http-client-server-api.html#get-unspents

That could be done in a new pull request.

PR #965, opened 2016-12-18
Consolidated all tests-related docs in one place

ttmc:
As suggested by @sohkai, I consolidated all the docs about writing and running all tests (unit tests and integration tests) in one place: `bigchaindb/tests/README.md`.

* Moved documentation from `PYTHON_STYLE_GUIDE.md` and `docs/server/source/dev-and-test/running-all-tests.md` to that file.
* Updated links in `CONTRIBUTING.md`
* Synthesized and edited the consolidated docs in `bigchaindb/tests/README.md`

To Do:

- [ ] I think the docs about running the tests using Docker Compose should be updated. @sbellem can you do that? Maybe by pushing to this branch, or in a follow-up pull request.


docs/server/source/dev-and-test/index.rst
```diff
@@ -5,4 +5,4 @@ Develop & Test BigchainDB Server
    :maxdepth: 1
 
    setup-run-node
-   running-unit-tests
+   running-all-tests
```
sohkai: Hmm, I think this section isn't so important (especially now that we're changing its link), but let's keep it for now.

I'm of the opinion that we don't need to add too much dev documentation in the readthedocs, since that's what our repo and Github is for.

tests/README.md
```diff
@@ -1,18 +1,91 @@
...
-## Test Structure
+## The bigchaindb/tests/ Folder
 
-Generally all tests are meant to be unit tests, with the exception of those in the [`integration/` folder](./integration/).
+The `bigchaindb/tests/` folder is where all the tests for BigchainDB Server live. Most of them are unit tests. Integration tests are in the [`bigchaindb/tests/integration/` folder](./integration/).
 
 A few notes:
 
-- [`common/`](./common/) contains self-contained tests only testing
+- [`bigchaindb/tests/common/`](./common/) contains self-contained tests only testing
```
sohkai: I actually think using the full path here is confusing, since you're already in the `tests/` folder. I think either:

- `tests/common/`; or
- `common/`

would be more clear.

tests/README.md
```diff
@@ -1,18 +1,91 @@
-# Tests
+# BigchainDB Server Tests
 
-## Test Structure
+## The bigchaindb/tests/ Folder
```
sohkai: I think

> ## The `tests/` Folder

might sound better, since you're already in the `bigchaindb/` repo (they might be confused that it means the *actual* `bigchaindb/` folder).

docs/server/source/dev-and-test/index.rst
ttmc: Yeah, we can probably remove this whole section some day, but let's keep it for now.

tests/README.md
ttmc: Okay I will change that as suggested.

sohkai:
> I think the docs about running the tests using Docker Compose should be updated. 

AFAIK, those instructions still work. If we wanted to, we could add a section on testing with mongo now.

PR #978, opened 2016-12-20
enable codecov for mongodb

r-marques:
Enable code coverage for mongodb code


r-marques:
I am not sure how codecov works because right now we run codecov 3 times. Two for rethinkdb (python3.4, python3.5) and once for mongodb.

When running the tests for rethinkdb, no tests are ran for mongodb so codecov should say that mongodb lines are misses and vice-versa.

I ran the locally coverage with mongodb and I do get 100% coverave for `bigchaindb/backend/mongodb/schema.py` will codecov in this pr only reports 34%

PR #982, opened 2016-12-21
Remove Asset and AssetLink

### LINKED ISSUE #974: Refactor common.Asset and common.AssetLink
Once #962 and #926 are merged, these objects aren't really useful anymore as they simply hold a dict and don't contain logic. We should be able to replace them with something simpler.

sohkai:
Fixes #974.

Moves some of the on-create validation to other model's init's and the one remaining static method to `Transaction` (where, arguably, it fits better).


bigchaindb/common/transaction.py
```diff
@@ -1126,6 +980,40 @@ def __str__(self):
         return Transaction._to_str(tx)
 
     @staticmethod
+    def get_asset_id(transactions):
```
sohkai: Moved from `Asset`.

bigchaindb/models.py
```diff
@@ -40,20 +40,28 @@ def validate(self, bigchain):
         inputs_defined = all([ffill.tx_input for ffill in self.fulfillments])
```
sohkai: There's a bit of duplication in what I added to this method, but as @r-marques and I discussed, this method is somewhat confused (it's doing too much). Until we get a handle on when and what we should be validating in the server, I think this is OK like this.

bigchaindb/common/transaction.py
```diff
@@ -647,6 +498,8 @@ def create(cls, owners_before, owners_after, metadata=None, asset=None):
             raise ValueError('`owners_before` list cannot be empty')
         if len(owners_after) == 0:
             raise ValueError('`owners_after` list cannot be empty')
+        if not (asset is None or isinstance(asset, dict)):
```
r-marques: Not sure if this check is needed since this check will be performed when calling `cls(...)`

bigchaindb/models.py
```diff
@@ -40,20 +40,28 @@ def validate(self, bigchain):
...
         if self.operation in (Transaction.CREATE, Transaction.GENESIS):
+            # validate asset
+            if self.asset['data'] is not None and not isinstance(self.asset['data'], dict):
+                raise TypeError(('`asset.data` must be a dict instance or '
+                                 'None for `CREATE` transactions'))
             # validate inputs
             if inputs_defined:
                 raise ValueError('A CREATE operation has no inputs')
-            # validate asset
-            amount = sum([condition.amount for condition in self.conditions])
```
r-marques: I guess this was not doing what I thought

bigchaindb/models.py
```diff
@@ -74,29 +82,26 @@ def validate(self, bigchain):
...
-            asset = bigchain.get_asset_by_id(asset_id)
-            # validate the asset
-            asset.validate_asset(amount=input_amount)
             # validate the amounts
-            output_amount = 0
-            for condition in self.conditions:
-                if condition.amount < 1:
-                    raise AmountError('`amount` needs to be greater than zero')
-                output_amount += condition.amount
+            if any(any(condition.amount < 1 for condition in conditions) for
```
r-marques: I don't think that this is needed.
You already validate the amounts of this transaction outputs in https://github.com/bigchaindb/bigchaindb/pull/982/files#diff-9f1ab5b5daf1b3072e2123b6939bfb07R51

And the input conditions where already validated when the input transation was submitted

sohkai: > You already validate the amounts

Ah, but that validation was only for `CREATE` and `GENESIS` transactions.

> And the input conditions where already validated when the input transaction was submitted

Hmmm, actually that's a good point... the input conditions *must* be valid since we get them from valid blocks.

r-marques:
In the end moving the types and amount validation code simplified the code 👍

bigchaindb/common/transaction.py
```diff
@@ -598,20 +458,11 @@ def __init__(self, operation, asset, fulfillments=None, conditions=None,
 
         self.version = version if version is not None else self.VERSION
         self.operation = operation
-        self.asset = asset or Asset()
+        self.asset = asset
```
libscott: Yay

PR #983, opened 2016-12-21
Initialize mongodb replica set

r-marques:
This pull request adds code to initialize a mongodb replica set when running `bigchaindb init`.
We always need to have a replica set even if we are only using one node so that we have an oplog to connect the changefeed to.

In order to be able to initialize a replica set if is necessary that the `mongod` instance is running with the `replSet` option enabled. This can be done in two ways:
- through the command line with `mongod --replSet=<any name>`
- through the config file by default located in `/etc/mongod.conf`:
```yaml
replication:
  replSetName: <any name>
```

How the initialization works:
- check if the replSet option is enable if not raise exception
- get the replica set name
- try to initialize
- if already initialized, skip
- if initializing for the first time wait until initialization is complete and the database is writeable

### DIFF
--- a/bigchaindb/backend/mongodb/schema.py
+++ b/bigchaindb/backend/mongodb/schema.py
@@ -1,8 +1,10 @@
 """Utils to initialize and drop the database."""
 
+import time
 import logging
 
 from pymongo import ASCENDING, DESCENDING
+from pymongo import errors
 
 from bigchaindb import backend
 from bigchaindb.common import exceptions
@@ -23,6 +25,9 @@ def create_database(conn, dbname):
     logger.info('Create database `%s`.', dbname)
     # TODO: read and write concerns can be declared here
     conn.conn.get_database(dbname)
+
+    # initialize the replica set
+    initialize_replica_set(conn)
 
 
 @register_schema(MongoDBConnection)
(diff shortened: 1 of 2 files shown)


bigchaindb/backend/mongodb/schema.py
```diff
@@ -85,3 +90,68 @@ def create_votes_secondary_index(conn, dbname):
...
+
+    # I did not find a better way to do this for now.
+    # To check if the database is ready we will poll the mongodb logs until
+    # we find the line that says the database is ready
+    logger.info('Waiting for mongodb replica set initialization')
+    while True:
+        logs = conn.conn.admin.command('getLog', 'rs')['log']
+        for line in logs:
+            if 'database writes are now permitted' in line:
+                return
```
sohkai: A cool trick I recently discovered that removes a layer of nesting (and may improve the readability?):

```python
logs = ...
if any('...' in line for line in logs):
    return
```

Should be just as performant, as `any` will terminate the generator on first `TRUE`.

bigchaindb/backend/mongodb/schema.py
```diff
@@ -85,3 +90,68 @@ def create_votes_secondary_index(conn, dbname):
...
+        conn.conn.admin.command('replSetInitiate', config)
+    except errors.OperationFailure as exc_info:
+        if exc_info.details['codeName'] == 'AlreadyInitialized':
+            logger.info('Replica set already initialized')
+            return
+        else:
+            raise
+
+    _wait_for_replica_set_initialization(conn)
+    logger.info('Initialized replica set')
```
sohkai: The try/except may or may not be more readable with an `else` on the `try`:

```python
try:
    conn.conn...
except errors.OperationFailure as exc_info:
    if exc_info.details['codeName'] == 'AlreadyInitialized':
        logger.info('Replica set already initialized')
    else:
        raise
else:
    _wait_for_replica_set_initialization(conn)
    logger.info('Initialized replica set')
```

tests/backend/mongodb/test_schema.py
```diff
@@ -104,3 +105,36 @@ def test_drop():
...
+
+def test_get_replica_set_name():
+    from pymongo.database import Database
+    from bigchaindb import backend
+    from bigchaindb.backend.mongodb.schema import _get_replica_set_name
+    from bigchaindb.common.exceptions import ConfigurationError
+
+    conn = backend.connect()
+
+    # no replSet option set
```
sohkai: I think this could be its own test, since it's testing an error condition.

r-marques:
No we don't catch those errors. I guess that would have be done at an higher level when calling `bigchaindb init` or `bigchaindb start`.

There are several ways it can raise exceptions right now: backends not running, misconfigurations, ...

PR #986, opened 2016-12-22
Rename config.database.name to config.database.dbname

### LINKED ISSUE #949: Rename configuration's database.name to database.dbname
`dbname` is a more obvious name than `name` for what this option actually does.

This will also allow us to fix the signature of `backend.connect()` so that `backend.connect()` is equivalent to `backend.connect(**bigchaindb.config['database'])` ([see related comment](https://github.com/bigchaindb/bigchaindb/pull/939#discussion_r92162224)).

sohkai:
Fixes #949.

Doing this rename also allows us to have a simpler implementation of `backend.connect()`, whereby we update the current node's config with any given keyword arguments and pass the result to instantiate a `Connection` instance.

As such, the `Connection` interface has been broadened to accept other arguments (as necessary; nothing takes advantage of this currently) and the `config.database` settings can be customized for each backend.

### DIFF
--- a/bigchaindb/backend/changefeed.py
+++ b/bigchaindb/backend/changefeed.py
@@ -44,11 +44,7 @@ class ChangeFeed(Node):
         self.prefeed = prefeed if prefeed else []
         self.table = table
         self.operation = operation
-        if connection:
-            self.connection = connection
-        else:
-            self.connection = bigchaindb.backend.connect(
-                **bigchaindb.config['database'])
+        self.connection = connection if connection else bigchaindb.backend.connect()
 
     def run_forever(self):
         """Main loop of the ``multipipes.Node``
--- a/bigchaindb/backend/changefeed.py
+++ b/bigchaindb/backend/changefeed.py
@@ -44,7 +44,7 @@ class ChangeFeed(Node):
         self.prefeed = prefeed if prefeed else []
         self.table = table
         self.operation = operation
-        self.connection = connection if connection else bigchaindb.backend.connect()
+        self.connection = connection or bigchaindb.backend.connect()
 
     def run_forever(self):
         """Main loop of the ``multipipes.Node``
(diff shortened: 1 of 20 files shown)


bigchaindb/backend/changefeed.py
```diff
@@ -44,11 +44,7 @@ def __init__(self, table, operation, *, prefeed=None, connection=None):
         self.prefeed = prefeed if prefeed else []
         self.table = table
         self.operation = operation
-        if connection:
-            self.connection = connection
-        else:
-            self.connection = bigchaindb.backend.connect(
-                **bigchaindb.config['database'])
+        self.connection = connection if connection else bigchaindb.backend.connect()
```
r-marques: Maybe use `connection or bigchaindb.backend.connect()`?

bigchaindb/core.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, public_key=None, private_key=None, keyring=[], connection=Non
         self.nodes_except_me = keyring or bigchaindb.config['keyring']
         self.backlog_reassign_delay = backlog_reassign_delay or bigchaindb.config['backlog_reassign_delay']
         self.consensus = BaseConsensusRules
-        self.connection = connection if connection else backend.connect(**bigchaindb.config['database'])
+        self.connection = connection if connection else backend.connect()
```
r-marques: same here `or` instead of `if else`

r-marques:
I think the docker files are using some env variables. We need to make sure that if they are using `BIGCHAINDB_DATABASE_NAME` they need to be updated

sbellem:
As mentioned in https://github.com/bigchaindb/bigchaindb/issues/949#issuecomment-268708953, I think we are mixing different concerns with this renaming task, and sacrificing the configuration from the end user point of view. The most obvious example is the variable name `'BIGCHAINDB_DATABASE_DBNAME'`. I mean, why?

sohkai:
Closing this until we revisit the configuration in detail later.

PR #988, opened 2016-12-22
Rename builtin-in input override in commands

### LINKED ISSUE #897: Avoid overriding the `input()` built-in in commands
See https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/commands/bigchain.py#L37.

We should rename this to something else, so as to avoid hiding the built-in and causing potential confusion.

sohkai:
Fixes #897.

Also allows the input override to not require a prompt (similar to how the built-in `input` behaves).


r-marques:
Looks good, you just need to fix the tests

👍

sohkai:
I swear all these tests were passing locally D:

PR #991, opened 2016-12-22
Rename any *util.py to *utils.py

### LINKED ISSUE #168: naming convention: util.py or utils.py
Without questioning why there are two util(s) modules, the naming could be consistent, that is, either the plural form is used, or the singular form. 

Or is there a reason for the different approaches?

Currently:

`bigchaindb/util.py`

and

`bigchaindb/db/utils.py`

sohkai:
Fixes #168.


sbellem:
:+1:

Issue #1010, opened 2016-12-30
docker build --tag local-bigchaindb . fails


josephwinston:
[Gist](https://gist.github.com/josephwinston/6a56be07f7f086dc52cc95db3f9faeb7) of failure.

josephwinston:
See https://github.com/bigchaindb/bigchaindb/pull/1011

r-marques:
Hi @josephwinston 

Thanks for spotting this issue.
I will keep this issue open until #1011 is merged

sohkai:
Closing as #1011 was merged.

PR #1011, opened 2016-12-30
Fix use of locales for python et al

josephwinston:
See https://github.com/bigchaindb/bigchaindb/issues/1010

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,16 @@
 FROM rethinkdb:2.3
 
-RUN apt-get update
+# From http://stackoverflow.com/a/38553499
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,16 @@
 FROM rethinkdb:2.3
 
-RUN apt-get update
+# From http://stackoverflow.com/a/38553499
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
```
ttmc: @josephwinston  Why did you do `apt-get update` and `DEBIAN_FRONTEND=noninteractive apt-get install -y locales` in the same line? It's not "wrong," I'm just curious if there's some reason that I don't grok.

josephwinston: Each command in the Dockerfile generates a new image.  I grouped those commands together as one so that a single image is produced if the command succeed.  This follows the general pattern found at https://docs.docker.com/engine/reference/builder/

ttmc: Ah, that makes sense. Thanks for explaining!

@sbellem is our local Docker expert and he's away on holidays for a little while longer, so don't worry if he doesn't review this PR right away.

r-marques:
This seems to be a problem in python-rapidjson https://github.com/kenrobbins/python-rapidjson/pull/52

PR #1014, opened 2017-01-03
update pysha3 to 1.0b1, sha3 NIST 2015 standard

### LINKED ISSUE #884: Should we increase the required version of pysha3?
Just noticed that pysha3 1.0b1 has been out for a while.  See some excerpts from their change log and readme. (from https://github.com/tiran/pysha3)

Changelog
=========

pysha3 1.0b1
------------

*Release date: 01-May-2016*

- Update backend to use the latest Keccak Code Package. pysha3 now implements
  the official NIST standard. The old Keccak hashes are available with
keccak prefix.

Updates since pysha 0.3
==================

**pysha3 1.0 is not compatible with pysha3 0.3!**

pysha3 < 1.0 used the old Keccak implementation. During the finalization of
SHA3, NIST changed the delimiter suffix from 0x01 to 0x06. The Keccak sponge
function stayed the same. pysha3 1.0 provides the previous Keccak hash, too.

libscott:
Fixes #884.

I wonder if we can get this into 0.9? For the sake of this blog post?

### DIFF
--- a/docs/server/source/appendices/cryptography.md
+++ b/docs/server/source/appendices/cryptography.md
@@ -15,8 +15,9 @@ algorithm provided by the
 which is a wrapper around the optimized reference implementation
 from [http://keccak.noekeon.org](http://keccak.noekeon.org).
 
-Here's the relevant code from `bigchaindb/bigchaindb/common/crypto.py`
-(as of 11 December 2016):
+**Important**: Since selecting the Keccak hashing algorithm for SHA-3 in 2012, NIST [released a new version](https://en.wikipedia.org/wiki/SHA-3#cite_note-14) of the hash using the same algorithm but slightly different parameters. As of version 0.9, BigchainDB is using the latest version. See below for an example output of the hash function.
+
+Here's the relevant code from `bigchaindb/bigchaindb/common/crypto.py:
 
 ```python
 import sha3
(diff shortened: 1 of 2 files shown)


docs/server/source/appendices/cryptography.md
```diff
@@ -15,8 +15,9 @@ algorithm provided by the
 which is a wrapper around the optimized reference implementation
 from [http://keccak.noekeon.org](http://keccak.noekeon.org).
 
-Here's the relevant code from `bigchaindb/bigchaindb/common/crypto.py`
-(as of 11 December 2016):
+**Important**: Since selecting the Keccak hashing algorithm for SHA-3 in 2012, NIST [released a new version](https://en.wikipedia.org/wiki/SHA-3#cite_note-14) of the hash using the same algorithm but slightly different parameters. As of version 0.9, BigchainDB is using the latest version. See below for an example output of the hash function.
```
ttmc: Maybe say which version of the `pysha3` Python package we are using as of BigchainDB Server version 0.9, i.e. `pysha3==1.0b1`

setup.py
```diff
@@ -61,7 +61,7 @@ def check_setuptools_features():
     # TODO Consider not installing the db drivers, or putting them in extras.
     'rethinkdb~=2.3',  # i.e. a version between 2.3 and 3.0
     'pymongo~=3.4',
-    'pysha3>=0.3',
+    'pysha3==1.0b1',
```
r-marques: Why not go with the latest version `1.1.0`

libscott: Good catch, i think i was confused by the changelog on https://pypi.python.org/pypi/pysha3, looks like they have an error there.

libscott: https://github.com/tiran/pysha3/issues/2

r-marques: yes, I think so

PR #1018, opened 2017-01-04
Change version in tx model to 0.9

libscott:
The idea is that we will go lockstep with server version until we hit 1.0, created a test to check that we follow this idea.


ttmc:
When we're developing the next version in `master`, the full version name has the `.dev` suffix (e.g. `0.9.0.dev`). I wonder if that will cause problems.

libscott:
I also changed it to be a string. I don't think there will be any problems but lets find out.

PR #1022, opened 2017-01-05
Collect http routes into single module

libscott:
The purpose here is to simplify the code (views files now only contain handlers and no use of flask Blueprints) and all routes are defined in a single file to make it easier to see the http api at a glance.


diminator:
Perfect
Get that travis going and you're good for merging

Issue #1021, opened 2017-01-05
Implement HTTP API changes for 0.9


libscott:
Here is the ticket for tracking the implementation of HTTP API (#830) changes for 0.9. 
- [x] Flask cleanup, single "routes" definition - #1022
- [ ] ~Canonical "Server API"~ shelved until use case
- [x] API root informational endpoint (/api/v1)
- [x] Transaction List
- [x] Statuses - #1026
- [x] /outputs endpoint #1066 
- [x] Merge docs from #830

PR #1024, opened 2017-01-06
1021/http api root

libscott:
`/` and `/api/v1` endpoints for #1021


diminator:
perfect

trailing slashes or not?

sohkai:
> trailing slashes or not?

I think we previously set it up to be usable on both (so it should still work with both)?

diminator:
it works with both `strict_slashes=False`

libscott:
I'm kind of in favor of redirecting to include the slash, being finicky about not having duplicate URLs these days, any opinions?

PR #1026, opened 2017-01-06
implement statuses endpoint

diminator:
implements `GET /statuses?tx_id={tx_id}|block_id={block_id}` according to #830 
- [x] status for transactions
- [x] status for blocks
- [x] tests

### DIFF
--- a/bigchaindb/web/views/statuses.py
+++ b/bigchaindb/web/views/statuses.py
@@ -1,0 +1,45 @@
+"""This module provides the blueprint for some basic API endpoints.
+
+For more information please refer to the documentation on ReadTheDocs:
+ - https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/
+   http-client-server-api.html
+"""
+from flask import current_app
+from flask_restful import Resource, reqparse
+
+from bigchaindb.web.views.base import make_error
+
+
+class StatusApi(Resource):
+    def get(self):
+        """API endpoint to get details about the status of a transaction or a block.
+
+        Return:
+            A ``dict`` in the format ``{'status': <status>}``, where
+            ``<status>`` is one of "valid", "invalid", "undecided", "backlog".
+        """
+        parser = reqparse.RequestParser()
+        parser.add_argument('tx_id', type=str)
+        parser.add_argument('block_id', type=str)
+
+        args = parser.parse_args(strict=True)
+
+        if sum(arg is not None for arg in args.values()) != 1:
+            return make_error(400, "Provide exactly one query parameter. Choices are: block_id, tx_id")
+
+        pool = current_app.config['bigchain_pool']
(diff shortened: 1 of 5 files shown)


libscott:
Looks pretty good! Just that one comment.

bigchaindb/web/views/statuses.py
```diff
@@ -0,0 +1,62 @@
...
+        Return:
+            A ``dict`` in the format ``{'status': <status>}``, where
+            ``<status>`` is one of "valid", "invalid", "undecided", "backlog".
+        """
+        parser = reqparse.RequestParser()
+        parser.add_argument('tx_id', type=str)
+        parser.add_argument('block_id', type=str)
+
+        args = parser.parse_args(strict=True)
+
```
TimDaub: I'd prefer if we extract `tx_id` and `block_id` as Python variables here already.

We can then:

- match the 400 error chase in a more readable fashion: `if tx_id is None and block_id is None: return error`
- We don't have to access `args` throughout the whole function call.

Your call if you want to change that, but I think it would increase readability.

bigchaindb/web/views/statuses.py
```diff
@@ -0,0 +1,62 @@
...
+        with pool() as bigchain:
+            if args['tx_id']:
+                status = bigchain.get_status(args['tx_id'])
+                links = {
+                    "tx": "/transactions/{}".format(args['tx_id'])
+                }
+
+            elif args['block_id']:
+                block = bigchain.get_block(block_id=args['block_id'])
```
TimDaub: `bigchain.get_block` allows to additionally fetch the status by setting `include_status=True` as a parameter:
`block, status = bigchain.get_block(block_id=args['block_id'], include_status=True)`

bigchaindb/web/views/statuses.py
```diff
@@ -0,0 +1,62 @@
...
+            A ``dict`` in the format ``{'status': <status>}``, where
+            ``<status>`` is one of "valid", "invalid", "undecided", "backlog".
+        """
+        parser = reqparse.RequestParser()
+        parser.add_argument('tx_id', type=str)
+        parser.add_argument('block_id', type=str)
+
+        args = parser.parse_args(strict=True)
+
+        if sum(arg is not None for arg in args.values()) != 1:
```
TimDaub: I think what you're looking for is an exclusive OR, right?

I don't have any experience with that in Python but perhaps you can do smth like:

- http://stackoverflow.com/a/433161/1263876
- http://stackoverflow.com/a/432844/1263876

Alternatively you can express an xor with the built in python operators.

bigchaindb/web/views/statuses.py
diminator: right, forgot about that

PR #1027, opened 2017-01-09
Feat/1020/fix tests mongodb

### LINKED ISSUE #1020: Make unittest work with mongodb
This issue is to make sure that all the fixtures and unit tests work with mongodb.

r-marques:
resolves #1020 

This pr introduces several changes to mongodb queries and some fixes to the mongodb changefeed.

Regarding the travis integration, travis currently [does not support mongodb 3.X](https://github.com/travis-ci/travis-ci/issues/3694).
Updated the travis scripts to download and run the mongodb 3.4.1 community edition binaries with the `replSet` option enabled.
The travis scripts are based on the [mongo-ruby driver .travis.yml](https://github.com/mongodb/mongo-ruby-driver/blob/master/.travis.yml) 

**Note**: I disabled the parallelization of tests for mongodb because there is a problem probably with the initialization of the replica set that causes the tests to fail if they are run in parallel. Created a separate issue to fix it #1028


.travis.yml
```diff
@@ -12,8 +12,6 @@ env:
 
 matrix:
   fast_finish: true
-  allow_failures:
-  - env: BIGCHAINDB_DATABASE_BACKEND=mongodb
```
sohkai: 👏

.ci/travis_script.sh
```diff
@@ -4,6 +4,8 @@ set -e -x
 
 if [[ -n ${TOXENV} ]]; then
   tox -e ${TOXENV}
+elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb ]]; then
+    pytest -v --database-backend=mongodb --cov=bigchaindb
```
sohkai: Some slightly odd tab formatting here; do we normally use 4 tabs for shell scripts?

r-marques: copy paste gone wrong

tests/backend/mongodb/test_schema.py
```diff
@@ -85,13 +86,19 @@ def test_create_secondary_indexes():
...
     # Votes table
     indexes = conn.conn[dbname]['votes'].index_information().keys()
     assert sorted(indexes) == ['_id_', 'block_and_voter']
 
 
+# The database is set up with a session scope.
+# If we run this test we will remove secondary indexes that are nedeed for
+# the rest of the tests
+@pytest.mark.skipif(reason='This will remove the secondary indexes needed'
+                           ' for the rest of the tests')
```
sbellem: You could use a dummy db instead, as in https://github.com/bigchaindb/bigchaindb/blob/master/tests/backend/rethinkdb/test_schema.py#L98

bigchaindb/backend/mongodb/query.py
```diff
@@ -70,33 +79,83 @@ def get_blocks_status_from_transaction(conn, transaction_id):
...
+    # get the txid of the create transaction for asset_id
+    cursor = conn.db['bigchain'].aggregate([
+        {'$match': {
+            'block.transactions.id': asset_id,
+            'block.transactions.operation': 'CREATE'
+        }},
+        {'$unwind': '$block.transactions'},
+        {'$match': {
+            'block.transactions.id': asset_id,
+            'block.transactions.operation': 'CREATE'
```
libscott: Shouldn't we have block.transactions.asset_id and TRANSFER here?

r-marques: This is only to retrieve the the create transaction. The transfer is done in a separate query

PR #1029, opened 2017-01-10
Implement votes endpoint

diminator:
implements GET /votes?block_id={block_id} according to #830

(only review after #1026 is implemented)

- [x] votes for blocks
- [x] tests


bigchaindb/web/views/votes.py
```diff
@@ -0,0 +1,36 @@
...
+        Return:
+            A list of votes voting for a block with ID ``block_id``.
+        """
+        parser = reqparse.RequestParser()
+        parser.add_argument('block_id', type=str)
+
+        args = parser.parse_args(strict=True)
+
+        if sum(arg is not None for arg in args.values()) != 1:
+            return make_error(400, "Provide the block_id query parameter.")
```
sohkai: Any reason not to just make the `block_id` argument required?

Then it could look like:

```python
        parser = reqparse.RequestParser()
        parser.add_argument('block_id', type=str, required=True)

        args = parser.parse_args(strict=True)

        pool = current_app.config['bigchain_pool']
        with pool() as bigchain:
            votes = list(backend.query.get_votes_by_block_id(bigchain.connection, args['block_id']))

        return votes
```

tests/web/test_votes.py
```diff
@@ -0,0 +1,65 @@
...
+    res = client.get(VOTES_ENDPOINT + "?block_id=" + block.id)
+    assert len(res.json) == 2
+    assert res.status_code == 200
+
+
+@pytest.mark.bdb
+def test_get_votes_endpoint_returns_empty_list_not_found(client):
+    res = client.get(VOTES_ENDPOINT + "?block_id=123")
+    assert [] == res.json
+    assert res.status_code == 200
```
sohkai: Also good to test an empty `block_id` parameter.

PR #1030, opened 2017-01-10
Feat/960/test mongodb queries

### LINKED ISSUE #960: Test mongodb queries
Create unit tests for mongodb queries

r-marques:
resolves #960 
to be reviewed and merged after #1027 

Created tests for mongodb queries.
Small fixes to some queries

### DIFF
--- a/tests/backend/mongodb/test_queries.py
+++ b/tests/backend/mongodb/test_queries.py
@@ -1,6 +1,6 @@
 import pytest
 
-pytestmark = [pytest.mark.bdb]
+pytestmark = pytest.mark.bdb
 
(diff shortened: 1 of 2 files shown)


tests/backend/mongodb/test_queries.py
```diff
@@ -0,0 +1,367 @@
+import pytest
+
+pytestmark = [pytest.mark.bdb]
```
sbellem: Just a small detail: the list is not needed for only one marker, i.e.:

```python
pytestmark = pytest.mark.bdb
```

would also work.

PR #1031, opened 2017-01-10
Implement blocks endpoint

diminator:
implements `GET /blocks/{block_id}` and `GET /blocks?tx_id={tx_id}` according to #830

(only review after #1026 is implemented)

- [x] `GET /blocks/{block_id}`
- [x] `GET /blocks?tx_id={tx_id}`
- [x] tests
- [x] update docs in #830

### DIFF
--- a/bigchaindb/web/views/blocks.py
+++ b/bigchaindb/web/views/blocks.py
@@ -58,8 +58,7 @@ class BlockListApi(Resource):
                 if status:
                     blocks = {k: v for k, v in blocks.items() if v == status}
                 blocks = list(blocks.keys())
-
-        if not blocks:
-            return make_error(404)
+            else:
+                blocks = []
 
         return blocks
--- a/bigchaindb/web/views/blocks.py
+++ b/bigchaindb/web/views/blocks.py
@@ -34,27 +34,28 @@ class BlockApi(Resource):
 
 class BlockListApi(Resource):
     def get(self):
-        """API endpoint to get the related blocks and statuses of a transaction.
+        """API endpoint to get the related blocks for a transaction.
 
         Return:
-            A ``dict`` in the format ``{'block_id': <status>}``, where
-            ``<status>`` is one of "valid", "invalid", "undecided", "backlog".
-            It's possible to return multiple keys of 'block_id'.
+            A ``list`` of ``block_id`` that may be filtered when provided
+            a status query parameter: "valid", "invalid", "undecided".
         """
         parser = reqparse.RequestParser()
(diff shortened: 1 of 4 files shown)


bigchaindb/web/views/blocks.py
```diff
@@ -0,0 +1,62 @@
...
+            block_id (str): the id of the block.
+
+        Return:
+            A JSON string containing the data about the block.
+        """
+
+        pool = current_app.config['bigchain_pool']
+
+        with pool() as bigchain:
+            block = bigchain.get_block(block_id=block_id)
```
TimDaub: According to [the specification](https://github.com/bigchaindb/bigchaindb/pull/830/files#diff-9cd1814988545a4c7fb7457a3f928312R308) this endpoint should take a query parameter `?status`.

For more detail see spec.

diminator: still updating the spec :-)

diminator: resolved

bigchaindb/web/views/blocks.py
```diff
@@ -0,0 +1,65 @@
...
+        status = args['status']
+
+        pool = current_app.config['bigchain_pool']
+
+        with pool() as bigchain:
+            blocks = bigchain.get_blocks_status_containing_tx(tx_id)
+            if blocks:
+                if status:
+                    blocks = {k: v for k, v in blocks.items() if v == status}
+                blocks = list(blocks.keys())
```
sohkai: You *could* reformulate this as a list comprehension, to avoid the nested `if`s (imo, makes it more complex):

```python
block_statuses = bigchain.get_blocks_status_containing_tx(tx_id)
blocks = [block_id for block_id, block_status in block_statuses.items() if not status or block_status == status]
```

tests/web/test_blocks.py
```diff
@@ -0,0 +1,156 @@
...
+        'message': 'Unknown arguments: foo'
+    }
+
+    res = client.get(BLOCKS_ENDPOINT + "?tx_id=123&status=123")
+    assert res.status_code == 400
+    assert res.json == {
+        'message': {
+            'status': '123 is not a valid choice'
+        }
+    }
```
sohkai: Another good test here would be an empty `tx_id`; I tried it and it doesn't complain (just returns an empty list) -- I assume the reqparser just returns an empty string for that argument.

E.g.:

```python
client.get(BLOCKS_ENDPOINT + "?tx_id=")
```

PR #1036, opened 2017-01-11
Some more tests for mongodb code

r-marques:
Better tests for the mongodb connection and schema.

### DIFF
--- a/tests/backend/mongodb/test_connection.py
+++ b/tests/backend/mongodb/test_connection.py
@@ -1,3 +1,7 @@
+import pytest
+
+from unittest import mock
+
 
 def test_get_connection_returns_the_correct_instance():
     from bigchaindb.backend import connect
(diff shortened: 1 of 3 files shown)


tests/backend/mongodb/test_connection.py
```diff
@@ -1,3 +1,7 @@
+import pytest
+
+from unittest import mock
+
```
sbellem: since `unittest` is part of the python standard library we can swap the order:

```python
from unittest import mock

import pytest
```

tests/backend/mongodb/test_connection.py
```diff
@@ -14,3 +18,21 @@ def test_get_connection_returns_the_correct_instance():
     conn = connect(**config)
     assert isinstance(conn, Connection)
     assert isinstance(conn, MongoDBConnection)
+
+
+@mock.patch('pymongo.MongoClient.__init__')
+@mock.patch('time.sleep')
+def test_connection_error(mock_sleep, mock_client):
+    from pymongo.errors import ConnectionFailure
```
sbellem: for code other than bigchaindb we can import at the following the PEP 8 style guide

tests/backend/mongodb/test_schema.py
```diff
@@ -1,5 +1,7 @@
 import pytest
-from unittest.mock import patch
+
+from unittest import mock
+from pymongo.database import Database
```
sbellem: pep 8 import, same as above, since unitttest is in the std lib:

```python
from unittest.mock import patch

import pytest
from pymongo.database import Database
```

tests/backend/mongodb/test_schema.py
```diff
@@ -149,5 +148,47 @@ def test_get_replica_set_name_config_file():
...
+            {'log': ['a line']},
+            {'log': ['database writes are now permitted']},
+        ]
+
+        # check that it returns
+        assert _wait_for_replica_set_initialization(conn) is None
+
+
+def test_initialize_replica_set():
+    from pymongo.errors import OperationFailure
```
sbellem: can go at the top

Issue #975, opened 2016-12-20
Change MongoDB Connection's interface to implement a .run()


sohkai:
From https://github.com/bigchaindb/bigchaindb/pull/890#pullrequestreview-13306930.

`pymongo`'s interface is very different than `rethinkdb`'s with queries being run directly through the mongodb connection object rather than a `.run()` method. We should discuss if `Connection`s should still be expected to have a .run() method.

r-marques:
I like what we do with `rethinkdb` because we have a single place to handle exceptions and deal with connection errors.

I have been looking at the `pymongo` documentation and I think it would be possible to do the same thing we do with `rethinkdb`.

`pymongo` [operations](http://api.mongodb.com/python/current/api/pymongo/operations.html) allows us to basically construct the queries without executing them and then we can use [bulk_write](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.bulk_write) to execute them on the database

sohkai:
From #977, when this is done, we should also port some of the RethinkDB connection tests to MongoDB's connection.

r-marques:
I moved this issue to a different milestone.
At this point I don't know exactly what needs to be done. `pymongo` handles database connections much differently from `rethinkdb` driver.

`pymongo` tries to reconnect by itself so we don't need to handle that like with `rethinkdb`.
Another big difference is that with rethinkdb you would connect to a specific node in a cluster. With mongodb this is different. You connect to the cluster itself and the driver takes care of connecting to a new node when the current node is down, etc...

This raises a lot of new questions that should be handled when you actually perform tests on a cluster.

r-marques:
A note that we got from MongoDB regarding our connection code:
> Retry loop not needed, as the driver will either block until a connection is available or timeout

From pymongo [documentation](http://api.mongodb.com/python/current/examples/high_availability.html#handling-failover):
> We get an AutoReconnect exception. This means that the driver was not able to connect to the old primary (which makes sense, as we killed the server), but that it will attempt to automatically reconnect on subsequent operations. When this exception is raised our application code needs to decide whether to retry the operation or to simply continue, accepting the fact that the operation might have failed.

> On subsequent attempts to run the query we might continue to see this exception. Eventually, however, the replica set will failover and elect a new primary (this should take no more than a couple of seconds in general). At that point the driver will connect to the new primary and the operation will succeed

[...]

PR #1063, opened 2017-01-18
Added replicaset name to bigchaindb config

### LINKED ISSUE #1056: Handle replica set name in bigchaindb configuration
Project #1054 

The replica set name is something that needs to always be available and should not change in a cluster.
BigchainDB needs this to add and remove nodes from the cluster (see #1055) and the driver needs this in order to setup a connection.

We should add this to the configuration file under `backend` for mongodb. E.g.

```json
{
    "backlog_reassign_delay": 120,
    "database": {
        "backend": "mongodb",
        "host": "localhost",
        "name": "bigchain",
        "port": 27017,
        "replicaset": "bigchain"
    },
}
```

We also need to update our `MongoDBConnection` initialization to take into account the replica set name. E.g.
```python
MongoClient('localhost', 27017, replicaset='bigchain')
```

r-marques:
resolves #1056 

This pr adds `replicaset` to `bigchaindb.config`.
Also updated the mongodb connection initialization to connect to a replica set, and moved the `initialize_replica_set` into `mongodb.connection`. The reason for this is that the replicaset needs to be initialized before we even try to setup a `MongoDBConnection`

### DIFF
--- a/bigchaindb/backend/mongodb/connection.py
+++ b/bigchaindb/backend/mongodb/connection.py
@@ -6,6 +6,7 @@ from pymongo.errors import ConnectionFailure
 
 import bigchaindb
 from bigchaindb.backend.connection import Connection
+from bigchaindb.backend import mongodb
 
 logger = logging.getLogger(__name__)
 
@@ -45,6 +46,10 @@ class MongoDBConnection(Connection):
     def _connect(self):
         for i in range(self.max_tries):
             try:
+                # we should only return a connection if the replica set is
+                # initialized. initialize_replica_set will check if the
+                # replica set is initialized else it will initialize it.
+                mongodb.schema.initialize_replica_set()
                 self.connection = MongoClient(self.host, self.port,
                                               replicaset=self.replicaset)
             except ConnectionFailure as exc:
--- a/bigchaindb/backend/mongodb/connection.py
+++ b/bigchaindb/backend/mongodb/connection.py
@@ -2,11 +2,11 @@ import time
 import logging
 
 from pymongo import MongoClient
-from pymongo.errors import ConnectionFailure
+from pymongo import errors
 
 import bigchaindb
+from bigchaindb.common import exceptions
 from bigchaindb.backend.connection import Connection
-from bigchaindb.backend import mongodb
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


bigchaindb/backend/mongodb/connection.py
```diff
@@ -41,9 +46,90 @@ def db(self):
...
+
+    # Setup a MongoDB connection
+    # The reason we do this instead of `backend.connect` is that
+    # `backend.connect` will connect you to a replica set but this fails if
+    # you try to connect to a replica set that is not yet initialized
+    conn = MongoClient(host=bigchaindb.config['database']['host'],
+                       port=bigchaindb.config['database']['port'])
+    _check_replica_set(conn)
+    config = {'_id': bigchaindb.config['database']['replicaset'],
+              'members': [{'_id': 0, 'host': 'localhost:27017'}]}
```
sbellem: I wonder: should we get the host and port from the config, for cases where the host and port differ from defaults?

bigchaindb/backend/mongodb/connection.py
```diff
@@ -41,9 +46,90 @@ def db(self):
...
+    """
+
+    # I did not find a better way to do this for now.
+    # To check if the database is ready we will poll the mongodb logs until
+    # we find the line that says the database is ready
+    logger.info('Waiting for mongodb replica set initialization')
+    while True:
+        logs = conn.admin.command('getLog', 'rs')['log']
+        if any('database writes are now permitted' in line for line in logs):
+                return
```
sbellem: indentation is one tab off

bigchaindb/backend/mongodb/connection.py
```diff
@@ -41,9 +46,90 @@ def db(self):
...
             try:
-                self.connection = MongoClient(self.host, self.port)
-            except ConnectionFailure as exc:
+                # we should only return a connection if the replica set is
+                # initialized. initialize_replica_set will check if the
+                # replica set is initialized else it will initialize it.
+                initialize_replica_set()
+                self.connection = MongoClient(self.host, self.port,
+                                              replicaset=self.replicaset)
+            except errors.ConnectionFailure as exc:
```
sbellem: we can discard the `exc` as it is not used

bigchaindb/backend/mongodb/connection.py
r-marques: yes

bigchaindb/backend/mongodb/connection.py
r-marques: copypasta failed

bigchaindb/backend/mongodb/connection.py
r-marques: prefer to handle that in #975

PR #1065, opened 2017-01-18
1021/http transactions list

libscott:
A `GET /api/v1/transactions/` endpoint, accepting params `asset_id` and `operation`.


bigchaindb/core.py
```diff
@@ -441,6 +417,19 @@ def get_owned_ids(self, owner):
 
         return owned
 
+    def get_transactions_filtered(self, asset_id=None, operation=None):
+        """
+        Get a list of transactions filtered on some criteria
+        """
+        if not asset_id:
+            raise ValueError("Need asset_id")
```
sbellem: Instead of checking for the `asset_id`, why not making it mandatory? It can still be a keyword argument:

```python
def get_transactions_filtered(self, *, asset_id, operation=None):
    # ...
```

libscott: Yea good point, that was a leftover from when it would have been filtered on asset ID OR public key.

bigchaindb/web/views/parameters.py
```diff
@@ -0,0 +1,30 @@
...
+        return key
+    raise ValueError("Invalid base58 ed25519 key")
+
+
+def valid_operation(op):
+    if op == 'CREATE':
+        return 'CREATE'
+    if op == 'TRANSFER':
+        return 'TRANSFER'
+    raise ValueError('Operation must be "CREATE" or "TRANSFER')
```
sbellem: we could use `choices` for this purpose -- see http://flask-restful.readthedocs.io/en/0.3.5/api.html#reqparse.Argument

> * **choices** – A container of the allowable values for the argument.

bigchaindb/web/views/parameters.py
```diff
@@ -0,0 +1,30 @@
...
+        return txid.lower()
+    raise ValueError("Invalid hash")
+
+
+def valid_bool(val):
+    if val == 'true':
+        return True
+    if val == 'false':
+        return False
+    raise ValueError('Boolean value must be "true" or "false" (lowercase)')
```
sbellem: Using `type=bool` in [`reqparse.RequestParser.add_argument(*args, **kwargs)`](http://flask-restful.readthedocs.io/en/0.3.5/api.html#reqparse.RequestParser.add_argument) may be enough?

tests/assets/test_digital_assets.py
```diff
@@ -90,95 +90,6 @@ def test_asset_id_mismatch(b, user_pk):
...
-    b.write_vote(vote)
-
-    asset_id = Transaction.get_asset_id([tx_create, tx_transfer])
-    txs = b.get_transactions_by_asset_id(asset_id)
-    assert len(txs) == 2
-
-    asset = b.get_asset_by_id(asset_id)
-    assert asset == tx_create.asset
-
-
```
sbellem: Why not update these tests instead of entirely removing them?

bigchaindb/web/views/parameters.py
libscott: I decided to leave the "boolean" and "operation" argument parsers, even though they could be supported by `choice` parameter in `parser.add_argument`, because having them separate promotes re-use. Put another way, if I wanted to have an "operation" parameter in another http endpoint, using `parser.add_argument` I'd have to re-formulate it, wheras having the `valid_operation` function separate it easily encapsulates the valid / invalid logic and the error message.

bigchaindb/web/views/parameters.py
libscott: See other [comment](https://github.com/bigchaindb/bigchaindb/pull/1065#discussion_r97033866)

bigchaindb/backend/query.py
```diff
@@ -318,3 +318,16 @@ def get_unvoted_blocks(connection, node_pubkey):
     """
 
     raise NotImplementedError
+
```
r-marques: the `get_txids_by_asset_id` singledispatch is no longer being implemented so we can just remove it

bigchaindb/web/views/parameters.py
TimDaub: > choices

https://www.youtube.com/watch?v=arqTu9Ay4Ig

bigchaindb/web/views/parameters.py
```diff
@@ -0,0 +1,30 @@
+import re
+
+
+def valid_txid(txid):
+    if re.match('^[a-fA-F0-9]{64}$', txid):
+        return txid.lower()
+    raise ValueError("Invalid hash")
+
+
+def valid_bool(val):
```
libscott: It's used in https://github.com/bigchaindb/bigchaindb/pull/1066/files#diff-96761b3426cf97be2817aa6a0ebff3f7R18 which actually includes all of the code in this PR.

bigchaindb/web/views/parameters.py
sbellem: Ok. 

It could be argued that the set of allowed operations should be represented by constants for re-use throughout the code as in https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L412, and then various parts of the code can reuse it including the http api layer. This would save code, and allow us to fall back on the existing mechanism of the framework we use, meanwhile still offering re-usability and modularity. E.g.:

```python
 parser.add_argument('operation', type=str, choices=ALLOWED_OPERATIONS_EXT)

File: tests/web/test_transactions.py [hunk 7]
```diff
@@ -180,3 +181,45 @@ def test_post_invalid_transfer_transaction_returns_400(b, client, user_pk):
         InvalidSignature.__name__, 'Transaction signature is invalid.')
     assert res.status_code == expected_status_code
     assert res.json['message'] == expected_error_message
+
+
+def test_transactions_get_list_good(client):
+    from functools import partial
```
sbellem: std lib imports can go at the top

PR #762, opened 2016-10-29
Pythonic cleanup issue #716

amirelemam:
Pythonic cleanup, according to issue #716, made to files:
benchmarking-tests/benchmark_utils.py
bigchaindb/core.py

Changes include:
- changing // for / and importing division for the correct behavior
- changing sequences from len(sequence) == 0 to not sequence
- some lines were adjusted to 80 columns only

PR #1074, opened 2017-01-22
Replace PR #762

ttmc:
This pull request replaces #762 by @amirelemam because that pull request has been sitting unfinished for a long time.

## In benchmarking-tests/benchmark_utils.py

* This PR doesn't add the `from __future__ import division` line, and it doesn't replace `tx_left = args.num_transactions // mp.cpu_count()` with `tx_left = args.num_transactions / mp.cpu_count()`. The reasons are that this code is expected to be run in a Python 3 environment, and we want `tx_left` to be an integer.
* This PR _does_ shorten many of the line lengths as suggested in PR #762 

## In bigchaindb/core.py

* The line `if len(validated_transactions) == 0:` was replaced with `if not validated_transactions:` as suggested in PR #762

Note: I used the latest `master` branch as my starting point, and this PR made only one _minor_ change to core code (and it's a syntax change, not a semantic change). Despite that, not all the tests passed: the ones that did fail all failed with the same error: `E       fixture 'caplog' not found` etc.

@amirelemam please take a look at this pull request if you have time, keeping in mind that we may merge it if you don't get back to us soon.


ttmc:
@amirelemam I will merge this pull request now. Thanks again for your help and suggestions. We'll give you credit as an external contributor in the upcoming version 0.9.0 release.

PR #1078, opened 2017-01-22
Better error handling in awsdeploy.sh

ttmc:
Added two more options to the `set -e` line at the top of `awsdeploy.sh`. See the comments in the code to understand what they do.

### DIFF
--- a/deploy-cluster-aws/awsdeploy.sh
+++ b/deploy-cluster-aws/awsdeploy.sh
@@ -1,8 +1,10 @@
-#! /bin/bash
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


krish7919:
LGTM.

sohkai:
It may also be interesting to set `IFS=$'\n\t'`; see http://redsymbol.net/articles/unofficial-bash-strict-mode/

ttmc:
@sohkai There are many interesting ideas in that blog post beyond setting `IFS=$'\n\t'`. I think I'll defer adding them for now.

PR #1080, opened 2017-01-23
short form 0.9, 0.10 etc tx version with no '.dev' suffix

libscott:
As in title. The TX version moves lock-step with server version UNTIL 1.0, and does not include `.dev` suffix.

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -410,7 +410,7 @@ class Transaction(object):
     TRANSFER = 'TRANSFER'
     GENESIS = 'GENESIS'
     ALLOWED_OPERATIONS = (CREATE, TRANSFER, GENESIS)
-    VERSION = bigchaindb.version.__version__
+    VERSION = bigchaindb.version.__short_version__[:-4]  # 0.9, 0.10 etc
 
     def __init__(self, operation, asset, inputs=None, outputs=None,
                  metadata=None, version=None):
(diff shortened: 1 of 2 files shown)


bigchaindb/common/transaction.py
```diff
@@ -410,7 +410,7 @@ class Transaction(object):
     TRANSFER = 'TRANSFER'
     GENESIS = 'GENESIS'
     ALLOWED_OPERATIONS = (CREATE, TRANSFER, GENESIS)
-    VERSION = bigchaindb.version.__version__
+    VERSION = bigchaindb.version.__short_version__[:-4]  # 0.9, 0.10 etc
```
sohkai: `[:4]` may be potentially brittle if we ever have alpha or beta (e.g. 0.9a1) releases in the future. Not sure if there's a better solution though (maybe only use the `<major>.<minor>` value?).

libscott: That's why the test compares it using a regular expression.

ttmc: I don't understand how a test can resolve the concern of @sohkai 

`[:-4]` means, "Include all characters except for the last four."

sohkai: Well, I think what @libscott is trying to say is that if the `[:4]` implementation ever breaks, the tests would catch it (and we can react accordingly).

Anyway, thinking about releases, this is likely to never be a problem since 1.0 will remove this bit of complexity.

libscott: Yea that's pretty much it @sohkai

PR #1082, opened 2017-01-23
Update configure command to support multiple backends

### LINKED ISSUE #1057: Add MongoDB support for bigchaindb configure
We need to update bigchaindb `configure` command to handle mongodb.

In interactive mode we need to ask for the database backend and we need to change the defaults depending on what backend the user chooses.

In non interactive mode `-y` we need to know what is the backend that we want maybe with a positional argument. E.g.
```bash
$ bigchaindb -y configure mongodb
```

r-marques:
resolves #1057 

This pr changes the `bigchaindb configure` command to support multiple backends.
The main change to the command is that it now requires a positional argument `backend`:
```bash
$ bigchaindb configure -h
usage: bigchaindb configure [-h] {rethinkdb,mongodb}

positional arguments:
  {rethinkdb,mongodb}  The backend to use. It can be either rethinkdb or
                       mongodb.
```
When run in interactive mode it updates the defaults based on the backend.

Another thing to note is that depending on the backend it populates only the config options required by that backend. For instance if we chose `rethinkdb` as the backend it will not have the `replicaset` option on the generated config file:
```bash
$ bigchaindb -y configure rethinkdb && bigchaindb show-config
Generating keypair
Generating default configuration for backend rethinkdb
Configuration written to /home/rodmar/.bigchaindb
Ready to go!
INFO:bigchaindb.config_utils:Configuration loaded from `/home/rodmar/.bigchaindb`
{
    "backlog_reassign_delay": 120,
    "database": {
        "backend": "rethinkdb",
        "host": "localhost",
        "name": "bigchain",
        "port": 28015
    },
    "keypair": {
        "private": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "public": "6G2GV3ZTe6ZgCb1hXjcEJYCW5Cq7d8LhKUbJw7omcABh"
    },
    "keyring": [],
    "server": {
        "bind": "localhost:9984",
        "threads": null,
        "workers": null
    },
    "statsd": {
        "host": "localhost",
        "port": 8125,
        "rate": 0.01
    }
}

$ bigchaindb -y configure mongodb && bigchaindb show-config
Generating keypair
Generating default configuration for backend mongodb
Configuration written to /home/rodmar/.bigchaindb
Ready to go!
INFO:bigchaindb.config_utils:Configuration loaded from `/home/rodmar/.bigchaindb`
{
    "backlog_reassign_delay": 120,
    "database": {
        "backend": "mongodb",
        "host": "localhost",
        "name": "bigchain",
        "port": 27017,
        "replicaset": "bigchain-rs"
    },
    "keypair": {
        "private": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "public": "BhRLBQsi3XuDjk7D7jdseMQpPP4g1rDHML3nSXZYC1Ls"
    },
    "keyring": [],
    "server": {
        "bind": "localhost:9984",
        "threads": null,
        "workers": null
    },
    "statsd": {
        "host": "localhost",
        "port": 8125,
        "rate": 0.01
    }
}
```

#### Todo:
- [x] Update documentation if this is accepted.

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -5,6 +5,20 @@ import os
 # PORT_NUMBER = reduce(lambda x, y: x * y, map(ord, 'BigchainDB')) % 2**16
 # basically, the port number is 9984
 
+_database_rethinkdb = {
+    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
+    'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
+    'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 28015)),
+    'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
+}
+
+_database_mongodb = {
+    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'mongodb'),
+    'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
+    'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 27017)),
+    'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
+    'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
+}
 
 config = {
     'server': {
@@ -14,13 +28,7 @@ config = {
         'workers': None,  # if none, the value will be cpu_count * 2 + 1
         'threads': None,  # if none, the value will be cpu_count * 2 + 1
     },
-    'database': {
-        'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
-        'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
-        'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 28015)),
-        'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
-        'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
-    },
+    'database': _database_rethinkdb,
     'keypair': {
         'public': None,
         'private': None,
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -20,6 +20,11 @@ _database_mongodb = {
     'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
 }
 
+_database_map = {
+    'mongodb': _database_mongodb,
+    'rethinkdb': _database_rethinkdb
+}
+
 config = {
     'server': {
         # Note: this section supports all the Gunicorn settings:
@@ -28,7 +33,9 @@ config = {
         'workers': None,  # if none, the value will be cpu_count * 2 + 1
         'threads': None,  # if none, the value will be cpu_count * 2 + 1
     },
-    'database': _database_rethinkdb,
+    'database': _database_map[
+        os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb')
+    ],
     'keypair': {
         'public': None,
         'private': None,
--- a/bigchaindb/backend/connection.py
+++ b/bigchaindb/backend/connection.py
@@ -40,6 +40,12 @@ def connect(backend=None, host=None, port=None, name=None, replicaset=None):
     host = host or bigchaindb.config['database']['host']
     port = port or bigchaindb.config['database']['port']
     dbname = name or bigchaindb.config['database']['name']
+    # Not sure how to handle this here. This setting is only relevant for
+    # mongodb.
+    # I added **kwargs for both RethinkDBConnection and MongoDBConnection
+    # to handle these these additional args. In case of RethinkDBConnection
+    # it just does not do anything with it.
+    replicaset = replicaset or bigchaindb.config['database'].get('replicaset')
 
     try:
         module_name, _, class_name = BACKENDS[backend].rpartition('.')
@@ -51,7 +57,7 @@ def connect(backend=None, host=None, port=None, name=None, replicaset=None):
         raise ConfigurationError('Error loading backend `{}`'.format(backend)) from exc
 
     logger.debug('Connection: {}'.format(Class))
-    return Class(host, port, dbname)
+    return Class(host, port, dbname, replicaset=replicaset)
 
 
 class Connection:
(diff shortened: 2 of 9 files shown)


bigchaindb/backend/connection.py
```diff
@@ -40,6 +40,12 @@ def connect(backend=None, host=None, port=None, name=None, replicaset=None):
     host = host or bigchaindb.config['database']['host']
     port = port or bigchaindb.config['database']['port']
     dbname = name or bigchaindb.config['database']['name']
+    # Not sure how to handle this here. This setting is only relevant for
+    # mongodb.
+    # I added **kwargs for both RethinkDBConnection and MongoDBConnection
+    # to handle these these additional args. In case of RethinkDBConnection
+    # it just does not do anything with it.
+    replicaset = replicaset or bigchaindb.config['database'].get('replicaset')
```
sohkai: > I added **kwargs

👍 , I think this is what we should be doing going forward.

Since we're only planning to add `replicaset` at this point, I passing this value into every `Connection` is fine for now.

The changes I made in https://github.com/bigchaindb/bigchaindb/pull/986 moved towards allowing arbitrary configuration properties (like `replicaset`), but it's unlikely we'll be able to have a nice way to do this until we rework the configuration at some point.

tests/backend/test_generics.py
```diff
@@ -69,27 +68,31 @@ def test_changefeed_class(changefeed_class_func_name, args_qty):
...
-    ('mongodb', 'MongoDBConnection'),
-    ('rethinkdb', 'RethinkDBConnection'),
-))
 @patch('bigchaindb.backend.schema.create_indexes',
        autospec=True, return_value=None)
 @patch('bigchaindb.backend.schema.create_tables',
        autospec=True, return_value=None)
 @patch('bigchaindb.backend.schema.create_database',
        autospec=True, return_value=None)
 def test_init_database(mock_create_database, mock_create_tables,
```
sohkai: Mmm, in some ways I wonder if this test would be better served by being split off into their own separate tests in `tests/backend/mongo/test_schema.py` and `tests/backend/rethink/test_schema.py`.

The fact that it requires an instance of the connection doesn't make it a *generic* test, and the duplication wouldn't be much different that what we have here already.

tests/commands/test_commands.py
```diff
@@ -31,8 +32,8 @@ def test_start_raises_if_command_not_implemented():
 
     with pytest.raises(NotImplementedError):
         # Will raise because `scope`, the third parameter,
-        # doesn't contain the function `run_configure`
-        utils.start(parser, ['configure'], {})
+        # doesn't contain the function `run_start`
+        utils.start(parser, ['start'], {})
```
sohkai: Oh haha, was this just a typo before?

tests/test_config_utils.py
```diff
@@ -36,6 +44,11 @@ def test_bigchain_instance_raises_when_not_configured(monkeypatch):
     # from existing configurations
     monkeypatch.setattr(config_utils, 'autoconfigure', lambda: 0)
 
+    # set the current backend so that Bigchain can create a connection
+    backend = request.config.getoption('--database-backend')
+    backend_conf = getattr(bigchaindb, '_database_' + backend)
+    bigchaindb.config['database'] = backend_conf
```
sohkai: We use this particular bit on setting up the database config a few times, maybe it's worth adding in a fixture?

bigchaindb/__init__.py
```diff
@@ -14,13 +28,7 @@
...
         'threads': None,  # if none, the value will be cpu_count * 2 + 1
     },
-    'database': {
-        'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
-        'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
-        'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 28015)),
-        'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
-        'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
-    },
+    'database': _database_rethinkdb,
```
sohkai: Rather than always default to the rethinkdb config, could we automatically use the setting that was given with `BIGCHAINDB_DATABASE_BACKEND`?

I guess something like:

```python
_database_map = {
    'mongodb': _database_mongodb,
    'rethinkdb': _database_rethinkdb,
}

config = {
    ...
   'database': _database_map[os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb')],
    ...
}
```

It might not be necessary because `autoconfigure()` should handle setting this properly, but from a glance, it does seem weird that `bigchaindb.config.database` might not reflect the setting of a running system.

r-marques:
> However, I think we're starting to come to the limits of what our current configuration mechanism can reasonably handle. At least for me, it's starting to get pretty confusing to keep track of how we create a running system's configuration.

I agree but improving the config system will be a project on its own.

tests/backend/test_generics.py
r-marques: In this case I will just remove this tests since we already test the creation of dbs tables and indexes in each backends `test_schema.py`

tests/commands/test_commands.py
r-marques: Its not a typo. Its just that configure now requires a positional argument and since you can use any command for this test I changed it to `start` to keep it simple.
Its mostly aesthetics.

bigchaindb/__init__.py
r-marques: I did it in another pull request https://github.com/bigchaindb/bigchaindb/pull/1088/files#diff-993779d6d8d7ec19551175f387ef2ea6R37

Although your solution looks cleaner. I will apply this changes to #1088

tests/test_config_utils.py
r-marques: I prefer to handle this on #1088 after I make the changes in the comment below.

r-marques:
@sohkai I prefer to do the changes you proposed in #1088 since these 2 prs overlap when it comes to configuration changes. See https://github.com/bigchaindb/bigchaindb/pull/1088#issuecomment-276308691

tests/test_config_utils.py
```diff
@@ -131,19 +131,35 @@ def test_autoconfigure_read_both_from_file_and_env(monkeypatch, request):
...
+        'port': 4242,
+        'name': 'test-dbname',
+    }
+    database_mongodb = {
+        'backend': 'mongodb',
+        'host': 'test-host',
+        'port': 4242,
+        'name': 'test-dbname',
+        'replicaset': 'bigchain-rs',
+    }
```
sohkai: Might be nice to set some of the variables like `test-host`, `4242`, and `test-dbname` as constants at the start of the test to avoid passing some magic strings around repeatedly.

bigchaindb/commands/bigchain.py
```diff
@@ -86,6 +86,16 @@ def run_configure(args, skip_if_exists=False):
     conf['keypair']['private'], conf['keypair']['public'] = \
         crypto.generate_key_pair()
 
+    # select the correct config defaults based on the backend
+    print('Generating default configuration for backend {}'
+          .format(args.backend))
+    database = {}
```
vrde: what about

```python
database = bigchaindb._database_map[args.backend]
```

It should not raise `KeyError`s because the `add_argument` has defined `choices`.

PR #1083, opened 2017-01-23
Add flake8-quotes checker and cleanup double quotes

libscott:
This is a flake8 extension to check for double quotes mis(?)use, and cleanup of the instances where we use double quotes instead of single quotes for inline string literals.

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -118,7 +118,7 @@ class Input(object):
             fulfillment = Fulfillment.from_uri(data['fulfillment'])
         except ValueError:
             # TODO FOR CC: Throw an `InvalidSignature` error in this case.
-            raise InvalidSignature('Fulfillment URI couldn\'t been parsed')
+            raise InvalidSignature("Fulfillment URI couldn't been parsed")
         except TypeError:
             # NOTE: See comment about this special case in
             #       `Input.to_dict`
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -1039,7 +1039,7 @@ class Transaction(object):
 
         if tx_body.get('operation') == Transaction.CREATE:
             if proposed_tx_id != tx_body['asset'].get('id'):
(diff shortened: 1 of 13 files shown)


bigchaindb/common/transaction.py
```diff
@@ -118,7 +118,7 @@ def from_dict(cls, data):
             fulfillment = Fulfillment.from_uri(data['fulfillment'])
         except ValueError:
             # TODO FOR CC: Throw an `InvalidSignature` error in this case.
-            raise InvalidSignature("Fulfillment URI couldn't been parsed")
+            raise InvalidSignature('Fulfillment URI couldn\'t been parsed')
```
ttmc: Sentences with internal single quotes, like this one, are one situation where I think double quotes can make the line more readable, but I'm not going to make a big fuss over it. Maybe we can just standardize on a single-quotes-always rule. If so, you should update this section of our Python style guide:   https://github.com/bigchaindb/bigchaindb/blob/master/PYTHON_STYLE_GUIDE.md#single-or-double-quotes

sbellem: I second @ttmc: double quotes are useful to surround text that includes single quotes

libscott:
> I added two small commits that address the two points above, since I was already looking into it. The two commits can be rebased, or squashed against the original commit.

Cool :)

@sbellem @sohkai  It looks like flake8-quotes accepts double quoted strings if they contain a single quote, but does this mean we are happy to merge single quoted strings with escaped single quotes, or must all string literals containing a single quote be double quoted? The reason I ask, is that the objective here is to avoid the need to manually verify this kind of thing in the future, so our tools can enforce our guidelines.

sbellem:
> but does this mean we are happy to merge single quoted strings with escaped single quotes, or must all string literals containing a single quote be double quoted? The reason I ask, is that the objective here is to avoid the need to manually verify this kind of thing in the future, so our tools can enforce our guidelines.

I am totally fine with both approaches: escaped or double-quoted, and the automatic checker tolerates both

PR #1085, opened 2017-01-24
Fix Makefile for new docs structure

sohkai:


Makefile
```diff
@@ -65,12 +65,11 @@ coverage: ## check code coverage quickly with the default Python
 		$(BROWSER) htmlcov/index.html
 
 docs: ## generate Sphinx HTML documentation, including API docs
-	rm -f docs/bigchaindb.rst
-	rm -f docs/modules.rst
-	sphinx-apidoc -o docs/ bigchaindb
```
sohkai: Not really sure if the above are still necessary, but it seems to work without them.

ttmc: I think that building `docs/bigchaindb.rst` was part of the BigchainDB release process, see https://github.com/bigchaindb/bigchaindb/blob/master/Release_Process.md

It's not there anymore, as far as I can see. Should something be _added_ to the release process?

sohkai: I don't think so... I think we've plugged into sphinx so that readthedocs will automatically generate any new documentation from the templates? @libscott might know better?

libscott: I don't know what they are tbh. Our hook into the docs build process is via `extensions` in `docs/server/source/conf.py`. I don't recognise any of the above. I think @sbellem is the person to ask.

sohkai: Turns out `sphinx-apidoc` is a way to autogenerate docstrings references (e.g. `automodule`s and the like).

We don't use this currently, but it might be nice to turn on in the future.

Issue #862, opened 2016-11-24
BigchainDB Events API specification


TimDaub:
### Definition

This project is deliberately called "Event APIs" as we didn't want to introduce a tool-driven bias (I'd consider "Websocket API" a biased way of naming it e.g.). What we mean by Event APIs are interfaces between

- a single node in BigchainDB; or
- multiple nodes in BigchainDB (e.g. in a cluster)

communicating with an external entity (two-way communication not a hard requirement). An external entity is defined as not being in any ways associated/connected with the entities running the multiple or single node(s).

Examples of such interfaces that were already mentioned:

- Communication primitives like WebSockets
- Pub-Sub patterns like Pubsubhubnub
- Federated event distribution systems (e.g. Jabber?)
- Intra-cluster-communication protocols of provisioned database (MongoDB, RethinkDB)
- (Please extend this list if you have more examples)

### Problem statement

For given actions that happen on a single or multiple nodes in a BigchainDB cluster, we'd like to PUSH information about that action in the form of an event to a client (assumed they've registered upfront stating that they'd like to receive that information).

### Collected user stories so far

This is a list of stories we've collected from users already. They're ordered by priority. If you know from users that are interested in a specific aspect, feel free to update this list without changing the prioritization. If you're a user and you're currently reading this ticket, feel free to comment below to let us know what's important to you.

#### Confirmed stories by actual users of BigchainDB (high priority)

1. As a BigchainDB/IPDB client I want to be notified for all transactions currently being persisted that follow a "specific schema"<sup>1</sup> so that I can process them centrally in my backend, but have them submitted in a decentralized fashion by my users without my involvement.
1. As a BigchainDB/IPDB client I want to be notified when and why (provided with a reason) the transactions I've sent get discarded by (a) node(s) so that I can construct them validly next time.
1. As a BigchainDB/IPDB client I want to be notified when transactions for a specific asset occur so that I can monitor the asset throughout its life.

#### Suggested stories by us (no confirmation for actual users YET; not mandatory for the completion of this project; lower priority)

1. As a BigchainDB/IPDB client I want to be notified when transactions occur that include a specific challenge to be resolved<sup>3</sup>.
1. As a BigchainDB/IPDB client I want to be notified when transactions I've sent have arrived in a nodes backlog.
1. As a BigchainDB/IPDB client I want to be notified when transactions I've sent get included into a block.
[...]

PR #1086, opened 2017-01-25
Event Stream API Spec

sohkai:
### Details left for implementation

There are a number of topics I didn't want to touch during this design phase, that deal more directly with an implementation. Some of these, however, do affect the design. This is left here as an reminder for the implementation phase.

* Web server
    * Keep using flask (and a WebSocket extension), but move from a threaded server to one that uses gevent or similar
    * Look at other servers, e.g. aiohttp
* Architecture
    * Bundle the WebSocket server with the HTTP server, and build a new pipeline for the sole purpose of filtering events
    * Separate the WebSocket API as a separate service (either ran on the same machine or other) with its own connection to the underlying DB

--------------

## Docs preview:
image

### DIFF
--- a/docs/server/source/drivers-clients/websocket-event-stream-api.rst
+++ b/docs/server/source/drivers-clients/websocket-event-stream-api.rst
@@ -1,0 +1,114 @@
+The WebSocket Event Stream API
+==============================
+
+.. important::
+    This is currently scheduled to be implemented in BigchainDB Server 0.10.
+
+BigchainDB provides real-time event streams over the WebSocket protocol with
+the Event Stream API.
+
+Connecting to an event stream from your application enables a BigchainDB node
+to notify you as events are processed, such as new `validated transactions <#valid-transactions>`_.
+
+
+Demoing the API
+---------------
+
+You may be interested in demoing the Event Stream API with the `WebSocket echo test <http://websocket.org/echo.html>`_
+to familiarize yourself before attempting an integration.
+
+
+Determining Support for the Event Stream API
+--------------------------------------------
+
+In practice, it's a good idea to make sure that the node you're connecting with
+has advertised support for the Event Stream API. To do so, send a HTTP GET
+request to the node's :ref:`Root URL <bigchaindb-root-url>` and check that the
+response contains a ``streams_<version>`` property in ``_links``::
+
+    {
+      "_links": {
+         "streams_v1": "ws://example.com:9984/api/v1/streams/"
+      }
+    }
+
+
+Connection Keep Alive
+~~~~~~~~~~~~~~~~~~~~~
+
+The Event Stream API requires clients to signal that they'd like their
+connection to stay open by sending "pings" across the open connection.
+BigchainDB nodes will automatically close any connections that haven't sent a
+ping in the last three minutes.
+
+.. note::
+
+    While three minutes is the limit before a BigchainDB server will terminate
+    a connection, we suggest sending a ping every 30 seconds for better
+    reliability.
+
+A "ping" consists of a message containing only the string ``"ping"``, for example
+in JavaScript:
+
+.. code-block:: javascript
+
+    new WebSocket("...").send("ping")
+
+If the BigchainDB node received the ping, it'll respond back with a message
+containing only the string ``"pong"``.
+
+
+Streams
+-------
+
+Each stream is meant as a unidirectional communication channel, where the
+BigchainDB node is the only party sending messages (except for `keep-alive
+pings <#connection-keep-alive>`_). Any messages sent to the BigchainDB node
+(except the keep-alive pings) will be ignored.
+
+Streams will always be under the WebSocket protocol (so ``ws://`` or
+``wss://``) and accessible as extensions to the ``/api/v<version>/streams/``
+API root URL (for example, `validated transactions <#valid-transactions>`_
+would be accessible under ``/api/v1/streams/valid_tx``). If you're running your
+own BigchainDB instance and need help determining its root URL, you can find
+more :ref:`here <determining-the-api-root-url>`.
+
+All messages sent in a stream are in the JSON format.
+
+.. note::
+
+    For simplicity, BigchainDB initially only provides a stream for all
+    validated transactions. In the future, we may provide streams for other
+    information, such as new blocks, new votes, or invalid transactions. We may
+    also provide the ability to filter the stream for specific qualities, such
+    as a specific ``output``'s ``public_key``.
+
+    If you have specific use cases that you think would fit as part of this
+    API, feel free to reach out via `gitter <https://gitter.im/bigchaindb/bigchaindb>`_
+    or `email <mailto:product@bigchaindb.com>`_.
+
+Valid Transactions
+~~~~~~~~~~~~~~~~~~
+
+``/valid_tx``
+
+Streams an event for any newly validated transactions. Message bodies contain
+the transaction's ID, associated asset ID, and containing block's ID.
+
+Example message::
+
+    {
+        "txid": "<sha3-256 hash>",
+        "assetid": "<sha3-256 hash>",
+        "blockid": "<sha3-256 hash>"
+    }
+
+
+.. note::
+
+    Transactions in BigchainDB are validated in batches ("blocks") and will,
+    therefore, be streamed in batches. Each block can contain up to a 1000
+    transactions, ordered by the time at which they were included in the block.
+    The ``/valid_tx`` stream will send these transactions in the same order
+    that the block stored them in, but this does **NOT** guarantee that you
+    will recieve the events in that same order.
--- a/docs/server/source/drivers-clients/websocket-event-stream-api.rst
+++ b/docs/server/source/drivers-clients/websocket-event-stream-api.rst
@@ -36,35 +36,17 @@ response contains a ``streams_<version>`` property in ``_links``::
 Connection Keep Alive
 ~~~~~~~~~~~~~~~~~~~~~
 
-The Event Stream API requires clients to signal that they'd like their
-connection to stay open by sending "pings" across the open connection.
-BigchainDB nodes will automatically close any connections that haven't sent a
-ping in the last three minutes.
-
-.. note::
-
-    While three minutes is the limit before a BigchainDB server will terminate
-    a connection, we suggest sending a ping every 30 seconds for better
-    reliability.
-
-A "ping" consists of a message containing only the string ``"ping"``, for example
-in JavaScript:
-
-.. code-block:: javascript
-
-    new WebSocket("...").send("ping")
-
-If the BigchainDB node received the ping, it'll respond back with a message
-containing only the string ``"pong"``.
-
+The Event Stream API initially does not provide any mechanisms for connection
+keep alive other than enabling TCP keepalive on each open WebSocket connection.
+In the future, we may add additional functionality to handle ping/pong frames
+or payloads designed for keep alive.
 
 Streams
 -------
 
 Each stream is meant as a unidirectional communication channel, where the
-BigchainDB node is the only party sending messages (except for `keep-alive
-pings <#connection-keep-alive>`_). Any messages sent to the BigchainDB node
-(except the keep-alive pings) will be ignored.
+BigchainDB node is the only party sending messages. Any messages sent to the
+BigchainDB node will be ignored.
 
 Streams will always be under the WebSocket protocol (so ``ws://`` or
 ``wss://``) and accessible as extensions to the ``/api/v<version>/streams/``
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -10,6 +10,7 @@ If you set up a BigchainDB node or reverse proxy yourself,
 and you're not sure what the API Root URL is,
 then see the last section of this page for help.
 
+.. _bigchaindb-root-url:
 
 BigchainDB Root URL
 -------------------
@@ -392,6 +393,8 @@ Votes
    :statuscode 200: A list of votes voting for a block with ID ``block_id`` was found and returned.
    :statuscode 400: The request wasn't understood by the server, e.g. just requesting ``/votes``, without defining ``block_id``.
 
+
+.. _determining-the-api-root-url:
 
 Determining the API Root URL
 ----------------------------
--- a/docs/server/source/drivers-clients/index.rst
+++ b/docs/server/source/drivers-clients/index.rst
@@ -15,6 +15,7 @@ community projects listed below.
    :maxdepth: 1
 
    http-client-server-api
+   websocket-event-stream-api
    The Python Driver <https://docs.bigchaindb.com/projects/py-driver/en/latest/index.html>
    Transaction CLI <https://docs.bigchaindb.com/projects/cli/en/latest/>
 


docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+contains a ``streams_<version>`` property in ``_links``::
+
+    {
+      "_links": {
+         "streams_v1": "ws://example.com:9984/api/v1/streams/"
+      }
+    }
+
+
+Connection Keep Alive
```
libscott: Why need to add the "ping" protocol?

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+
+``/valid_tx``
+
+Streams the IDs of any newly validated transactions. Message bodies will also
+contain the block ID containing the validated transaction.
+
+Example message::
+
+    {
+        "txid": "<sha3-256 hash>",
```
libscott: I'd prob include the asset_id in here too. Or just full transaction bodies. Otherwise people will have no way to know if they want to see the contents other than to do a `GET /api/v1/transactions/...`.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: This is really awkward, I thought I wrote comments on Github explaining some of these decisions but they seem to have been lost... I'll write them up again.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: `asset_id` is a good idea; that'll really help people to filter out if they care about that transaction for now.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+================================
+
+BigchainDB provides real-time event streams over the WebSockets protocol with
+the Events Stream API.
+
+Connecting to a Events Stream from your application enables a BigchainDB node
+to notify you as events are processed, such as new `validated transactions <#valid-transactions>`_.
+
+
+Determining Support for the Events Stream API
```
sohkai: Added this in case we decide to implement this API as an optional component that's separated from the current flask server.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: I added this because I didn't want the server to keep pushing updates on connections that might have died. Most WebSocket implementations require some sort of keep-alive ping/pong, although most of the time this is only necessary during periods of inactivity to avoid a firewall or etc from closing the connection prematurely.

For us, I wanted to eagerly close connections (if the client doesn't ping us back) to avoid leaving dead connections open. Not sure how easy this is to actually implement, and is something we can descope during the implementation (or remove if it's not a concern). Regardless, establishing some sort of keep-alive protocol is useful for clients that want to ensure their connection stays alive.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
+The WebSockets Events Stream API
```
sohkai: Good point, I kind of made up "Events Stream API" as I went, but we can change the naming to "Events API" or the file name (or, really, any other option) as we go.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+
+The Events Stream API requires clients to signal that they'd like their
+connection to stay open by sending "pings" across the open connection.
+BigchainDB nodes will automatically close any connections that haven't sent a
+ping in the last three minutes.
+
+.. note::
+
+    While three minutes is the limit before a BigchainDB server will terminate
+    a connection, we suggest sending a ping every 30 seconds for better
```
sohkai: 30 seconds seems pretty common, I've seen it around in a few places. Seems like ~1min or so of inactivity causes some connections to automatically close.

3 minutes is totally arbitrary, I just made that up as a starting place. Apparently TCP can take up to 10min to determine that a connection's been broken, but the [specifics of that are beyond me](http://blog.stephencleary.com/2009/05/detection-of-half-open-dropped.html).

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+        "txid": "<sha3-256 hash>",
+        "blockid": "<sha3-256 hash>"
+    }
+
+
+.. note::
+
+    Transactions in BigchainDB are validated in batches ("blocks") and may,
+    therefore, be streamed in batches. Each block can contain up to a 1000
+    transactions, which may all be sent at once without any required ordering.
```
sohkai: This is a really good point, since I was wanting to put the `blockid` in the payload.

Semantically, this would change the behaviour to *valid blocks* rather than purely *valid transactions*, and people would have to filter out the `txid_list` to find out if there were any transactions they would be interested in (but this would've happened anyway, just more transparently).

On @libscott's point about `asset_id`, it might also be interesting to use a structure like:

```json
{
    "blockid": "<sha3-256 hash>",
    "txs": [{
        "id": "<sha3-256 hash>",
        "asset_id": "<asset id>",
    }, ... ]
}
```

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+
+All messages sent in a stream are in the JSON format.
+
+.. note::
+
+    For simplicity, BigchainDB initially only provides a stream for all
+    validated transactions. In the future, we may provide streams for other
+    information, such as new blocks, new votes, or invalid transactions. We may
+    also provide the ability to filter the stream for specific qualities, such
+    as a specific ``output``'s ``public_key``.
```
TimDaub: I think it would make sense to include a note here saying something like:

"If you have specific use cases that you think fit the WebSockets Events Stream API's functionality, feel free to reach out via [insert communication channel]"

docs/server/source/drivers-clients/websocket-event-stream-api.rst
libscott: > I wanted to eagerly close connections (if the client doesn't ping us back) to avoid leaving dead connections open

TCP takes care of this for you, when you push a message, either the client responds with the appropriate TCP acknowledgements (which includes consuming the data) or the connection goes into an error state at our end. Trying to write on such a connection will throw an error.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+.. note::
+
+    Transactions in BigchainDB are validated in batches ("blocks") and may,
+    therefore, be streamed in batches. Each block can contain up to a 1000
+    transactions, which may all be sent at once without any required ordering.
+
+
+Tips and Tricks
```
TimDaub: I'd remove that headline for now until we actually provide Tips and Tricks in this document.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
TimDaub: > I wanted to eagerly close connections (if the client doesn't ping us back) to avoid leaving dead connections open.

:+1: 

> Not sure how easy this is to actually implement, and is something we can descope during the implementation (or remove if it's not a concern).


I think it's good to outline this behavior specifically in the spec here as it outlines a useful requirement for the implementation phase later on. How this will then be implemented is TBD (when we implement). It could turn out that some Python WebSockets implementations in fact implement this behavior already (they probably should as the WebSockets ietf rfc also talks about this behavior: https://tools.ietf.org/html/rfc6455#section-5.5.2).

For the implementation, I'd say we should push this as close to the websocket protocol as possible.
For this spec, this note is appropriate.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
TimDaub: I'd leave them at 30secs for now and then change the spec according to our findings in the implementation and deployment phase.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
TimDaub: Instead of posting only hashes, would it make sense to post relative URIs to `GET /transactions/txid` and `GET /blocks/blockid`?

I don't have a technical angle on this (any technical objections?) but I think that might be a nice touch in terms of UX.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
TimDaub: > For transactions in the same block, it seems arbitrary, yet it might matter for some users.

I'm not aware of any user that cares about the transaction order inside of a block. As the endpoint is currently proposed, global order of transactions is provided by including the block's hash in the payload. That's great and also good enough.
I'll let you know immediately when someone reaches out about real-time updates of transactions' order in a block (and we can think about a `valid blocks` WebSocket API endpoint).

docs/server/source/drivers-clients/websocket-event-stream-api.rst
krish7919: > I wanted to eagerly close connections (if the client doesn't ping us back) to avoid leaving dead connections open

Kernel can run out of FDs if there are long-lived connections maintained by it. Theoretically, we have 65k, but it can soon jump over. Ping/Pong is definitely needed, sooner or later. Also, useful for a long-living system like ours.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
krish7919: @TimDaub: We will need a code change here every time the API changes. Is the tradeoff for UX required here? Kind of in two minds about it.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
krish7919: @sohkai : With TCP and the exponential backoff, realistically it takes about 17mins for TCP to realize the connection is broken if keepalive is enabled. Hence, FDs need to be realeased for recycling in the kernel. See ing/Pong comment above too. :)

docs/server/source/drivers-clients/websocket-event-stream-api.rst
TimDaub: > @TimDaub: We will need a code change here every time the API changes. Is the tradeoff for UX required here? Kind of in two minds about it.

Changing two static strings (`../transactions/` and `../blocks`) of API routes that ideally do not change again from now on is definitely something we can afford.

> I'd prob include the asset_id in here too

That might be useful. I'd add it. 

> Or just full transaction bodies.

I'd leave it to the client to determine with information to download.

krish7919: @sohkai : Websocket != http for metering. If there are full transactions being sent over websocket, we would also need time and efforts to meter it.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: It's important to note that the WebSockets IETF spec for ping/pong is different from how I've tried to structure it here; theirs is at the *protocol* level while what I've written is supposed to be implemented at the *application* level.

I think the main reason most people implement their own version is because the *protocol* way of handling pings/pongs is usually hidden from the user; I think for a lot of WebSockets implementations, the user is never exposed to the *protocol* ping/pongs (and if they ever happen).

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: @krish7919 We might have to decide on a strategy for metering WebSockets.

For the most part, I'd imagine anyone interested in the feed is/was actually sending transactions, so we might not need to add additional billing based on what they've received via this API. In case there's a lot of followers who are just listening, I guess we could also force people to pay for the privilege of using this API?

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: See 946567a

docs/server/source/drivers-clients/websocket-event-stream-api.rst
sohkai: See da54972.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
krish7919: @sohkai: The current strategy is to meter/bill the number of bytes transferred in and out of the node.
If websockets send data out, albeit small, then someone (consortium or BigchainDB) has to absorb the cost of data transfer out (esp. in AWS, Azure, etc.) The small bytes soon add up.
And it's a totally different way to allow metering with websockets versus http/s.

TimDaub: Metering/billing is out of scope of this PR getting merged.

krish7919: @TimDaub : I don't think it would be fair if we give extra info to the user for better UX, which he can easily derive at his end, and then charge for it. It directly affects the events API spec now.
We might have to spend cycles to remove the relative URI later, if we add it now.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,109 @@
...
+and accessible as extensions to the ``/api/v<version>/streams/`` API root URL
+(for example, `validated transactions <#valid-transactions>`_ is accessible
+under ``/api/v1/streams/valid_tx``). If you're running your own BigchainDB
+instance and need help determining its root URL, you can look :ref:`here <determining-the-api-root-url>`.
+
+All messages sent in a stream are in the JSON format.
+
+.. note::
+
+    For simplicity, BigchainDB initially only provides a stream for all
```
diminator: Does this mean we are sending 1000txs each time a full block is created?

If so, what are the BW limitations ifo the number of connections per node?

docs/server/source/drivers-clients/websocket-event-stream-api.rst
diminator: > Or just full transaction bodies.

That'll probably congest the BW imho...

diminator: > Changing two static strings (../transactions/ and ../blocks) of API routes that ideally do not change again from now on is definitely something we can afford.

:+1: for hypermedia

docs/server/source/drivers-clients/websocket-event-stream-api.rst
diminator: > I'm not aware of any user that cares about the transaction order inside of a block.

:+1:

diminator: > Therefore, it seems to me that the messages about newly-valid things should always and only be messages about newly-valid blocks, with a list of all the transaction IDs in each block, i.e. an example message would be:

:+1:

docs/server/source/drivers-clients/websocket-event-stream-api.rst
vrde: > I'd prob include the asset_id in here too. Or just full transaction bodies. Otherwise people will have no way to know if they want to see the contents other than to do a GET /api/v1/transactions/....

+1 for `asset_id`. I'd like to have the full body as well but this might require having filtered event streams first.

> Changing two static strings (../transactions/ and ../blocks) of API routes that ideally do not change again from now on is definitely something we can afford.

I don't think it's a good idea. A transaction has an `id` that is it's hash. If we put the URL, then who is receiving events might need to "unpack" the URL to have the `id`. Also, a URL is tied to a host, and we have a federation of hosts, so which hosts would you use?

TimDaub: > Also, a URL is tied to a host, and we have a federation of hosts, so which hosts would you use?

Note the use of relative URLs: `../<xyz>`.


> If we put the URL, then who is receiving events might need to "unpack" the URL to have the id.

You can argue the other way saying that for querying the full body of a transaction, you need to format it with the ledger's host name and port.

Let's discuss this in the review meeting this week.

sohkai:
Changes as per review meeting @ 15/02/2017:

* Formalize as `Event Stream` API
* Remove explicit handling of timeouts, and rely on TCP keep-alive (until experience shows otherwise)
* Add `assetid` to payload
* Explicitly declare that transactions in the same block will be delivered in-order on a best-effort basis

krish7919:
Is it too late to consider (gRPC)[http://www.grpc.io/] rather than websockets?

TimDaub:
> Is it too late to consider gRPC rather than websockets?
> Docs here: http://www.grpc.io/docs/.

Since this PR is fulfills all requirements from external users enumerated [here](https://github.com/bigchaindb/bigchaindb/issues/862), there is no need to explore another option anymore.
It's not to late, but it's currently unnecessary to explore gRPC until we find use cases for why we should consider it a solution to a problem.

krish7919:
@TimDaub : Well, I was of the opinion that we were not discussing 'implementation' so far.
Websockets _is_ an implementation. Of it were a spec, we would just be discussing the messages going across and such. I had a Q on #862 about whether we have selected a strategy.

docs/server/source/drivers-clients/websocket-event-stream-api.rst
```diff
@@ -0,0 +1,96 @@
...
+--------------------------------------------
+
+In practice, it's a good idea to make sure that the node you're connecting with
+has advertised support for the Event Stream API. To do so, send a HTTP GET
+request to the node's :ref:`Root URL <bigchaindb-root-url>` and check that the
+response contains a ``streams_<version>`` property in ``_links``::
+
+    {
+      "_links": {
+         "streams_v1": "ws://example.com:9984/api/v1/streams/"
```
krish7919: We have another issue for not exposing the protocol://host:port from the response. #1141 
The user is already aware of this; it is implicit. Unless we are asking the user to connect to another host altogether.

sohkai: Given that it's a different protocol `ws` vs `http`, I don't particularly mind having the full `protocol://host:port` in there. It's similar to the discussion on relative links earlier.

krish7919: That's implementation specific, isn't it?
My opinion is to allow users to connect to any node in the federation, without specifying an explicit host or a port.
We will go back to the discussion we had yesterday about how do we support overloaded nodes and distribute requests.

TimDaub: @krish7919 and I concluded that for the scope of this PR it is OK to ignore this issue for now and handle it in #1141. For that matter I also submitted a comment on the ticket.

PR #1087, opened 2017-01-25
Return 202 on successful transaction POST

libscott:
This brings implementation into line with docs that specify 202 response code on TX POST.

https://github.com/bigchaindb/bigchaindb/blob/master/docs/server/generate_http_server_api_documentation.py#L70


TimDaub:
LGMT

PR #1088, opened 2017-01-25
BigchainDB commands to add and remove nodes from the replica set

### LINKED ISSUE #1055: Add BigchainDB commands to add and remove nodes from a replica set for MongoDB
Project #1054 

This would be the equivalent of `set-replicas` with RethinkDB.
The difference is that with MongoDB we are actually adding and removing nodes to the cluster instead of simply changing the cluster configuration.

This can be achieved with [pymongo](http://api.mongodb.com/python/current/examples/high_availability.html#initializing-the-set) using the `c.admin.command` with the [replSetReconfig](https://docs.mongodb.com/manual/reference/command/replSetReconfig/#dbcmd.replSetReconfig) command.

We also need to decide how the command would look like. E.g.:
```bash
$ bigchaindb add-replica <hostname> <port>
$ bigchaindb remove-replica <hostname> <port>
```
I find it difficult to have a unified interface between different backends since they behave quite differently.
I would be ok if in the `singledispatch` instead of raising a `NotImplemented` exception we could just output something like `this command can only be used with mongodb backend`.

r-marques:
resolves #1055

This pr adds two new bigchaindb commands to add and remove nodes from a MongoDB replicaset:
```bash
$ bigchaindb add-replicas -h
usage: bigchaindb add-replicas [-h] replicas [replicas ...]

positional arguments:
  replicas    A list of space separated hosts to add to the replicaset. Each
              host should be in the form `host:port`.

$ bigchaindb remove-replicas -h
usage: bigchaindb remove-replicas [-h] replicas [replicas ...]

positional arguments:
  replicas    A list of space separated hosts to remove from the replicaset.
              Each host should be in the form `host:port`.

# usage example
$ bigchaindb add-replicas localhost:27018 localhost:27019
$ bigchaindb remove-replicas localhost:27018 localhost:27019
```

#### Todo:
- [x] Update the documentation if this accepted


r-marques:
Some changes to make:
- https://github.com/bigchaindb/bigchaindb/pull/1082#discussion_r98494145
- https://github.com/bigchaindb/bigchaindb/pull/1082#discussion_r98497400

bigchaindb/backend/mongodb/admin.py
```diff
@@ -0,0 +1,68 @@
...
+    # the set and continue from there.
+    cur_id = max([member['_id'] for member in conf['config']['members']])
+
+    # add the nodes to the members list of the replica set
+    for replica in replicas:
+        cur_id += 1
+        conf['config']['members'].append({'_id': cur_id, 'host': replica})
+
+    # increase the configuration version number
+    conf['config']['version'] += 1
```
sohkai: Ahh... is the configuration "version" the version of the actual document (i.e. a "revision" number) and not the version of the expected schema? They really made this confusing.

bigchaindb/backend/mongodb/admin.py
```diff
@@ -0,0 +1,68 @@
...
+        filter(lambda member: member['host'] not in replicas,
+               conf['config']['members'])
+    )
+
+    # increase the configuration version number
+    conf['config']['version'] += 1
+
+    # apply new configuration
+    try:
+        return connection.conn.admin.command('replSetReconfig', conf['config'])
```
sohkai: Curious as to what the return value of `conn.admin.command()` is. If it's not that useful, we could avoid returning it, otherwise, we could document it.

bigchaindb/backend/mongodb/admin.py
r-marques: Yes. When you reconfigure, mongodb expects a version number higher then what it currently has

bigchaindb/backend/mongodb/admin.py
```diff
@@ -0,0 +1,68 @@
...
+    except OperationFailure as exc:
+        raise DatabaseOpFailedError(exc.details['errmsg'])
+
+
+@register_admin(MongoDBConnection)
+def remove_replicas(connection, replicas):
+    """Remove a set of replicas from the replicaset
+
```
r-marques: forgot to update the docstrings

bigchaindb/backend/mongodb/admin.py
r-marques: It does not return anything valuable. Something like `{"ok": 1}`

bigchaindb/commands/utils.py
```diff
@@ -95,6 +96,34 @@ def start(parser, argv, scope):
     return func(args)
 
 
+def mongodb_host(host):
```
r-marques: Not sure where to put this. This is basically a special type validation method for `ArgumentParser`

tests/commands/test_commands.py
```diff
@@ -345,3 +376,57 @@ def test_calling_main(start_mock, base_parser_mock, parse_args_mock,
...
+    mock_remove_replicas.return_value = None
+    assert run_remove_replicas(args) is None
+
+    # test add_replicas with `DatabaseOpFailedError`
+    mock_remove_replicas.side_effect = DatabaseOpFailedError()
+    assert run_remove_replicas(args) is None
+
+    # test add_replicas with `NotImplementedError`
+    mock_remove_replicas.side_effect = NotImplementedError()
+    assert run_remove_replicas(args) is None
```
r-marques: `called_once` is new in python 3.6.
Since we are still use python 3.4 in tests and the documentation I will use `call_count == 1` and `reset_mock`

tests/commands/test_commands.py
```diff
@@ -345,3 +376,57 @@ def test_calling_main(start_mock, base_parser_mock, parse_args_mock,
...
+def test_mongodb_host_type():
+    from bigchaindb.commands.utils import mongodb_host
+
+    # bad port provided
+    with pytest.raises(ArgumentTypeError):
+        mongodb_host('localhost:11111111111')
+
+    # no port information provided
+    with pytest.raises(ArgumentTypeError):
+        mongodb_host('localhost')
```
r-marques: MongoDB `uri_parser` is limited in that regard. The best I can do is make sure that the host is not empty.

bigchaindb/commands/utils.py
sohkai: Mmm yeah, and I guess we'd be importing `pymongo` anyway since the command parser needs it.

Maybe in the future we can have backend-specific commands (and some way to hook them up to the main command parser).

tests/backend/mongodb/test_admin.py
```diff
@@ -30,14 +30,22 @@ def mock_replicaset_config():
     }
 
 
-def test_add_replicas(mock_replicaset_config):
+@pytest.fixture
+def connection():
```
sohkai: I would just rename this to something like `connected_connection` to be painfully obvious that it's connected ;)

bigchaindb/commands/utils.py
vrde: If the problem is the import at the top of the file, there is nothing wrong to put it inside the `mongodb_host` function.

PR #1090, opened 2017-01-25
Parallel testing script

libscott:
Script to run test suite in many parallel processes.

Uses `docker-compose` and separate networks rather than the pytest-xdist method, because of the need for isolation in filehandles.


libscott:
Closing, goals are unclear, will return to it later.

PR #1043, opened 2017-01-16
duplicate tx.id into tx.asset.id in CREATE transactions

libscott:
Fixes #1033 

All CREATE transactions must have `id` duplicated into `asset.id`.

```json
{
  "asset": {
    "data": null,
    "id": "7a21b877db0b1e3cbf7cd8967be8f990be1d6ebb09e2967f15577b4b23c8e3b8"
  },
  "id": "7a21b877db0b1e3cbf7cd8967be8f990be1d6ebb09e2967f15577b4b23c8e3b8",
```

@ttmc 
screenshot from 2017-01-16 15-21-51

Issue #970, opened 2016-12-19
Proposal: Signing of partial tx vs signing of txid


libscott:
Currently, in order to sign each input to a TX we do a little dance whereby for each input, the payload to be signed is a partial transaction is created containing only that input + the other contents of the transaction:

## Option A (current)

```python
for index, input_ in enumerate(self.inputs):                         
    # NOTE: We clone the current transaction but only add the output 
    #       and input we're currently working on plus all            
    #       previously signed ones.                                  
    tx_partial = Transaction(self.operation, self.asset, [input_],   
                             self.outputs, self.metadata,            
                             self.version)                           
                                                                     
    tx_partial_dict = tx_partial.to_dict()                           
    tx_partial_dict = Transaction._remove_signatures(tx_partial_dict)
    payload = Transaction._to_str(tx_partial_dict)             
    self._sign_input(input_, index, payload, key_pairs)        
```

This has the properties:

* ID of the transaction being signed is **not** part of the signature.
* Person signing cannot verify the other inputs to their transaction, since the transaction is malleable in it's inputs.
* "I authorize this transaction with these outputs and this metadata to spend this output" - does not cover the whole transaction.
* Signature for one output cannot be used for another output.

## Option B (proposed, hash then sign)

The other option would be that for each input, the payload to be signed is a tuple of `(txid, fulfills)`:

```python
for index, input_ in enumerate(self.inputs):
    payload = serialize((txid, input_.fulfills))
    self._sign_input(input_, index, payload, key_pairs)
```

Has the properties:

* ID of the transaction being signed **is** part of the signature.
* Person signing can therefore sign given only the `(txid, fulfills)` tuple, if they accept.
* Person signing can also validate the entire transaction body since they must have the txid in order to sign.
* Signature for one output cannot be used for another output.
* Signatures depend on transaction ID and thus nothing can change.

diminator:
I'm in for Option B - as it manages expectations better...

maybe include @TimDaub since he has implemented the common portion?

sohkai:
Also vote for option b, we had a brief discussion about it earlier but couldn't think anything reason to make it not feasible (again, @TimDaub will know best why it was implemented this way...).

[...]

PR #1098, opened 2017-01-25
Keep asset ID when signing fulfillments

### LINKED ISSUE #1089: Don't remove asset.id before signing transaction
common/transaction.py#L959 should be separated out into a different method to avoid popping the asset id during transaction signing.

sohkai:
Fixes #1089.

Fixes regression in https://github.com/bigchaindb/bigchaindb/pull/1043/files#diff-78ff9e91be50c5bfbd702d4e0b22fb4fR958 that removed an asset's ID in CREATE transactions when signing fulfillments.

-----------

@TimDaub:

This (old) behaviour (of keeping the asset's ID) may be awkward in cases where we sign partial transactions; assets used to have their own IDs (separate from a transaction's ID), but now they're the same. This poses a problem in the case of partial transactions, as the asset ID that will be included in the fulfillment URI won't be the same as the partial transaction's ID (the partial transaction's ID is different from the complete transaction).

#970 may address this concern, but this is something we missed when redoing the asset IDs for 0.9.

Thoughts?

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -925,14 +925,34 @@ class Transaction(object):
             'version': self.version,
         }
 
-        tx_no_signatures = Transaction._remove_signatures(tx)
-        tx_serialized = Transaction._to_str(tx_no_signatures)
-        tx['id'] = Transaction._to_hash(tx_serialized)
+        tx['id'] = Transaction._generate_tx_id(tx)
         if self.operation == Transaction.CREATE:
             # Duplicate asset into asset for consistency with TRANSFER
             # transactions
             tx['asset']['id'] = tx['id']
         return tx
+
+    @staticmethod
+    def _generate_tx_id(tx):
+        """Generates an ID from a Transaction dictionary
+
+        Args:
+            tx (dict): The Transaction (as a dict) to get an ID from
+
+        Returns:
+            string: The Transaction's ID
+        """
+        # Remove any parts that are not used when generating the tx
+        # This includes any existing ID, signatures (as signing a transaction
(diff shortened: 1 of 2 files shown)


tests/common/test_transaction.py
```diff
@@ -482,6 +483,26 @@ def test_add_output_to_tx_with_invalid_parameters(asset_definition):
...
+    from cryptoconditions import Ed25519Fulfillment
+    from cryptoconditions.crypto import Ed25519SigningKey
+    utx_dict = utx.to_dict()
+    utx_dict['inputs'][0]['fulfillment'] = None
+    message = rapidjson.dumps(
+        utx_dict,
+        skipkeys=False,
+        ensure_ascii=False,
+        sort_keys=True,
+    ).encode()
```
sbellem: Just as a piece of information, for the purpose of testing, `json` is sufficient, i.e., `rapidjson` is not essential, except for its better performance.

bigchaindb/common/transaction.py
```diff
@@ -925,16 +925,36 @@ def to_dict(self):
...
 
     @staticmethod
+    def _generate_tx_id(tx):
+        """Generates an ID from a Transaction dictionary
+
+        Args:
+            tx (dict): The Transaction (as a dict) to get an ID from
+
+        Returns:
+            string: The Transaction's ID
```
sbellem: I think the return type needs to be an object, and if so in this case, it should be `str` instead of `string`.

sohkai:
Superceded by #1101, but I'll reopen another after it's merged to keep the clean up changes made here.

PR #1100, opened 2017-01-26
[WIP] Add run interface to MongoDB connection (aka: lazy queries)

vrde:
With RethinkDB, building a query and executing it are two different concerns. This is pretty handy because we can have on one side one or more module containing queries, and on the other side a function (in our case the `Connection.run` method) to execute them and catch some common exceptions such as disconnections.

The idea is to have a similar approach for MongoDB (see #975). To achieve that, I've tried a couple of ideas:
 - [operations API](https://github.com/bigchaindb/bigchaindb/issues/975#issuecomment-274523556)
 - [bulk API](https://github.com/bigchaindb/bigchaindb/issues/975#issuecomment-275144883)

But they both provide a limited set of operations. This PR adds a new utility class called `Lazy`.
An instance of `Lazy` collects all the method calls on it, supports chaining, and it's able to run the sequence of calls on a specific object.

```python
from bigchaindb.utils import Lazy
l = Lazy()
l.split(',')[1].split(' ').pop(1).strip()
result = l.run('Like humans, cats tend to favor one paw over another')
assert result == 'cats'
```

If this proposal meets the requirements, I'll refine this PR adding:
 - [ ] better docs
 - [ ] more tests


r-marques:
looks good to me

sohkai:
Honestly surprised python doesn't provide some sort of `Lazy`, but this is a nice idea!

bigchaindb/backend/mongodb/connection.py
```diff
@@ -60,6 +64,10 @@ def _connect(self):
                     time.sleep(2**i)
 
 
+def table(name):
```
sohkai: Might be more canonical to name this `collection`, given that that's what mongo names them (and this is a mongo specific implementation).

vrde: Good point!

vrde:
> Honestly surprised python doesn't provide some sort of Lazy, but this is a nice idea!

Thanks! I'll refine the implementation, docs, and tests.

bigchaindb/backend/mongodb/connection.py
```diff
@@ -43,32 +60,75 @@ def conn(self):
...
+                                                      replicaset=self.replicaset,
+                                                      serverselectiontimeoutms=self.connection_timeout)
+
+            # `initialize_replica_set` might raise `ConnectionFailure` or `OperationFailure`.
+            except (pymongo.errors.ConnectionFailure,
+                    pymongo.errors.OperationFailure) as exc:
+                logger.warning('Attempt %s/%s. Connection to %s:%s failed after %sms.',
+                               attempt, self.max_tries if self.max_tries != 0 else '∞',
+                               self.host, self.port, self.connection_timeout)
+                if attempt == self.max_tries:
```
r-marques: Can't this be moved outside the for loop?
`if attempt == self.max_tries` seems to be doing the same as `for i in self.max_tries_counter`

bigchaindb/backend/mongodb/query.py
```diff
@@ -70,17 +81,20 @@ def get_transaction_from_block(conn, transaction_id, block_id):
...
 
 @register_query(MongoDBConnection)
 def get_blocks_status_from_transaction(conn, transaction_id):
-    return conn.db['bigchain']\
-            .find({'block.transactions.id': transaction_id},
-                  projection=['id', 'block.voters'])
+    return conn.run(
+        conn.collection('bigchain')
+        .find({'block.transactions.id': transaction_id},
+              projection=['id', 'block.voters']))
```
r-marques: I think `get_txids_filtered` was added recently and it also needs to use connection.run

bigchaindb/backend/mongodb/query.py
```diff
@@ -113,115 +127,133 @@ def get_txids_filtered(conn, asset_id, operation=None):
...
-    last_voted = conn.db['votes']\
-                  .find({'node_pubkey': node_pubkey},
-                        sort=[('vote.timestamp', -1)])
+    last_voted = conn.run(
+            conn.collection('votes')
+            .find({'node_pubkey': node_pubkey},
+                  sort=[('vote.timestamp', -1)]))
 
     # pymongo seems to return a cursor even if there are no results
     # so we actually need to check the count
```
r-marques: `get_unvoted_blocks` is not using `conn.run`

bigchaindb/backend/mongodb/changefeed.py
```diff
@@ -27,12 +27,14 @@ def run_forever(self):
 
         while True:
             try:
+                # XXX: hack to force reconnection,
+                #      the correct way to fix this is to manage errors
+                #      for cursors.conn
+                self.connection.connection = None
```
r-marques: Can you explain why do we need to do this?

r-marques:
One thing that we are not doing and that `RethinkDBConnection.run` does is retrying a query if something goes wrong most importantly in case of a `AutoReconnect`

bigchaindb/backend/mongodb/connection.py
vrde: I check it there because it's the context where I can access `exc`, so I can throw the original exception together with ours.

bigchaindb/backend/mongodb/query.py
vrde: Yes! I missed it, thanks!

bigchaindb/backend/mongodb/query.py
vrde: :+1:

PR #1102, opened 2017-01-26
Fix spend input twice bug 

### LINKED ISSUE #1099: Multiple fulfillments for the same condition leads to double spends
From @jmduque on gitter:

> We can created a transaction using the same condition twice for a divisible asset and we were successfully able to create an output where the total amount was expected to be the sum of both inputs and consequently created money!

> last condition:

blob

> fullfilments for next transaction:
> notice both fulfillment's reference same transaction and same cid

blob

> Output Conditions (notice they are double the amount of latest condition):

blob

#### Todo:
- [ ] Recreate the bug
- [ ] Fix it
- [ ] Incorporate the changes into 0.8.1 and do hotfix release 0.8.2
- [ ] Incorporate the changes into master

libscott:
Fixes #1099

### DIFF
--- a/bigchaindb/models.py
+++ b/bigchaindb/models.py
@@ -88,6 +88,11 @@ class Transaction(Transaction):
                 if output.amount < 1:
                     raise AmountError('`amount` needs to be greater than zero')
 
+            # Validate that all inputs are distinct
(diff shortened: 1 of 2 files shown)


r-marques:
The same needs to be done on the on the `0.8.1` branch so that we can release an hot fix

PR #1108, opened 2017-01-26
Consensus Plugin

tymlez:
We are return possibility to connect a consensus plugin, which was removed in the commit:
https://github.com/bigchaindb/bigchaindb/commit/50b0b3cef217bc412dba5a062de1db2d1adcd514
Possibility to extend basic consensus rules by custom logic.


libscott:
Succeeded by #1209

PR #1109, opened 2017-01-27
[WIP] Use lambda to make things lazy

vrde:
This is an approach I've tried but discarded in favor of #1100. It basically solves #975, but with a different approach. I'm revamping this approach after @libscott poked me.

### DIFF
--- a/bigchaindb/backend/mongodb/query.py
+++ b/bigchaindb/backend/mongodb/query.py
@@ -17,7 +17,10 @@ register_query = module_dispatch_registrar(backend.query)
 @register_query(MongoDBConnection)
 def write_transaction(conn, signed_transaction):
     try:
-        return conn.db['backlog'].insert_one(signed_transaction)
+        return conn.run(
+            lambda db:
+            db['backlog']
+            .insert_one(signed_transaction))
     except errors.DuplicateKeyError:
         return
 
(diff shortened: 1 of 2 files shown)


bigchaindb/backend/mongodb/query.py
```diff
@@ -17,7 +17,10 @@
 @register_query(MongoDBConnection)
 def write_transaction(conn, signed_transaction):
     try:
-        return conn.db['backlog'].insert_one(signed_transaction)
+        return conn.run(
+            lambda db:
```
libscott: Couldn't it be:

```python
return conn.run(lambda db:
    db['backlog'].insert_one(signed_transaction))

Comment (libscott) [inline bigchaindb/backend/mongodb/query.py]: Couldn't it be, to save a little space:

```python
return conn.run(lambda db:
    db['backlog'].insert_one(signed_transaction))
```

Another idea: Use a decorator like `@retry` and then just return the lambda directly :)

vrde: The linter complains, so no.

We can have one of the following:
```python
return conn.run(lambda db:
                db['backlog'].insert_one(signed_transaction))
```

or 

```python
return conn.run(
    lambda db:
    db['backlog'].insert_one(signed_transaction))
```

I've chosen the latter because this:
```python
def get_transaction_from_block(conn, transaction_id, block_id):
    try:
        return conn.run(
            lambda db:
            db['bigchain']
            .aggregate([
                {'$match': {'id': block_id}},
                {'$project': {
                    'block.transactions': {
                        '$filter': {
                            'input': '$block.transactions',
                            'as': 'transaction',
                            'cond': {
                                '$eq': ['$$transaction.id', transaction_id]
                            }
                        }
                    }
                }}]).next()['block']['transactions'].pop())
    except (StopIteration, IndexError):
        # StopIteration is raised if the block was not found
        # IndexError is returned if the block is found but no transactions
        # match
        return
```

looks better than this:
```python
def get_transaction_from_block(conn, transaction_id, block_id):
    try:
        return conn.run(lambda db:
                        db['bigchain']
                        .aggregate([
                            {'$match': {'id': block_id}},
                            {'$project': {
                                'block.transactions': {
                                    '$filter': {
                                        'input': '$block.transactions',
                                        'as': 'transaction',
                                        'cond': {
                                            '$eq': ['$$transaction.id', transaction_id]
                                        }
                                    }
                                }
                            }}]).next()['block']['transactions'].pop())
    except (StopIteration, IndexError):
        # StopIteration is raised if the block was not found
        # IndexError is returned if the block is found but no transactions
        # match
        return
```

> Another idea: Use a decorator like @retry and then just return the lambda directly :)

not sure on how this can work.

libscott: > not sure on how this can work.

Actually it doesnt even need lambda:

```python
@retrylogic
def get_transaction_from_block(conn, transaction_id, block_id):
    return conn.query("get block please")
```

`@retrylogic` then handles stale connections.

OR

Take advantage of single dispatch to apply retry logic.

vrde:
Closing in favor of #1100.

PR #1113, opened 2017-01-30
Break out of the loop once a connection is established

### LINKED ISSUE #1068: [backend] break out of the loop once a connection is established
e.g.:
instead of:
```python
    def _connect(self):
        for i in range(1, self.max_tries + 1):
            try:
                self.conn = r.connect(host=self.host, port=self.port, db=self.dbname)
            except r.ReqlDriverError:
                if i == self.max_tries:
                    raise
                wait_time = 2**i
                time.sleep(wait_time)
```
add else at the end of the `try/except` clause:

```python
    def _connect(self):
        for i in range(1, self.max_tries + 1):
            try:
                self.conn = r.connect(host=self.host, port=self.port, db=self.dbname)
            except r.ReqlDriverError:
                if i == self.max_tries:
                    raise
                wait_time = 2**i
                time.sleep(wait_time)
            else:
                break
```

same goes for mongodb backend

vrde:
Closes #1068

### DIFF
--- a/bigchaindb/backend/rethinkdb/connection.py
+++ b/bigchaindb/backend/rethinkdb/connection.py
@@ -77,3 +77,5 @@ class RethinkDBConnection(Connection):
                 wait_time = 2**i
                 logging.debug('Error connecting to database, waiting %ss', wait_time)
                 time.sleep(wait_time)
+            else:
+                break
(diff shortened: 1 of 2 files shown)


vrde:
@sbellem related to the `MongoDBConnection`, I'll need to change some connection things on #975, so I'll tackle the problem on that PR.

PR #1119, opened 2017-01-31
Recommend Ubuntu 16.04 for now

### LINKED ISSUE #1077: Proposal: Recommend server users use Ubuntu 16.04, for now
This is just for BigchainDB Server.

We use a variety of operating systems internally, but I think it would be best if we standardized on recommending one for our users, for now, and reflecting that choice throughout our documentation, installation scripts, etc.

In the future, we can explicitly support multiple operating systems, but for now Ubuntu 16.04 is probably sufficient, and if someone wants to use some other Ubuntu version or Linux distro, they can look at the Ubuntu 16.04 stuff and figure out what to do from that. (If they want to use Windows or macOS, we can tell them to use a virtual machine or Docker.)

Today, our docs and code have a mix of Ubuntu 14.04, 16.04 and Fedora stuff. I'd like to clean that up.

If we opt to Dockerize everything (for deployment), then it's fine to have just one OS inside the container, and the Docker machine / host can be running anything that works with Docker.

This is just a proposal that I'm putting out there for comment.

ttmc:
I updated the docs to be more heavily tilted towards helping users of Ubuntu 16.04 (although there are still some Fedora instructions that I left).

I kept `sudo apt-get blah` rather than using `sudo apt blah` because the `apt-get` form will work with earlier versions of Ubuntu as well as 16.04.

Bonus: I also updated the Terraform and Ansible stuff to work with Ubuntu 16.04.

This pull request resolves #1077


r-marques:
looks good to me although I did not test the terraform or ansible scripts

krish7919:
Looks goot to me, although I have not worked with Terraform/Ansible/LXC directly.

Issue #1105, opened 2017-01-26
Refactor the Dockerfile


krish7919:
We need to add apt-get update to the second apt-get install in line 14.

```
  5 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
  6 
  7 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
  8     echo 'LANG="en_US.UTF-8"'>/etc/default/locale && \
  9     dpkg-reconfigure --frontend=noninteractive locales && \
 10     update-locale LANG=en_US.UTF-8
 11 
 12 ENV LANG en_US.UTF-8
 13 
 14 RUN apt-get -y install python3 python3-pip libffi-dev
 15 RUN pip3 install --upgrade pip
 16 RUN pip3 install --upgrade setuptools
 ```
Because of https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

sohkai:
Fixed with #1121.

PR #1121, opened 2017-01-31
Refactor Dockerfile

krish7919:
The `apt-get update` command executed with the install instructions
should not use a locally cached storage layer.
Solves #1105.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,9 @@ RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
 
 ENV LANG en_US.UTF-8
 
-RUN apt-get -y install python3 python3-pip libffi-dev
-RUN pip3 install --upgrade pip
-RUN pip3 install --upgrade setuptools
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -11,9 +11,9 @@ RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
 
 ENV LANG en_US.UTF-8
 
-RUN apt-get -y install python3 python3-pip libffi-dev
-RUN pip3 install --upgrade pip
-RUN pip3 install --upgrade setuptools
+RUN apt-get update && apt-get -y install python3 python3-pip libffi-dev \
```
ttmc: I'm not sure it's necessary to refresh the list of packages again (using apt-get update), since we just did that back in line 5, and it probably hasn't changed in the meantime. But I guess it can't hurt either.

ttmc: Maybe add a comment why it's a good idea to `apt-get update` _again_, maybe with a link to https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

PR #1130, opened 2017-02-01
fix wrong logic in validate_block

### LINKED ISSUE #519: wrong logic in validate_block
The initial logic of validate_block was to run all the rules to check if a block is valid and either raise an exception if something is invalid or return the block if valid.

https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/core.py#L421
If a block was already voted on this returns the block regardless of its validity

`validate_block` should be independent of the vote and this logic should be moved outside
https://github.com/bigchaindb/bigchaindb/pull/515/files#diff-dbd823dc078bde30c4d6ab30e335853bR46

### LINKED ISSUE #660: Block validation checks federation list multiple times per validation
In `bigchaindb/pipelines/votes.py`:

``` python
    def validate_block(self, block):
        if not self.bigchain.has_previous_vote(block['id'], block['block']['voters']):
        ...
        try:
                self.consensus.validate_block(self.bigchain, block)
```

This eventually calls to `bigchaindb/models/block.py`:

``` python
    def validate(self, bigchain):
        ...
        if bigchain.has_previous_vote(self.id, self.voters):
        ...
```

Where `self` in this context is the block.  Seems like we're checking this twice for no reason?

r-marques:
resolves #519 
resolves #660 

The line in question is in `models.Block.validate` https://github.com/bigchaindb/bigchaindb/blob/9762b4b96854cd5896f4b71c7fbfcc82d6f24f5c/bigchaindb/models.py#L207
```python
if bigchain.has_previous_vote(self.id, self.voters):
    return self
```
This pr removes this check and the associated test.

This is wrong for two reasons:
1. `Block.validate` should run the block validation logic regardless of the voting, the same way that `Transaction.validate` runs the transaction validation logic regardless of if the transaction already exists or not. It's up to the block creation pipeline to make sure that the transaction is unique https://github.com/bigchaindb/bigchaindb/blob/9762b4b96854cd5896f4b71c7fbfcc82d6f24f5c/bigchaindb/pipelines/block.py#L63, the same way that is up to the voting pipeline to make sure that the block was not already voted on https://github.com/bigchaindb/bigchaindb/blob/9762b4b96854cd5896f4b71c7fbfcc82d6f24f5c/bigchaindb/pipelines/vote.py#L52.
2. `Block.validate` returns `self` if the block is valid or raises an exception otherwise. With the line above it would return `self` if it had already been voted on regardless of whether the block had been voted valid or invalid.

### DIFF
--- a/bigchaindb/models.py
+++ b/bigchaindb/models.py
@@ -202,11 +202,6 @@ class Block(object):
             OperationError: If a non-federation node signed the Block.
             InvalidSignature: If a Block's signature is invalid.
         """
-
-        # First, make sure this node hasn't already voted on this block
-        if bigchain.has_previous_vote(self.id, self.voters):
-            return self
-
         # Check if the block was created by a federation node
         possible_voters = (bigchain.nodes_except_me + [bigchain.me])
         if self.node_pubkey not in possible_voters:
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -163,16 +163,3 @@ class TestBlockModel(object):
 
         public_key = PublicKey(b.me)
         assert public_key.verify(expected_block_serialized, block.signature)
-
-    def test_validate_already_voted_on_block(self, b, monkeypatch):
-        from unittest.mock import Mock
-        from bigchaindb.models import Transaction
-
-        tx = Transaction.create([b.me], [([b.me], 1)])
-        block = b.create_block([tx])
-
-        has_previous_vote = Mock()
-        has_previous_vote.return_value = True
-        monkeypatch.setattr(b, 'has_previous_vote', has_previous_vote)
-        assert block == block.validate(b)
-        assert has_previous_vote.called is True


r-marques:
Not sure what is wrong with codecov.
Removed 2 lines from `Block.validate` and `Block.validate` still has 100% coverage

libscott:
Sounds right. Strictly speaking, this:

https://github.com/bigchaindb/bigchaindb/blob/9762b4b96854cd5896f4b71c7fbfcc82d6f24f5c/bigchaindb/pipelines/vote.py#L52

Should not be in a method called `validate_block` either. But since it's not part of an API it doesn't make alot of difference.

ttmc:
Does this pull request also resolve issue #660?

r-marques:
Yes it does. Seems like we had some duplicated issues for that

Issue #1126, opened 2017-02-01
Test that secondary indexes are being used


r-marques:
Write unit tests to check that our queries are using secondary index.
With MongoDB this can be accomplished with [Cursor.explain](http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.explain) and [Database.command](https://api.mongodb.com/python/current/examples/aggregation.html#aggregation-framework) for aggregation operations.

Rethinkdb has a similar query profiler.

PR #1131, opened 2017-02-01
Fix asset id index

libscott:
Asset ID index got broke when removing the "transaction" wrapper, this PR fixes and adds a test.

I would have tested the query `get_txids_filtered` directly, but:

1. the cursors are hidden inside a generator
1. refactoring `get_txids_filtered` out into `_get_txids_filtered_cursor` was going to work, but then
1. You can't called `.explain()` on a cursor returned by `.aggregate`, you must use `Database.command` instead.
1. We therefore can't easily test the indexes used by that database cursor, so it's tested separately.

Relates to https://github.com/bigchaindb/bigchaindb/issues/1126

### DIFF
--- a/tests/backend/mongodb/test_indexes.py
+++ b/tests/backend/mongodb/test_indexes.py
@@ -1,0 +1,23 @@
+import pytest
+from unittest.mock import MagicMock
+
+pytestmark = pytest.mark.bdb
+
+
+def test_asset_id_index():
+    from bigchaindb.backend.mongodb.query import get_txids_filtered
+    from bigchaindb.backend import connect
+
+    # Passes a mock in place of a connection to get the query params from the
+    # query function, then gets the explain plan from MongoDB to test that
+    # it's using certain indexes.
+
+    m = MagicMock()
+    get_txids_filtered(m, '')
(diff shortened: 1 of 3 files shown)


tests/backend/mongodb/test_indexes.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+pytestmark = pytest.mark.bdb
+
+
+def test_asset_id_index():
```
r-marques: I don't really like this. The main reason is that we are not actually testing the queries that we have.

All the points you mentioned in https://github.com/bigchaindb/bigchaindb/pull/1131#issue-204602113 are valid.
Maybe the Lazy object that @vrde is working on will help or maybe we need to find a different to solution to test the secondary indexes.

I propose that we postpone secondary index tests and handle them correctly in #1126

vrde: The commit is not there anymore, so I don't have context.

libscott: I've changed it to use MagicMock

libscott:
@r-marques looks ok now?

r-marques:
👍

PR #1136, opened 2017-02-02
Bug/1132/vote pipeline validates transactions twice

### LINKED ISSUE #1132: Vote pipeline validates transactions twice
see #1005 

`Vote.validate_block` calls `Block.validate` which validates all transactions and then calls  `Vote.validate_tx` which validates all the transactions again.

### LINKED ISSUE #1133: Consolidate block validation logic
see #1005 

In `models.Block` the block validation logic is done in both `from_dict` and `validate`.
We should consolidate all of the validation logic into `validate` while `from_dict` should only validate schema and instantiate a `Block` object.

r-marques:
resolves #1132 
resolves #1133 

Initially this pr was supposed to address the duplicate validation of transactions in the votes pipeline.
The pipeline was calling `Block.validate` (which also validates the transactions) and then `validate_tx` which validates the transactions again.
Now we want to use `validate_tx` because it parallelizes the validation of transactions.

I solved this by subdividing `Block.validate` into `Block._validate_block` and `Block._validate_block_transactions`.
This way the votes pipeline can simply call `Block._validate_block` and then validate the transactions separately.

While refactoring the `Block.validate` method I found another problem and I non-intentionally addressed #1133 in this pr.

Here the problem is that we had duplicated code for the validation of the signature.
When calling `Block.from_dict(); Block.validate()` which we usually want to do we were validating the signature twice.

So there were two solutions to remove the duplication:
1. Validate the signature in `from_dict`
2. Validate the signature in `validate`

The first approach is tricky. Before `from_dict` would work if the document did not have a signature. This is because we separate the creation of the document from the signing. If we were to do validation of the signature only on `from_dict` we would need to enforce that the signature is present always, meaning that we would no longer be able to do `from_dict` in an unsigned block.

The second approach works similar to the transaction model. `from_dict` does not validate the signature but it does include it in the model if a signature is provided.
And then the validation of the signature happens only when calling `Block.validate`

### DIFF
--- a/bigchaindb/models.py
+++ b/bigchaindb/models.py
@@ -203,6 +203,12 @@ class Block(object):
             InvalidSignature: If a Block's signature is invalid.
         """
 
+        self._validate_block(bigchain)
+        self._validate_block_transactions(bigchain)
+
+        return self
+
+    def _validate_block(self, bigchain):
         # First, make sure this node hasn't already voted on this block
         if bigchain.has_previous_vote(self.id, self.voters):
             return self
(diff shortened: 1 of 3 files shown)


bigchaindb/models.py
```diff
@@ -202,23 +202,28 @@ def validate(self, bigchain):
             OperationError: If a non-federation node signed the Block.
             InvalidSignature: If a Block's signature is invalid.
         """
+
+        self._validate_block(bigchain)
+        self._validate_block_transactions(bigchain)
+
+        return self
+
+    def _validate_block(self, bigchain):
```
ttmc: It might be worthwhile to add a comment to note that the hash (id) is validated in the `from_dict()` method, not here (in case someone wonders if/when that ever gets checked).

bigchaindb/models.py
```diff
@@ -268,34 +273,19 @@ def from_dict(cls, block_body):
             InvalidSignature: If the block's signature is not corresponding
                 to it's data or `node_pubkey`.
```
ttmc: I guess the `from_dict()` method won't raise `InvalidSignature` any more.

Issue #1141, opened 2017-02-03
Remove host and port from URLs exposed in Bigchaindb http endpoints


krish7919:
We should remove all references to the protocol, host and the port at which BigchainDB is available.
For one, the users already know about it if they can hit the root URL.
Secondly, BigchainDB will not be aware where is it running in the future - it might be frontended by a load balancer, a firewall or pretty much anything else.

Planning to remove all references to `http://localhost:59984` in the following output:
```
curl -X GET http://localhost:59984/
{
  "_links": {
    "api_v1": "http://localhost:59984/api/v1/", 
    "docs": "https://docs.bigchaindb.com/projects/server/en/0.9.dev/"
  }, 
  "keyring": [], 
  "public_key": "6F5Bi9HEgaFicPiT1FNDCPJgQBhAr8SJVUX7P2uBKLRY", 
  "software": "BigchainDB", 
  "version": "0.9.0.dev"
}
curl -X GET http://localhost:59984/api/v1
{
	"_links": {
		"docs": "https://docs.bigchaindb.com/projects/server/en/0.9.dev/drivers-clients/http-client-server-api.html",
		"statuses": "http://localhost:59984/api/v1/statuses/",
		"transactions": "http://localhost:59984/api/v1/transactions/",
		"self": "http://localhost:59984/api/v1/"
	}
}
```

TimDaub:
For the completion of this ticket, it is also necessary to handle this for the Events API specification and implementation.

TimDaub:
Consider that node operators will want to set relative paths (or as @krish7919 called it "prefixes") for `api_v1`. Description here: https://github.com/bigchaindb/bigchaindb/issues/1141

krish7919:
@TimDaub 
To reiterate, the proposal suggests stripping all the `protocol://host:port` from all the exposed URLs from our root endpoint.
I do not get the point of configuring them. The users are already aware of the information. It is overhead on our side if we go with configuring these details, not to mention the chance of a security risk if the user hits `https://host1:port1/` and the return value has `"api_v1": "https://host2:port2/api/v1/"`.

TimDaub:
Moved to "BDB 1.0: Breaking HTTP API changes, 2nd iteration HTTP API" milestone. Will handle in #1522

TimDaub:
[...]

PR #1146, opened 2017-02-03
Removed base_url()

krish7919:
- This will ensure we do not show protocol://host:port when someone hits
the root URL or root_url/api/v1

Solves #1141

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ ntools/one-m/ansible/ansible.cfg
 # Just in time documentation
 docs/server/source/schema
 docs/server/source/drivers-clients/samples
+
+#ignore .idea
+.idea/
--- a/bigchaindb/web/views/base.py
+++ b/bigchaindb/web/views/base.py
@@ -3,7 +3,7 @@ Common classes and methods for API handlers
 """
 import logging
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -81,3 +81,6 @@ ntools/one-m/ansible/ansible.cfg
 # Just in time documentation
 docs/server/source/schema
 docs/server/source/drivers-clients/samples
+
+#ignore .idea
+.idea/
```
ttmc: What's `.idea/`? If it's something you want your local Git to ignore, but it's not something _everyone_ should ignore, then put it in `bigchaindb/.git/info/exclude`

sohkai: Or a global gitignore in your home folder.

krish7919:
@TimDaub Will wait for the second iteration of the HTTP API.
@sbellem , @ttmc and I discussed this and are of the opinion that we should not expose the api endpoint info to the outside world as:
1.  That's incorrect info - not applicable in all execution environments.
2.  That the world shouldn't know where we are running the bigchaindb cluster.
3.  This only provides the endpoint info of the node the request is being serviced by, which cannot be the same the next time in a clustered setup.

Closing this PR for now.

PR #1148, opened 2017-02-03
Make sure BlockPipeline.validate_tx handles all exceptions

### LINKED ISSUE #989: Transaction validation should not throw exception in pipeline
https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/pipelines/block.py#L62

During block creation, transaction validation can throw an exception if the transaction is invalid. This should be handled because:

* A single invalid transaction should not abort creation of a block, and
* The pipeline processes do not have a restart mechanism (though even if they did, the bad transaction would need to be reassigned to avoid the same thing happening until another node considers the transaction to be stale).

r-marques:
resolves #989 

`BlockPipeline` was not handling the exceptions that can occur when calling `Transaction.from_dict`.

### DIFF
--- a/tests/pipelines/test_block_creation.py
+++ b/tests/pipelines/test_block_creation.py
@@ -39,6 +39,37 @@ def test_validate_transaction(b, create_tx):
     assert block_maker.validate_tx(valid_tx.to_dict()) == valid_tx
 
 
+def test_validate_transaction_handles_exceptions(b, signed_create_tx):
+    """
+    This test makes sure that `BlockPipeline.validate_tx` handles possible
+    exceptions from `Transaction.from_dict`.
+    """
+    from bigchaindb.pipelines.block import BlockPipeline
+    block_maker = BlockPipeline()
+
+    # Test SchemaValidationError
+    tx_dict = signed_create_tx.to_dict()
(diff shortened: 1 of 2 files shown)


tests/pipelines/test_block_creation.py
```diff
@@ -39,6 +39,37 @@ def test_validate_transaction(b, create_tx):
...
+    # Test InvalidSignature when we pass a bad fulfillment
+    tx_dict = signed_create_tx.to_dict()
+    tx_dict['inputs'][0]['fulfillment'] = 'cf:0:aaaaaaaaaaaaaaaaaaaaaaaaa'
+    assert block_maker.validate_tx(tx_dict) is None
+
+    # Test AmountError
+    signed_create_tx.outputs[0].amount = 0
+    tx_dict = signed_create_tx.to_dict()
+    # set the correct value back so that we can continue using it
+    signed_create_tx.outputs[0].amount = 1
```
sohkai: Could we just change the amount in `tx_dict` instead?

r-marques: that would force me to recalculate the `id`

sohkai: Ahh... good point.

PR #1152, opened 2017-02-06
Upgrade pysha3 to 1.0.2 to fix usage on Python 3.6 on macOS

### LINKED ISSUE #1128: sha3 missing from install_requires (macOS10.12.3/python3.6)
It appears pysha3 is currently not enough of a dependency

```bash
MAC:~ MAC$ bigchaindb
  File "/usr/local/bin/bigchaindb", line 7, in <module>
    from bigchaindb.commands.bigchain import main
  File "/usr/local/lib/python3.6/site-packages/bigchaindb/__init__.py", line 40, in <module>
    from bigchaindb.core import Bigchain  # noqa
  File "/usr/local/lib/python3.6/site-packages/bigchaindb/core.py", line 7, in <module>
    from bigchaindb.common import crypto, exceptions
  File "/usr/local/lib/python3.6/site-packages/bigchaindb/common/crypto.py", line 3, in <module>
    import sha3
  File "/usr/local/lib/python3.6/site-packages/pysha3-1.0.0-py3.6-macosx-10.12-x86_64.egg/sha3.py", line 8, in <module>
    from _sha3 import keccak_224, keccak_256, keccak_384, keccak_512
ImportError: cannot import name 'keccak_224'
```

however when installing sha3 (pip3 install sha3) the error disappears

sohkai:
Fixes #1128.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ install_requires = [
     # TODO Consider not installing the db drivers, or putting them in extras.
     'rethinkdb~=2.3',  # i.e. a version between 2.3 and 3.0
     'pymongo~=3.4',
-    'pysha3==1.0.0',
+    'pysha3~=1.0.2',
     'cryptoconditions>=0.5.0',
     'python-rapidjson>=0.0.8',
     'logstats>=0.2.1',


setup.py
```diff
@@ -63,7 +63,7 @@ def check_setuptools_features():
     # TODO Consider not installing the db drivers, or putting them in extras.
     'rethinkdb~=2.3',  # i.e. a version between 2.3 and 3.0
     'pymongo~=3.4',
-    'pysha3==1.0.0',
+    'pysha3==1.0.2',
```
sbellem: From [Python Packaging User Guide](https://packaging.python.org/requirements/#install-requires):

We could use the `~` instead:

```python
pysha3~=1.0`
```

or 

```python
pysha3~=1.0.2`
```

sohkai: Yeah... not sure if there was a reason behind pinning it in the first place, but `~=` sounds good to me. @libscott do you think we should keep pinning it?

libscott: I think we should eventually pin everything for perhaps that's for post 1.0. Fine with unpinning for now :+1:

PR #1164, opened 2017-02-07
Various modifications to the Release_Process.md

ttmc:
* Clarify that the 0.X branch is for all 0.X.Y releases (tags), including 0.X.0, 0.X.1, 0.X.2, etc.
* Don't merge the 0.X.0 release into the `master` branch. The CHANGELOG.md file was already updated before branching off the master branch.
* In the patch release process, added a step to cherry-pick the commit to update the CHANGELOG.md file.
* In the common steps, added an optional `make clean` step before the PyPI steps.

Any comments or suggestions?

### DIFF
--- a/Release_Process.md
+++ b/Release_Process.md
@@ -41,11 +41,12 @@ These steps are common between minor and patch releases:
 1. Fill in the details:
    - Tag version: version number preceeded by 'v', e.g. "v0.9.1"
    - Target: the release branch that was just pushed
-   - Title: Same as tag name
-   - Description: The body of the changelog entry (Added, Changed etc)
-1. Publish the release on GitHub
-1. (Optional) Do a `make clean` in the repository root directory (`bigchaindb/`)
-1. Generate the release tarball with `python setup.py sdist`. Upload the release to PyPI, e.g. `twine upload dist/BigchainDB-0.9.0.tar.gz` (requires you to have a `~/.pypirc` file)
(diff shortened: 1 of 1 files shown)


Release_Process.md
```diff
@@ -42,7 +43,8 @@ These steps are common between minor and patch releases:
    - Target: the release branch that was just pushed
    - Title: Same as tag name
    - Description: The body of the changelog entry (Added, Changed etc)
-1. Publish the release on Github
+1. Publish the release on GitHub
+1. (Optional) Do a `make clean` in the repository root directory (`bigchaindb/`)
```
sbellem: Perhaps, to prevent troubles, it would be best to remove the "(Optional)".

Also, it is important that the PyPI release matches the tagged version in github. For this reason, an extra line, before, the `make clean` instructing to checkout the tag would be necessary, e.g.:

```
$ git fetch <remote-name>
$ git checkout v0.9.1
```
and then `make clean`, `python setup.py sdist`, ...

sbellem: @ttmc You may like a more compact approach via the `Makefile` as in https://github.com/bigchaindb/bigchaindb/pull/1165

We could then replace the three steps (`make clean, ..., twine upload`) by one: `make release`

ttmc: Good points and ideas:  I'll add a line about checking out the tag generated by the GitHub release, and a line about `make release`.

Issue #1153, opened 2017-02-06
Decide what Docker containers to create & create them


ttmc:
We need to decide what Docker containers to create (i.e. what should they contain) for different purposes:
* for people to try out BigchainDB Server on Windows, macOS, or whatever
* for development
* for internal testing
* for production

Ideally, the same containers can be used for many or all of these use cases.

krish7919:
I would suggest having 2 containers:
1.  BigchainDB Server
2.  MongoDB has [released official containers](https://hub.docker.com/r/library/mongo/)
3.  BigchainDB driver.

We can use all of them for dev, internal testing, dev and prod.
Decoupling DB from Bigchain is better as:
1.  Upgrade to bigchaindb does not affect mongodb and vice versa.
2.  BigchainDB crash does not affect mongodb and vice versa.
3.  We can support multiple transaction models with the same db backend in the future. In case our transaction schema changes, we can have a mixture of old + new transaction schema versions running over the same database cluster, before the entire cluster is migrated.

ttmc:
We shouldn't forget RethinkDB.

krish7919:
@ttmc : RethinkDB has official containers too. https://hub.docker.com/_/rethinkdb/
This is what we use as base image for bigchaindb docker container currently.

trentmc:
This is for IPDB right? IPDB is going to be deployed with MongoDB. So why bother with RethinkDB?

ttmc:
Because IPDB isn't our only user? Everyone, up til now, has been using RethinkDB. Some may opt to continue using it.

trentmc:
OK, this isn't just for IPDB then (that was my first question). No problem then:)

krish7919:
We have decided to decouple the backend database from bigchaindb when running in containers.
We will use the official rethinkdb/mongodb container images hosted in dockerhub.

We will also use the docker0 vnic (Ubuntu 16.04, docker 1.13) to connect the backend DB and bigchainDB together.

The general workflow is:

##### Configure BDB:
```
docker run \
  --rm \
  -v "$HOME/bigchaindb_docker:/data" \
  bigchaindb/krish-test \
  -y configure mongodb
```
OR
```
docker run \
  --rm \
  -v "$HOME/bigchaindb_docker:/data" \
  bigchaindb/krish-test \
  -y configure rethinkdb
```

##### Run backend DB:
```
docker run \
  --name=mongodb \
  --publish=172.17.0.1:27017:27017 \
[...]

PR #1174, opened 2017-02-08
Decouple backend DB and BigchainDB Docker containers

krish7919:
Solved #1153.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rethinkdb:2.3
+FROM ubuntu:xenial
 
 # From http://stackoverflow.com/a/38553499
 
@@ -34,8 +34,6 @@ ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 # but maybe our Docker or Docker Compose stuff does?
 # ENV BIGCHAINDB_API_ENDPOINT http://bigchaindb:9984/api/v1
 
-ENTRYPOINT ["bigchaindb", "--dev-start-rethinkdb", "--dev-allow-temp-keypair"]
+ENTRYPOINT ["bigchaindb"]
 
 CMD ["start"]
-
-EXPOSE 8080 9984 28015 29015
--- a/docs/server/source/appendices/run-with-docker.md
+++ b/docs/server/source/appendices/run-with-docker.md
@@ -49,9 +49,6 @@ Let's analyze that command:
 * `bigchaindb/bigchaindb` the image to use. All the options after the container name are passed on to the entrypoint inside the container.
 * `-y configure` execute the `configure` sub-command (of the `bigchaindb`
  command) inside the container, with the `-y` option to automatically use all the default config values
-* `--dev-allow-temp-keypair` specifies that this is a dev environment and
-  enables creation of temporary key pairs to be written to the configuration
-  file
 * `mongodb` or `rethinkdb` specifies the database backend to use with bigchaindb
 
 
--- a/docs/server/source/appendices/run-with-docker.md
+++ b/docs/server/source/appendices/run-with-docker.md
@@ -27,7 +27,7 @@ docker run \
   --tty \
   --volume "$HOME/bigchaindb_docker:/data" \
   bigchaindb/bigchaindb \
-  -y configure --dev-allow-temp-keypair \
+  -y configure \
   [mongodb|rethinkdb]
 
 Generating keypair
@@ -58,7 +58,7 @@ Let's analyze that command:
 ### Run the backend database
 From v0.9 onwards, you can run either RethinkDB or MongoDB.
 
-We use the virtual interface created by the docker daemon to allow
+We use the virtual interface created by the Docker daemon to allow
 communication between the BigchainDB and database containers.
 It has an IP address of 172.17.0.1 by default.
 
--- a/docs/server/source/appendices/run-with-docker.md
+++ b/docs/server/source/appendices/run-with-docker.md
@@ -15,13 +15,21 @@ In a terminal shell, pull the latest version of the BigchainDB Docker image usin
 docker pull bigchaindb/bigchaindb
 ```
 
-then do a one-time configuration step to create the config file; we will use
+### Configuration
+A one-time configuration step is required to create the config file; we will use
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -34,8 +34,6 @@ ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 # but maybe our Docker or Docker Compose stuff does?
 # ENV BIGCHAINDB_API_ENDPOINT http://bigchaindb:9984/api/v1
 
-ENTRYPOINT ["bigchaindb", "--dev-start-rethinkdb", "--dev-allow-temp-keypair"]
+ENTRYPOINT ["bigchaindb"]
```
ttmc: I don't understand why you can remove the `--dev-allow-temp-keypair` flag.

Dockerfile
```diff
@@ -34,8 +34,6 @@ ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
...
 
 CMD ["start"]
-
-EXPOSE 8080 9984 28015 29015
```
ttmc: I don't understand why port 9984 isn't exposed. (The other ones are for RethinkDB, which is gone from this container, so that much is clear.)

docs/server/source/appendices/run-with-docker.md
```diff
@@ -30,45 +38,84 @@ Ready to go!
...
  you can read more in the [official Docker
  documentation](https://docs.docker.com/engine/tutorials/dockervolumes/#/mount-a-host-directory-as-a-data-volume)
-* `-t` allocate a pseudo-TTY
-* `-i` keep STDIN open even if not attached
-* `bigchaindb/bigchaindb` the image to use
-* `-y configure` execute the `configure` sub-command (of the `bigchaindb` command) inside the container, with the `-y` option to automatically use all the default config values
+* `bigchaindb/bigchaindb` the image to use. All the options after the container name are passed on to the entrypoint inside the container.
+* `-y configure` execute the `configure` sub-command (of the `bigchaindb`
+ command) inside the container, with the `-y` option to automatically use all the default config values
+* `--dev-allow-temp-keypair` specifies that this is a dev environment and
```
ttmc: The `--dev-allow-temp-keypair` option generates one temporary keypair if none is found, but it's usually used with the `bigchaindb start` command, not the `bigchaindb configure` command, e.g. see: https://docs.bigchaindb.com/projects/server/en/master/server-reference/bigchaindb-cli.html#bigchaindb-start  

Doesn't `bigchaindb -y configure` generate a keypair?

docs/server/source/appendices/run-with-docker.md
```diff
@@ -30,45 +38,84 @@ Ready to go!
...
+  enables creation of temporary key pairs to be written to the configuration
+  file
+* `mongodb` or `rethinkdb` specifies the database backend to use with bigchaindb
+
+
+### Run the backend database
+From v0.9 onwards, you can run either RethinkDB or MongoDB.
+
+We use the virtual interface created by the docker daemon to allow
```
ttmc: This is a small thing, but "Docker" is a proper noun and should be capitalized when used in regular English prose. The exception is when you're referring specifically to the `docker` command.

docs/server/source/appendices/run-with-docker.md
```diff
@@ -30,45 +38,84 @@ Ready to go!
...
+communication between the BigchainDB and database containers.
+It has an IP address of 172.17.0.1 by default.
```
ttmc: This is interesting! Can you link to the Docker documentation about 172.17.0.1? I can't help but wonder if there's a less arcane way to do this.

Dockerfile
krish7919: As I understand, `--dev-allow-temp-keypair` flag is used to create a [temp keypair](https://github.com/bigchaindb/bigchaindb/blob/e6d99d74c022e6ca7954786b7db106f9fe36d98a/bigchaindb/commands/bigchain.py#L183).
If we go into production with this, we should be able to set a keypair ourselves rather than let bigchaindb generate it.
We also need to do key management and periodic recycling of the keys.
I have removed it as a mandatory `entrypoint` option, and let devs specify it explicitly during testing, as is documented below.

Dockerfile
krish7919: This is just a hint to docker to expose ports for container networking.
The real option required is --publish over here. We were exposing non-required ports before.

docs/server/source/appendices/run-with-docker.md
krish7919: We would want a temp keypair created when we set up the configuration file.
Is it really require a `--dev-allow-temp-keypair` option? It can fail even if no conf file is found.

docs/server/source/appendices/run-with-docker.md
krish7919: We will not use the docker vnic in production, as both containers need routable IPs.
From [here](https://docs.docker.com/engine/userguide/networking/):
`
The bridge network represents the docker0 network present in all Docker installations. Unless you specify otherwise with the docker run --network=<NETWORK> option, the Docker daemon connects containers to this network by default. You can see this bridge as part of a host’s network stack by using the ifconfig command on the host.
`

docs/server/source/appendices/run-with-docker.md
ttmc: Here's what happens when the file `$HOME/.bigchaindb` file does not exist:
```text
$ bigchaindb -y configure rethinkdb
Generating keypair
Generating default configuration for backend rethinkdb
Configuration written to /home/troy/.bigchaindb
Ready to go!
```
It generates a keypair.

The `--dev-allow-temp-keypair` option exists (as an option to go with `bigchaindb start`) for the cases when you want to be able to do `bigchaindb start` _without_ doing a `bigchaindb -y configure <backend>` first. Most BigchainDB configuration variables have default values: the only exception is the keypair: there is no default keypair.

ttmc: Anyway, I see you removed the `--dev-allow-temp-keypair` from the actual `docker` command (above); this comment about `--dev-allow-temp-keypair` should also be removed.

PR #1184, opened 2017-02-11
Added some MongoDB ports to docs Notes for Firewall Setup

ttmc:
Based on https://docs.mongodb.com/manual/reference/default-mongodb-port/


docs/server/source/appendices/firewall-notes.md
```diff
@@ -63,6 +64,11 @@ You may want to have Gunicorn and the reverse proxy running on different servers
 Port 28015 is the default port used by RethinkDB client driver connections (TCP). If your BigchainDB node is just one server, then Port 28015 only needs to listen on localhost, because all the client drivers will be running on localhost. Port 28015 doesn't need to accept inbound traffic from the outside world.
 
 
+## Port 28017
```
r-marques: Not sure what we should mention about this port.
The HTTP interface is disabled by default and in the documentation they say that it should not be enabled in production.
They also have a not saying that it is deprecated since version 3.2 and a warning saying that the authentication mechanisms are not support on the HTTP interface

https://docs.mongodb.com/v3.2/core/security-mongodb-configuration/

ttmc: Oh! I didn't know that was deprecated since version 3.2. I removed all mentions of that port.

PR #1193, opened 2017-02-15
Add stepping pipeline (test tool) and fix issue #1191

### LINKED ISSUE #1191: MongoDB changefeed not fetching updated doc
In the mongodb changefeed, when the operation is an update, the query to return a record returns `None`. The pipelines code does not notice the error because `None` is silently discarded.

https://github.com/bigchaindb/bigchaindb/blob/0c8927dbbe80471f59241b556b24e31dc4396c3f/bigchaindb/backend/mongodb/changefeed.py#L88

The code is:

```python
doc = self.connection.conn[dbname][table].find_one(
    {'_id': record['o2']},                         
    {'_id': False}                                 
)                                                  
```

The `record` is:

    {'ts': Timestamp(1487159111, 2), 't': 1, 'h': 5958958684939816142, 'o2': {'_id': ObjectId('58a43f4791e1123be4bc91bb')}, 'o': {'$set': {'assignment_timestamp': 1487159111.4288526}}, 'ns': 'bigchain_test.backlog', 'op': 'u', 'v': 2}

So instead of `{'_id': record['o2']}`, it should be `{'_id': record['o2']['_id']}`

libscott:
This PR adds the stepping pipeline as a module in the test suite and uses it to check that the changefeed produces updates as expected.

* Add MultipipesStepper
* Fix #1191 
* Make Bigchain class accept `backlog_reassign_delay=0` (it was not making distinction between 0 and None previously)

Fixes #1191

### DIFF
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -55,7 +55,9 @@ class Bigchain(object):
         self.me = public_key or bigchaindb.config['keypair']['public']
         self.me_private = private_key or bigchaindb.config['keypair']['private']
         self.nodes_except_me = keyring or bigchaindb.config['keyring']
-        self.backlog_reassign_delay = backlog_reassign_delay or bigchaindb.config['backlog_reassign_delay']
+        if backlog_reassign_delay is None:
+            backlog_reassign_delay = bigchaindb.config['backlog_reassign_delay']
+        self.backlog_reassign_delay = backlog_reassign_delay
         self.consensus = BaseConsensusRules
         self.connection = connection if connection else backend.connect(**bigchaindb.config['database'])
         if not self.me or not self.me_private:
--- a/tests/pipelines/stepping.py
+++ b/tests/pipelines/stepping.py
@@ -2,6 +2,33 @@
 Pipeline stepping is a way to advance the asynchronous data pipeline
 deterministically by exposing each step separately and advancing the states
 manually.
+
+The multipipes.Pipeline class implements a pipeline that advanced
+asynchronously and concurrently. This module provides an interface to the
+BigchainDB pipelines that is static, ie, does not advance without prompting.
+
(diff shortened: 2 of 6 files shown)


bigchaindb/core.py
```diff
@@ -55,7 +55,9 @@ def __init__(self, public_key=None, private_key=None, keyring=[], connection=Non
         self.me = public_key or bigchaindb.config['keypair']['public']
         self.me_private = private_key or bigchaindb.config['keypair']['private']
         self.nodes_except_me = keyring or bigchaindb.config['keyring']
-        self.backlog_reassign_delay = backlog_reassign_delay or bigchaindb.config['backlog_reassign_delay']
+        if backlog_reassign_delay is None:
+            backlog_reassign_delay = bigchaindb.config['backlog_reassign_delay']
+        self.backlog_reassign_delay = backlog_reassign_delay
```
ttmc: Do these three lines do something different from what the original one line did?

libscott: Yea, previously it would not accept a 0 second timeout (thinking it was None)

sohkai:
Given that you've fixed:

> * Fix #1191
> * Make Bigchain class accept backlog_reassign_delay=0 (it was not making distinction between 0 and None previously)

It might be nice to pull out the stepper from this PR so that it only gets reviewed in #1199.

tests/pipelines/stepping.py
```diff
@@ -0,0 +1,138 @@
...
+from contextlib import contextmanager
+from unittest.mock import patch
+
+import bigchaindb.core
+from bigchaindb.backend.changefeed import ChangeFeed
+import bigchaindb.pipelines.block
+import bigchaindb.pipelines.stale
+
+
+class MultipipesStepper:
```
r-marques: Can you document this class? It would make it easier to review

PR #1195, opened 2017-02-16
Docs: clarified how block id & signature are calculated

ttmc:
The docs about how the block id and signature are calculated weren't strictly _wrong_, they were just a bit vague, so I added some more details and clarifications. I also added links to the page in the Appendices about how we do hashes and signatures in BigchainDB.

### DIFF
--- a/docs/server/source/data-models/block-model.rst
+++ b/docs/server/source/data-models/block-model.rst
@@ -27,7 +27,7 @@ A block has the following structure:
       It's the list of federation nodes which can cast a vote on this block.
       This list can change from block to block, as nodes join and leave the federation.
 
-- ``signature``: :ref:`Cryptographic signature <Signature Algorithm and Keys>` of the block by the node that created the block. To generate the signature, the node builds a dict including the ``id``, the inner ``block`` & the ``signature`` (with a value of ``None``), it serializes that dict, and then signs *that* with its private key.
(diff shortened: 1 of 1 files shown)


docs/server/source/data-models/block-model.rst
```diff
@@ -27,7 +27,7 @@ A block has the following structure:
       It's the list of federation nodes which can cast a vote on this block.
       This list can change from block to block, as nodes join and leave the federation.
 
-- ``signature``: Cryptographic signature of the block by the node that created the block. (To create the signature, the node serializes the block contents and signs it with its private key.)
+- ``signature``: :ref:`Cryptographic signature <Signature Algorithm and Keys>` of the block by the node that created the block. To generate the signature, the node builds a dict including the ``id``, the inner ``block`` & the ``signature`` (with a value of ``None``), it serializes that dict, and then signs *that* with its private key.
```
r-marques: > signature of the block by the node that created the block (`node_pubkey`)

It may be worth to say that the block needs to be signed by `node_pubkey` 

> To generate the signature, the node builds a dict including the ``id``, the inner ``block`` & the ``signature`` (with a value of ``None``)

This is wrong. The signature is only done on the inner block (`timestamp`, `transactions`, `node_pubkey`, `voters`) https://github.com/bigchaindb/bigchaindb/blob/ae029c7bcfebbe8cfcb2058a642c5ddddcf641b1/bigchaindb/models.py#L281

ttmc: I added a note about `node_pubkey`.

Good catch on how the block signature is calculated. Somehow I missed that line. I've fixed it.

It's nice to know that the `id` and `signature` are based on the same serialized message: the inner `block`.

r-marques: Yes, I also prefer it this way

PR #1196, opened 2017-02-16
[WIP] Watchdog

libscott:
Watchdog is a tool for checking pipeline correctness which tracks MongoDB database operations using the oplog, and for each operation, checks that it does not violate some set of correctness assertions.

The scope of this issue / PR is to implement tests to show that the watchdog can catch all violations related to correctness **of the transaction graph**. This means that transaction structural correctness is not considered in detail.

- [x] BLOCK ELECTION STATUS CHANGE
- [x] DOUBLE SPEND SINGLE BLOCK
- [x] DOUBLE SPEND ACROSS BLOCKS
- [x] DUPLICATE TX SINGLE BLOCK
- [x] DUPLICATE TX ACROSS BLOCKS
- [x] INPUT DOES NOT EXIST
- [x] TX BALANCE ERROR (sum inputs == sum outputs)
- [ ] All fulfillments must contain signature
- [ ] All signatures are verified using Tx identity and crypto-condition of output
- [ ] Threshold and Ed25519 condition is supported (up to some size limit)
- [ ] We are able to provide non embarrassing timestamp


libscott:
Being put on the backburner for now. Will re-open later.

PR #1199, opened 2017-02-17
No double inclusion

### LINKED ISSUE #1150: Transaction already in valid block is not checked
Solves #1140 

During the process of creating a block, we do not check if transactions have already been included in a valid block. This is not an issue for TRANSFER transactions because a double spend would be detected, but it is an issue with CREATE transactions because they do not generate double spends.

Edit: It COULD be an issue for transfers, where the same transfer is submitted twice and not identified as a double spend. See #592 

Outstanding question: Should this be part of the `models.Transaction.validate()` method? Should validate() called on an existing, already included transaction check that it has not been included? The description is then incorrect, it would be more correct to then call the method: `is_eligible_for_inclusion`.

libscott:
This PR addresses the problem of multiple inclusion of a transaction, since the vote pipeline was not previously performing this check.

This also includes the PR #1193 which is not yet merged.

Fixes #1150


bigchaindb/pipelines/vote.py
```diff
@@ -101,6 +101,15 @@ def validate_tx(self, tx, block_id, num_tx):
             Three values are returned, the validity of the transaction,
             ``block_id``, ``num_tx``.
         """
+
+        # If tx in any undecided or valid blocks that are not this block, say
+        # not valid
+        validity = self.bigchain.get_blocks_status_containing_tx(tx.id)
```
ttmc: Could you give `validity` a better name? What is it, exactly?

ttmc: `block_statuses` is much better :+1:

tests/pipelines/stepping.py
```diff
@@ -0,0 +1,148 @@
...
+from unittest.mock import patch
+
+import bigchaindb.core
+from bigchaindb.backend.changefeed import ChangeFeed
+import bigchaindb.pipelines.block
+import bigchaindb.pipelines.stale
+import bigchaindb.pipelines.vote
+
+
+class MultipipesStepper:
```
sohkai: This module's really cool!

I just wish there were more comments (maybe on each function?) to explain how it hooks up to multipipes' nodes and what some of the state is useful for (e.g. `self.queues` vs. `self.tasks` vs. `self.processes`) and what they expect to contain (e.g. `self.tasks` expects things like `{pipeline_name}_{node_name}`). But maybe most of this is crystal clear to @vrde and others who have a better memory of multipipes than I do?

vrde: I like this as well (I was trying to achieve something similar inside Multipipes, it was called [*sequential strategy*](https://github.com/bigchaindb/multipipes/blob/master/tests/test_sequential_strategy.py) but it was considering the pipeline as a whole). Would be nice to have some of this code in the `multipipes` module.

The only thing I don't like that much is the signature of the method, that is `stepper.<pipeline_name>_<pipeline_function>`, why not having `stepper.<pipeline_name>.<pipeline_function>`?

libscott: Thanks! :D I did it kinda quickly. I should put a block comment in `MultipipesStepper`, it's basically an analogue of `Pipeline` for deterministic stepping of the pipeline.

libscott:
> The only thing I don't like that much is the signature of the method, that is stepper.<pipeline_name>_<pipeline_function>, why not having stepper.<pipeline_name>.<pipeline_function>?

That would take more code :smile: 

I agree it would look more intuitive. IN THEORY.

tests/pipelines/stepping.py
libscott: Comments have been added in #1193

libscott:
Closing this, #1199 has been superceeded by #1206. 

Testing approach (stepping / watchdog, cleanup tests tech debt) will be addressed separately.

PR #1201, opened 2017-02-20
Docs: Improvements to the Quickstart page

ttmc:
Suggested by Trent

### DIFF
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -55,4 +55,22 @@ $ bigchaindb start
 
 That's it!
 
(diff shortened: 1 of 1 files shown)


ttmc:
Trent gave me some more feedback (verbally), and as a result, I:

* removed the choice of installing RethinkDB or MongoDB. (It's a Quickstart page, not an exhaustive exploration of all the options and how to use them.) Now the page just tells them to install MongoDB and gives MongoDB-specific instructions.
* made it so there's only one obvious "next step" link at the end of the page.

PR #1213, opened 2017-02-23
Identify voters + Voting class

### LINKED ISSUE #1217: Block state change INVALID -> VALID due to previous_block
Given a 3 node setup:

1. One node vote for block C with previous block A. Status is UNDECIDED.
2. Another node votes for block C with previous block B. Majority now says VALID but disagrees on previous block. Status is INVALID.
3. The last node votes for block C with previous block B. Majority now agrees on previous block. Status is VALID.

https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/core.py#L689

libscott:
The integration of this feature (to replace `Bigchain.block_election_status`) is being done in another PR: #1215, it is split in two to make it easier to review.

In order to fix some bugs* and make the voting logic more robust and testable, a new module is introduced `bigchaindb.voting`. The module is being created with the goals of correctness, testability and readability.

Scope:

- [x] Create module `bigchaindb.voting` with pure functions for vote counting
- [x] Implement test for vote counting that uses hardcoded table of thresholds
- [x] Votes with incorrect schema are considered eligible, as long as signature is correct.
- [x] Eligible voters are part of block voter list AND node keyring
- [x] Contains all functions for vote verification
- [x] Fix #1217 
- [x] Implement test for voting that checks state transitions

### DIFF
--- a/bigchaindb/voting.py
+++ b/bigchaindb/voting.py
@@ -70,7 +70,7 @@ class Voting:
         * Votes must agree on previous block, otherwise they become invalid.
 
         note:
-            The sum of votes returned by this function does not neccesarily
+            The sum of votes returned by this function does not necessarily
             equal the length of the list of votes fed in. It may differ for
             example if there are found to be multiple votes submitted by a
             single voter.
--- a/bigchaindb/voting.py
+++ b/bigchaindb/voting.py
@@ -1,4 +1,5 @@
 import collections
+from bigchaindb.backend.exceptions import BigchainDBCritical
 from bigchaindb.common.schema import SchemaValidationError, validate_vote_schema
 from bigchaindb.common.utils import serialize
 from bigchaindb.common.crypto import PublicKey
@@ -28,11 +29,12 @@ class Voting:
         """
         Calculate the election status of a block.
         """
-        eligible_voters = set(block['voters']) & set(keyring)
+        eligible_voters = set(block['block']['voters']) & set(keyring)
         eligible_votes, ineligible_votes = \
             cls.partition_eligible_votes(votes, eligible_voters)
+        n_voters = len(eligible_voters)
         results = cls.count_votes(eligible_votes)
-        results['status'] = cls.decide_votes(results['counts'])
+        results['status'] = cls.decide_votes(n_voters, **results['counts'])
         results['ineligible'] = ineligible_votes
         return results
 
@@ -48,12 +50,11 @@ class Voting:
             voter_eligible = vote.get('node_pubkey') in eligible_voters
             if voter_eligible:
                 try:
-                    cls.verify_vote_signature(vote)
+                    if cls.verify_vote_signature(vote):
+                        eligible.append(vote)
+                        continue
                 except ValueError:
                     pass
-                else:
-                    eligible.append(vote)
-                    continue
             ineligible.append(vote)
 
         return eligible, ineligible
@@ -94,17 +95,19 @@ class Voting:
                 n_invalid += 1
                 continue
 
-            prev_blocks[vote['vote']['previous_block']] += 1
             if vote['vote']['is_block_valid']:
+                prev_blocks[vote['vote']['previous_block']] += 1
                 n_valid += 1
             else:
                 n_invalid += 1
+
+        n_prev = prev_blocks.most_common()[0][1] if prev_blocks else 0
 
         return {
             'counts': {
                 'n_valid': n_valid,
                 'n_invalid': n_invalid,
-                'n_agree_prev_block': prev_blocks.most_common()[0][1],
+                'n_agree_prev_block': n_prev,
             },
             'cheat': cheat,
             'malformed': malformed,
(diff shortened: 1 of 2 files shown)


bigchaindb/voting.py
```diff
@@ -0,0 +1,152 @@
...
+        n_voters = len(eligible_voters)
+        results = cls.count_votes(eligible_votes)
+        results['status'] = cls.decide_votes(n_voters, **results['counts'])
+        results['ineligible'] = ineligible_votes
+        return results
+
+    @classmethod
+    def partition_eligible_votes(cls, votes, eligible_voters):
+        """
+        Filter votes from unknown nodes or nodes that are not listed on
```
r-marques: In this case an ineligible vote is not only a vote from an unknown node but also an incorrectly constructed vote from a known node.

bigchaindb/voting.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @classmethod
+    def verify_vote_signature(cls, vote):
+        """
+        Verify the signature of a vote
+        """
+        signature = vote.get('signature')
+        pk_base58 = vote.get('node_pubkey')
+
+        if not (type(signature) == str and type(pk_base58) == str):
```
r-marques: shouldn't this be detected by the schema validator?

bigchaindb/voting.py
libscott: Well yes; with the exception that, if there is a badly constructed / wrong previous block vote but correctly signed, it counts towards the INVALID count. This reduces the chances that a block gets stuck as UNDECIDED.

bigchaindb/voting.py
libscott: It would be, but signature checking happens before schema validation.

tests/test_voting.py
```diff
@@ -0,0 +1,203 @@
...
+
+    votes = [Bigchain(v).vote('block', 'a', True) for v in voters]
+    votes[2]['vote']['is_block_valid'] = False
+    # Incorrect previous block subtracts from n_valid and adds to n_invalid
+    votes[-1]['vote']['previous_block'] = 'z'
+
+    assert TestVoting.count_votes(votes) == {
+        'counts': {
+            'n_valid': 9,    # 9 kosher votes
+            'n_invalid': 4,  # 1 cheat, 1 invalid, 1 malformed, 1 rogue prev block
```
vrde: `votes` has `14` elements, but `counts.n_valid + counts.n_invalid` yields `13`. As far as I understand this is because a node voted twice and only one vote is considered. What if the sum of `counts` is always equal to `len(votes)`? Would it make sense to add `n_cheats`?

vrde: Another idea is to change (and flatten) the structure of the dict to something like:
```python
{
    'voters_for_valid`: 9,
    'voters_for_invalid`: 4,
    'cheat': [],
    # ...
}
```

so it's clear that the number returned by the function is related to the **voters**, not to the **votes**.

Hope it doesn't sound super picky, but I actually spent a bit of time trying to understand that.

tests/test_voting.py
```diff
@@ -0,0 +1,203 @@
...
+        'malformed': [],
+        'cheat': [],
+    }
+
+
+################################################################################
+# Tests for vote decision making
+
+
+DECISION_TESTS = [dict(
```
vrde: What about having just the result here? It's dumber but easier to process by a brain.

```python
[{'n_invalid': 1, 'n_valid': 1, 'n_voters': 1},
 {'n_invalid': 1, 'n_valid': 2, 'n_voters': 2},
 {'n_invalid': 2, 'n_valid': 2, 'n_voters': 3},
 {'n_invalid': 2, 'n_valid': 3, 'n_voters': 4},
 {'n_invalid': 3, 'n_valid': 3, 'n_voters': 5},
 {'n_invalid': 3, 'n_valid': 4, 'n_voters': 6},
 {'n_invalid': 4, 'n_valid': 4, 'n_voters': 7},
 {'n_invalid': 4, 'n_valid': 5, 'n_voters': 8}]
```

bigchaindb/voting.py
```diff
@@ -0,0 +1,152 @@
...
+from bigchaindb.common.utils import serialize
+from bigchaindb.common.crypto import PublicKey
+
+
+VALID = 'valid'
+INVALID = 'invalid'
+UNDECIDED = 'undecided'
+
+
+class Voting:
```
vrde: all methods here are `classmethod`s, would it make sense to remove the class and just have functions?

tests/test_voting.py
libscott: Sure; having just the table made it quite easy to look at the numbers and see what they meant though.

bigchaindb/voting.py
libscott: I started out with just functions, but then I used classmethods so it's easy to create a derived implementation that overrides one function (to modify the schema for example).

tests/test_voting.py
libscott: > As far as I understand this is because a node voted twice and only one vote is considered

That's right. I added a comment to the test to make it more clear, hope that's ok.

PR #1215, opened 2017-02-23
Voting class integration

### LINKED ISSUE #1166: MultipleVotesError when there are votes from unknown nodes
See [example](https://github.com/bigchaindb/bigchaindb/commit/89d816055418dcb3c3d3227a03015f8c870df9f7)

Each block contains a list of public keys which are the voting nodes for that block. If there are more votes than voters in the block, an exception is thrown:

> bigchaindb.common.exceptions.MultipleVotesError: Block cfbfb0b125e1f6cba22bf97ecf5ae2c10df76deb525216cbc79137e204347907 has 11 votes cast, but only 1 voters

Well behaved nodes should not vote for a block where they are not part of the voters set. But this is a sybill attack vector which we can protect against.

### LINKED ISSUE #1168: Possible to create valid block via sybill
[Example](https://github.com/bigchaindb/bigchaindb/commit/39f689e8de95042ef2ee4586ee799d00207ffe04)

Relates to #1166 

Problems:

* Possible to create block listing fake nodes as voters which is then voted VALID, ie, voters don't validate block voter lists against their own voters
* Invalid votes (wrong signature) are counted such that they affect voting outcome, so any node with database access can DOS the system.

Solution:

Block is VALID if:
* Majority (>50%) of ELIGIBLE votes say VALID & agree on previous block
* OR: Dependent (later) block is valid (a vote on a block is an acknowledgement of validity of previous blocks) (this is new)

Vote is ELIGIBLE if:
* Is only vote from node n for block b
* Signature matches pubkey and payload
* Is in block voter list
* Is in federation keyring (this is new)

### LINKED ISSUE #1217: Block state change INVALID -> VALID due to previous_block

libscott:
SEE FIRST: #1213

This is the integration of Voting class, split into a separate PR to as to make it easier to review.

Scope:

- [x] Thorough testing of voting functions
- [x] Moving / fixing / removing tests that get broken
- [x] ~Revisit @ttmc 's [proposal](https://github.com/bigchaindb/bigchaindb/issues/1168#issuecomment-281295382) about block voter list and keyring constraint~ (will be addressed in separate issue, #1224 )
- [x] Validate votes in `has_previous_vote`

Fixes #1166 
Fixes #1168
Fixes #1217


bigchaindb/core.py
```diff
@@ -494,36 +486,20 @@ def validate_block(self, block):
...
-        has_previous_vote = False
-        if votes:
-            if utils.verify_vote_signature(voters, votes[0]):
-                has_previous_vote = True
-            else:
-                raise exceptions.ImproperVoteError('Block {block_id} already has an incorrectly signed vote '
-                                                   'from public key {me}'.format(block_id=block_id, me=self.me))
-
-        return has_previous_vote
+        el, _ = self.consensus.voting.partition_eligible_votes(votes, [self.me])
```
r-marques: This can allow a node to vote multiple times in the same block.
For example if a node created one vote with an invalid signature (ineligible) this will return false meaning that the voting pipeline will cast a new vote. This may fail with mongodb if #1258 is accepted.

The question is: should a node be able to only vote once, or should a node be able to vote multiple times as long as only one of the votes is eligible?

libscott: Would this happen in practice? Vote pipeline listens to block changefeed, so it shouldn't get itself into a loop. Also, should we even be calling has_previous_vote from the vote pipeline? is there a case where it could return True and so a vote is not cast and the pipeline is functioning correct with respect to `previous_block`?

r-marques: Good question. I don't know, don't remember the need to check if a block already has a vote in the voting pipeline

CsterKuroi:
There are 3 nodes A,B,C. Dishonest node A creates a block with an Illegal transaction ,and modifies block['voters']=[pubkey_A]    (Normally there will be three public keys)
image
then A votes TRUE for the block. Finally ,this block will be valid.

libscott:
@CsterKuroi, well spotted! We do actually have [an issue](https://github.com/bigchaindb/bigchaindb/issues/1224) for this so it's definitely on our radar. Nice to know there are other diligent eyes on the code though! :+1:

PR #1218, opened 2017-02-24
Refactor commands

sohkai:
A few clean ups for `bigchaindb/commands`.

Builds on #1211.

Highlights:

* Moved `input_on_stderr()` to `commands/utils.py` (591b42c)
* Use `stderr` for informational messages (51720e9)
* Move `--dev-allow-temp-keypair` and `--dev-start-rethinkdb` to be only under `bigchaindb start` since it's only used there (e902c8c)
* Use a decorator to set up configuration before running each command (ccfb5c0)
* Move tests for `commands.utils.py` to `tests/commands/test_utils.py` (235c202)

### DIFF
--- a/bigchaindb/commands/bigchain.py
+++ b/bigchaindb/commands/bigchain.py
@@ -287,16 +287,6 @@ def create_parser():
         description='Control your BigchainDB node.',
         parents=[utils.base_parser])
 
-    parser.add_argument('--dev-start-rethinkdb',
-                        dest='start_rethinkdb',
-                        action='store_true',
-                        help='Run RethinkDB on start')
-
-    parser.add_argument('--dev-allow-temp-keypair',
-                        dest='allow_temp_keypair',
-                        action='store_true',
-                        help='Generate a random keypair on start')
-
     # all the commands are contained in the subparsers object,
     # the command selected by the user will be stored in `args.command`
     # that is used by the `main` function to select which other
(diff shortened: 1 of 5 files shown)


bigchaindb/commands/bigchain.py
```diff
@@ -243,63 +238,55 @@ def run_load(args):
     workers.start()
 
 
+@configure_bigchaindb
 def run_set_shards(args):
```
sohkai: As far as I can tell, `run_set_shards()` and `run_set_replicas()` should have been autoconfiguring before (since they require a db connection, which may be affected by the config).

Unless this was intended to be this way (to not configure)? Maybe @sbellem or @r-marques you might be able to confirm.

sbellem: +1 for the configuration step

bigchaindb/commands/utils.py
```diff
@@ -3,18 +3,39 @@
...
+import bigchaindb.config_utils
 from bigchaindb import backend
 from bigchaindb.common.exceptions import StartupError
 from bigchaindb.version import __version__
 
 
+def configure_bigchaindb(command):
+    @functools.wraps(command)
+    def configure(args):
+        bigchaindb.config_utils.autoconfigure(filename=args.config, force=True)
```
sbellem: Oh! @sohkai What if `args` has no `config`?! :smirk:

sbellem: :duck:

bigchaindb/commands/utils.py
```diff
@@ -3,18 +3,39 @@
...
 from pymongo import uri_parser
 
 import bigchaindb
+import bigchaindb.config_utils
 from bigchaindb import backend
 from bigchaindb.common.exceptions import StartupError
 from bigchaindb.version import __version__
 
 
+def configure_bigchaindb(command):
```
vrde: Nice. Would be even cooler if instead of having a new function `configure_bigchaindb`, we make `autoconfigure` working both as a function and as a decorator (less cognitive overload).

sbellem: Ah sure. Good idea. Could we offload this to a different ticket though?

sbellem: Also, it may be important to point out that the idea of the decorator was also to include logging configuration and setup. So the goal of the decorator goes beyond the functionalities of `autoconfigure`. @sohkai already did the work: https://github.com/bigchaindb/bigchaindb/pull/1219/commits/a6584daf78e6011c1410e78d463b578acf62c081#diff-8f049242225a1f66c973d24911b5179aR31

vrde: oh I didn't notice that. Good point! Let's keep things as they are now.

sbellem: Just as a side note: perhaps we can find a better name, since `configure_bigchaindb` does sound like it could do just what `autoconfigure` does.

PR #1223, opened 2017-02-27
Docs on using special SSH keypair with ACS/Kubernetes

ttmc:
We already had a section in the Appendices about how to generate an SSH key pair for use with AWS virtual machines. I refactored that section out into a new stand-alone page in the Appendices, because...

In the page about how to deploy a Kubernetes cluster on Azure, I added a link to the new page about generating an SSH key pair, and I showed how to deploy a cluster so that all the VMs in the cluster have the just-generated SSH public key.

I also removed the `--location` option from the `az acs create` command because it's not necessary: the cluster location is inherited from the resource group, which is already specified.

I also added a short section on how to ssh to the "master" node.

### DIFF
--- a/docs/server/source/appendices/generate-key-pair-for-ssh.md
+++ b/docs/server/source/appendices/generate-key-pair-for-ssh.md
@@ -1,0 +1,34 @@
+# Generate a Key Pair for SSH
+
+This page describes how to use `ssh-keygen`
+to generate a public/private RSA key pair
+that can be used with SSH.
+(Note: `ssh-keygen` is found on most Linux and Unix-like
+operating systems; if you're using Windows,
+then you'll have to use another tool,
+such as PuTTYgen.)
+
+By convention, SSH key pairs get stored in the `~/.ssh/` directory.
+Check what keys you already have there:
+```text
+ls -1 ~/.ssh/
+```
+
+Next, make up a new key pair name (called `<name>` below).
+Here are some ideas:
+
+* `aws-bdb-2`
+* `tim-bdb-azure`
+* `chris-bcdb-key`
+
+Next, generate a public/private RSA key pair with that name:
+```text
+ssh-keygen -t rsa -C "<name>" -f ~/.ssh/<name>
+```
+
+It will ask you for a passphrase.
+You can use whatever passphrase you like, but don't lose it.
+Two keys (files) will be created in `~/.ssh/`:
+
+1. `~/.ssh/<name>.pub` is the public key
+2. `~/.ssh/<name>` is the private key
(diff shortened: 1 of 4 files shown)


docs/server/source/appendices/generate-key-pair-for-ssh.md
```diff
@@ -0,0 +1,34 @@
...
+Next, make up a new key pair name (called `<name>` below).
+Here are some ideas:
+
+* `aws-bdb-2`
+* `tim-bdb-azure`
+* `chris-bcdb-key`
+
+Next, generate a public/private RSA key pair with that name:
+```text
+ssh-keygen -t rsa -C "<name>" -f ~/.ssh/<name>
```
krish7919: Is `-b 4096` required? Or are we good with 2048 bits?

docs/server/source/appendices/generate-key-pair-for-ssh.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+It will ask you for a passphrase.
+You can use whatever passphrase you like, but don't lose it.
```
krish7919: Will a passphrase protected key work with kubectl?

docs/server/source/cloud-deployment-templates/template-kubernetes-azure.rst
```diff
@@ -100,4 +112,32 @@ and click on the one you created
...
+
+where you can get the IP address or hostname
+of a master node from the Azure Portal.
+Note how the default username is ``azureuser``.
+
+The "agent" nodes don't get public IP addresses or hostnames,
+so you can't SSH to them *directly*,
+but you can first SSH to the master
+and then SSH to an agent from there 
+(using the *private* IP address of the agent node).
```
krish7919: private IP or hostname of the agents?
As it will be a pain to go to the portal and drill down to each node and get their IP, whereas the hostnames are listed directly at the resource groups page.

docs/server/source/appendices/generate-key-pair-for-ssh.md
ttmc: 2048 bits is probably fine. From `man ssh-keyghen`: "Generally, 2048 bits is considered sufficient."

docs/server/source/appendices/generate-key-pair-for-ssh.md
ttmc: kubectl uses HTTPS.

docs/server/source/cloud-deployment-templates/template-kubernetes-azure.rst
ttmc: I will add that.

docs/server/source/appendices/generate-key-pair-for-ssh.md
krish7919: Yes, the doc says that coz the NSA has already broken it. ;)

PR #1225, opened 2017-02-28
Sign whole transaction body instead of partial transaction

### LINKED ISSUE #970: Proposal: Signing of partial tx vs signing of txid

libscott:
This PR changes the message that is signed to sign a transaction input:

```python
tx_dict = self.to_dict()                                         
tx_dict = Transaction._remove_signatures(tx_dict)                
tx_serialized = Transaction._to_str(tx_dict)                     
for i, input_ in enumerate(self.inputs):                         
    self.inputs[i] = self._sign_input(input_, tx_serialized, key_pairs)
```

So the message is just the serialized transaction ie `"{tx_serialized}"`.

* Includes the transaction ID so the signer knows which tx to look for
* Includes the transaction body so the signer knows what they are signing
* **Upon signing a transaction with a given key pair, the signer is confirming their approval for that transaction to spend any outputs secured with that key pair.**

Fixes #970


TimDaub:
As an FYI for everyone: We're currently discussing _when_ to ideally merge this PR (as it breaks all clients). More on that soon!

libscott:
An idea, the payload could be a sentence, eg: https://gist.github.com/libscott/b86509e8af16f4a5f94c1a85c57b10a1

Thoughts? @r-marques @ttmc @TimDaub

r-marques:
@dimi Using only the id is dangerous because it would allow a client to fulfill an input without having to know what the transaction is about. Requiring a signature over the entire transaction forces the client to have the entire transaction before creating a fulfillment.

Using the same message for all inputs means that once you provide a fulfillment for one input. The same fulfillment maybe be used for all other inputs. This may not be desirable.

libscott:
> Using the same message for all inputs means that once you provide a fulfillment for one input. The same fulfillment maybe be used for all other inputs. This may not be desirable.

This is what we're trying to figure out. If you sign the whole transaction without specifying an input number, the implication is that you've authorized the transaction. This seems sensible. It should be noted however that each input contains a separate fulfillment, even though in practice many fulfillments inside a transaction could be the same, and thus, if a transaction has two inputs controlled by the same private key, a fulfillment of one of them can be used to fulfill the other too. But, the transaction cannot be mutated, so additional outputs that are not specified in the message cannot be spent.

libscott:
Code has been updated to sign transaction body alone @diminator

PR #1237, opened 2017-03-02
Avoid dupe txs in block

### LINKED ISSUE #1229: Validate repeated tx in block
Validate that a transaction does not appear multiple times in a block, in models.py.

libscott:
Fixes #1229.

- [x] Transaction collector to dedupe transactions in a block.
- [x] DuplicateTransaction exception if dupe tx found in block validation
- [x] Stepping pipeline step calls return output
- [x] Remove old integration test
- [x] ~Catch DuplicateTransaction in Vote pipeline~ Not needed after #1243

### DIFF
--- a/bigchaindb/pipelines/block.py
+++ b/bigchaindb/pipelines/block.py
@@ -31,7 +31,7 @@ class BlockPipeline:
     def __init__(self):
         """Initialize the BlockPipeline creator"""
         self.bigchain = Bigchain()
-        self.txs = []
+        self.txs = tx_collector()
 
     def filter_tx(self, tx):
         """Filter a transaction.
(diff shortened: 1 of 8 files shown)


bigchaindb/pipelines/block.py
```diff
@@ -134,6 +133,27 @@ def delete_tx(self, block):
         return block
 
 
+def tx_collector():
```
vrde: Why a coroutine here? It's smart, but it requires time to understand it.

libscott: Didn't do it to be smart. I wanted to keep the deduplication logic separate from the Block pipeline. Coroutine just occurred to me before object, so i dids it... Also, thinking is good for your brain ;)

PR #1239, opened 2017-03-03
Enhancements to run with docker

krish7919:


ttmc:
I don't fully understand all the changes, especially re/ volumes. Could you add some explanation?

krish7919:
The `--restart=always ` will cause docker to keep track of a container's life-cycle and restart it when it goes down, such as in case of machine reboots and container deaths.

And when a container restarts, we need data to be persisted outside the container.
The `--volume "$HOME/bigchaindb_docker:/data"` ensures this for rethinkdb.
The  `--volume=/tmp/mongodb_docker/db:/data/db` and the `--volume=/tmp/mongodb_docker/configdb:/data/configdb` ensures this for mongodb.

However, mongo is a little bit tricky. We need some extra work as described below.
The mongo daemon inside the container runs as user `mongodb`, and the daemon writes all its configs and files as user `mongodb` and group `mongodb`, with a uid and gid of 999. It creates this user and group [here](https://github.com/docker-library/mongo/blob/4a81205a13fefc418355248f750551e4f7c62361/3.4/Dockerfile#L4).
Now, for the host to persist these files, the host instances needs to have a user and a group with uid 999 and gid 999 respectively. Otherwise, the host discards the files.
Hence the big `Note:` section in the doc.

PR #1240, opened 2017-03-03
Run bigchaindb container + RDB/MDB container in a k8s StatefulSet

krish7919:
This PR include the config file to run BDB+RDB (as in v 0.9.1) in a single pod. It uses statefulsets to expose them as services to the external world.

Further plan of action is to split this pod to contain two images: bdb + rdb/mdb.
That will come after this is merged so that we can keep track of the evolution in case we need to refer to it.


ttmc:
Thanks @krish7919, I'll test this out with the PersistentVolumeClaim I created earlier and add some docs about how to do all that: I'll add those changes & docs as commits on this branch/PR.

k8s/node-mdb-ss.yaml
```diff
@@ -76,12 +81,6 @@ spec:
             port: bdb-port
           initialDelaySeconds: 15
           timeoutSeconds: 10
-#        readinessProbe:
-#          httpGet:
-#            path: /
-#            port: bdb-port
-#          initialDelaySeconds: 15
-#          timeoutSeconds: 10
```
ttmc: Why did you remove this readinessProbe from the BigchainDB container?

krish7919: We already have liveness probe every 15 secs.
The service is exposed outside only when it is alive.

ttmc: That's not my understanding of what a readinessProbe is for.

"The kubelet uses readiness probes to know when a Container is ready to start accepting traffic. A Pod is considered ready when all of its Containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers." from https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ 

Suppose a container is live (i.e. it's running fine), and it's been handling traffic fine ("ready") for a long time. Then something goes wrong, the container is still running fine (i.e. it's livenessProbe is working), but it's no longer handling incoming traffic properly for some reason. It's readinessProbe would fail, and that would tell all Services to stop sending traffic to that pod.

I wonder if we should use an exec probe instead of an httpGet probe for livenessProbes, i.e. a more OS-level check that doesn't rely on some webserver working inside the container: that's what the readinessProbe should be checking. The above link has a nice example using `cat`.

ttmc:
@krish7919 - We're exposing some sensitive ports here, e.g. MongoDB's port 27017 and RethinkDB's ports 29015 & 8080. We need to be careful to safeguard them somehow.

PR #1243, opened 2017-03-03
Make ValidationError a superclass of all validation errors and use it

### LINKED ISSUE #1232: Bigchain.is_valid_transaction does not catch the AssetIdMismatch exception
If an `AssetIdMismatch` exception is raised while the pipeline is validating a transaction, the exception is not handled and the pipeline breaks.

libscott:
Fixes #1232.

Scope:

* Make all validation errors subclasses of ValidationError and catch that when validating transactions and blocks.
* Don't use DoubleSpend if there's a problem with the chain, it's a validation error not a critical error
* Don't use OperationError in validation, it's vague
* Don't use TypeError and ValueError in validation
* Remove is_valid_transaction, catch ValidationError in block / vote pipelines.
* Log invalid transactions in block and vote pipeline

### DIFF
--- a/bigchaindb/backend/exceptions.py
+++ b/bigchaindb/backend/exceptions.py
@@ -15,11 +15,3 @@ class OperationError(BackendError):
 
 class DuplicateKeyError(OperationError):
     """Exception raised when an insert fails because the key is not unique"""
-
-
-class CriticalDoubleSpend(BigchainDBError):
-    """Data integrity error that requires attention"""
-
-
-class CriticalDoubleInclusion(BigchainDBError):
-    """Data integrity error that requires attention"""
--- a/bigchaindb/backend/exceptions.py
+++ b/bigchaindb/backend/exceptions.py
@@ -17,5 +17,9 @@ class DuplicateKeyError(OperationError):
     """Exception raised when an insert fails because the key is not unique"""
 
 
-class BigchainDBCritical(Exception):
-    """Unhandleable error that requires attention"""
+class CriticalDoubleSpend(BigchainDBError):
+    """Data integrity error that requires attention"""
+
+
+class CriticalDoubleInclusion(BigchainDBError):
+    """Data integrity error that requires attention"""
--- a/bigchaindb/common/exceptions.py
+++ b/bigchaindb/common/exceptions.py
@@ -42,6 +42,11 @@ class OperationError(BigchainDBError):
 
 ################################################################################
 # Validation errors
+#
+# All validation errors (which are handleable errors, not faults) should
+# subclass ValidationError. However, where possible they should also have their
+# own distinct type to differentiate them from other validation errors,
+# especially for the purposes of testing.
 
 
 class ValidationError(BigchainDBError):
--- a/bigchaindb/common/exceptions.py
+++ b/bigchaindb/common/exceptions.py
@@ -7,46 +7,24 @@ class ConfigurationError(BigchainDBError):
     """Raised when there is a problem with server configuration"""
 
 
-class OperationError(BigchainDBError):
-    """Raised when an operation cannot go through"""
-
-
-class TransactionDoesNotExist(BigchainDBError):
-    """Raised if the transaction is not in the database"""
-
-
-class TransactionOwnerError(BigchainDBError):
-    """Raised if a user tries to transfer a transaction they don't own"""
-
-
-class DoubleSpend(BigchainDBError):
-    """Raised if a double spend is found"""
-
-
-class ValidationError(BigchainDBError):
-    """Raised if there was an error in validation"""
-
-
-class InvalidHash(ValidationError):
-    """Raised if there was an error checking the hash for a particular
-    operation"""
-
-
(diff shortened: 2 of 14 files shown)


bigchaindb/web/views/transactions.py
```diff
@@ -93,15 +80,7 @@ def post(self):
...
-            except (ValueError,
-                    OperationError,
-                    TransactionDoesNotExist,
-                    TransactionOwnerError,
-                    DoubleSpend,
-                    InvalidHash,
-                    InvalidSignature,
-                    TransactionNotInValidBlock,
-                    AmountError) as e:
+            except ValidationError as e:
```
vrde: thanks

bigchaindb/common/exceptions.py
```diff
@@ -7,85 +7,98 @@ class ConfigurationError(BigchainDBError):
     """Raised when there is a problem with server configuration"""
 
 
-class OperationError(BigchainDBError):
-    """Raised when an operation cannot go through"""
-
-
-class TransactionDoesNotExist(BigchainDBError):
-    """Raised if the transaction is not in the database"""
+class DatabaseAlreadyExists(BigchainDBError):
```
vrde: Why this exception is defined here? Shouldn't it be in `bigchaindb.backend.exceptions`?

bigchaindb/common/exceptions.py
```diff
@@ -7,85 +7,98 @@ class ConfigurationError(BigchainDBError):
...
+    """Raised when trying to create the database but the db is already there"""
 
 
-class TransactionOwnerError(BigchainDBError):
-    """Raised if a user tries to transfer a transaction they don't own"""
+class DatabaseDoesNotExist(BigchainDBError):
```
vrde: Same for this exception.

bigchaindb/common/exceptions.py
```diff
@@ -7,85 +7,98 @@ class ConfigurationError(BigchainDBError):
...
+class AssetIdMismatch(ValidationError):
     """Raised when multiple transaction inputs related to different assets"""
 
 
-class AmountError(BigchainDBError):
+class AmountError(ValidationError):
     """Raised when there is a problem with a transaction's output amounts"""
+
+
+class TransactionDoesNotExist(ValidationError):
```
vrde: Why this is a `ValidationError`?

bigchaindb/pipelines/vote.py
```diff
@@ -105,7 +105,13 @@ def validate_tx(self, tx, block_id, num_tx):
         if not new:
             return False, block_id, num_tx
 
-        valid = bool(self.bigchain.is_valid_transaction(tx))
+        try:
+            tx.validate(self.bigchain)
+            valid = True
+        except exceptions.ValidationError as e:
+            logger.warning('Invalid tx: %s' % e)
```
vrde:
```
logger.warning('Invalid tx: %s', e)
```

bigchaindb/backend/exceptions.py
```diff
@@ -17,5 +17,9 @@ class DuplicateKeyError(OperationError):
     """Exception raised when an insert fails because the key is not unique"""
 
 
-class BigchainDBCritical(Exception):
-    """Unhandleable error that requires attention"""
+class CriticalDoubleSpend(BigchainDBError):
```
vrde: AFAIK `CriticalDoubleSpend` is used in `core.py`, so I wonder why this exception is defined in `bigchaindb.backend.exceptions` and not in `bigchaindb.exceptions`.

bigchaindb/backend/exceptions.py
```diff
@@ -17,5 +17,9 @@ class DuplicateKeyError(OperationError):
...
+    """Data integrity error that requires attention"""
+
+
+class CriticalDoubleInclusion(BigchainDBError):
```
vrde: Same as [previous comment](https://github.com/bigchaindb/bigchaindb/pull/1243/files#r104628021).

bigchaindb/common/exceptions.py
libscott: Yes but thats not the scope of this PR. It's just highlighted because the order was switched around.

bigchaindb/common/exceptions.py
libscott: Renamed to InputDoesNotExist. Comment was wrong too.

bigchaindb/backend/exceptions.py
libscott: Cool, I moved it.

bigchaindb/backend/exceptions.py
libscott: Moved it

bigchaindb/web/views/transactions.py
libscott: yw

bigchaindb/common/exceptions.py
r-marques: I am so confused by this diff

PR #1257, opened 2017-03-07
Schemas for CREATE and TRANSFER transactions

libscott:
It was found that a CREATE transaction with multiple inputs could be created (some tests were doing this), and that a CREATE transaction specifying an asset ID could be created. Separate transaction schemas to check constraints are introduced here to have a place to put these things.

Scope

* Create separate schemas for CREATE and TRANSFER transactions
* Validate that CREATE transaction has 1 input with no `fulfills`, via schema validation
* Check that transaction asset ID is not specified in CREATE
* Check asset structure of TRANSFER transaction
* Check implementation of determinable txid via json/sha3


bigchaindb/common/schema/__init__.py
```diff
@@ -29,6 +29,8 @@ def _load_schema(name):
 
 
 TX_SCHEMA_PATH, TX_SCHEMA = _load_schema('transaction')
```
r-marques: Should we rename this schema from `transaction` to something like `transaction_common` so that it is easier to understand what is going on?

bigchaindb/common/schema/__init__.py
```diff
@@ -43,6 +45,10 @@ def _validate_schema(schema, body):
 def validate_transaction_schema(tx):
     """ Validate a transaction dict """
```
r-marques: This would be a good place to explain why the 3 different schemas

PR #1258, opened 2017-03-07
No duplicate vote inserts with mongodb

r-marques:
This pr adds a unique compound index on the votes collection with `node_pubkey` and `vote.voting_for_block`. This prevents duplicated votes for the same public key for the same block.

Created and fixed tests.
Removed some unnecessary tests.


libscott:
This turns out to be kind of critical. A duplicate vote could take a block from being decided to not being decided anymore. This can't be mitigated by the server, the server can only detect when the database has incorrectly allowed a duplicate vote and throw a critical error.

libscott:
LGTM

PR #1286, opened 2017-03-15
transaction.outputs.amount as string

### LINKED ISSUE #1012: Rounding of large (54bit) amounts in Output
Currently we use an integer for our amount, but the safety of this is debatable, since JSON numbers are 64 bit floats but integers have only a [53 effective bits](http://www.2ality.com/2012/07/large-integers.html). The effect of this can be seen in `nodejs` and in the browser:

```shell
> JSON.stringify([9007199254740992])
'[9007199254740992]'
> JSON.stringify([9007199254740992 + 1])
'[9007199254740992]'
> 2400000000000000
```

The rounding errors do not always happen, and probably won't happen on our servers. However, we don't have a way to allow clients to guard against these errors.

We could allow string amounts to be submitted, as well as numbers.

Options:

* A: Don't fix
* B: Allow string, but also number of any size
* C: Allow string, disallow numbers beyond a certain size 
* D: Allow only string. `"amount": "1"`, etc.

libscott:
Fixes #1012

This is done so that transactions with large amounts can be safely constructed in Javascript.

* transaction.outputs.amount is now a string
* Capped at 9000000000000000000 (9 with 18 zeros) as per @ttmc's suggestion

To be merged with #1225


bigchaindb/common/transaction.py
```diff
@@ -229,6 +229,8 @@ def __init__(self, fulfillment, public_keys=None, amount=1):
             raise TypeError('`amount` must be an int')
         if amount < 1:
             raise AmountError('`amount` must be greater than 0')
+        if amount > 9 * 10 ** 18:
```
r-marques: Can we make this a const somewhere so that we don't need to recalculate this value everytime?

libscott: Sure

tests/validation/test_transaction_structure.py
```diff
@@ -0,0 +1,159 @@
...
+# Asset
+
+def test_transfer_asset_schema(signed_transfer_tx):
+    tx = signed_transfer_tx.to_dict()
+    validate(tx)
+    tx['asset']['data'] = {}
+    validate_raises(tx)
+    del tx['asset']['data']
+    tx['asset']['id'] = 'b' * 63
+    validate_raises(tx)
```
r-marques: Can the asset have no `data` field?
Is this raising because asset has no `data` field or because it has a malformed hash?

bigchaindb/common/transaction.py
```diff
@@ -229,8 +231,8 @@ def __init__(self, fulfillment, public_keys=None, amount=1):
             raise TypeError('`amount` must be an int')
         if amount < 1:
             raise AmountError('`amount` must be greater than 0')
-        if amount > 9 * 10 ** 18:
-            raise AmountError('`amount` must be <= 9000000000000000000')
+        if amount > self.MAX_AMOUNT:
+            raise AmountError('`amount` must be <= %s' % self.MAX_AMOUNT)
```
r-marques: Not that it matters much but shouldn't it be `%d`

libscott: It could be, but I've never done it that way. Always worked with %s.

tests/validation/test_transaction_structure.py
libscott: Transfers don't have `asset.data`.

PR #1294, opened 2017-03-16
_sign_threshold now signs all subconditions for a public key

### LINKED ISSUE #1291: Threshold conditions with duplicated public keys not signed correctly
When a transaction is constructed with a threshold conditions containing the same public key multiple times only the first subcondition is fulfilled.

r-marques:
resolves #1291 

This pr makes sure that all subconditions for a given public key are fulfilled.

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -768,20 +768,19 @@ class Transaction(object):
                 key_pairs (dict): The keys to sign the Transaction with.
         """
         input_ = deepcopy(input_)
-        for owner_before in input_.owners_before:
-            try:
-                # TODO: CC should throw a KeypairMismatchException, instead of
-                #       our manual mapping here
+        for owner_before in set(input_.owners_before):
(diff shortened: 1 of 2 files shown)


bigchaindb/common/transaction.py
```diff
@@ -768,20 +768,19 @@ def _sign_threshold_signature_fulfillment(self, input_, index,
...
+
+            # TODO FOR CC: Naming wise this is not so smart,
+            #              `get_subcondition` in fact doesn't return a
+            #              condition but a fulfillment
+
+            # TODO FOR CC: `get_subcondition` is singular. One would not
+            #              expect to get a list back.
+            ccffill = input_.fulfillment
+            subffills = ccffill.get_subcondition_from_vk(owner_before)
+            if not subffills:
```
TimDaub: This might be worth creating a ticket on Cryptoconditions that `get_subcondition_from_vk` should throw :thinking:
Thoughts?

r-marques: Sure. All 3 todos in this method should be cc issues

Issue #1281, opened 2017-03-14
Add MongoDB connection options for SSL and authentication


tomconte:
I found that requirement while testing with Azure DocumentDB, however I think it can be useful in case anyone needs to use a MongoDB server that requires SSL and/or authentication.

krish7919:
That is the plan. We just got started with a replica set on azure.
We are aiming for whitelisting IP addresses and then TLS connections.

r-marques:
@krish7919 can we close this?

krish7919:
@r-marques : I am unaware how this was tested with a secure MongoDB instance. I would keep it open till we can test and confirm that all works as expected, which will be till the time we add TLS support to MongoDB instance itself, and configure BigchainDB to use secure communication channel.

r-marques:
@krish7919 what about now?

krish7919:
@r-marques Fixed in PRs #1299, #1510, #1552

PR #1299, opened 2017-03-17
Add SSL & auth support to MongoDB client

tomconte:
This a first stab at providing additional MongoDB connection options (issue #1281).

Turning SSL on is just an option to add in the `MongoClient()` constructor.

Authentication needs to be performed at the database level, so that is a separate call to `authenticate()`.

Note: currently the MongoDB backend requires admin access (not sure this is a good idea BTW, this might not be possible in shared/hosted environments), so the user must have the right privileges/role.

Here is how to setup a test MongoDB instance:

Run MongoDB with auth:

```
docker run --name dev-mongo -d mongo --auth --replSet bigchain-rs
```

Create root user:

```
docker exec -it dev-mongo mongo admin
rs.initiate()
db.createUser({ user: 'tconte', pwd: 'Pass1234', roles: [ { role: "root", db: "admin" } ] });
```

(exit mongo shell)

```
docker exec -it dev-mongo mongo -u tconte -p Pass1234 admin
use bigchain
db.createUser({ user: 'bcdb', pwd: 'Pass1234', roles: [ { role: "root", db: "admin" } ] });
```

Now the `bcdb` user in the `bigchain` database has `root` role.

A couple (optional) parameters can be added to `.bigchaindb` to provide the new values, e.g.

```
        "ssl": false,
        "login": "bcdb",
        "password": "Pass1234",
```

### DIFF
--- a/bigchaindb/backend/mongodb/connection.py
+++ b/bigchaindb/backend/mongodb/connection.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 
 class MongoDBConnection(Connection):
 
-    def __init__(self, replicaset=None, ssl=False, **kwargs):
+    def __init__(self, replicaset=None, ssl=False, login=None, password=None, **kwargs):
         """Create a new Connection instance.
 
         Args:
@@ -30,6 +30,8 @@ class MongoDBConnection(Connection):
         self.replicaset = replicaset or bigchaindb.config['database']['replicaset']
         self.ssl = bigchaindb.config['database'].get('ssl') if bigchaindb.config['database'].get('ssl') is not None \
             else ssl
+        self.login = login or bigchaindb.config['database'].get('login')
+        self.password = password or bigchaindb.config['database'].get('password')
 
     @property
     def db(self):
@@ -73,15 +75,20 @@ class MongoDBConnection(Connection):
             # we should only return a connection if the replica set is
             # initialized. initialize_replica_set will check if the
             # replica set is initialized else it will initialize it.
-            initialize_replica_set(self.host, self.port, self.connection_timeout, self.ssl)
+            initialize_replica_set(self.host, self.port, self.connection_timeout, self.dbname, self.ssl, self.login, self.password)
 
             # FYI: this might raise a `ServerSelectionTimeoutError`,
             # that is a subclass of `ConnectionFailure`.
-            return pymongo.MongoClient(self.host,
-                                       self.port,
-                                       replicaset=self.replicaset,
-                                       serverselectiontimeoutms=self.connection_timeout,
-                                       ssl=self.ssl)
+            client =  pymongo.MongoClient(self.host,
+                                          self.port,
+                                          replicaset=self.replicaset,
+                                          serverselectiontimeoutms=self.connection_timeout,
+                                          ssl=self.ssl)
+            
+            if self.login is not None and self.password is not None:
+                client[self.dbname].authenticate(self.login, self.password)
+
+            return client
 
         # `initialize_replica_set` might raise `ConnectionFailure` or `OperationFailure`.
         except (pymongo.errors.ConnectionFailure,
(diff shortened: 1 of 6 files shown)


krish7919:
@tomconte  To clarify, this is between the pymongo driver in BigchainDB and the MongoDb instance?

tomconte:
@krish7919 yes, so that BigchainDB can connect to a secured MongoDB instance.

ttmc:
I'm not super-familiar with the code for establishing the connection to MongoDB, so I think I'll ask @vrde to take a look at this pull request too.

We'll also have to update the docs page about the BigchainDB Server settings, i.e. https://docs.bigchaindb.com/projects/server/en/latest/server-reference/configuration.html 
I could do that if you like.

Codecov is also complaining about a decline in unit test coverage...

tomconte:
Looks like I need to work on these tests :)

bigchaindb/backend/mongodb/connection.py
```diff
@@ -71,22 +75,29 @@ def _connect(self):
...
+
+            if self.login is not None and self.password is not None:
+                client[self.dbname].authenticate(self.login, self.password)
+
+            return client
 
         # `initialize_replica_set` might raise `ConnectionFailure` or `OperationFailure`.
         except (pymongo.errors.ConnectionFailure,
                 pymongo.errors.OperationFailure) as exc:
             raise ConnectionError() from exc
```
vrde: Unfortunately `pymongo` doesn't provide a specific exception when the authentication fails, but a generic `OperationFailure`. This might trigger a retry, but it's not the correct behavior: if credentials are wrong we want BigchainDB to fail and give an explicit error message to the user.

You **don't have to care about** it in this PR, some of us can do it after we merge your PR :smile:

See #1336

vrde:
@tomconte thanks for updating your PR. If you [allow us to change your PR](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) I can fix the test coverage and get this merged soon.

tomconte:
@vrde many thanks! The "Allow edits from maintainers" option is enabled on this PR.

vrde:
> A couple (optional) parameters can be added to .bigchaindb to provide the new values, e.g.

I addressed it in 699e615, but I noticed I was adding too much scope, so I reverted it in 0471080. I've created #1346 to keep track of this issue.

chrisckchang:
Hi folks, not sure if this is the right place for this comment but the best practice for connecting to replica sets is to include the host&port information for each primary-electable replica set member (no [arbiter](https://docs.mongodb.com/manual/core/replica-set-arbiter/) or [hidden secondary](https://docs.mongodb.com/v3.4/core/replica-set-hidden-member/) nodes) in the connection string.

You may want to consider supporting the MongoDB connection URI, which can be passed directly into the driver: https://docs.mongodb.com/manual/reference/connection-string/. For example, Meteor.js allows users to pass an entire URI via CLI: https://github.com/meteor/meteor/blob/master/tools/cli/commands.js#L374. You can then use the pymongo uri_parser lib if you need specific parts of the URI: http://api.mongodb.com/python/current/api/pymongo/uri_parser.html.

ttmc:
@vrde Maybe we could create a new issue to look into using the MongoDB connection URI as suggested by @chrisckchang (so that idea doesn't get lost in the comments of a merged PR)? ( https://github.com/bigchaindb/bigchaindb/pull/1299#issuecomment-297077895 )

vrde:
@chrisckchang wrote:
> Hi folks, not sure if this is the right place for this comment but the best practice for connecting to replica sets is to include the host&port information for each primary-electable replica set member (no arbiter or hidden secondary nodes) in the connection string.

AFAIK having multiple servers in the connection string allows the client to have multiple choices if one server is not reachable. Since BigchainDB is a federated system, when the driver fails to connect to the `mongod` it is supposed to talk with, then the driver should not try to connect to other instances, and BigchainDB should not fail to start.

PR #1304, opened 2017-03-20
NGINX frontend for MongoDB and BigchainDB

krish7919:
* Added NGINX deployment to frontend both BDB and MDB.
* Whitelisting source IPs for MongoDB based on ConfigMap entries. Nginx is configured with a whitelist which is read from a ConfigMap to allow only other MDB nodes in the cluster to communicate with it.
* Documentation for deploying NGINX
* Addressed a typo in PYTHON_STYLE_GUIDE.md


docs/server/source/cloud-deployment-templates/add-node-on-kubernetes.rst
```diff
@@ -161,3 +161,14 @@ zero downtime during updates.
 
 You can SSH to an existing BigchainDB instance and run the ``bigchaindb
 show-config`` command to check that the keyring is updated.
+
+
+Step 7: Run NGINX as a Deployment - Optional
+--------------------------------------------
+
+This step is required only if you are want to control access to the BigchainDB
```
ttmc: It says Step 7 is optional, but the updated YAML files for the MongoDB Service and the BigchainDB Service make them so they are no longer `type: LoadBalancer`, i.e. there is an implicit assumption of NGINX (or something else) between those Service endpoints and the public Internet now, so this step is not optional.

Option 1: Make this step non-optional.

Option 2: Add documentation to change
```text
type: ClusterIP
clusterIP: None
```
to `type: LoadBalancer` in the Service definition YAML files for BigchainDB and MongoDB.

I prefer Option 1.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -195,9 +195,10 @@ which can also be obtained using the ``az account list-locations`` command.
 You can also try to assign a name to an Public IP in Azure before starting
 the process, or use ``nslookup`` with the name you have in mind to check
 if it's available for use.
-In the rare chance that name in the ``data.fqdn`` field is not available,
-you must create a ConfigMap with a unique name and restart the
-MongoDB instance.
+
+In the rare chance that the name specified in the ``data.fqdn`` field is not
+available, we will need to create a ConfigMap with a unique name and
```
ttmc: Please only use "we" to mean "we, BigchainDB GmbH, the company." In this line, the "we" should be "you"

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -195,9 +195,10 @@ which can also be obtained using the ``az account list-locations`` command.
 You can also try to assign a name to an Public IP in Azure before starting
 the process, or use ``nslookup`` with the name you have in mind to check
 if it's available for use.
-In the rare chance that name in the ``data.fqdn`` field is not available,
-you must create a ConfigMap with a unique name and restart the
-MongoDB instance.
+
+In the rare chance that the name specified in the ``data.fqdn`` field is not
+available, you will need to create a ConfigMap with a unique name and
+restart the MongoDB instance.
```
ttmc: I don't think they have a MongoDB instance running at this point, so there's nothing to restart.

krish7919: Should I phrase it as "In the rare chance that the name specified in the `data.fqdn` field is not available when MongoDB starts, you will need...."
The reason is because there is a time lag (although small) between reserving the name, populating the config map and then actually assigning the name to Mongo instance.

ttmc: You should just tell them to pick a value for `data.fqdn` that they're _certain_ nobody else is using, or will ever use.

ttmc:
There's a line in `node-on-kubernetes.rst` that will have to be updated, since the MongoDB Service was renamed from `mdb` to `mdb-svc`:

screenshot from 2017-03-20 17-05-42

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
krish7919: Handled in the latest commit.

ttmc:
Step 10 will have to be revised. Step 10.1 won't work anymore because neither BigchainDB nor MongoDB are exposed publicly yet (and MongoDB never will be, from now on).

<hr>

screenshot from 2017-03-20 17-25-42

<hr>

Maybe move the external testing of BigchainDB to a new Step 12? (And delete the external testing of MongoDB.)

krish7919:
You are right, the external testing is covered post the nginx installation.
Now that nginx install isn't optional, we do not need to test the node externally.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -460,14 +411,65 @@ Create the NGINX deployment using:
...
+themselves.
+
+Step 11.2: Testing Externally
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Try to access the ``<dns/ip of your exposed bigchaindb service endpoint>:80``
+on your browser. You must receive a json output that shows the BigchainDB
+server version among other things.
+
+Try to access the ``<dns/ip of your exposed mongodb service endpoint>:27017``
```
ttmc: Once they have NGINX set up with a whitelist, they won't be able to get any response from port 27017, will they? Maybe an HTTP 4xx or 5xx error code?

krish7919: Good catch.
I will elaborate to test with both allowed and disallowed IP addresses to test this setup.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -367,48 +367,111 @@ Create the required Deployment using:
...
+.. code:: bash
+   
+   $ wget https://raw.githubusercontent.com/bigchaindb/bigchaindb/master/k8s/nginx/nginx-cm.yaml
 
-Step 10.2: Testing Internally
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The IP address whitelist can be explicitly configured in ``nginx-cm.yaml``
+file. You will need a list of the IP addresses of all the other MongoDB 
+instances in the cluster. If the MongoDB intances specify a hostname, then this
+needs to be resolved to the corresponding IP addresses. If the IP address of
```
ttmc: You could provide a bit more detail here. For example, maybe say that you have to edit the `nginx-cm.yaml` file and change the list of `allowed-hosts` to be a colon-separated list of IP addresses (of the other MongoDB instances in the BigchainDB cluster). A concrete example of a realistic-looking `  allowed-hosts:` line would go farther than 1000 words about general principles.

ttmc: The current line:
```text
  allowed-hosts: 192.168.0.0/16:10.0.2.0/24
```
doesn't seem very realistic: those are private networks.

ttmc: Actually, it's a list of CIDR blocks, not just IP addresses.

ttmc:
I wonder if it would make sense to move the ConfigMaps to their own subdirectory of the `k8s` directory (or just in the `k8s` directory itself), because the idea of ConfigMaps is that they're shared by and accessible by everything in the k8s cluster. For example, it seems weird that the `mongodb-whitelist` ConfigMap is in the `k8s/nginx/nginx-cm.yaml` file.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
ttmc: Looking at [the NGINX docs](https://nginx.org/en/docs/http/ngx_http_access_module.html), it seems that single IP addresses are also allowed. For example:
```text
    deny  192.168.1.1;
    allow 192.168.1.0/24;
    allow 10.1.1.0/16;
    allow 2001:0db8::/32;
    deny  all;
```

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -367,48 +367,111 @@ Create the required Deployment using:
...
+   # curl -X GET http://ngx-svc:27017
+
+The ``nslookup`` commands should output the configured IP addresses of the
+services in the cluster
+
+The ``dig`` commands should return the port numbers configured for the
+various services in the cluster.
+
+Finally, the ``curl`` commands test the availability of the services
+themselves.
```
ttmc: When I did the last curl command, i.e. `# curl -X GET http://ngx-svc:27017`, I got the response:
```text
curl: (56) Recv failure: Connection reset by peer
```

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
```diff
@@ -367,48 +367,111 @@ Create the required Deployment using:
...
+
+Step 11.2: Testing Externally
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Try to access the ``<dns/ip of your exposed bigchaindb service endpoint>:80``
```
ttmc: This should be changed to the `dns / external-ip of your ngx-svc Service` (which you can get using `kubectl get services`).

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
krish7919: @ttmc: Also, there will be no error codes sent as we are blocking TCP traffic. NGINX will just close the socket connection and the client will get a 'connection reset' error message.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
krish7919: 1. I put private IP blocks here so that we can commit this file to Github.
2. Anyone working with this (engineers) would be aware of how IP addresses are specified.
3. The explicit specification of a /32 is much better than not specifying it, as it will be easier to debug issues and immediately find out whether it's a block or a single IP that the use is allowing access to. Either way, it's a minor documentation change.

krish7919:
> I wonder if it would make sense to move the ConfigMaps to their own subdirectory of the k8s directory (or just into the k8s directory itself), because the idea of ConfigMaps is that they're shared by and accessible to everything in the k8s cluster. For example, it seems weird that the mongodb-whitelist ConfigMap is in the k8s/nginx/nginx-cm.yaml file.

This should be done, but the goal here is to make separate things work independently before we have a consolidated config.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
krish7919: You might not have allowed access to connect from within the k8s cluster.
If you think this is not the issue, we can debug together.

docs/server/source/cloud-deployment-templates/node-on-kubernetes.rst
krish7919: We have multiple ways to get this IP:
1. The cloud provider portal. Eg: Azure.
2. The k8s UI/dashboard, under Services.
3. The `kubectl get services` CLI.

If someone were to test connectivity between MongoDB cluster instances, one would need an IP/DNS name to the remote instance.

If someone were to test connectivity to a BigchainDB instance, one would need an IP/DNS name to the remote instance, where BigchainDB is exposed. Hence, an `dns/ip of your exposed bigchaindb ...`.

ttmc:
I don't know why codecov is complaining. I wouldn't worry about it though.

ttmc:
@krish7919 I added the `k8s` directory to the ignore list in `codecov.yml`. Maybe that will appease it.

PR #1309, opened 2017-03-20
Voting module raises CriticalDuplicateVote if there's a duplicate vote

### LINKED ISSUE #1306: Duplicate vote can make decided block undecidable
Following on from #1258, it's necessary to treat duplicate votes (per node and block) as a critical error, as there's no way to decide which is the correct vote to count.

libscott:
Fixes #1306
(Kind of) goes along with #1258

### DIFF
--- a/tests/test_voting.py
+++ b/tests/test_voting.py
@@ -1,4 +1,5 @@
 import pytest
+from unittest.mock import patch
 from collections import Counter
 
 from bigchaindb.core import Bigchain
@@ -235,11 +236,8 @@ def test_block_election(b):
     }
 
 
+@patch('bigchaindb.voting.Voting.verify_vote_signature', return_value=True)
 def test_duplicate_vote_throws_critical_error(b):
-    class TestVoting(Voting):
-        @classmethod
-        def verify_vote_signature(cls, vote):
-            return True
     keyring = 'abc'
     block = {'id': 'xyz', 'block': {'voters': 'ab'}}
     votes = [{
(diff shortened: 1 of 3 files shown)


tests/test_voting.py
```diff
@@ -230,8 +229,22 @@ def verify_vote_schema(cls, vote):
...
         'malformed': [],
         'previous_block': 'a',
         'other_previous_block': {},
     }
+
+
+def test_duplicate_vote_throws_critical_error(b):
+    class TestVoting(Voting):
+        @classmethod
+        def verify_vote_signature(cls, vote):
```
r-marques: its common practice to just mock this kind of behaviour

r-marques:
This looks good to me but I think this is only helpful for rethinkdb since for mongodb the unique secondary index should prevent this.

libscott:
> This looks good to me but I think this is only helpful for rethinkdb since for mongodb the unique secondary index should prevent this.

I agree, but the goal of this PR is to simplify the voting logic since it no longer tries to handle the case where there are duplicate votes from a node. Previously, it would collect duplicates together and consider them to be "cheats", but this could cause illegal block result transitions if nodes submitted multiple votes. Now, it doesn't try to handle this case, but since the data structure returned by the database is a list of votes and not a dictionary of votes indexed by node pubkey, it performs this check for safety. It's not just for the benefit of Rethinkdb but any other backend, and also MongoDB if we make a mistake in the future.

tests/test_voting.py
libscott: It's consistent with the testing in the rest of the module.

r-marques: we expect consistency across the entire code base

libscott: Ok I'll change it here then, suppose it doesn't hurt.

PR #1311, opened 2017-03-21
[PROPOSAL][WIP] Add CORS to webserver for localhost

diminator:
When consuming the API from the browser, CORS issues arise due to `same-origin` policy

This makes it more tedious to start building webapps on generic BDB backends.
In this tentative PR, I've added CORS for BDB as `localhost`, but maybe you'll want to investigate/add a generic CORS implementation?

I know one can also put CORS at a higher level like on the reverse-proxy level. But this is additional complexity


TimDaub:
I think adding support for this makes sense.
Especially when building demo apps in the browser, it's very likely that you're not running your web server under the same port as BigchainDB.

Instead of fixing the allowed domain to `localhost or 127.0.0.1`, I'd rather add a configuration option defining the domain allowed to request though!

diminator:
@TimDaub @vrde @sbellem can we merge this? 

or are there any implications re:
- security
- deployment
?

PR #1282, opened 2017-03-14
Move tx structural validation tests

libscott:
Scope is to organise transaction validation tests into distinct modules:

* Structural validation in `tests/validation/test_transaction_structure.py`
* Unify all nonspecific schema tests into tests/common/test_schema.py
* Move transaction structure tests from:
    - tests/common/schema/test_transaction_schema.py
    - tests/common/test_transaction.py
    - tests/test_models.py
    - tests/assets/test_digital_assets.py
    - tests/assets/test_divisible_assets.py
* Move block model tests from tests/test_models.py to tests/test_block_model.py

PR #1319, opened 2017-03-21
Federation tests

libscott:
This PR tests election properties at the level of a federation.

Also includes code from #1282.


tests/common/test_schema.py
```diff
@@ -67,3 +79,37 @@ def test_drop_descriptions():
     }
     drop_schema_descriptions(node)
     assert node == expected
+
+
+################################################################################
+# Test call transaction schema
+
+
+def test_validate_transaction_create(create_tx):
```
r-marques: should we assert something on these tests?

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+def federation(n):
+    """
+    Return a list of Bigchain objects and pipeline steppers to represent
+    a BigchainDB federation
+    """
+    keys = [generate_key_pair() for _ in range(n)]
+    config_orig = bigchaindb.config
+
+    @contextmanager
+    def make_nodes(i):
```
r-marques: This could use some comments. The code is difficult to read.

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+        nonlocal keys
```
r-marques: What does `nonlocal` do here?

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+        if i == 0:
```
r-marques: what is the reason to skip the first one?

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+            yield []
+        else:
+            config = deepcopy(config_orig)
+            keys = [keys[-1]] + keys[:-1]
```
r-marques: Wouldn't it be simpler to just use the index `i`?

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+            config['keypair']['private'] = keys[0][0]
+            config['keypair']['public'] = keys[0][1]
+            config['keyring'] = list(list(zip(*keys[1:]))[1])
```
r-marques: this would be more readable with a single list comprehension `[k.public_key for k in keys[1:]]`

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+def test_elect_dupe_vote(federation_3):
+    from bigchaindb.exceptions import CriticalDuplicateVote
+    [bx, (s0, s1, s2)] = federation_3
+    tx = input_single_create(bx[0])
+    process_tx(s0)
+    process_tx(s1)
+    process_tx(s2)
+    vote = process_vote(s0, True)
+    # Drop the unique index and write the vote again
+    bx[0].connection.db.votes.drop_index('block_and_voter')
```
r-marques: Is this a realistic test? Are we assuming that the database schema will change?

tests/integration/test_federation.py
```diff
@@ -0,0 +1,215 @@
...
+@pytest.mark.genesis
+def test_elect_sybill(federation_3):
+    [bx, (s0, s1, s2)] = federation_3
+    tx = input_single_create(bx[0])
+    process_tx(s0)
+    process_tx(s1)
+    process_tx(s2)
+    # What we need is some votes from unknown nodes!
+    for s in [s0, s1, s2]:
+        s.vote.bigchain.me_private = generate_key_pair()[0]
```
r-marques: This will only make the vote signatures invalid, not make the nodes ineligible which I think its what we want to test.

tests/pipelines/stepping.py
```diff
@@ -106,7 +106,7 @@ def _enqueue(self, name, item):
         for item in items:
             if type(item) != tuple:
                 item = (item,)
-            queue.append(item)
+            queue.append(list(item))
```
r-marques: what is reason to check if `item` is a tuple and then cast it to a list?

tests/common/test_schema.py
libscott: The test passes if no validation error is thrown. Any exception will fail a test.

tests/integration/test_federation.py
libscott: It's like `global` in that it allows you to assign to a variable in a non local scope.

tests/integration/test_federation.py
libscott: It's a recursive function, it is called with n where n is the number of nodes you want to simulate. When `i` is 0, there are no remaining nodes to be created so it returns.

tests/integration/test_federation.py
libscott: This is really to enable getting the keyring via `keys[1:]`.

tests/pipelines/stepping.py
libscott: It goes tuple -> list

Alternately, it would be written as:

```python
if type(item) == tuple:
    item = list(item)
else:
    item = [item]
```

tests/integration/test_federation.py
libscott: Incorrectly signed votes are also ineligible. Eligibility is correctly signed by known node, this is tested in detail in test_voting.py. I added a comment anyway.

r-marques: What is the goal of the test? Isn't it to test that votes from unknown nodes are not considered?

tests/integration/test_federation.py
libscott: Added comment to explain what function is doing

tests/integration/test_federation.py
libscott: Yes, but the election code makes no distinction between a badly signed vote and a vote from an unknown node. They are both considered ineligible in the same way.

PR #1322, opened 2017-03-22
Clarify allowed maximum complexity of conditions

### LINKED ISSUE #356: Put a limit on how complicated a complex (crypto-) condition can be
The more complicated the condition, the more CPU work it takes to validate the fulfillment. A client could mess with BigchainDB by submitting transaction sequences with extremely complex conditions, tying up CPUs in validating the fulfillments. It's a nerdy and probably not very effective kind of denial-of-service attack.

I gather that there's an easy way to put a limit on how complex a condition (or fulfillment?) can be. Let's do that.

r-marques:
resolves #356 

Changed the documentation to clarify how we can limit the the complexity of all cryptoconditions in a transaction.

### DIFF
--- a/docs/server/source/data-models/inputs-outputs.rst
+++ b/docs/server/source/data-models/inputs-outputs.rst
@@ -22,7 +22,12 @@ One can also put different weights on the inputs to a threshold condition, along
 
 The (single) output of a threshold condition can be used as one of the inputs of other threshold conditions. This means that one can combine threshold conditions to build complex logical expressions, e.g. (x OR y) AND (u OR v).
 
-When one creates a condition, one can calculate its fulfillment length (e.g. 96). The more complex the condition, the larger its fulfillment length will be. A BigchainDB federation can put an upper limit on the allowed fulfillment length, as a way of capping the complexity of conditions (and the computing time required to validate them).
+When one creates a condition, one can calculate its fulfillment length (e.g.
+96). The more complex the condition, the larger its fulfillment length will be.
+A BigchainDB federation can put an upper limit on the complexity of the
+conditions, either directly by setting an allowed maximum fulfillment length,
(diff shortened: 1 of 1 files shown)


docs/server/source/data-models/inputs-outputs.rst
```diff
@@ -22,7 +22,12 @@ One can also put different weights on the inputs to a threshold condition, along
 
 The (single) output of a threshold condition can be used as one of the inputs of other threshold conditions. This means that one can combine threshold conditions to build complex logical expressions, e.g. (x OR y) AND (u OR v).
 
-When one creates a condition, one can calculate its fulfillment length (e.g. 96). The more complex the condition, the larger its fulfillment length will be. A BigchainDB federation can put an upper limit on the allowed fulfillment length, as a way of capping the complexity of conditions (and the computing time required to validate them).
+When one creates a condition, one can calculate its fulfillment length (e.g.
+96). The more complex the condition, the larger its fulfillment length will be.
+A BigchainDB federation can put an upper limit on the complexity of the
+conditions, either directly by setting an allowed maximum fulfillment length,
+or indirectly by setting a maximum allowed transaction size which would limit
+the overall complexity accross all inputs and outputs of a transaction.
```
ttmc: I can understand how one might limit the overall transaction size (e.g. using an NGINX `client_max_body_size` setting), but I don't think there's any easy way to set an allowed maximum fulfillment length (e.g. by changing a setting in the BigchainDB config file).

r-marques: I think cryptoconditions allows us to set a maximum fulfillment length.
We could also add it as a validation rule.

But the strategy that we agreed was to limit the overall complexity across all cryptoconditions of a transaction using by limiting the transaction size limit.

ttmc: ok

PR #1332, opened 2017-03-23
New docs page about updating all the software on a BigchainDB node on Kubernetes

ttmc:
This is a first draft of some documentation about how to update all the software associated with a BigchaindB node running on Kubernetes, including:

- the host operating systems
- Docker
- Kubernetes
- BigchainDB-related software

As with all these BDB-node-on-Kubernetes docs, I expect they'll eventually be moved out of this section to some docs about Production BigchainDB Node Administration.

### DIFF
--- a/docs/server/source/cloud-deployment-templates/upgrade-on-kubernetes.rst
+++ b/docs/server/source/cloud-deployment-templates/upgrade-on-kubernetes.rst
@@ -1,0 +1,105 @@
+Kubernetes Template: Upgrade all Software in a BigchainDB Node
+==============================================================
+
+This page outlines how to upgrade all the software associated
+with a BigchainDB node running on Kubernetes,
+including host operating systems, Docker, Kubernetes,
+and BigchainDB-related software.
+
+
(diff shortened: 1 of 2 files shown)


docs/server/source/cloud-deployment-templates/upgrade-on-kubernetes.rst
```diff
@@ -0,0 +1,105 @@
...
+   Once you are in an SSH session with a host,
+   the ``docker info`` command is a handy way to detemine the
+   host OS (including version) and the Docker version.
+
+When you want to upgrade the software on a Kubernetes node,
+you should "drain" the node first,
+i.e. tell Kubernetes to gracefully terminate all pods
+on the node and mark it as unscheduleable
+(so no new pods get put on the node during its downtime).
+
```
krish7919: We will usually need to over-provision by 1 node, for a 1 node update so that the jobs drained are rescheduled somewhere else. We need to test this workflow.

krish7919:
I have approved this for now, although I would like to discuss immutable infrastructure approach. :)

PR #1348, opened 2017-03-28
Add missing config options

### LINKED ISSUE #1346: Add ssl, login, and password fields to configuration
Related to #1299.
Part of this issue has been done in 699e615.

vrde:
This PR completes #1299 adding the missing configuration options to the CLI `bigchaindb configure`, see also [this comment](https://github.com/bigchaindb/bigchaindb/pull/1299#issuecomment-289764898)
Close #1346.

/cc @tomconte

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -5,23 +5,50 @@ import os
 # PORT_NUMBER = reduce(lambda x, y: x * y, map(ord, 'BigchainDB')) % 2**16
 # basically, the port number is 9984
 
-_database_rethinkdb = {
-    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
+
+_base_database_rethinkdb = {
     'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
     'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 28015)),
     'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
-    'connection_timeout': 5000,
-    'max_tries': 3,
 }
 
-_database_mongodb = {
-    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'mongodb'),
+# The following variable is used by `bigchaindb configure` to
+# prompt the user for database values. We cannot rely on
+# _base_database_rethinkdb.keys() or _base_database_mongodb.keys()
+# because dicts are unordered. I tried to configure
+
+_base_database_rethinkdb_keys = ('host', 'port', 'name')
+
+_base_database_mongodb = {
     'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
     'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 27017)),
     'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
     'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
+    'ssl': bool(os.environ.get('BIGCHAINDB_DATABASE_SSL', False)),
+    'login': os.environ.get('BIGCHAINDB_DATABASE_LOGIN'),
+    'password': os.environ.get('BIGCHAINDB_DATABASE_PASSWORD')
+}
+
+_base_database_mongodb_keys = ('host', 'port', 'name', 'replicaset',
+                               'ssl', 'login', 'password')
+
+_database_rethinkdb = {
+    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
     'connection_timeout': 5000,
     'max_tries': 3,
+}
+_database_rethinkdb.update(_base_database_rethinkdb)
+
+_database_mongodb = {
+    'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'mongodb'),
+    'connection_timeout': 5000,
+    'max_tries': 3,
+}
+_database_mongodb.update(_base_database_mongodb)
+
+_database_keys_map = {
+    'mongodb': _base_database_mongodb_keys,
+    'rethinkdb': _base_database_rethinkdb_keys
 }
 
 _database_map = {
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -17,7 +17,10 @@ _base_database_rethinkdb = {
 # _base_database_rethinkdb.keys() or _base_database_mongodb.keys()
 # because dicts are unordered. I tried to configure
 
-_base_database_rethinkdb_keys = ('host', 'port', 'name')
+_database_keys_map = {
+    'mongodb': ('host', 'port', 'name', 'replicaset'),
+    'rethinkdb': ('host', 'port', 'name')
+}
 
 _base_database_mongodb = {
     'host': os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost'),
@@ -28,9 +31,6 @@ _base_database_mongodb = {
     'login': os.environ.get('BIGCHAINDB_DATABASE_LOGIN'),
     'password': os.environ.get('BIGCHAINDB_DATABASE_PASSWORD')
 }
-
-_base_database_mongodb_keys = ('host', 'port', 'name', 'replicaset',
-                               'ssl', 'login', 'password')
 
 _database_rethinkdb = {
     'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb'),
@@ -45,11 +45,6 @@ _database_mongodb = {
     'max_tries': 3,
 }
 _database_mongodb.update(_base_database_mongodb)
-
-_database_keys_map = {
-    'mongodb': _base_database_mongodb_keys,
-    'rethinkdb': _base_database_rethinkdb_keys
-}
 
 _database_map = {
     'mongodb': _database_mongodb,
--- a/bigchaindb/commands/utils.py
+++ b/bigchaindb/commands/utils.py
@@ -74,12 +74,50 @@ def start_logging_process(command):
     return start_logging
 
 
+def _convert(value, default=None, convert=None):
+    def convert_bool(value):
+        if value.lower() in ('true', 't', 'yes', 'y'):
+            return True
+        if value.lower() in ('false', 'f', 'no', 'n'):
+            return False
+        raise ValueError('{} cannot be converted to bool'.format(value))
+
+    if value == '':
+        value = None
+
+    if convert is None:
+        if default is not None:
+            convert = type(default)
+        else:
+            convert = str
+
+    if convert == bool:
+        convert = convert_bool
+
+    if value is None:
+        return default
+    else:
+        return convert(value)
+
+
 # We need this because `input` always prints on stdout, while it should print
 # to stderr. It's a very old bug, check it out here:
 # - https://bugs.python.org/issue1927
-def input_on_stderr(prompt=''):
+def input_on_stderr(prompt='', default=None, convert=None):
+    """Output a string to stderr and wait for input.
+
+    Args:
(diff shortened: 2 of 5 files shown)


bigchaindb/__init__.py
```diff
@@ -5,24 +5,51 @@
...
     'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 27017)),
     'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
     'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
+    'ssl': bool(os.environ.get('BIGCHAINDB_DATABASE_SSL', False)),
+    'login': os.environ.get('BIGCHAINDB_DATABASE_LOGIN'),
+    'password': os.environ.get('BIGCHAINDB_DATABASE_PASSWORD')
+}
+
+_base_database_mongodb_keys = ('host', 'port', 'name', 'replicaset',
+                               'ssl', 'login', 'password')
```
sbellem: Why not simply

```python
tuple(_base_database_mongodb.keys())
```

vrde: > Why not simply

Because of [this comment](https://github.com/bigchaindb/bigchaindb/pull/1348/files#diff-993779d6d8d7ec19551175f387ef2ea6R15).

> otherwise they may fall out of sync

Hopefully [not](https://github.com/bigchaindb/bigchaindb/pull/1348/files#diff-347ee4bd1d043b58203b19a913c7c342R22).

sbellem: Oh sorry, I overlooked the comment. Well, then why not using [`OrderedDict`](https://docs.python.org/3.6/library/collections.html#collections.OrderedDict)?

> An _OrderedDict_ is a dict that remembers the order that keys were first inserted.

vrde: > Oh sorry, I overlooked the comment.

No problem :dancer: 

I was thinking the same, but building one is quite awful because you have to pass a list of "pairs" unfortunately :sob:

```python
# from the docs

>>> # regular unsorted dictionary
>>> d = {'banana': 3, 'apple': 4, 'pear': 1, 'orange': 2}

>>> # dictionary sorted by key
>>> OrderedDict(sorted(d.items(), key=lambda t: t[0]))
OrderedDict([('apple', 4), ('banana', 3), ('orange', 2), ('pear', 1)])

>>> # dictionary sorted by value
>>> OrderedDict(sorted(d.items(), key=lambda t: t[1]))
OrderedDict([('pear', 1), ('orange', 2), ('banana', 3), ('apple', 4)])

>>> # dictionary sorted by length of the key string
>>> OrderedDict(sorted(d.items(), key=lambda t: len(t[0])))
OrderedDict([('pear', 1), ('apple', 4), ('orange', 2), ('banana', 3)])
```

For us it would be:
```python
OrderedDict([
    ('backend', os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'rethinkdb')),
    ('host', os.environ.get('BIGCHAINDB_DATABASE_HOST', 'localhost')),
    ('port', int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 28015))),
    ('name', os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'))
])
```

It might be fine tho, but I really don't like the ﻿ａｅｓｔｈｅｔｉｃ.

sbellem: Ah ok.

It may be useful to observe, if I am not mistaken, that from Python 3.6 onwards we could instantiate an `OrderedDict` with keyword arguments, and rely on their order to be preserved. (See [release notes](https://docs.python.org/3.6/whatsnew/3.6.html#pep-468-preserving-keyword-argument-order).)

As a side note, in Python 3.6, the new `dict` implementation preserves the key order, but it is currently considered to be an implementation detail that should not be relied on. See release note [here](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation).

> The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon (this may change in the future, but it is desired to have this new dict implementation in the language for a few releases before changing the language spec to mandate order-preserving semantics for all current and future Python implementations; this also helps preserve backwards-compatibility with older versions of the language where random iteration order is still in effect, e.g. Python 3.5).

bigchaindb/commands/utils.py
```diff
@@ -74,12 +74,50 @@ def start_logging(args):
     return start_logging
 
 
+def _convert(value, default=None, convert=None):
```
sbellem: It would be nice to use the "keyword-only arguments" approach as outlined in PEP 3102. That is:

```python
def _convert(value, *, default=None, convert=None)
```

This helps making the code clearer.

bigchaindb/commands/utils.py
```diff
@@ -74,12 +74,50 @@ def start_logging(args):
...
+
+    if convert == bool:
+        convert = convert_bool
+
+    if value is None:
+        return default
+    else:
+        return convert(value)
+
+
```
sbellem: Perhaps, just a question of personal preferences, but I would simplify the above in a few ways. As an example, a possible simplification could yield:

```python
def _convert_bool(value):
    if value.lower() in ('true', 't', 'yes', 'y'):
        return True
    if value.lower() in ('false', 'f', 'no', 'n'):
        return False
    raise ValueError('{} cannot be converted to bool'.format(value))


def _convert(value, default=None, convert=None):
    if value == '':
        return default

    if convert is None and default is None:
        return value

    if convert is None and default is not None:
        convert = type(default)

    if convert == bool:
        convert = _convert_bool

    return convert(value)
```

bigchaindb/commands/utils.py
```diff
@@ -74,12 +74,50 @@ def start_logging(args):
...
 # We need this because `input` always prints on stdout, while it should print
 # to stderr. It's a very old bug, check it out here:
 # - https://bugs.python.org/issue1927
-def input_on_stderr(prompt=''):
+def input_on_stderr(prompt='', default=None, convert=None):
```
sbellem: Here too, making the keyword arguments only keyword arguments would be nice.

tests/commands/test_utils.py
```diff
@@ -13,6 +13,33 @@ def reset_bigchaindb_config(monkeypatch):
...
+    assert _convert('3.14', default=1.0) == 3.14
+    assert _convert('TrUe', default=False) is True
+
+    with pytest.raises(ValueError):
+        assert _convert('TRVE', default=False)
+
+    with pytest.raises(ValueError):
+        assert _convert('ಠ_ಠ', convert=int)
+
+
```
sbellem: This test could fail for multiple different reasons, and upon failure it wouldn't be straightforward to know what is actually failing. The test could be split into three separate tests:

```python
def test_input_on_stderr():
    from bigchaindb.commands.utils import input_on_stderr

    with patch('builtins.input', return_value='I love cats'):
        assert input_on_stderr() == 'I love cats'


# input_on_stderr uses `_convert` internally, from now on we will
# just use that function
def test_convert():
    from bigchaindb.commands.utils import _convert
    assert _convert('hack the planet') == 'hack the planet'
    assert _convert('42') == '42'
    assert _convert('42', default=10) == 42
    assert _convert('', default=10) == 10
    assert _convert('42', convert=int) == 42
    assert _convert('True', convert=bool) is True
    assert _convert('False', convert=bool) is False
    assert _convert('t', convert=bool) is True
    assert _convert('3.14', default=1.0) == 3.14
    assert _convert('TrUe', default=False) is True


def test_convert_raises():
    from bigchaindb.commands.utils import _convert
    with pytest.raises(ValueError):
        assert _convert('TRVE', default=False)

    with pytest.raises(ValueError):
        assert _convert('ಠ_ಠ', convert=int)
```

sbellem: We could also use pytest parametrization:

```python
@pytest.mark.parametrize('value,default,convert,expected_value,op', (
    ('hack the planet', None, None, 'hack the planet', operator.eq),
    ('42', None, None, '42', operator.eq),
    ('42', 10, None, 42, operator.eq),
    ('', 10, None, 10, operator.eq),
    ('42', None, int, 42, operator.eq),
    ('True', None, bool, True, operator.is_),
    ('False', None, bool, False, operator.is_),
    ('t', None, bool, True, operator.is_),
    ('3.14', 1.0, None, 3.14, operator.eq),
    ('TrUe', False, None, True, operator.is_),
))
def test_convert(value, default, convert, expected_value, op):
    from bigchaindb.commands.utils import _convert
    assert op(
        _convert(value, default=default, convert=convert), expected_value)


@pytest.mark.parametrize('value,default,convert', (
    ('TRVE', False, None), ('ಠ_ಠ', None, int)
))
def test_convert_raises(value, default, convert):
    from bigchaindb.commands.utils import _convert
    with pytest.raises(ValueError):
        _convert(value, default=default, convert=convert)
```

tests/test_config_utils.py
```diff
@@ -19,6 +19,15 @@ def clean_config(monkeypatch, request):
...
 
+def test_ordered_keys_match_database_config():
+    import bigchaindb
+
+    assert set(bigchaindb._base_database_rethinkdb.keys()) ==\
+        set(bigchaindb._base_database_rethinkdb_keys)
+    assert set(bigchaindb._base_database_mongodb.keys()) ==\
+        set(bigchaindb._base_database_mongodb_keys)
+
+
```
sbellem: Duplicate entries could be shown to the user. In the code below, from the command line module:

```python
        for key in database_keys:
            val = conf['database'][key]
            conf['database'][key] = input_on_stderr('Database {}? (default `{}`): '.format(key, val), val)
```

duplicate elements (keys) will trigger a prompt to the user.

We can check the length:

```python
def test_ordered_keys_match_database_config():
    from bigchaindb import (
        _base_database_rethinkdb, _base_database_rethinkdb_keys,
        _base_database_mongodb, _base_database_mongodb_keys,
    )
    assert len(_base_database_rethinkdb.keys()) ==\
        len(_base_database_rethinkdb_keys)
    assert len(_base_database_mongodb.keys()) ==\
        len(_base_database_mongodb_keys)
    assert set(_base_database_rethinkdb.keys()) ==\
        set(_base_database_rethinkdb_keys)
    assert set(_base_database_mongodb.keys()) ==\
        set(_base_database_mongodb_keys)
```

ttmc:
We should also have docs about these new configuration settings (for MongoDB only). Do you want to add that to this PR, or do it in a separate issue/PR? Note that there's a potential for a merge conflict with PR #1338 (which added docs for `database.connection_timeout` and `database.max_tries`).

vrde:
> We should also have docs about these new configuration settings (for MongoDB only). Do you want to add that to this PR, or do it in a separate issue/PR? Note that there's a potential for a merge conflict with PR #1338 (which added docs for database.connection_timeout and database.max_tries).

I'll update the docs and fix the conflicts.

tests/commands/test_utils.py
vrde: If I can avoid using `@pytest.mark.parametrize` I'm happy. For me it adds cognitive overload, especially when mixed with `operator`.

vrde: > This test could fail for multiple different reasons, and upon failure it wouldn't be straightforward to know what is actually failing. The test could be split into three separate tests:

`_convert` is called only by `input_on_stderr`, that's why I decided to put everything in the same test.

vrde:
> I'll update the docs and fix the conflicts.

@ttmc after talking with @r-marques, we've decided to keep those options undocumented until we support them fully, i.e. BigchainDB should configure MongoDB accordingly to the configuration.

vrde:
@sbellem @r-marques: if you don't have any further comments and if Travis is happy, I'll merge it tomorrow :1st_place_medal:

ttmc:
@vrde You wrote:

>  ... we've decided to keep those options undocumented until we support them fully, i.e. BigchainDB should configure MongoDB accordingly to the configuration, but now this is not happening, so it might be confusing for them.

Okay, I'm not sure I understand, but I trust your judgement. Maybe make a new issue describing what needs to be done so we don't forget.

r-marques:
👍

vrde:
@ttmc I've just created #1371

Issue #1134, opened 2017-02-01
[http api] query parameters are "tx_id" meanwhile the payloads are "txid"


sbellem:
For instance, the query parameter for the transaction id on the statuses endpoint is `tx_id`. Given that we support querying the block id via the parameter `block_id` and that `blockid` would be awkward, then the first question is do we wish to have some kind of consistency between the query parameters and the payloads. If yes, then perhaps we could consider renaming the field `txid` to `tx_id`?

Although this may be viewed as an annoying detail, consistency can help making interfaces simpler and easier to understand.

TimDaub:
#1134 is overwritten by #1531 and is according to what was decided by the team in #1522.

I'll close this issue and we'll handle all changes in #1531. 

@r-marques: OK?

r-marques:
Yes. I was looking for an already open issue but didn't found this one

r-marques:
Yes

PR #1349, opened 2017-03-28
Events API project

vrde:
This PR is the hub to all the sub-PRs related to #1086 and specifically the [WebSocket event stream API](https://github.com/bigchaindb/bigchaindb/blob/master/docs/server/source/drivers-clients/websocket-event-stream-api.rst).


bigchaindb/web/websocket_server.py
```diff
@@ -0,0 +1,173 @@
...
+    Args:
+        queue: a queue
+        value: the value to put
+    """
+    while True:
+        try:
+            queue.put_nowait(value)
+            return
+        except asyncio.QueueFull:
+            queue.get_nowait()
```
sbellem: A `try except else` pattern could be used here, for readability and style reasons, since obviously not much can go wrong with the `return` statement :).

```python
try:
    queue.put_nowait(value)
except asyncio.QueueFull:
    queue.get_nowait()
else:
    return
```

tests/pipelines/test_election.py
```diff
@@ -199,3 +199,27 @@ def test_full_pipeline(b, user_pk):
...
+    from bigchaindb.events import setup_events_queue, EventTypes
+
+    events_queue = setup_events_queue()
+    e = election.Election(events_queue=events_queue)
+    block_id = 'a' * 64
+
+    assert events_queue.qsize() == 0
+
+    # no event should be emited in case a block is undecided
+    e.handle_block_events({'status': Bigchain.BLOCK_UNDECIDED}, block_id)
```
sbellem: small typo in comment: emited --> emitted

tests/pipelines/test_election.py
```diff
@@ -199,3 +199,27 @@ def test_full_pipeline(b, user_pk):
...
+    assert events_queue.qsize() == 0
+
+    # put an invalid block event in the queue
+    e.handle_block_events({'status': Bigchain.BLOCK_INVALID}, block_id)
+    event = e.event_handler.get_event()
+    assert event.type == EventTypes.BLOCK_INVALID
+
+    # put an valid block event in the queue
```
sbellem: small typo: an valid --> a valid

tests/pipelines/test_election.py
```diff
@@ -199,3 +199,27 @@ def test_full_pipeline(b, user_pk):
...
+    e.handle_block_events({'status': Bigchain.BLOCK_VALID}, block_id)
+    event = e.event_handler.get_event()
+    assert event.type == EventTypes.BLOCK_VALID
```
sbellem: We can treat this in a separate issue, as a clean up task, but I'll take the opportunity to share some thoughts on one possible approach regarding the imports in tests. There's this idea of not importing a module-under-tests (MUT) at module scope. One helpful reference for this approach is at http://pylonsproject.org/community-unit-testing-guidelines.html

The module-under-test (MUT) can be a function, class, or method for instance. In the test above one could say that the MUT is the class `Election`. As per the recommendation of the approach being discussed here, one would then import the `Election` class under the test testing it. The reason that is given for doing so is:

> Import failures in the module-under-test (MUT) should cause individual test cases to fail, and they should never prevent those tests from being run. Depending on the test runner, import problems may be much harder to distinguish at a glance than normal test failures.

So that's it. We do not need to be maniacal about it though :smile: as this is most likely debatable.

tests/web/test_websocket_server.py
```diff
@@ -154,15 +156,16 @@ def test_websocket_string_event(test_client, loop):
...
+def test_websocket_block_event(b, _block, test_client, loop):
     from bigchaindb import events
     from bigchaindb.web.websocket_server import init_app, POISON_PILL, EVENTS_ENDPOINT
 
     event_source = asyncio.Queue(loop=loop)
     app = init_app(event_source, loop=loop)
     client = yield from test_client(app)
     ws = yield from client.ws_connect(EVENTS_ENDPOINT)
-    block = create_block(b, 10).to_dict()
+    block = _block.to_dict()
```
sbellem: General note for this commit: 

Using a fixture is proposed here instead of a function because we already have some tests that also rely on having a block, and these tests also have their own function to create the dummy block. See https://github.com/bigchaindb/bigchaindb/blob/master/tests/db/test_bigchain_api.py#L26-L31 for instance.

We now have the opportunity to take care of this, hence this proposed change. The fixture could perhaps renamed differently, and moved up into the main `conftest` module.

vrde: OK but I believe the previous solution was more understandable[:](https://twitter.com/vrde/status/852075949835382784)

```python
def create_block(b, total=1):
    # ...


def test_websocket_block_event(b, test_client, loop):
    # ...
    block = create_block(b, 10)
    # ...
```

vs

```python
@pytest.fixture
def _block(b, request):
    total = getattr(request, 'param', 1)
    # ...


@pytest.mark.parametrize('_block', (10,), indirect=('_block',), ids=('block',))
def test_websocket_block_event(b, _block, test_client, loop):
    # ...
    block = _block.to_dict()
    # ...
```

tests/web/test_websocket_server.py
```diff
@@ -0,0 +1,228 @@
...
+
+    bridge = threading.Thread(target=_multiprocessing_to_asyncio,
+                              args=(sync_queue, async_queue, loop),
+                              daemon=True)
+    bridge.start()
+
+    sync_queue.put('we')
+    sync_queue.put('are')
+    sync_queue.put('the')
+    sync_queue.put('robots')
```
sbellem:
```python
    sync_queue.put('dreaming')
    sync_queue.put('we')
    sync_queue.put('are')
    sync_queue.put('humans')
```

tests/pipelines/test_election.py
libscott: I'm not too hot on this requirement. I've yet to see a circumstance where having the imports at the top of the file is masking a problem.

bigchaindb/__init__.py
```diff
@@ -59,6 +59,10 @@
         'workers': None,  # if none, the value will be cpu_count * 2 + 1
         'threads': None,  # if none, the value will be cpu_count * 2 + 1
     },
+    'wsserver': {
+        'host': os.environ.get('BIGCHAINDB_WSSERVER_HOST') or 'localhost',
```
libscott: My *personal* preference would be to `s/wsserver/websocket/g`, little bit more readable, but that's personal.

bigchaindb/events.py
```diff
@@ -0,0 +1,33 @@
...
+        self.data = event_data
+
+
+class EventHandler:
+
+    def __init__(self, events_queue):
+        self.events_queue = events_queue
+
+    def put_event(self, event, timeout=None):
+        # TODO: handle timeouts
```
libscott: Will these TODOs be addressed before merging? I don't think we should have "TODO" comments in master, but there should be issues addressing the requirement.

bigchaindb/web/views/info.py
```diff
@@ -30,16 +31,19 @@ def get(self):
...
             '/drivers-clients/http-client-server-api.html',
         ]
-        return {
+        return flask.jsonify({
             '_links': {
                 'docs': ''.join(docs_url),
                 'self': api_root,
                 'statuses': api_root + 'statuses/',
                 'transactions': api_root + 'transactions/',
+                # TODO: The version should probably not be hardcoded
```
libscott: I think it should be hardcoded, the point is to have a stable API.

bigchaindb/web/websocket_server.py
```diff
@@ -0,0 +1,174 @@
...
+
+import aiohttp
+from aiohttp import web
+
+from bigchaindb import config
+from bigchaindb.events import EventTypes
+
+
+logger = logging.getLogger(__name__)
+POISON_PILL = 'POISON_PILL'
```
libscott: :bat:

bigchaindb/web/websocket_server.py
```diff
@@ -0,0 +1,174 @@
...
+
+            if isinstance(event, str):
+                str_buffer.append(event)
+
+            elif event.type == EventTypes.BLOCK_VALID:
+                block = event.data
+
+                for tx in block['block']['transactions']:
+                    asset_id = tx['id'] if tx['operation'] == 'CREATE' else tx['asset']['id']
+                    data = {'blockid': block['id'],
```
libscott: I'd prefer `block_id`, `asset_id`, `tx_id`, though I've been guilty of using `txid` in the past.

bigchaindb/web/websocket_server.py
```diff
@@ -0,0 +1,174 @@
...
+
+logger = logging.getLogger(__name__)
+POISON_PILL = 'POISON_PILL'
+EVENTS_ENDPOINT = '/api/v1/streams/valid_tx'
+
+
+def _put_into_capped_queue(queue, value):
+    """Put a new item in a capped queue.
+
+    If the queue reached its limit, get the first element
```
libscott: So, to clarify, this queue will lose messages rather than block if the consumer doesn't empty it fast enough? If that's the case, why is that the desired behaviour?

tests/web/test_info.py
```diff
@@ -31,5 +31,6 @@ def test_api_v1_endpoint(client):
             'self': 'http://localhost/api/v1/',
             'statuses': 'http://localhost/api/v1/statuses/',
             'transactions': 'http://localhost/api/v1/transactions/',
+            'streams_v1': 'ws://localhost:9985/api/v1/streams/valid_tx',
```
libscott: Perhaps a trailing forwardslash to be consistent with the other paths?

libscott:
Please address codecov :+1:

sbellem:
fyi: I am fixing the codecov problem

tests/pipelines/test_election.py
vrde: I personally like the MUT approach, it makes the unit–test self contained. I'd defer it to another PR tho.

bigchaindb/__init__.py
vrde: I noticed that the `ws` prefix is used quite a lot to name libraries and variables. I was thinking if using `websocket` or `wsserver`. At the end I thought that `websocket` is the name of a protocol, while `wsserver` is more aligned with the configuration key `server`. I'd like to rename `server` to `apiserver` actually :sweat_smile:.

bigchaindb/events.py
vrde: @libscott: hybrid approach, it's fine to have `TODO`s as long as they reference an issue in our tracker. What do you think?

@r-marques: can you write an issue for this `TODO` please?

sbellem: I do not mind `TODO`s.

bigchaindb/__init__.py
libscott: Oh ok.

bigchaindb/web/websocket_server.py
vrde: I just followed [the specification](https://github.com/bigchaindb/bigchaindb/blob/9307c21baa1677ee0a06a2ba3d283fde83bcd3c3/docs/server/source/drivers-clients/websocket-event-stream-api.rst)... but :100:% agree here :dancer: :man_dancing:

Lemme ask quickly to the team.

## Question
Given the original message:
```json
{
    "txid": "<sha3-256 hash>",
    "assetid": "<sha3-256 hash>",
    "blockid": "<sha3-256 hash>"
}
```

**Do you agree in changing the message using those new keys?**
```json
{
    "tx_id": "<sha3-256 hash>",
    "asset_id": "<sha3-256 hash>",
    "block_id": "<sha3-256 hash>"
}
```

Yes: :+1:
No: :-1:

tests/web/test_websocket_server.py
sbellem: Yes, I understand. 

For this specific case if we make the default number of transactions in a block to be `10` then the decorator part is not needed. Moreover the fixture could simplified, such that we end up with something like:

```python
@pytest.fixture
def block(b, request):
    total = getattr(request, 'param', 10)
    # ...

def test_websocket_block_event(b, block, test_client, loop):
    # ...
    block_dict = block.to_dict()
    # ...
```

The whole point with this approach is that it provides us with a fixture we can re-use, and when we need to fine tune the quantity of transactions that the block contains we can do so with [pytest's `indirect` parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments) mechanism.

I think that the complexity is worth it in this case given what it provides us with, but that is my opinion. Lastly, tests in which this pattern is encountered is not un-common.

bigchaindb/events.py
libscott: Personally i use TODO to indicate something i need to address in a PR. If there are TODOs in master its no problem, i will choose another random word to use.

bigchaindb/web/websocket_server.py
sbellem: :-1:

sbellem: I am voting down because our current HTTP api uses `txid` in multiple response payloads.

I think we should be consistent, such that if we use `tx_id` we use it in all responses.

sbellem: In broader terms, I see that as an API design specification question, and it has already been answered for the HTTP API, and in some way also for the websocket API. It is too late right now to change it I think.

bigchaindb/web/websocket_server.py
vrde: Uhm, I see what you mean. I don't have a good answer for that. My assumption is that creating blocks is computationally more intense than dispatching messages to the connected clients, but I've never tested it. Let's address this issue when we have more tools to do stress testing.

bigchaindb/web/websocket_server.py
sbellem: If we wish to change the specifications (HTTP and websocket APIs) we  can:

1. do so after the planned release v0.10
2. change everything now (both HTTP and websocket APIs)
3. only change the websocket API now for `v0.10` and change the HTTP API for `v0.11`

In any case, a change to the HTTP API will break clients.

tests/web/test_info.py
vrde: *There is a bit of personal interpretation in the following statement.*

The difference is that `http://localhost/api/v1/transactions/` represents a "directory". Under it you can find transactions that are the "files".

On the other side `ws://localhost:9985/api/v1/streams/valid_tx` is a "file" since it's the stream endpoint.

We could argue if `streams_v1` should point to `ws://localhost:9985/api/v1/streams/valid_tx` or `http://localhost:9985/api/v1/streams/`

bigchaindb/web/websocket_server.py
ttmc: We're already inconsistent with `txid` and `tx_id`. We use `tx_id` in HTTP API endpoints, but `txid` inside transactions (e.g. `inputs.fulfills.txid`); see issue #1134

sbellem: link to HTTP API docs (grep `txid`): https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/http-client-server-api.html

TimDaub: > We're already inconsistent with txid and tx_id. We use tx_id in HTTP API endpoints, but txid inside transactions (e.g. inputs.fulfills.txid); see issue #1134

Yup. How about putting a stake in the ground now:

- Long term we want to separate all our object IDs with `object_id`.
- #1134 is completed by renaming all ids to the above schema `object_id`
- For this PR, we change the implementation now and it's specification.

Can we agree on that?

libscott: I agree with using `tx_id` in place of `txid` from hereon, at least as far as APIs are concerned (local variables are another thing).

sbellem: Updated my comments and vote. Basically going for:

> 3. only change the websocket API now for v0.10 and change the HTTP API for v0.11

ttmc: Yes, I agree that consistency is better (at least in the publicly-visible stuff), and `tx_id` is easier to read.

bigchaindb/web/websocket_server.py
libscott: I think since we don't have the problem of high throughput right now, it might be safer to make it a blocking queue, ie, doesn't drop messages. If a client is not reading fast enough, then the server may get blocked waiting for their socket to become writeable. In that case, the slow client socket should be somehow dealt with. If it isn't dealt with, a slow client may interfere with what other clients see just by connecting and reading slowly.

bigchaindb/web/websocket_server.py
```diff
@@ -0,0 +1,174 @@
...
+                for tx in block['block']['transactions']:
+                    asset_id = tx['id'] if tx['operation'] == 'CREATE' else tx['asset']['id']
+                    data = {'blockid': block['id'],
+                            'assetid': asset_id,
+                            'txid': tx['id']}
+                    str_buffer.append(json.dumps(data))
+
+            for _, websocket in self.subscribers.items():
+                for str_item in str_buffer:
+                    websocket.send_str(str_item)
```
libscott: Are these blocking calls?

tests/web/test_info.py
libscott: Oh, that kinda makes sense then

bigchaindb/web/websocket_server.py
vrde: Nope! Weird, right? I didn't have time to dig into the implementation details. I guess that when you call [`send_str`](http://aiohttp.readthedocs.io/en/stable/web.html?highlight=send_str#websockets) you are writing to a buffer that will be eventually emptied.

bigchaindb/web/websocket_server.py
vrde: > In that case, the slow client socket should be somehow dealt with. If it isn't dealt with, a slow client may interfere with what other clients see just by connecting and reading slowly.

AFAIK this should be handled by aiohttp itself. If a slow client cannot keep up with the stream, the output buffer will be [drained](https://github.com/aio-libs/aiohttp/blob/9b7af75cf47f1566615b65ad3d5c85469be4c9ee/aiohttp/http_websocket.py#L506).

I'll investigate a bit more and comment again.

vrde: @libscott: I spent quite some time on it but I'm still not 100% convinced.

I've opened an issue in the `aiohttp` repo:
- aio-libs/aiohttp/issues/1823

Issue #1038, opened 2017-01-12
GET /transaction/ID need status flag


diminator:
Currently it returns all transaction statuses (`backlog`, `undecided`, `valid`)

By default it should only return `valid`

The other statuses should be provided explicitly
`GET /transactions/{tx_id}?status={backlog|undecided}`

also affects the docs #830

sbellem:
> also affects the docs #830

addressed in 22806907a8ef46eddf0d83d22557836f6ac4e3c9

sbellem:
@diminator If a transaction has made it to a valid block, and is also in an undecided block, should a `GET /transactions/txi_id?status=undecided` return the transaction?

sbellem:
Here is a rough outline of some possible behaviours. That being said I strongly advise we work from real use cases to guide the decision making process.

### Return status in response payload
No status query parameter. `GET /transactions/id` returns the transaction if it is in:
* the `BACKLOG`, or
* in an `UNDECIDED` block, or
* in a `VALID` block.

**This requires adding a `status` field in the response payload.**

### Query exactly what the API consumer requests
With a `status` query parameter, such that `GET /transactions/id?status=(backlog,undecided,valid)` will query exactly what is being asked for. This means that if a transaction is both in an undecided and valid block, and an API consumer makes the calls `GET /transactions/id?status=undecided` or `GET /transactions/id?status=valid` in both cases the transaction will be returned.

**This requires a slightly modified version of `core.Bigchain.get_transaction()` that would take `status` as a parameter.**

libscott:
No, it shouldn't. A transaction that is in a valid block and also in an undecided block is valid. That undecided block is in fact no longer valid, since it would contain a double spend.

sbellem:
@libscott What about the case in which a transaction is in the backlog as well as in a valid block and/or undecided block, and GET /transactions/id?status=backlog is requested?

libscott:
I think these questions are outside the scope of this issue, the method `Bigchain.get_transaction` returns the transaction status, which is compared with the status provided by the user.

In which case the question becomes, what should `get_transaction` return as the status if the transaction is in the backlog and also in a block.

For the use case:

1. User posts transaction
1. User requests `/api/v1/transactions/[txid]?status=BACKLOG`
1. User expects to see transaction

The problem here is the user:

* They are racing against the server, which may remove the transaction from the backlog faster than they request it from the backlog, or
[...]

PR #1360, opened 2017-03-30
Fixed documentation about transactions endpoint.

r-marques:
relates to #1038 

The `/transactions` endpoint says that a transaction is returned if it was included in a valid block. This is not true. A transaction is returned as long as it was not included in an invalid block.

So this enpoint will return a transaction as long as it is in the `UNDECIDED`, `BACKLOG` and `VALID` state.

### DIFF
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -46,8 +46,10 @@ Transactions
 
    Get the transaction with the ID ``tx_id``.
 
-   This endpoint returns a transaction only if a ``VALID`` block on
-   ``bigchain`` exists.
+   This endpoint returns a transaction if it was included in a ``VALID`` block,
+   if it is still waiting to be processed ``BACKLOG`` or is still in an
+   undecided block ``UNDECIDED``. BigchainDB does not keep state about invalid
+   transactions. A request to an invalid transaction will return a ``404 Not Found``.
 
    :param tx_id: transaction ID
    :type tx_id: hex string
--- a/docs/server/source/drivers-clients/http-client-server-api.rst
+++ b/docs/server/source/drivers-clients/http-client-server-api.rst
@@ -49,8 +49,9 @@ Transactions
    This endpoint returns a transaction if it was included in a ``VALID`` block,
    if it is still waiting to be processed (``BACKLOG``) or is still in an
    undecided block (``UNDECIDED``). All instances of a transaction in invalid
-   blocks are ignored and treated as if they don't exist. A request to an
-   invalid transaction will return a ``404 Not Found``.
+   blocks are ignored and treated as if they don't exist. If a request is made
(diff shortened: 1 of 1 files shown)


TimDaub:
Talked to @r-marques: We'll link to #1038 in the description :+1:

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -46,8 +46,10 @@ Transactions
 
    Get the transaction with the ID ``tx_id``.
 
-   This endpoint returns a transaction only if a ``VALID`` block on
-   ``bigchain`` exists.
+   This endpoint returns a transaction if it was included in a ``VALID`` block,
+   if it is still waiting to be processed ``BACKLOG`` or is still in an
+   undecided block ``UNDECIDED``. BigchainDB does not keep state about invalid
```
ttmc: I would put parentheses around `BACKLOG` and `UNDECIDED`, otherwise those words break up the flow of the sentence.

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -46,8 +46,10 @@ Transactions
...
+   transactions. A request to an invalid transaction will return a ``404 Not Found``.
```
ttmc: It's actually more complicated than that, because, for example, there might be three copies of a transaction in three different blocks:

* one copy in invalid block abc123...
* one copy in invalid block xyz543...
* one copy in undecided block u4u5u6...

It would be more accurate to say that all instances of a transaction in invalid blocks are ignored and treated as if they don't exist (by this endpoint).

TimDaub: > It would be more accurate to say that all instances of a transaction in invalid blocks are ignored and treated as if they don't exist (by this endpoint).

Makes sense.

r-marques:
codecov is drunk

docs/server/source/drivers-clients/http-client-server-api.rst
```diff
@@ -46,8 +46,11 @@ Transactions
 
    Get the transaction with the ID ``tx_id``.
 
-   This endpoint returns a transaction only if a ``VALID`` block on
-   ``bigchain`` exists.
+   This endpoint returns a transaction if it was included in a ``VALID`` block,
+   if it is still waiting to be processed (``BACKLOG``) or is still in an
+   undecided block (``UNDECIDED``). All instances of a transaction in invalid
+   blocks are ignored and treated as if they don't exist. A request to an
+   invalid transaction will return a ``404 Not Found``.
```
ttmc: This last sentence bothers me because it's not clear what you mean by an "invalid transaction." The mere fact that a transaction is in a decided-invalid block doesn't mean that it's an invalid transaction. Maybe say:

If a request is made for a transaction and copies of that transaction are found only in invalid blocks, then the response will be `404 Not Found`.

PR #1365, opened 2017-04-02
Treat --log-level argument as case-insensitive

### LINKED ISSUE #1316: [log] make cmd line option for log level case insensitive

jackric:
Resolves #1316 

`ArgumentParser.add_argument` takes a callable in the `type` kwarg, which we exploit to uppercase the log level name before it's checked against `choices`

### DIFF
--- a/bigchaindb/commands/utils.py
+++ b/bigchaindb/commands/utils.py
@@ -198,7 +198,7 @@ base_parser.add_argument('-c', '--config',
                               '(use "-" for stdout)')
 
 base_parser.add_argument('-l', '--log-level',
-                         type=lambda l: l.upper(),  # case insensitive conversion
+                         type=str.upper  # convert to uppercase for comparison to choices
                          choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
                          default='INFO',
                          help='Log level')
--- a/bigchaindb/commands/utils.py
+++ b/bigchaindb/commands/utils.py
@@ -198,7 +198,7 @@ base_parser.add_argument('-c', '--config',
                               '(use "-" for stdout)')
 
 base_parser.add_argument('-l', '--log-level',
-                         type=str.upper  # convert to uppercase for comparison to choices
+                         type=str.upper, # convert to uppercase for comparison to choices
                          choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
                          default='INFO',
                          help='Log level')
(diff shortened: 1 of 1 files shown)


ttmc:
Thanks @jackric --- I sent you an email with a special block of text to paste in the comments here. Once you do that, the CLA hurdle will be passed.

bigchaindb/commands/utils.py
```diff
@@ -198,6 +198,7 @@ def mongodb_host(host):
                               '(use "-" for stdout)')
 
 base_parser.add_argument('-l', '--log-level',
+                         type=lambda l: l.upper(),  # case insensitive conversion
```
ttmc: Why not:
```python
                         type=str.upper,  # convert to uppercase for comparison to choices
```
The relevant docs about `type=` in argparse: https://docs.python.org/3.5/library/argparse.html#type 

More discussion: https://stackoverflow.com/questions/27616778/case-insensitive-argparse-choices

jackric:
BEGIN BLOCK

YUSygnoPHBiTeZsnMaAhtCfK4aPocAXxamiuXPRbUWWgesmu51YMfyhx0kYrSFi

END BLOCK

bigchaindb/commands/utils.py
jackric: Agreed, `str.upper` is nicer, I will amend

bigchaindb/commands/utils.py
```diff
@@ -198,7 +198,7 @@ def mongodb_host(host):
                               '(use "-" for stdout)')
 
 base_parser.add_argument('-l', '--log-level',
-                         type=lambda l: l.upper(),  # case insensitive conversion
+                         type=str.upper  # convert to uppercase for comparison to choices
```
ttmc: Oops, you missed the comma.

ttmc:
@jackric Travis CI is complaining. The Flake8 test is failing. When I go into Travis to get the details, it looks like they're loading, but they never do... Anyway, I suspect that the problem is the single space before the comment. According to the [Inline Comments section of PEP 8](http://legacy.python.org/dev/peps/pep-0008/#inline-comments):

"Inline comments should be separated by at least two spaces from the statement."

so there should be two spaces between the comma and the #.

bigchaindb/commands/utils.py
```diff
@@ -198,7 +198,7 @@ def mongodb_host(host):
                               '(use "-" for stdout)')
 
 base_parser.add_argument('-l', '--log-level',
-                         type=str.upper, # convert to uppercase for comparison to choices
+                         type=str.upper,  # convert to uppercase for comparison to choices
```
sbellem: you could omit the comment, as the code speak for itself in this case

ttmc: I'm going to approve and merge this pull request. It has been through enough.

PR #1372, opened 2017-04-05
block_order table + integration

libscott:
This PR introduces a `block_order` table, to add an explicit block order to the system.

It piggybacks on `Bigchain.write_block` in order to write to the `block_order` table whenever a block is written into the `bigchain` table.

The primary key used is the simple numeric block height, and the block id and previous block id are included in the payload. This has the properties:

* Multiple writes with the same block height will fail
* The chain can be validated because each item points to the previous item.


libscott:
Putting this on hold - there's a better solution: Counters table where nodes contend to CAS a single record containing block height, BEFORE writing their block.

* Blocks point to previous blocks
* Mechanism is easy to replace
* Single node is always block order leader and that can be changed easily by changing the ID of the block height counter record

PR #1384, opened 2017-04-11
[proposal] Fixture for acceptance tests

vrde:
I was trying to build an integration test for the Events API (see [`test integration from webapi to websocket`](https://github.com/bigchaindb/bigchaindb/pull/1349/files#diff-fa4976ef7face09bacee7b51842924fcR169) in  #1349) but I soon noticed that it was quite a different test from the ones we have in our code base, and that's the reason why I'm creating a PR for that.

The idea was to test the new Events API using BigchainDB as a *blackbox*:
 1. connect to the web socket endpoint
 2. create a transaction
 3. push the transaction to the API
 4. check if the transaction pops up in the web socket

This kind of testing can be defined in many ways:
- Acceptance testing
- Black box testing
- Integration testing
- Validation testing

This kind of testing also doesn't increment code coverage, but it might be useful to describe and test the behavior of BigchainDB.

Thoughts?

Update: *works on my machine™* but Travis-ci is complaining.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ Tasks:
 
 import os
 import copy
+import json
 import random
 
 import pytest
@@ -454,3 +455,34 @@ def mocked_setup_pub_logger(mocker):
 def mocked_setup_sub_logger(mocker):
     return mocker.patch(
         'bigchaindb.log.setup.setup_sub_logger', autospec=True, spec_set=True)
+
+
+@pytest.fixture
+def bigchaindb_blackbox():
+    def wait_for_connection(url):
+        while True:
+            try:
+                requests.get(url)
+            except requests.exceptions.ConnectionError:
+                sleep(0.1)
+            else:
+                break
+
+    import tempfile
+    import requests
+    from time import sleep
+    from subprocess import Popen, PIPE
+    from bigchaindb import config
+
+    with tempfile.NamedTemporaryFile(mode='w') as config_file:
+        json.dump(config, config_file)
+        config_file.flush()
+
+        proc = Popen(['bigchaindb', '-c', config_file.name, 'start'],
+                     stdin=PIPE)
+
+        wait_for_connection('http://{bind}'.format(**config['server']))
+
+        yield config
+
+        proc.kill()
--- a/tests/web/test_integration.py
+++ b/tests/web/test_integration.py
@@ -1,0 +1,28 @@
+def test_integration_web_api(bigchaindb_blackbox):
+    import random
+    import time
+
+    from bigchaindb_driver import BigchainDB, exceptions
+    from bigchaindb_driver.crypto import generate_keypair
+
+    bdb = BigchainDB('http://localhost:9984')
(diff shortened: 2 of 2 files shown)


r-marques:
I understand what you are trying to do and I agree that it would be useful to have a fixture that would start the bigchaindb process for some tests. But the test itself does not represent the usefulness of such a fixture since its basically only testing the driver which should be done on the driver project itself. From the backend we already test the views.

tests/conftest.py
```diff
@@ -454,3 +455,34 @@ def mocked_setup_pub_logger(mocker):
 def mocked_setup_sub_logger(mocker):
     return mocker.patch(
         'bigchaindb.log.setup.setup_sub_logger', autospec=True, spec_set=True)
+
+
+@pytest.fixture
+def bigchaindb_blackbox():
+    def wait_for_connection(url):
+        while True:
+            try:
```
r-marques: We should probably only try a couple of times so that this does not run forever.

tests/conftest.py
```diff
@@ -454,3 +455,34 @@ def mocked_setup_pub_logger(mocker):
...
+    import requests
+    from time import sleep
+    from subprocess import Popen, PIPE
+    from bigchaindb import config
+
+    with tempfile.NamedTemporaryFile(mode='w') as config_file:
+        json.dump(config, config_file)
+        config_file.flush()
+
+        proc = Popen(['bigchaindb', '-c', config_file.name, 'start'],
```
r-marques: We should probably use this with fixtures that we already have to clean up the database before doing this.
Also I don't think that we this fixture the tests can be paralelized because we will be trying to start the api several times on the same port.

r-marques:
All jobs are timing out. Either the bigchaindb process is not starting correctly or the is not being shutdown correctly.

vrde:
> I understand what you are trying to do and I agree that it would be useful to have a fixture that would start the bigchaindb process for some tests. But the test itself does not represent the usefulness of such a

vrde:
> I understand what you are trying to do and I agree that it would be useful to have a fixture that would start the bigchaindb process for some tests. But the test itself does not represent the usefulness of such a fixture since its basically only testing the driver which should be done on the driver project itself. From the backend we already test the views.

The test can be described as: in normal circumstances if you push a transaction through the Web API, you should see it validated at one point. It was the simplest integration test that came up to my mind while creating the fixture. The real test scenario I'm trying to tackle is [`test integration from webapi to websocket`](https://github.com/bigchaindb/bigchaindb/pull/1349/files#diff-fa4976ef7face09bacee7b51842924fcR169), but since #1349 is not merged yet, I don't have the code to test :smiley:.

> All jobs are timing out. Either the bigchaindb process is not starting correctly or is not being shutdown correctly.

Yep, I noticed it, and I still don't know why this is happening since *it works on my machine™*. I'll spend some time on it later.

r-marques:
ping

vrde:
> ping

Pong! I've set a reminder to review this at the end of next week.

PR #1389, opened 2017-04-12
Resume voting on blocks in order

### LINKED ISSUE #1264: Call to has_previous_vote in Vote pipeline shouldn't be there
Since the voting pipeline is reacting to changefeed events on the blocks table, and the blocks table does not experience updates, there should not in theory be a duplicated opportunity to vote on a block. Nodes must vote on the same blocks in the same order, the order dictated by the primary. If there is a previous vote then this is a critical error that requires investigation; the correct behaviour should be a duplicate key exception (given #1258).

@ttmc @roderik

### LINKED ISSUE #1325: get_unvoted_blocks doesn't reflect changelog order
When the vote pipeline starts up, it calls `get_unvoted_blocks` to get blocks that it missed voting on. This has some issues:

* The order of the blocks that is returned doesn't reflect changefeed order
* This is alot of useless work for a node joining the federation, their votes won't count since they wont be part of block voters list

Possible solutions:

* Remove this; if nodes miss out on changefeed events, they won't vote on blocks they missed
* Have nodes walk existing votes to extrapolate correct changefeed order via `previous_block`.
* Order by block timestamp; won't be 100% the same as changefeed order all the time but it'll be pretty good.

This is non critical really... It's incorrect but it shouldn't result in an invalid state.

cc @vrde @roderik @ttmc

libscott:
Fixes #1325
Fixes #1264

This PR addresses voting block order.

* Using MongoDB, changefeed is resumed from last voted block so block voting order is correct
* Using MongoDB, dropped connection in changefeed will not result in missed blocks
* Using RethinkDB, missed blocks in changefeed will not be voted on (before, they were voted on using an initial feed but the block order was potentially incorrect).

Note regarding implementation; it would have been quite tricky to get achieve this using the existing Changefeed class (I really tried), in the end the easiest way to extend the behaviour of the changefeed was to split the operational logic of the changefeed from the multipipes Node.


bigchaindb/backend/query.py
```diff
@@ -298,3 +284,8 @@ def get_txids_filtered(connection, asset_id, operation=None):
     """
 
     raise NotImplementedError
+
+
+@singledispatch
+def get_new_blocks_feed(connection, start_block_id):
```
r-marques: docstring

bigchaindb/backend/mongodb/changefeed.py
```diff
@@ -104,3 +75,37 @@ def get_changefeed(connection, table, operation, *, prefeed=None):
...
+            cursor = conn.run(query)
+            logging.debug('Tailing oplog at %s/%s', namespace, last_ts)
+            while cursor.alive:
+                try:
+                    record = cursor.next()
+                    yield record
+                    last_ts = record['ts']
+                except StopIteration:
+                    if _FEED_STOP:
+                        return
```
r-marques: Since this is being used as an iterator shouldn't we raise a `StopIteration` at this point?
Whatever is consuming this may not be expecting it to return `None`

bigchaindb/backend/rethinkdb/changefeed.py
```diff
@@ -42,6 +33,19 @@ def run_changefeed(self):
...
 
 
+def run_changefeed(connection, table):
+    """Encapsulate operational logic of tailing changefeed from RethinkDB
+    """
+    while True:
+        try:
+            for change in connection.run(r.table(table).changes()):
+                yield change
+            break
```
r-marques: raise a `StopIteration` here?

bigchaindb/backend/mongodb/changefeed.py
libscott: The function is a generator, because it uses `yield`, in this case you don't need to raise StopIteration manually and the caller does not expect a return value. If you try you get a "SyntaxError: 'return' with argument inside generator". Calling a generator function returns an iterator which raises StopIteration when the generator returns.

bigchaindb/backend/rethinkdb/changefeed.py
libscott: See other comment

PR #1391, opened 2017-04-13
Repaired broken hyperlinks in the server docs

ttmc:
I fixed all the errors caused by Sphinx's `make linkcheck`, and also updated some of the hyperlinks that were redirecting (but not causing errors).


ttmc:
I didn't bother requesting reviewers on this one as it was mostly just trivial changes.

PR #1403, opened 2017-04-18
Dispatcher is a consumer, no capped queue needed

vrde:
My previous implementation of the Events API was based on the wrong assumption that `Dispatcher` is not a consumer, but it is. This PR is to remove the useless code.

### DIFF
--- a/bigchaindb/web/websocket_server.py
+++ b/bigchaindb/web/websocket_server.py
@@ -29,26 +29,6 @@ POISON_PILL = 'POISON_PILL'
 EVENTS_ENDPOINT = '/api/v1/streams/valid_tx'
 
 
-def _put_into_capped_queue(queue, value):
-    """Put a new item in a capped queue.
-
-    If the queue reached its limit, get the first element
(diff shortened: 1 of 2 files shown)


bigchaindb/web/websocket_server.py
```diff
@@ -60,7 +40,7 @@ def _multiprocessing_to_asyncio(in_queue, out_queue, loop):
 
     while True:
         value = in_queue.get()
-        loop.call_soon_threadsafe(_put_into_capped_queue, out_queue, value)
+        loop.call_soon_threadsafe(out_queue.put_nowait, value)
```
libscott: What happens if the queue is full?

vrde: Forgot to remove the `maxsize`.

vrde: The queue is unbounded, see:
 - https://github.com/bigchaindb/bigchaindb/pull/1403/files#diff-d0bb39e8d0b6c010c762ab9744cad775L164

PR #1411, opened 2017-04-19
Fast unspents

libscott:
This PR introduces a new "fastquery" module which joins blocks on votes from a single node.

This means that the number of database queries required to get a set of valid transactions is just 2; one to select the transactions, and another to select the votes for the corresponding blocks to determine validity. Getting the unspents is therefore just 4 transactions: 2 to get the valid outputs, and 2 more to check if they are spent already.

The benchmark results show that the time to get the unspent transactions is greatly improved, and has much better scaling characteristics. The numbers can be further improved by tuning the database indexes.

## Benchmark

A [very simple](https://gist.github.com/libscott/a2dabde429c436f2b8ccd62476585e24) benchmark was performed. Given a test size `n`:

1) Insert n CREATE transactions and vote them valid in separate blocks
2) insert n TRANSFER transactions, spending all the CREATES, and vote them valid in separate blocks
3) Get the number of unspent transactions (and assert that it's equal to `n`).

n | master | fast-unspents
-- | ------------ | -------------
1 | 28.11ms | 2.81ms
10 | 267.78ms | 5.70ms
100 | 3216.10ms | 29.57ms
1000 | 1-2 mins | 274.96ms
10000 | ???? | 3214.42ms

### DIFF
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -207,7 +207,6 @@ def get_votes_for_blocks_by_voter(connection, block_ids, pubkey):
     Returns:
         A cursor of votes matching given votes.
     """
-
     raise NotImplementedError
 
 
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -141,6 +141,19 @@ def get_spent(connection, transaction_id, condition_id):
 
 
 @singledispatch
+def get_spending_transactions(connection, inputs):
+    """Return transactions which spend given inputs
+
+    Args:
+        inputs (list): list of {txid, output}
+
+    Returns:
+        List of transactions that spend given inputs
+    """
+    raise NotImplementedError
+
+
+@singledispatch
 def get_owned_ids(connection, owner):
     """Retrieve a list of `txids` that can we used has inputs.
 
@@ -178,6 +191,21 @@ def get_votes_by_block_id_and_voter(connection, block_id, node_pubkey):
 
     Returns:
         A cursor for the matching votes.
+    """
+
+    raise NotImplementedError
+
+
+@singledispatch
+def get_votes_for_blocks_by_voter(connection, block_ids, pubkey):
+    """Return votes for many block_ids
+
+    Args:
+        block_ids (set): block_ids
+        pubkey (str): public key of voting node
+
+    Returns:
+        A cursor of votes matching given votes.
     """
 
     raise NotImplementedError
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -148,7 +148,8 @@ def get_spending_transactions(connection, inputs):
         inputs (list): list of {txid, output}
 
     Returns:
-        List of transactions that spend given inputs
+        Iterator of (block_ids, transaction) for transactions that
+        spend given inputs.
     """
     raise NotImplementedError
 
@@ -161,9 +162,9 @@ def get_owned_ids(connection, owner):
         owner (str): base58 encoded public key.
 
     Returns:
-        A cursor for the matching transactions.
+        Iterator of (block_id, transaction) for transactions
+        that list given owner in conditions.
     """
-
     raise NotImplementedError
 
 
@@ -205,7 +206,7 @@ def get_votes_for_blocks_by_voter(connection, block_ids, pubkey):
         pubkey (str): public key of voting node
 
     Returns:
-        A cursor of votes matching given votes.
+        A cursor of votes matching given block_ids and public key
     """
     raise NotImplementedError
 
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -4,11 +4,10 @@ from time import time
 from bigchaindb import exceptions as core_exceptions
 from bigchaindb.common import crypto, exceptions
 from bigchaindb.common.utils import gen_timestamp, serialize
-from bigchaindb.common.transaction import TransactionLink
 
 import bigchaindb
 
-from bigchaindb import backend, config_utils, utils, fastquery
+from bigchaindb import backend, config_utils, fastquery
 from bigchaindb.consensus import BaseConsensusRules
 from bigchaindb.models import Block, Transaction
 
@@ -375,51 +374,6 @@ class Bigchain(object):
 
         # Either no transaction was returned spending the `(txid, output)` as
         # input or the returned transactions are not valid.
-
-    def get_outputs(self, owner):
-        """Retrieve a list of links to transaction outputs for a given public
-           key.
-
-        Args:
-            owner (str): base58 encoded public key.
-
-        Returns:
-            :obj:`list` of TransactionLink: list of ``txid`` s and ``output`` s
-            pointing to another transaction's condition
-        """
-        # get all transactions in which owner is in the `owners_after` list
-        response = backend.query.get_owned_ids(self.connection, owner)
-        return [
-            TransactionLink(tx['id'], index)
-            for tx in response
-            if not self.is_tx_strictly_in_invalid_block(tx['id'])
-            for index, output in enumerate(tx['outputs'])
-            if utils.output_has_owner(output, owner)
-        ]
-
-    def is_tx_strictly_in_invalid_block(self, txid):
-        """
-        Checks whether the transaction with the given ``txid``
-        *strictly* belongs to an invalid block.
-
-        Args:
-            txid (str): Transaction id.
-
-        Returns:
-            bool: ``True`` if the transaction *strictly* belongs to a
-            block that is invalid. ``False`` otherwise.
-
-        Note:
-            Since a transaction may be in multiple blocks, with
-            different statuses, the term "strictly" is used to
-            emphasize that if a transaction is said to be in an invalid
-            block, it means that it is not in any other block that is
-            either valid or undecided.
-
-        """
-        validity = self.get_blocks_status_containing_tx(txid)
-        return (Bigchain.BLOCK_VALID not in validity.values() and
-                Bigchain.BLOCK_UNDECIDED not in validity.values())
 
     def get_owned_ids(self, owner):
         """Retrieve a list of ``txid`` s that can be used as inputs.
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -8,7 +8,7 @@ from bigchaindb.common.transaction import TransactionLink
 
 import bigchaindb
 
-from bigchaindb import backend, config_utils, utils
+from bigchaindb import backend, config_utils, utils, fastquery
 from bigchaindb.consensus import BaseConsensusRules
 from bigchaindb.models import Block, Transaction
 
@@ -433,14 +433,17 @@ class Bigchain(object):
         """
         return self.get_outputs_filtered(owner, include_spent=False)
 
+    @property
+    def fastquery(self):
+        return fastquery.FastQuery(self.connection, self.me)
+
     def get_outputs_filtered(self, owner, include_spent=True):
         """
         Get a list of output links filtered on some criteria
         """
-        outputs = self.get_outputs(owner)
+        outputs = self.fastquery.get_outputs_by_pubkey(owner)
         if not include_spent:
-            outputs = [o for o in outputs
-                       if not self.get_spent(o.txid, o.output)]
+            outputs = self.fastquery.filter_spent_outputs(outputs)
         return outputs
 
     def get_transactions_filtered(self, asset_id, operation=None):
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -395,7 +395,7 @@ class Bigchain(object):
         """
         Get a list of output links filtered on some criteria
         """
-        outputs = self.fastquery.get_outputs_by_pubkey(owner)
+        outputs = self.fastquery.get_outputs_by_public_key(owner)
         if not include_spent:
             outputs = self.fastquery.filter_spent_outputs(outputs)
         return outputs
--- a/bigchaindb/fastquery.py
+++ b/bigchaindb/fastquery.py
@@ -9,8 +9,7 @@ class FastQuery:
     Database queries that join on block results from a single node.
 
     * Votes are not validated for security (security is replication concern)
-    * Votes come from only one node, and as such, fault tolerance is not provided
-      (elected Blockchain table not yet available)
+    * Votes come from only one node, and as such, fault tolerance is reduced
 
     In return, these queries offer good performance, as it is not neccesary to validate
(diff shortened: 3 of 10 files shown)


bigchaindb/backend/query.py
```diff
@@ -141,6 +141,19 @@ def get_spent(connection, transaction_id, condition_id):
...
 
 @singledispatch
+def get_spending_transactions(connection, inputs):
+    """Return transactions which spend given inputs
+
+    Args:
+        inputs (list): list of {txid, output}
+
+    Returns:
+        List of transactions that spend given inputs
```
r-marques: Better to describe what is actually returned. In this case a tuple (block_id, transaction).
And the same for `get_owned_ids`

bigchaindb/backend/query.py
```diff
@@ -184,6 +197,20 @@ def get_votes_by_block_id_and_voter(connection, block_id, node_pubkey):
...
 @singledispatch
+def get_votes_for_blocks_by_voter(connection, block_ids, pubkey):
+    """Return votes for many block_ids
+
+    Args:
+        block_ids (set): block_ids
+        pubkey (str): public key of voting node
+
+    Returns:
+        A cursor of votes matching given votes.
```
r-marques: Should it be something like `matching given block ids and pubkey`?

bigchaindb/core.py
```diff
@@ -433,14 +387,17 @@ def get_owned_ids(self, owner):
         """
         return self.get_outputs_filtered(owner, include_spent=False)
 
+    @property
```
r-marques: Can we make this so that we don't need to instantiate a new `FastQuery` instance every time we need to use it, or move it under `__init__`?

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,62 @@
...
+        self.me = me
+
+    def filter_block_ids(self, block_ids, include_undecided=True):
+        """
+        Given block ids, filter the invalid blocks
+        """
+        block_ids = list(set(block_ids))
+        votes = query.get_votes_for_blocks_by_voter(
+                    self.connection, block_ids, self.me)
+        votes = {v['vote']['voting_for_block']: v['vote']['is_block_valid'] for v in votes}
```
r-marques: Lets leave a note here for future reference as to why its enough to look at a single notes vote instead of the general consensus about a block.

bigchaindb/core.py
libscott: For what reason? I don't think there's a performance incentive.

bigchaindb/fastquery.py
libscott: Updated, i put it at class level though.

tests/backend/mongodb/test_queries.py
```diff
@@ -417,3 +417,52 @@ def test_get_txids_filtered(signed_create_tx, signed_transfer_tx):
...
+    conn.db.bigchain.insert_one(block.to_dict())
+
+    links = [inputs[0].fulfills.to_dict(), inputs[2].fulfills.to_dict()]
+    res = list(query.get_spending_transactions(conn, links))
+
+    # tx3 not a member because input 1 not asked for
+    assert res == [(block.id, tx2.to_dict()), (block.id, tx4.to_dict())]
+
+
+def test_get_votes_for_blocks_by_voter():
```
TimDaub: It seems `get_votes_for_blocks_by_voter` does not check the retrieved vote's signature (which should be corresponding to the node's own public/private keys).

As the votes table is replicated, this would allow other nodes to overwrite a node's vote without going noticed for a limited amount of time.

I'd say the retrieving node must check its own vote's signature (at least).
Not sure if there is any other long-hanging-fruit security adjustments we could make. If there are, lets do them (as long as they don't significantly influence performance).

bigchaindb/core.py
```diff
@@ -433,14 +387,17 @@ def get_owned_ids(self, owner):
...
+    def fastquery(self):
+        return fastquery.FastQuery(self.connection, self.me)
+
     def get_outputs_filtered(self, owner, include_spent=True):
```
TimDaub: I had the suspicion that `get_outputs` or `get_outputs_filtered` would be used in the consensus logic.

I checked through the code, but neither:

- `get_spent`
- `get_owned_ids`
- `get_transaction`

uses `get_output_filtered` when validating. Meaning that `get_outputs_filtered` does not influence the consensus logic. Good.


However, this leads me to the question of: How can we make sure that a "unsafe" method like `get_outputs_filtered` will never get used by any collaborator accidentally (e.g. because they don't know that it's not checking the block's election status).

A few options come to mind:

1. Have `get_outputs_filtered` return results based on the federation's block election status by default (only use the unsafe method in the HTTP API with a big warning.
  - Add a parameter `safe=True|False` to the method?
2. Put a big warning comment above the method explaining that it should never be used in consensus functions (however: what about e.g. `get_owned_ids`, which is a proxy of `get_outputs_filtered`? It should also never be used
3. ...?

Anyone have best practices for this?

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+        return [b for b in block_ids if votes.get(b, include_undecided)]
+
+    def filter_valid_blocks(self, blocks, key=lambda b: b[0]):
+        """
+        Given things with block ids, remove the invalid ones.
+        """
+        blocks = list(blocks)
```
TimDaub: Technically, these are still tuples of (block_id, transaction).

I'd suggest to rename.

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Given block ids, filter the invalid blocks.
+        """
+        block_ids = list(set(block_ids))
+        votes = query.get_votes_for_blocks_by_voter(
+                    self.connection, block_ids, self.me)
+        votes = {v['vote']['voting_for_block']: v['vote']['is_block_valid'] for v in votes}
+        return [b for b in block_ids if votes.get(b, include_undecided)]
+
+    def filter_valid_blocks(self, blocks, key=lambda b: b[0]):
```
TimDaub: I'm not a big fan of `key=lambda b: b[0]`.
As I wasn't familiar with the fact that `query.get_owned_ids` returns a tuple of `(block_id, transaction)`, it took me several minutes to figure this simple looking method out.

As it's called `filter_valid_blocks`, and it's parameters are `(..., blocks, ...)`. I was expecting blocks.

I'm not sure I have an ideal improvement suggestion:

How about submitting only blocks or block_ids in the first place?

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Given things with block ids, remove the invalid ones.
+        """
+        blocks = list(blocks)
+        valid_block_ids = set(self.filter_block_ids(key(b) for b in blocks))
```
TimDaub: I think calling `filter_block_ids` with it's default parameter `include_undecided=True` (which should not even be the option IMO anyways) is potentially dangerous.

**Requirement:**
https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#transaction-outputs doesn't state whether or not it returns transactions from invalid blocks. We should however have the endpoint return only transactions from valid blocks applying a best-effort & non-embarrassing performance strategy upon retrieval.


I'd either change `self.filter_block_ids(key(b)` it to include `include_undecided=False`, or remove the parameter from `filter_block_ids`.

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def filter_block_ids(self, block_ids, include_undecided=True):
+        """
+        Given block ids, filter the invalid blocks.
+        """
+        block_ids = list(set(block_ids))
+        votes = query.get_votes_for_blocks_by_voter(
+                    self.connection, block_ids, self.me)
+        votes = {v['vote']['voting_for_block']: v['vote']['is_block_valid'] for v in votes}
+        return [b for b in block_ids if votes.get(b, include_undecided)]
```
TimDaub: https://github.com/bigchaindb/bigchaindb/pull/1411/files#r113227054 See here.

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Given things with block ids, remove the invalid ones.
+        """
+        blocks = list(blocks)
+        valid_block_ids = set(self.filter_block_ids(key(b) for b in blocks))
+        return [b for b in blocks if key(b) in valid_block_ids]
+
+    def get_outputs_by_pubkey(self, pubkey):
+        """ Get outputs for a public key """
+        res = list(query.get_owned_ids(self.connection, pubkey))
```
TimDaub: Btw. I don't think this was necessarily introduced in this PR, but I'm confused that `query.get_owned_ids` and `bigchaindb.get_owned_ids` have nothing in common anymore except their name.

If anyone of the reviewers can untangle this, please do so and create a ticket. Thx.

tests/backend/mongodb/test_queries.py
libscott: See also: #1328.

bigchaindb/fastquery.py
libscott: To?

bigchaindb/fastquery.py
libscott: I don't see a problem exactly. Perfect is the enemy of the good etc.

bigchaindb/fastquery.py
libscott: This is mimicking the previous behavior: https://github.com/bigchaindb/bigchaindb/pull/1411/files#diff-7781ca6bb8e2e208778e4612964d804aL395

I don't mind changing it. Should we triage the decision?

tests/backend/mongodb/test_queries.py
TimDaub: Ah cool! :+1:

bigchaindb/fastquery.py
TimDaub: if we just submit block ids to the function, then `block_ids`

libscott: But that's what the above function, `filter_block_ids` does.

bigchaindb/fastquery.py
sbellem: Could we please use a better variable name than "`b`"? In this case, simply `block_id`, since that is what it is:

```python
return [block_id for block_id in block_ids if votes.get(block_id, include_undecided)]
```

sbellem:
*Some general thoughts.*

#### `public_key` or `pubkey`?
In some parts of the code base we name some function/method arguments `public_key`.  Could we stick to this simply to make the code more consistent?

#### What does `b` stand for?
- [ ] A `Bigchain` class instance?
- [ ] A block?
- [ ] Some block ids?
- [ ] A mongodb document?

Could we simply use explicit variable names, unless it is super duper obvious, or the variable is somewhat used as a dummy, or to be thrown away?

As far as I know, `b` was introduced to represent a `Bigchain` instance. If we stick to this tradition, then it could help a lot to keep `b` reserved for `Bigchain` instances whenever possible.

#### Can we try to name things for what they are?
As an example we currently see the name `blocks` used to represent different things. This is simply confusing.

#### About abbreviations.
As far as I know it is considered to be a good practice to name things explicitly instead of using all kinds of abbreviations that may be obvious for the programmer and/or the people closely involved with the code at the time it is being written, but will be somewhat obscure to the outsiders, and even to the very person who wrote the code, later on.

bigchaindb/fastquery.py
```diff
@@ -0,0 +1,65 @@
...
+        txs = [tx for _, tx in self.filter_valid_blocks(res)]
+        return [TransactionLink(tx['id'], i)
+                for tx in txs
+                for i, o in enumerate(tx['outputs'])
+                if output_has_owner(o, pubkey)]
```
sbellem: Could we please use/preserve the variable names `index` and `output` instead of `i` and `o` as this helps making the code clearer and quicker to understand.

libscott:
@TimDaub 

> Have retrieving node check their vote's signature according to comment above.

Lets chat offline, we could do this with replication.

> Add warning mechanism to get_outputs_filtered for prosperity

Please clarify?

> Only return transactions from blocks where the retrieving node has voted with valid already

I'm consistent with previous behaviour* from the API perspective, perhaps there could be a separate issue if that needs to be changed?
* https://github.com/bigchaindb/bigchaindb/pull/1411/files#diff-7781ca6bb8e2e208778e4612964d804aL395

TimDaub:
> Lets chat offline, we could do this with replication.

Agreed.

> Please clarify? I changed the method names for clarity, does that fix?

You also changed the default behavior to not include undecided blocks. That's great. These changes address my requests appropriately. Yes, it fixes it, thanks!

> I'm consistent with previous behaviour* from the API perspective, perhaps there could be a separate issue if that needs to be changed?

OK. We should indeed have at least a ticket for this (whatever the solution might be - can be no action at all, also). Can please you create it? I'm open to create it if you want.

jmduque:
Hi, not a BigchainDB developer, but as a company trying to use it, I still have concerns about this "fix".

I am happy to see that performance has been dramatically improved for unspents querying, but I still feel the improvements displayed below are not necessary enough:
image

I consider any API that may take more than 500ms as un-acceptable and with your current fix ~2000 transactions a user starts to become painfully slow to handle. As much as ~2000 transactions are, there is several use cases where it may and most likely will definitely happen.

While this is an step in the right direction, somehow it tries to ignore the root of the problem: An always-growing list of outputs. I think to fix the problem you may consider about creating a method to "update" a user's private/public keys. What do I mean by that? If a user's unspents becomes too slow, we application providers on top of the blockchain should have a method to create new keys for the user and request all unspents of the user to be transferred to the new public key. Since the new public key will have no output-history beyond the list of unspents it should dramatically improve the querying speed for the user's available unspents. While this can be done on top of current APIs, I think it would be a lot faster and safe to be provided as a blockchain method itself to ensure all the outputs are transferred within the same block in order to avoid potential conflicts.

libscott:
> I am happy to see that performance has been dramatically improved for unspents querying, but I still feel the improvements displayed below are not necessary enough:

Hi @jmduque,

Of course this is correct, 500ms to return 2000 transactions doesn't represent great performance and we don't expect this to satisfy the future needs of all users who use BigchainDB in production.

Querying of BigchainDB transactions in an aggregate mode is still a nascent topic and there are many problems to solve; getting unspents is obviously a fundamental one, unspents representing the "current state" of an application if you like, and there are a few different solutions that we are considering. Specialised indexes, and eventually user-definable indexes, are likely to come into play here, and it's also likely that our database storage model will evolve towards something that's easier to query.

Cheers

PR #1412, opened 2017-04-19
Docker build test in travis

krish7919:


Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN apt-get -qq update \
     && apt-get -y upgrade \
-    && pip install --no-cache-dir -e . \
+    && pip install --no-cache-dir . \
```
ttmc: Oh, good catch!

.travis.yml
```diff
@@ -1,3 +1,10 @@
+dist: trusty
+sudo: required
+services:
+  - docker
+git:
+  depth: 2
```
ttmc: I looked up what this means and found:

https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth

but I don't understand the reason for going with depth 2 rather than 5 or some other value.

krish7919: > Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.

This means that one can test only the last 2 commits to a branch and not more.
It is also related to: https://docs.travis-ci.com/user/customizing-the-build#Building-only-the-latest-commit

So of we have a branch where a lot of commits happen, only the last 2 are picked up.

krish7919:
We have to move to ubuntu trusty anyway due to this: https://blog.travis-ci.com/2017-04-17-precise-EOL

r-marques:
It's nice to have some feedback on the docker image but with this setup the docker build is tests for all the jobs under travis. We should have a separate travis job only for docker.

What we are doing here is also done on dockerhub. Is there any integration between github and dockerhub?

krish7919:
1. There is no integration between github and docker hub. Also, we do get an email when a build fails on docker hub which is post-merge always, rather than as a part of the tests/pre-merge.

2. I did not find another way to get this done. Can you be more specific on `separate travis job only for docker`?
> this setup the docker build is tests for all the jobs under travis. We should have a separate travis job only for docker.

krish7919:
@sbellem @r-marques Any suggestions on this?

PR #1416, opened 2017-04-20
Remove `editable` flag in Dockerfile

krish7919:
From `pip install --help`:
```
-e, --editable <path/url>   Install a project in editable mode (i.e.
setuptools "develop mode") from a local project path or a VCS url.
```


krish7919:
Yes, I separated it out to keep the changes logically separated.
This can wait, and does not need a new version.

sbellem:
> Do you think it's worth the hassle of creating a new version of BigchainDB once this pull request is merged? (I don't.)

no

PR #1418, opened 2017-04-20
Single node setup

krish7919:


k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -4,46 +4,28 @@
...
     metadata:
       labels:
-        app: bdb-dep
+        app: bdb-instance-0-dep
     spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: bigchaindb
-        image: bigchaindb/bigchaindb:master
+        image: bigchaindb/bigchaindb:0.10.1
```
ttmc: The comment at the top of this file says "This config file runs bigchaindb:master as a k8s Deployment...". Maybe just change that comment to "This config file runs BigchainDB Server as a k8s Deployment..." (i.e. with no explicit mention of a version number).

k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -59,8 +41,15 @@ spec:
...
-          value: ""
+        - name: BIGCHAINDB_DATABASE_MAXTRIES
+          value: "3"
+        - name: BIGCHAINDB_DATABASE_CONNECTION_TIMEOUT
+          value: "120"
+        - name: BIGCHAINDB_LOG_LEVEL_CONSOLE
+          value: debug
+        # The following env var is not required for the bootstrap/first node
+        #- name: BIGCHAINDB_KEYRING
+        #  value: ""
```
ttmc: There are several BigchainDB Server configuration settings that can be left at their default values besides the BIGCHAINDB_KEYRING. I figured you were including _all_ of the settings just to be explicit, but I suppose the list of settings has gotten rather long lately... I would be okay with _only_ including the ones which should be changed from their default values. The configuration settings and their defaults are all documented on the page: https://docs.bigchaindb.com/projects/server/en/v0.10.1/server-reference/configuration.html

k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -59,8 +41,15 @@ spec:
           value: 3CjmRhu718gT1Wkba3LfdqX5pfYuBdaMPLd7ENUga5dm
```
ttmc: It's probably not a good idea to include a valid default keypair, because someone might use it. Maybe just put something like `<insert a valid public key here>` and `<insert a valid private key corresponding to the above public key>`.

k8s/configuration/config-map.yaml
```diff
@@ -0,0 +1,36 @@
+#######################################################
+# This YAML file desribes a ConfigMap for the cluster #
```
ttmc: The opening comment indicates there is one ConfigMap for the whole cluster.

But then there are four different ConfigMaps in this file. Why not just have one ConfigMap for the entire Kubernetes cluster? (A ConfigMap can contain several key-value pairs.)

k8s/bigchaindb/bigchaindb-dep.yaml
krish7919: We can handle this in a separate PR about configuring BigchainDB. We will also have to move to Secrets in a separate PR.
Since the list of config options are huge now, we will trim these too, esp for the first node setup.

k8s/bigchaindb/bigchaindb-dep.yaml
krish7919: Agreed. I will change it to a placeholder string.

k8s/configuration/config-map.yaml
krish7919: I wanted a single one with the key value pair.
However, I would have had to change variable names all the other files in case for that to work.
Hence, this was an easy way out.
Also, 2 of them for the Cloud Manager api keys will move to secrets.
I will handle a single config map with changes in the cluster in a separate PR.

k8s/bigchaindb/bigchaindb-dep.yaml
krish7919: Fixed.

Issue #1337, opened 2017-03-24
Initialize timeout when creating a RethinkDB connection


vrde:
Right now the `connection_timeout` configuration parameter is used only by `MongoDBConnection`. The RethinkDB driver supports a similar option called [`timeout`](https://rethinkdb.com/api/python/connect/), but we are not using it.

lavinasachdev3:
What is the expected timeout ?

tdsgit:
is this right way of testing? 
`py.test -s test_connection.py`
if yes 3 attempts to connect port 28015 fails after 5000ms.

vrde:
@lavinasachdev3: sorry for now replying earlier, I missed the notification on GitHub

@tdsgit: AFAIK we are not testing if the `timeout` parameter is honored by the driver. The time should be cumulative, so if it's `3` tries, the total time is `15000ms`. To run the tests you should do `pytest tests/backend/rethinkdb/test_connection.py`.

tdsgit:
@vrde 
tried testing mongodb connection to see what's going right with mongodb first.
but bigchaindb/backend/connection.py retaining old config values and trying to connect to rethinkdb!
tried clearing tests/__pycache__
the only file which has old keypairs was tests/conftest.py .

tdsgit:
update (april 20):
the mongodb test need to be run as 
`pytest -q tests/backend/mongodb/test_connection.py --database-backend="mongodb"`
and not as
`pytest -q tests/backend/mongodb/test_connection.py`
this snippet assigns tiemout to serverselectiontimeoutms 
`           client = pymongo.MongoClient(self.host,
                                         self.port,
                                         replicaset=self.replicaset,
                                         serverselectiontimeoutms=self.connection_timeout,
                                         ssl=self.ssl)
`
however in rethinkdb the parameter to be passed is timeout ( not connection_timeout)
`r.connect(host=self.host, port=self.port, db=self.dbname, timeout=5000 )`
mock_connect.assert_called_once_with may assert
original_connect also should pass timeout

this concludes the test scripts passing timeout to rethinkdb socket.

PR #1434, opened 2017-04-21
Bug/1337/init timeout rethinkdb conn

tdsgit:
fix for timeout not initialized on rethinkdb connection - issue 1337


ttmc:
@tdsgit Thanks for tackling issue #1337 

I got an email when you agreed to the BigchainDB Contributor License Agreement. I will email you with instructions to finish that process by copying a block of text here.

ttmc:
1. The test named `test_reconnect_when_connection_lost_tries_n_times` is failing when testing with the RethinkDB backend.
2. Please edit the docs page in the file `/docs/server/source/server-reference/configuration.md`. You can delete the sentence, "Note: At the time of writing, this setting was only used by MongoDB; there was an open issue to make RethinkDB use it as well." (It's referring to the `database.connection_timeout` setting and issue #1337.)

tdsgit:

lytsOGTWWuBW0dNPVIfCe48ALgkLPgrlkmU6RUYHVDHlyZkMuNbFPoc4hmzNMCC

ttmc:
Thanks @tdsgit - the CLA is no longer a blocker.

vrde:
Hi @tdsgit, thanks for your PR. The `timeout` can be handled directly by the RethinkDB Driver:
 - https://rethinkdb.com/api/python/connect/

By passing the `timeout` parameter to the driver, you can also simplify a lot the code.

tdsgit:
@vrde 
mongodb has pymongo driver which uses connection_timeout to wait/retry till timeout.
`bigchaindb -y configure mongodb`
`bigchaindb show-config`

the timeout when passed to rethinkdb driver directly was not waiting/retrying till timeout . it immediately exits if connection fails.
`>>> import rethinkdb as r`
`>>> r.connect("localhost",port,timeout=x)`

rethinkdb too has a default connection_timeout parameter similar to mongodb.
`bigchaindb -y configure rethinkdb`
`bigchaindb show-config`

hence the change

vrde:
@tdsgit said:
> the timeout when passed to rethinkdb driver directly was not waiting/retrying till timeout . it immediately exits if connection fails.

When connecting to a socket, you can have those two errors:
- **connection error**: when the socket cannot be open because the target host is actively refusing the connection
- **timeout error** when:
  - the connection is established but the requested operation (e.g. a connection to the database) is not accomplished within the `timeout` time frame; or
  - the target host is dropping your packets.

If RethinkDB **is not** running and you try to connect, you'll get a `ReqlDriverError`. You can simulate this by **not** running RethinkDB on your machine, and trying to connect to localhost.

If RethinkDB **is** running but it is super slow in establishing the connection (it takes more than the predefined `timeout`), you'll get a `ReqlTimeoutError`. You can simulate this by running `nc -l 28015` on your machine, and trying to connect to localhost.

For the use case of #1337: connection refused and connection timeout should lead to a retry (note that this is handled already by [`bigchaindb.backend.connection.Connection::connect`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/connection.py#L130)).

The steps to achieve that are changing the module `bigchaindb.backend.rethinkdb.connection` and:
1. add the `timeout` parameter to `r.connect` [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/rethinkdb/connection.py#L43)
2. add a new `except` to handle the `ReqlTimeoutError` [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/rethinkdb/connection.py#L44)
3. (ideally, but we can take care of that) add a unit test for the timeout

tdsgit:
@vrde wrote:

> add the timeout parameter to r.connect here

have mentioned adding timeout as extra argument in call from bigchaindb [issue](https://github.com/bigchaindb/bigchaindb/issues/1337#issuecomment-295727770) and mentioned the consequences
>however in rethinkdb the parameter to be passed is timeout ( not connection_timeout)
>r.connect(host=self.host, port=self.port, db=self.dbname, timeout=5000 )
>mock_connect.assert_called_once_with may assert
>original_connect also should pass timeout

later realized there was a iteration error from test_reconnect_when_connection_lost_tries_n_times and so on. 
moreover the main issue is rethinkdb driver is not honouring timeout when **DB is not running** which is already quoted in above comment (code inside **python prompt**). its only applicable for second test case `nc -l 28015`.
mongodb waits for timeout even when DB is not running i.e for both test cases.

@vrde wrote:

>add a new except to handle the ReqlTimeoutError here

this can be added as

`except (r.ReqlDriverError,r.ReqlTimeoutError) as exc:`

vrde:
The problem with your solution is that you are trying to connect multiple times in a tight loop. The reconnection logic is handled by the super class [`bigchaindb.backend.connection.Connection`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/connection.py#L139). A solution would be to add a `time.sleep` if the connection fails, and then raise the exception. Please tell me if you have other ideas on how to solve this.

tdsgit:
@vrde 

agree with sleep. have added .

vrde:
Hi, sorry for my late reply. This implementation is still not compatible to the current behavior of the super class [`bigchaindb.backend.connection.Connection`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/connection.py#L139). The `connect` method is the one that is responsible to retry. In your implementation if the first connection fails, there is anyway a second attempt.

I don't understand the `if str(exc) == 'mock'` in the exception handling, but I assume it's something related to the test suite we have. If that's the case, I bet [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) has something we can use to avoid putting this statement in the code base.

I'm pretty busy with another project at the moment, I'll try to push some code on this PR soon.

tdsgit:
@vrde 
`connect` leads to `_connect` which then leads to `rethinkdb.connect` 
the issue is not with bigchaindb implementation but with the rethinkdb driver which does not take timeout into account. have explained it in my previous comment. although have added, `r.ReqlTimeoutError` is not thrown currently by rethinkdb driver.

as far as mock is concerned, exception message is thrown by mocktests during testing.

r-marques:
closed in favour of #1512

PR #1439, opened 2017-04-24
WebSocket support in NGINX

krish7919:
* Open port in Dockerfile
* Added the BIGCHAINDB_WS_FRONTEND_PORT and BIGCHAINDB_WS_BACKEND_PORT as parameters.
* No support for `wss://` yet.

### DIFF
--- a/k8s/nginx/container/Dockerfile
+++ b/k8s/nginx/container/Dockerfile
@@ -7,5 +7,5 @@ RUN apt-get update \
     && apt-get clean
 COPY nginx.conf.template /etc/nginx/nginx.conf
 COPY nginx_entrypoint.bash /
-EXPOSE 80 443 27017
+EXPOSE 80 81 443 444 27017
 ENTRYPOINT ["/nginx_entrypoint.bash"]
--- a/k8s/nginx/container/README.md
+++ b/k8s/nginx/container/README.md
@@ -40,6 +40,8 @@ docker run \
 --env "BIGCHAINDB_FRONTEND_PORT=<port where nginx listens for BigchainDB connections>" \
 --env "BIGCHAINDB_BACKEND_HOST=<ip/hostname of instance where BigchainDB is running>" \
 --env "BIGCHAINDB_BACKEND_PORT=<port where BigchainDB is listening for connections>" \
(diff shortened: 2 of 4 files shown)


k8s/nginx/container/Dockerfile
```diff
@@ -7,5 +7,5 @@ RUN apt-get update \
     && apt-get clean
 COPY nginx.conf.template /etc/nginx/nginx.conf
 COPY nginx_entrypoint.bash /
-EXPOSE 80 443 27017
+EXPOSE 80 81 443 444 27017
```
ttmc: You seem to be tilting in favor of ports 81 and 444 for WebSocket traffic.

I'm new to WebSocket so I looked around and it seems that it's more standard to use 80 and 443 (i.e. the same as the standard ports for HTTP and HTTPS), e.g. see https://stackoverflow.com/questions/28522857/shall-i-use-websocket-on-ports-other-than-80

krish7919: Since we use 2 different ports for Websocket and HTTP/API traffic, I have decided to map 80/9984 for the API and 81/9985 for WebSocket, and correspondingly 443/9984 for HTTPS/API and 444/9985 for Websocket Secure.

libscott: Out of interest, how come it could not be a separate `location` block within the same `server` as the HTTP API?

krish7919: That's a good question.
Currently we run the aiohttp on `9985` and the flask server at `9984`.

```
netstat -nalp | grep LIST
tcp        0      0 0.0.0.0:9984            0.0.0.0:*               LISTEN      54/python3.6    
tcp        0      0 0.0.0.0:9985            0.0.0.0:*               LISTEN      55/python3.6  
```

We can have one public port exposed to the outside world and filter by URL and forward to different bdb ports internally. For example, all URLs with `/stream` can be forwarded to the backend at port `9985`.

I decided against that because:

1.  `wss` is treated as a `stream` in NGINX, and hence it would be a non-trivial change when we move from `ws` to `wss`.

2.  I wanted to map the deployment to be similar to the code-base underneath. This will ensure that there are no surprises when users with no deployment experience have a look at it.

krish7919:
Yup. Http before https, and ws before wss. :)
Will merge now.

PR #1441, opened 2017-04-26
Updated Drivers & Clients docs page, including link to Ruby driver

ttmc:
Updated the Drivers & Clients docs page, including a link to the Ruby driver by license.rocks.

Also edited the top of the page too, to simplify it and make it clear that the "BigchainDB Team" maintains the Python Driver and the Transaction CLI.

It's a bit goofy to have a directory with a single index.rst file in it, with no "toc" but I'm keeping it that way because it will keep existing links unbroken and in the future we may want to have separate pages for different libraries, tools, clients or drivers.

### DIFF
--- a/docs/server/source/drivers-clients/index.rst
+++ b/docs/server/source/drivers-clients/index.rst
@@ -1,29 +1,27 @@
 Drivers & Clients
 =================
 
-Currently, the only language-native driver is written in the Python language.
+Libraries and Tools Maintained by the BigchainDB Team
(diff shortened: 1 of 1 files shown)


sbellem:
> It's a bit goofy to have a directory with a single index.rst file in it, with no "toc" but I'm keeping it that way because it will keep existing links unbroken and in the future we may want to have separate pages for different libraries, tools, clients or drivers.

No worries!

PR #1442, opened 2017-04-26
Config settings for MongoDB Backup Agent

krish7919:


ttmc:
Does this work?

krish7919:
No errors so far. Backups are active on Cloud Manager and I can see logsthat data is being sync'ed.

k8s/mongodb-backup-agent/container/mongodb_backup_agent_entrypoint.bash
```diff
@@ -12,9 +12,10 @@ if [[ -z "${mms_api_key}" ]]; then
 fi
 
 sed -i '/mmsApiKey/d'  $MONGODB_BACKUP_CONF_FILE
+sed -i '/mothership/d'  $MONGODB_BACKUP_CONF_FILE
 
 echo "mmsApiKey="${mms_api_key} >> $MONGODB_BACKUP_CONF_FILE
+echo "mothership=api-backup.eu-west-1.mongodb.com" >> $MONGODB_BACKUP_CONF_FILE
```
ttmc: Is this subdomain specific to us? Should it be parametrized or should we mention that they should change it (in the docs)?

krish7919: The domain cannot be specific to us. It is bad design to have a domain for every customer.
The API keys (which are parameterized) are used to distinguish users of MongoDB Cloud Manager.
Hence, the hardcoded value and not the parameterized value.

Now that you point it out, we might want to parameterize it if some of our users want that in EastUS, rather than EU.
I will file an issue for it.

ttmc: :+1:

PR #1451, opened 2017-05-04
fix for 1374

### LINKED ISSUE #1374: Better checks during startup
I forgot to create the `BIGCHAINDB_DATABASE_NAME` database in mongodb and the associated tables.
It only checks for the existence of the replica set today.

Expectation: BigchainDB should verify that the underlying schema is set up correctly, instead of silently starting up. Things to check include: existence of replica set, collections/tables and their indexes.

tdsgit:
check server replicaset database tables collections and print indexes ( for now ). 

This PR resolves #1374

### DIFF
--- a/bigchaindb/commands/utils.py
+++ b/bigchaindb/commands/utils.py
@@ -123,6 +123,55 @@ def input_on_stderr(prompt='', default=None, convert=None):
     return _convert(value, default, convert)
 
 
+def start_mongodb():
+    """Start mongo shell as a process and check if mongodb server is started
+    with replicaSet. If yes it checks if database is ready.
+    If yes it prints out the indexes in each of the collections of database
+
+    Raises:
+        :class:`~bigchaindb.common.exceptions.StartupError` if
+            mongo shell cannot be started.
+    """
+    import pymongo
+    import logging
+
+    logging.basicConfig(level=logging.INFO)
+    logger = logging.getLogger(__name__)
+    host = str(bigchaindb.config['database']['host'])
+    port = str(bigchaindb.config['database']['port'])
+    hostport = host + ':' + port
+    dbname = bigchaindb.config['database']['name']
+    replname = bigchaindb.config['database']['replicaset']
+    proc = subprocess.Popen(['mongo', '--host', host, '--port', port],
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT,
+                            universal_newlines=True)
+    line = ''
+
+    for line in proc.stdout:
+        if "MongoDB server version" in line:
+                logger.info(line)
+                try:
+                    mongoclient = pymongo.MongoClient(host=[hostport], document_class=dict,
+                                                      tz_aware=False, connect=True)
+                    mongoreplicaset = pymongo.mongo_client.MongoClient(hostport, replicaSet=replname)
+                    if mongoclient.server_info()['ok'] == 1.0 and \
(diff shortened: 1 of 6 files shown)


tdsgit:
codecov failing since test cases for mongodb missing

bigchaindb/commands/utils.py
```diff
@@ -123,6 +123,55 @@ def input_on_stderr(prompt='', default=None, convert=None):
...
+    import logging
+
+    logging.basicConfig(level=logging.INFO)
+    logger = logging.getLogger(__name__)
+    host = str(bigchaindb.config['database']['host'])
+    port = str(bigchaindb.config['database']['port'])
+    hostport = host + ':' + port
+    dbname = bigchaindb.config['database']['name']
+    replname = bigchaindb.config['database']['replicaset']
+    proc = subprocess.Popen(['mongo', '--host', host, '--port', port],
```
ttmc: Can you comment on why you found it necessary to open a mongo shell?

krish7919:
Hi @tdsgit,
Thanks for giving #1374 a shot!

I see that you have implemented a `--dev-start-mongo` flag.
Although that is good, but that is not the startup check in the usual flow.

The design I had was to have it as a part of the startup section when we find that it's a mongodb backend and then try to verify everything is set up right.

This should happen before starting the current connection pool.
So, at a high level, an initial single connection which connects to the MongoDB instance using the pymongo driver to check/verify the schema (similar to what happens [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/mongodb/connection.py#L106)).

Let me know your thoughts on this approach.

bigchaindb/commands/utils.py
tdsgit: @ttmc : sorry for delay. was out for a bit. 
used shell to check or bind to already running instance. 
mongod has bind_ip option but it seems not able to overcome the lock.

tdsgit:
@krish7919,
afaik there are two things happening. 
at startup  the schema is created if it does not exist . since issue was to check connection and schema, have not used that. 
the dev mode checks server , db , tables , collections , indexes in that order.

bigchaindb/commands/utils.py
krish7919: @tdsgit You might not always have mongo shell available where bigchaindb is run. We use the `database.host` and`database.port` args with the pymongo driver to connect to the mongodb instance.

krish7919:
Yes, the schema is created only if it does not exist.
However, if a partial schema exists (say database exists, but no tables or indexes exist), `init_database` would still return with a `DatabaseAlreadyExists` exception, without checking for tables and indexes.

This set of checks needs to happen whether it is dev mode or prod mode and hence it makes sense to have a normal check (without mongo shell or explicit CLI arg) during startup to sanitize the environment during startup.

bigchaindb/commands/utils.py
tdsgit: Pymongo driver is used for connections in the `start_mongodb` in subsequent checks.
mongo can be replaced if not needed.

tdsgit:
well, did not look into partial case. will look into it and link it to start

krish7919:
I would also suggest to check the schema setup before startup in  `schema.py` for the mongodb backend, as every backend might need to have different checks.
We can even move the replica-set check for mongodb from `connection.py` to `schema.py`.

Let me know if you have any questions and I will be happy to guide you. :)

tdsgit:
since requirements have changed, closing this PR

PR #1454, opened 2017-05-06
1374 fix

### LINKED ISSUE #1374: Better checks during startup

tdsgit:
PR resolves #1374

### DIFF
--- a/bigchaindb/backend/mongodb/connection.py
+++ b/bigchaindb/backend/mongodb/connection.py
@@ -70,6 +70,9 @@ class MongoDBConnection(Connection):
                 fails.
         """
         import inspect
+        import logging
+        from bigchaindb.common import exceptions
+        logger = logging.getLogger(__name__)
 
         try:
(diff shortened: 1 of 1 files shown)


tdsgit:
the normal start checks the server and then initializes replicaset if its not already.
the fix checks if database is found and throws exception if its not.
then it prints if it contains collections and indexes further, if not prints the same.( exception for collections and indexes yet to be written in common )

this flow-check happens only with `bigchaindb start` command and not otherwise.

TimDaub:
Currently don't have capacity to handle this PR. Sorry.

I'll close for now and reopen once this comes up again.

PR #1460, opened 2017-05-11
Feat/1275/asset decoupling

### LINKED ISSUE #1275: Decouple transaction from asset
# Proposal
This document explores the benefits in decoupling asset and transaction, and proposes a simple technical solution to implement it.

## Current implementation

In the [current implementation](https://docs.bigchaindb.com/projects/server/en/latest/data-models/transaction-model.html) of BigchainDB the asset is embedded inside the transaction model.

```json
{
    "id": "<hash of transaction, excluding signatures (see explanation)>",
    "version": "<version number of the transaction model>",
    "inputs": ["<list of inputs>"],
    "outputs": ["<list of outputs>"],
    "operation": "<string>",
    "asset": "<digital asset description (explained in the next section)>",
    "metadata": "<any JSON document>"
}
```

`asset` can have one of the following schemas:
 - if `operation` is `CREATE`, then
```json
{ "data": "<json document>" }
```
 - if `operation` is `TRANSFER`, then
```json
{ "id": "<asset's CREATE transaction ID (sha3-256 hash)>" }
```

## Decoupling transactions and assets
This proposal is to move the `asset` **outside** the transaction model to store it in another collection, and to make it content addressable. In the transaction model, `asset` will just contain the hash of the actual asset.

Moreover, decoupling transaction and asset makes a clear distinction between the permissioning layer (handled by *conditions* and *fulfillments* with *Crypto Conditions*) and the data layer. Think about it this way:
 - Each document in the new `assets` collection is just the asset as you would store it in a regular database.
 - The `bigchain` collection contains all the metadata (permissioning, provenance, auditing, and more) about the digital asset.

Note: in this proposal, when issuing a `CREATE` transaction, including the asset is considered **optional**.

## Benefits
Decoupling unlocks some interesting features for BigchainDB, specifically:
 - New assets can be `CREATE`d without revealing the content of the asset.
 - Because of the previous point, we can enable the tracking of private payloads in the blockchain: since no content is revealed, a user can encrypt it and share it off–chain, while the original hash can still be used to verify that the shared content is the expected one.
 - Since the asset is not stored in the `block` model anymore, it can be **removed** from BigchainDB without breaking the chain of hashes.
 - The content of the asset is not buried in `block['block']['transactions'][<index>]['asset']` since it would live in its own collection, hence it's simpler to write queries).
 - To improve queryability (and indexes) and speed we can even decide to store specific kind of payloads in different collections :1st_place_medal:

# Implementation

## Transaction model
The only difference in the transaction model is for the `asset` key. Instead of holding the value of the asset, it holds the hash of the asset.

```json
{
    "id": "<hash of transaction, excluding signatures (see explanation)>",
    "version": "<version number of the transaction model>",
    "inputs": ["<list of inputs>"],
    "outputs": ["<list of outputs>"],
    "operation": "<string>",
    "asset": "<hash of the asset>",
    "metadata": "<any JSON document>"
}
```
At this point, both `CREATE` and `TRANSFER` operations will just point to the hash of the asset, so there is no need to separate the two using `data` or `id`.

## API
`POST`ing data:
- The API should expose a new endpoint to upload assets. An endpoint will store the asset if and only if there is a valid `CREATE` transaction `tx` where `tx['asset'] == hash(asset)`.
- The API might extend the current [`POST /api/v1/transactions`](https://docs.bigchaindb.com/projects/server/en/v0.9.3/drivers-clients/http-client-server-api.html#post--api-v1-transactions) to accept a *transaction* and an *asset* together, to simplify the UX.

`GET`ting data:
- Symmetrically, the API should expose a new endpoint to download assets.
- Symmetrically, the API should expose a new endpoint to get a *transaction* and an *asset* together.

# :woman_astronaut: Advanced ideas :woman_astronaut:

mind-blown

## Using `CID`s
What if I told you that `asset` in the transaction can be a [CID](https://github.com/ipld/cid)? If so, we 
would be able to manage *binary assets* uploaded to other systems. (Eventually, we can ask @multiformats to add @ipdb to the [`protocols.csv`](https://github.com/multiformats/multiaddr/blob/master/protocols.csv)).

## Different collections with different indexes
To improve queryability, we can have different collections providing different indexes. Let's say we have a collection better suitable for temporal data, another one optimized for full text search. A user might be able to "route" an asset to the preferred collection.

r-marques:
resolves #1275 

This pr takes a different approach that does not require any change to the data models.
Everything stays the same, the only difference is when writing the block the the bigchain table the assets are decoupled from the block and stored separately.  When retrieving a block from the database the assets are also retrieved and block is reconstructed.

The asset is store in the `assets` table as is with the addition of the the `id` key which is the `txid` of the `CREATE` transaction that created the asset. There is a unique index on the `id` key to make sure the asset is only written once even if a transaction is written to multiple blocks due to invalid blocks.

There is another idea that is to add the `block_id` with the asset to make it easier to check it the asset belongs to a valid transaction. I created an issue #1461 for this and we can consider when implementing querying.

Main changes:
- addition of a new `assets` table to store the assets (schema, indexes, queries read/write)
- Some helper functions on the `Block` and `Transactions` model to decouple the assets from the `CREATE` transactions when writing the block to the database and to reconstruct the block when reading from the database
- Created tests for the new methods
- Fixed and updated tests and fixtures.


bigchaindb/backend/query.py
```diff
@@ -212,6 +212,33 @@ def get_block(connection, block_id):
...
+    """
+    raise NotImplementedError
+
+
+@singledispatch
+def get_assets(connection, asset_ids):
+    """Get a list of assets from the assets table.
+
+    Args:
+        asset_ids (list): a of list of ids for the assets to be retrieved from
```
sbellem: a of list --> a list?

bigchaindb/backend/query.py
```diff
@@ -259,15 +286,15 @@ def get_genesis_block(connection):
...
     """Get the last voted block for a specific node.
 
     Args:
         node_pubkey (str): base58 encoded public key.
 
     Returns:
-        The last block the node has voted on. If the node didn't cast
-        any vote then the genesis block is returned.
+        The last block id the node has voted on. If the node didn't cast
+        any vote then the genesis block id is returned.
```
sbellem: Really small detail, but it seems to me that it would be more accurate to write something like:

> The id of the last block the node has voted on.

Otherwise, "the last block id the node has voted on" sounds like the vote is being done on the id rather than on the block itself.

@ttmc thoughts?

bigchaindb/core.py
```diff
@@ -616,3 +639,26 @@ def block_election_status(self, block):
...
+
+    def get_assets(self, asset_ids):
+        """
+        Return a list of assets that match the asset_ids
+
+        Args:
+            asset_ids (:obj:`list` of :obj:`str`): A list of asset_ids to
+                retrieve from the database.
+
+        Returs:
```
sbellem: typo: Returs --> Returns :smile:

bigchaindb/backend/query.py
ttmc: That would be more clear.

bigchaindb/models.py
```diff
@@ -84,6 +86,31 @@ def from_dict(cls, tx_body):
...
+                database.
+
+        Returns:
+            :class:`~Transaction`
+
+        """
+        if tx_dict['operation'] in [Transaction.CREATE, Transaction.GENESIS]:
+            # TODO: Maybe replace this call to a call to get_asset_by_id
+            asset = list(bigchain.get_assets([tx_dict['id']]))[0]
+            asset.pop('id')
```
sbellem:
```python
del asset['id']
```
would be more precise given that we do not need the popped item :smile:

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+                asset = transaction.pop('asset')
+                asset.update({'id': transaction['id']})
+                assets.append(asset)
+
+        return (assets, block_dict)
+
+    @staticmethod
+    def couple_assets(block_dict, assets):
+        """
+        Give a block_dict with not assets (as returned from a database call)
```
sbellem: Give --> Given ?

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+        and a list of assets, reconstruct the original block by puting the
```
sbellem: puting --> putting

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+        assets back into the `CREATE` transactions in the block.
```
sbellem: I guess that is a Markdown versus ReStructuredText thing:

\`CREATE\` --> \``CREATE\``

Single ticks will render as italics with Sphinx.

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+
+        """
+        asset_ids = cls.get_asset_ids(block_dict)
+        assets = bigchain.get_assets(asset_ids)
+        block_dict = cls.couple_assets(block_dict, assets)
+        return cls.from_dict(block_dict)
+
+    def decouple_assets(self):
+        """
+        Extracts the assets from the `CREATE` transactions in the block.
```
sbellem: I guess that is a Markdown versus ReStructuredText thing:

\`CREATE\` --> \``CREATE\``

Single ticks will render as italics with Sphinx.

bigchaindb/models.py
sbellem: not --> no ?

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+            dict: The dict of the reconstructed block.
+        """
+        # create a dict with {'<txid>': asset}
+        assets = {asset.pop('id'): asset for asset in assets}
+        # add the assets to the block transactions
+        for transaction in block_dict['block']['transactions']:
+            if transaction['operation'] in [Transaction.CREATE,
+                                            Transaction.GENESIS]:
+                transaction.update({'asset': assets.get(transaction['id'],
+                                                        None)})
```
sbellem: If `id` is not in `transaction`, `get` will return `None` by default, so it could be omitted here.


> `get(key[, default])`
>  Return the value for _key_ if _key_ is in the dictionary, else _default_. If _default_ is not given, it defaults to `None`, so that this method never raises a [KeyError](https://docs.python.org/3.6/library/exceptions.html#KeyError).


docs: https://docs.python.org/3.6/library/stdtypes.html#dict.get

bigchaindb/models.py
```diff
@@ -300,5 +327,96 @@ def to_dict(self):
...
+        return block_dict
+
+    @staticmethod
+    def get_asset_ids(block_dict):
+        """
+        Given a block_dict return all the asset_ids for that block (the txid
+        of CREATE transactions). Usefull to know which assets to retrieve
```
sbellem: Usefull --> Useful :smile:

tests/backend/mongodb/test_queries.py
```diff
@@ -466,3 +470,46 @@ def test_get_votes_for_blocks_by_voter():
...
+    ]
+
+    # write the assets
+    conn.db.assets.insert_many(deepcopy(assets), ordered=False)
+
+    # read only 2 assets
+    cursor = query.get_assets(conn, [1, 3])
+
+    assert cursor.count() == 2
+    assert list(cursor.sort('id', pymongo.ASCENDING)) == assets[::2]
```
sbellem: Cool slice step!

tests/test_block_model.py
```diff
@@ -153,3 +154,137 @@ def test_block_dupe_tx(self, b):
...
+        # create the block
+        block = Block(txs)
+        # decouple assets
+        assets_from_block, block_dict = block.decouple_assets()
+
+        # reconstruct the block
+        block_dict_reconstructed = Block.couple_assets(block_dict,
+                                                       assets_from_block)
+
+        # check that the reconstructed block is the as the original block
```
sbellem: ... is the as ... ---> ... is the same as ... ?

sbellem:
@r-marques there's a small conflict to resolve

```bash
$ git rebase master 
First, rewinding head to replay your work on top of it...
Applying: Initial implementation to decouple assets from transactions.
Using index info to reconstruct a base tree...
M       bigchaindb/backend/mongodb/query.py
M       bigchaindb/backend/query.py
M       bigchaindb/backend/rethinkdb/query.py
M       bigchaindb/core.py
M       tests/backend/mongodb/test_queries.py
M       tests/backend/test_generics.py
M       tests/db/test_bigchain_api.py
Falling back to patching base and 3-way merge...
Auto-merging tests/db/test_bigchain_api.py
Auto-merging tests/backend/test_generics.py
CONFLICT (content): Merge conflict in tests/backend/test_generics.py
Auto-merging tests/backend/mongodb/test_queries.py
Auto-merging bigchaindb/core.py
Auto-merging bigchaindb/backend/rethinkdb/query.py
Auto-merging bigchaindb/backend/query.py
Auto-merging bigchaindb/backend/mongodb/query.py
error: Failed to merge in the changes.
Patch failed at 0001 Initial implementation to decouple assets from transactions.
The copy of the patch that failed is found in: .git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".

$ git diff
diff --cc tests/backend/test_generics.py
index 581a0b3,6a1e944..0000000
--- a/tests/backend/test_generics.py
+++ b/tests/backend/test_generics.py
@@@ -36,8 -36,8 +36,13 @@@ def test_schema(schema_func_name, args_
      ('get_votes_by_block_id_and_voter', 2),
      ('update_transaction', 2),
      ('get_transaction_from_block', 2),
++<<<<<<< 6aaa4503dd8407a228908d886713f1e54bad78c2
 +    ('get_votes_for_blocks_by_voter', 2),
 +    ('get_spending_transactions', 1),
++=======
+     ('write_assets', 1),
+     ('get_assets', 1),
++>>>>>>> Initial implementation to decouple assets from transactions.
  ))
  def test_query(query_func_name, args_qty):
      from bigchaindb.backend import query
```

CsterKuroi:
https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/schema.py#L22

maybe `TABLES = ('bigchain', 'backlog', 'votes') ` shoulde append `'asset'` though the variable `'TABLES'`is not used

r-marques:
Hi @CsterKuroi 

Thanks for noticing that. `TABLES` is actually used by the rethinkdb backend when setting sharding and replication. Would you like to make a pr for it?

PR #1469, opened 2017-05-12
Text search support for mongodb backend

### LINKED ISSUE #1462: Add full text search support for querying assets
Add full text support for assets.
This is feature will only be available if running with mongodb as the backend.

The goal is to provide an endpoint that returns all assets belonging to valid transactions that match a given search string.

r-marques:
resolves #1462 
depends on #1460 (incorporates all the changes of #1460)

This pr adds text search support for the assets when using mongodb as the backend.

Main changes:
- added a wildcard text index to the `assets` collection
- added a mongodb only backend query `text_search` that implements all the functionality provided by mongodb

For usage examples check this [unit test](https://github.com/bigchaindb/bigchaindb/pull/1469/files#diff-f5cfd44f63d3651ae15a85d31bd9861cR469) and the mongodb documentation for the [$text operator](https://docs.mongodb.com/manual/reference/operator/query/text/#examples)

Not sure that we should expose the entire text search functionality on the API because goal is to provide querying on the assets and not necessarily text search.
This is just a first cut of adding querying support.

Todo:
- Filter out assets from invalid transactions
- Handle the fact that rethinkdb does not support text search (this is the first case where we create a query that only works on one of the backends). `NotImplementedError` provided by `singledispatch` is not an accurate error in the case.
- Create the api endpoint (can be done in a separate pr)
- Update client driver
- Documentation


bigchaindb/backend/mongodb/query.py
```diff
@@ -327,3 +327,28 @@ def get_unvoted_blocks(conn, node_pubkey):
...
+                '$caseSensitive': case_sensitive,
+                '$diacriticSensitive': diacritic_sensitive}},
+              {'score': {'$meta': 'textScore'}, '_id': False})
+        .sort([('score', {'$meta': 'textScore'})])
+        .limit(limit))
+
+    if text_score:
+        return cursor
+    else:
+        return (_remove_text_score(asset) for asset in cursor)
```
sbellem: just a side note: the `else` is not necessary

bigchaindb/backend/query.py
```diff
@@ -328,8 +330,28 @@ def get_txids_filtered(connection, asset_id, operation=None):
 
 
 @singledispatch
-def text_search(conn, search, language='english', case_sensitive=False,
+def text_search(conn, search, *, language='english', case_sensitive=False,
```
sbellem: :+1:

bigchaindb/backend/query.py
```diff
@@ -328,8 +330,28 @@ def get_txids_filtered(connection, asset_id, operation=None):
...
-    # TODO: docstring
+    """Return all the assets that match the text search.
 
-    raise NotImplementedError
+    The results are sorted by text score.
+
+    Args:
+        search (str): Text search string to query the text index
+        language (str, optional): The language for the search and the rules for
+            stemmer and tokenizer. If the language is `None` text search uses
```
sbellem: double ticks for `None`:

\`None\` --> \``None\``

bigchaindb/backend/query.py
```diff
@@ -328,8 +330,28 @@ def get_txids_filtered(connection, asset_id, operation=None):
...
+            simple tokenization and no stemming.
+        case_sensitive (bool, optional): Enable or disable case sensitive
+            search.
+        diacritic_sensitive (bool, optional): Enable or disable case sensitive
+            diacritic search.
+        text_score (bool, optional): If `True` returns the text score with
+            each document.
```
sbellem: double ticks for `True`:

\`True\` --> \``True\``

bigchaindb/backend/query.py
```diff
@@ -328,8 +330,28 @@ def get_txids_filtered(connection, asset_id, operation=None):
...
+        limit (int, optional): Limit the number of returned documents.
+
+    Returns:
+        :obj:`list` of :obj:`dict`: a list of assets
+    """
+
+    raise OperationError('This query is only supported when running '
+                         'BigchainDB with MongoDB as the backend.')
```
sbellem: Not sure what is the ideal way to document this, but it would be nice to have some pointers, or brief explanations about stemming.

We could perhaps simply point to the MongoDB relevant docs, such as https://docs.mongodb.com/manual/reference/operator/query/text/#stemmed-words

I was a bit puzzled to see results with "baking" returned for the word "bake". The stemming behaviour explains it though:

> For case insensitive and diacritic insensitive text searches, the $text operator matches on the complete stemmed word. So if a document field contains the word blueberry, a search on the term blue will not match. However, blueberry or blueberries will match.

I think that is an important detail.

tests/db/test_bigchain_api.py
```diff
@@ -213,6 +213,89 @@ def test_get_transaction_in_invalid_and_valid_block(self, monkeypatch, b):
...
+
+        # get the assets through text search
+        # this query only works with MongoDB
+        try:
+            assets = list(b.text_search('bigchaindb'))
+        except OperationError as exc:
+            assert not isinstance(b.connection, MongoDBConnection)
+            return
+
+        assert len(assets) == 3
```
sbellem: Oh! Here's a nice use case for the `else` clause after a `try/except`:

```python
try:
    assets = list(b.text_search('bigchaindb'))
except OperationError as exc:
    assert not isinstance(b.connection, MongoDBConnection)
else:
    assert len(assets) == 3
```

tests/db/test_bigchain_api.py
```diff
@@ -213,6 +213,89 @@ def test_get_transaction_in_invalid_and_valid_block(self, monkeypatch, b):
...
+            return
+
+        # should only return one asset
+        assert len(assets) == 1
+        # should return the asset created by tx1
+        assert assets[0] == {
+            'data': {'msg': 'Hello BigchainDB!'},
+            'id': tx1.id
+        }
+
```
sbellem: Just as a side note, I wondered whether monkeypatching `time` could be problematic, or invite for troubles, and I came across [freezegun](https://github.com/spulec/freezegun/).

Since our current approach works, we don't have to bother, but just in case we run into troubles, it may be useful to know about alternatives.

bigchaindb/backend/query.py
r-marques: I am not sure what type of functionality we want to provide. The idea is to provide querying, not necessarily text search.

I don't know if we should expose the full text search functionality to the api.

sbellem: Ah ok, then maybe we could simply add a note about the stemming behaviour?

PR #1480, opened 2017-05-17
Update bigchaindb default deployment version

krish7919:
Updated to the latest release, version 0.10.2.

### DIFF
--- a/k8s/bigchaindb/bigchaindb-dep.yaml
+++ b/k8s/bigchaindb/bigchaindb-dep.yaml
@@ -18,7 +18,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: bigchaindb
-        image: bigchaindb/bigchaindb:0.10.1
+        image: bigchaindb/bigchaindb:0.10.2
         imagePullPolicy: IfNotPresent
         args:
         - start


ttmc:
I guess we should add a step to the `Release_Process.md` file, because this will have to be updated as part of every release from now on?

Release_Process.md
```diff
@@ -59,3 +59,12 @@ These steps are common between minor and patch releases:
         (without the 'v' in front) is _not_ active.
      3. Make sure the **stable** branch is _not_ active.
      4. Scroll to the bottom of the page and click the Submit button.
+1. Update all deployments with the currently released version accordingly.
```
ttmc: It seems like this point is too late in the release process to be updating the BigchainDB version number in `k8s/bigchaindb/bigchaindb-dep.yaml`. Shouldn't that be done around the same time as the update of the version number in `bigchaindb/version.py` (before the release on GitHub and PyPI)?

krish7919: Shouldn't this is to be done post the release. That's when we update the deployments.
So, it would be a `release commit` for one, and a `update deployment` commit for the other.
Or should it all be in one commit and then just add docs to remind users to update their deployment in the end?

krish7919: Also, I prefer 2 separate commits as the release might succeed but deployments might need to be changed. For example, open new ports, modify NGINX accordingly, update k8s yaml files, enable/disable security.
We have separate concerns targeted at each commit that way.

ttmc: I'd rather not mention anything about updating actual running instances of BigchainDB Server ("deployments" with a lowercase 'd') in the Release Process document. The Release Process is mostly about the steps to make a release (version X.Y.Z) of BigchainDB Server. We could have a whole other document about how to update a running BigchainDB Server instance. Indeed, that's partly what this page is about: https://docs.bigchaindb.com/projects/server/en/latest/cloud-deployment-templates/upgrade-on-kubernetes.html

krish7919: That's right. I will revert this one.

ttmc:
Let's wait for Travis to finish. It was failing earlier for some reason. Then we can merge.

krish7919:
> Let's wait for Travis to finish. It was failing earlier for some reason. Then we can merge.

Travis is broken because of the CORS support addition. It will fail again.

PR #1482, opened 2017-05-17
Add steps to Release_Process.md, to update Docker image version for k8s

ttmc:
When deploying BigchainDB Server in a Kubernetes cluster, they should pull the latest bigchaindb/bigchaindb image from Docker Hub. I added a step to update the version number in `k8s/bigchaindb/bigchaindb-dep.yaml` as part of the BigchainDB Server release process.

@krish7919 and @sbellem Do you agree with the point where I added the step to update the version number of the Docker image?

Note: I _didn't_ add a step at the very end saying how to update all running BigchainDB Server instances. That's generally something for others to do, not us (even though we also maintain some running instances of BigchainDB Server). (`Release_Process.md` is for us.)

### DIFF
--- a/Release_Process.md
+++ b/Release_Process.md
@@ -10,6 +10,7 @@ that [major version 0.x does not export a stable API](http://semver.org/#spec-it
 A minor release is preceeded by a feature freeze and created from the 'master' branch. This is a summary of the steps we go through to release a new minor version of BigchainDB Server.
 
 1. Update the `CHANGELOG.md` file in master
+1. In `k8s/bigchaindb/bigchaindb-dep.yaml`, find the line of the form `image: bigchaindb/bigchaindb:0.8.1` and change the version number to the new version number, e.g. `0.9.0`. (This is the Docker image that Kubernetes should pull from Docker Hub.) Commit that change to master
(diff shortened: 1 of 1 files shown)


sbellem:
I am wondering whether we could simply pull the `latest` image, given that we always update the version number to the latest release (in https://github.com/bigchaindb/bigchaindb/blob/master/k8s/bigchaindb/bigchaindb-dep.yaml#L21).

In the following, given that release `0.10.2` is the latest

```yaml
spec:
  spec:
    containers:
    - name: bigchaindb
      image: bigchaindb/bigchaindb:0.10.2
```
The lines:

```yaml
      image: bigchaindb/bigchaindb:0.10.2
```

```yaml
      image: bigchaindb/bigchaindb:latest
```

```yaml
      image: bigchaindb/bigchaindb
```

are all equivalent. 

So if we use either one of the last two we will not need to update the yaml file each time we do a new release.

krish7919:
Deployments always have to be explicit in specifying version, instead of implicit. It also helps in rollback, as you cannot roll back the the `latest` version.
Also, some practices here: http://blog.ploeh.dk/2013/12/10/semantic-versioning-with-continuous-deployment/

PR #1485, opened 2017-05-18
Minor docs fixes

ttmc:
1. In `docs/server/source/quickstart.md`, I told them to run MongoDB using `sudo mongod --replSet=bigchain-rs` because if they use `mongod --replSet=bigchain-rs` they will get an error like "exception in initAndListen: 20 Attempted to create a lock file on a read-only directory: /data/db, terminating". There are many ways to avoid that error, but since this is a Quickstart page, I figured running as root would suffice.

2. In `tests/README.md`, I told them to run MongoDB with the default replica set name, `bigchain-rs`, rather than `rs0` (when setting up to run the pytest tests), i.e. `mongod --replSet=bigchain-rs`

### DIFF
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -8,7 +8,7 @@ A. Install MongoDB as the database backend. (There are other options but you can
 
 B. Run MongoDB. Open a Terminal and run the command:
 ```text
-$ mongod --replSet=bigchain-rs
+$ sudo mongod --replSet=bigchain-rs
 ```
 
 C. Ubuntu 16.04 already has Python 3.5, so you don't need to install it, but you do need to install some other things:
--- a/tests/README.md
+++ b/tests/README.md
@@ -27,8 +27,7 @@ BigchainDB from source. The [`CONTRIBUTING.md` file](../CONTRIBUTING.md) has
 instructions for how to do that.
 
 Next, make sure you have RethinkDB or MongoDB running in the background. You
-can run RethinkDB using `rethinkdb --daemon` or MongoDB using `mongod
---replSet=rs0`.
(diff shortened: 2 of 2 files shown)


docs/server/source/quickstart.md
```diff
@@ -8,7 +8,7 @@ A. Install MongoDB as the database backend. (There are other options but you can
 
 B. Run MongoDB. Open a Terminal and run the command:
 ```text
-$ mongod --replSet=bigchain-rs
+$ sudo mongod --replSet=bigchain-rs
```
r-marques: Did anyone complain about this?
I don't think that running mongod with sudo is a good solution

sbellem: Hmmm ... I don't know. With docker, we run it as `root` by default (without the explicit `sudo`).

This is a MongoDB issue in any case, and ideally we could simply point the user to the appropriate documentation.

The MongoDB docs I quickly looked up does not mention the usage of `sudo`: https://docs.mongodb.com/manual/tutorial/manage-mongodb-processes/

If we give instructions I guess we can mention two alternatives:

* changing the file permissions of `/data/db` (`sudo chown u+w /data/db`).
* specifying a different data directory (e.g.: `mongod --dbpath /srv/mongodb/`).

ttmc: Today if you take the Quickstart page literally and just do `mongod --replSet=bigchain-rs` then you will get an error like "exception in initAndListen: 20 Attempted to create a lock file on a read-only directory: /data/db, terminating". That's not a great user experience.

@sbellem suggested two other fixes which would also work. We should just tell them to do one though; it's a Quickstart page, not an exhaustive list of all possible ways to get going with BigchainDB.

krish7919: @r-marques : I think MongoDB needs sudo to chown & chgrp to 'mongod' during boot.
That's the default setup, which is what happens in docker too.

PR #1488, opened 2017-05-19
Disable threads in web workers

libscott:
Threading in the frontend negatively impacts performance due to transaction validation and the GIL; it is advantageous to have one thread per web worker process and parallelise requests across processes.

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -62,7 +62,6 @@ config = {
         'loglevel': logging.getLevelName(
             log_config['handlers']['console']['level']).lower(),
         'workers': None,  # if none, the value will be cpu_count * 2 + 1
-        'threads': None,  # if none, the value will be cpu_count * 2 + 1
     },
     'wsserver': {
         'host': os.environ.get('BIGCHAINDB_WSSERVER_HOST') or 'localhost',
--- a/bigchaindb/web/server.py
+++ b/bigchaindb/web/server.py
@@ -48,7 +48,7 @@ class StandaloneApplication(gunicorn.app.base.BaseApplication):
         return self.application
 
 
-def create_app(*, debug=False, threads=4):
+def create_app(*, debug=False, threads=1):
     """Return an instance of the Flask application.
 
     Args:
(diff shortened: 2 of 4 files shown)


docs/server/source/server-reference/configuration.md
```diff
@@ -152,7 +152,7 @@ These settings are for the [Gunicorn HTTP server](http://gunicorn.org/), which i
 [Gunicorn's documentation](http://docs.gunicorn.org/en/latest/settings.html#loglevel)
 for more information.
 
-`server.workers` is [the number of worker processes](http://docs.gunicorn.org/en/stable/settings.html#workers) for handling requests. If `None` (the default), the value will be (cpu_count * 2 + 1). `server.threads` is [the number of threads-per-worker](http://docs.gunicorn.org/en/stable/settings.html#threads) for handling requests. If `None` (the default), the value will be (cpu_count * 2 + 1). The HTTP server will be able to handle `server.workers` * `server.threads` requests simultaneously.
+`server.workers` is [the number of worker processes](http://docs.gunicorn.org/en/stable/settings.html#workers) for handling requests. If `None` (the default), the value will be (cpu_count * 2 + 1). If `None` (the default), the value will be 1. The HTTP server will be able to handle `server.workers` requests simultaneously.
```
ttmc: You can remove the sentence
```text
If `None` (the default), the value will be 1.
```
Maybe add a comment that each worker process will have one thread?

bigchaindb/__init__.py
```diff
@@ -62,7 +62,6 @@
         'loglevel': logging.getLevelName(
             log_config['handlers']['console']['level']).lower(),
         'workers': None,  # if none, the value will be cpu_count * 2 + 1
-        'threads': None,  # if none, the value will be cpu_count * 2 + 1
```
r-marques: Not sure if we should remove the option to set threads.

libscott: Actually I didn't; I just hid it. If you set the parameter it'll still have effect.

bigchaindb/web/server.py
```diff
@@ -102,7 +102,10 @@ def create_server(settings):
         settings['workers'] = (multiprocessing.cpu_count() * 2) + 1
 
     if not settings.get('threads'):
```
r-marques: If we do remove the config option for threads this check is not needed since `threads` will never be set

libscott: See above

PR #1492, opened 2017-05-22
Transaction schema validation pipeline tweaks

libscott:
This branch parallelises transaction schema validation in Vote pipeline by deferring `Transaction.from_dict`.


bigchaindb/models.py
```diff
@@ -347,7 +348,8 @@ def from_db(cls, bigchain, block_dict):
         asset_ids = cls.get_asset_ids(block_dict)
         assets = bigchain.get_assets(asset_ids)
         block_dict = cls.couple_assets(block_dict, assets)
-        return cls.from_dict(block_dict)
+        kwargs = from_dict_kwargs or {}
```
r-marques: Can't we just remove this line and do `from_dict_kwargs={}` in the method parameters?

libscott: Yea, but mutable objects as default method parameters are a bit of a code smell IMO.

Issue #1505, opened 2017-05-29
Security: MongoDB <-> BigchainDB TLS connectivity


krish7919:
* Ensure authentication between BigchainDB and MongoDB instances.
* This is the first step to enable authorization later.

* Implementation details:

- [x] Add configuration options to .bigchaindb. Add a block as:
```
"ssl": {
   "ca":"absolute path to the CA certificate",
   "key":"absolute path the the consolidated public and private key file"
}
```

- [x] Sanity checks: Read only permissions and ownership of the files.

- [x] Add support for TLS in `PyMongo.MongoClient` in `bigchaindb/backend/mongodb/connection.py`.

- [x] Add support for tests with SSL support.
- [x] Add support for Travis CI

r-marques:
For reference can probably reuse some of the work done in this issue to resolve #1013

krish7919:
@r-marques :
The configuration options aren't as described above; I have added them as:
```
"database": {
        "ca_cert": "/ssl/ca.crt",
        "certfile": "/ssl/bdb-instance-0.crt",
        "keyfile": "/ssl/bdb-instance-0.key",
        "keyfile_passphrase": "",
        "crlfile": "/ssl/crl.pem"
}
```
As I thought more about it, we can have a generic `ssl` block proposed above when we support both RethinkDB and MongoDB.

r-marques:
@krish7919 can we close this?

krish7919:
:+1:

PR #1510, opened 2017-05-31
Add SSL support for MongoDB connections

krish7919:
As per #1505


.ci/travis_script.sh
```diff
@@ -5,7 +5,9 @@ set -e -x
 if [[ -n ${TOXENV} ]]; then
   tox -e ${TOXENV}
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb ]]; then
-  pytest -v --database-backend=mongodb --cov=bigchaindb
+  pytest -sv --database-backend=mongodb --cov=bigchaindb
```
krish7919: I added `-s` for debugging Travis builds. I would prefer to leave it as is to provide as much debug info for failed tests as possible. Cool?

bigchaindb/__init__.py
```diff
@@ -30,7 +30,6 @@
     'port': int(os.environ.get('BIGCHAINDB_DATABASE_PORT', 27017)),
     'name': os.environ.get('BIGCHAINDB_DATABASE_NAME', 'bigchain'),
     'replicaset': os.environ.get('BIGCHAINDB_DATABASE_REPLICASET', 'bigchain-rs'),
-    'ssl': bool(os.environ.get('BIGCHAINDB_DATABASE_SSL', False)),
```
krish7919: Moved the MongoDB 'ssl' param from `_base_database_mongodb` to `_database_mongodb` as it is not base config for us in my opinion. The `login` and the `password` params can remain to authenticate as another user.
(My hunch is we might use it in the a soon to be subsequent PR).

bigchaindb/backend/connection.py
```diff
@@ -66,13 +73,15 @@ def connect(backend=None, host=None, port=None, name=None, max_tries=None,
...
+                 replicaset=replicaset, ssl=ssl, login=login, password=password,
+                 ca_cert=ca_cert, certfile=certfile, keyfile=keyfile,
+                 keyfile_passphrase=keyfile_passphrase, crlfile=crlfile)
 
 
 class Connection:
     """Connection class interface.
 
-    All backend implementations should provide a connection class that
+    All backend implementations should provide a connection class that inherits
```
krish7919: Fixed a typo here.

bigchaindb/backend/mongodb/connection.py
```diff
@@ -12,12 +12,17 @@
                                            AuthenticationError)
 from bigchaindb.backend.connection import Connection
 
+from ssl import CERT_REQUIRED
+
```
krish7919: From the [docs](http://api.mongodb.com/python/current/api/pymongo/mongo_client.html): 
> ssl_cert_reqs: Specifies whether a certificate is required from the other side of the connection, and whether it will be validated if provided. It must be one of the three values ssl.CERT_NONE (certificates ignored), ssl.CERT_REQUIRED (certificates required and validated), or ssl.CERT_OPTIONAL (the same as CERT_REQUIRED, unless the server was configured to use anonymous ciphers). If the value of this parameter is not ssl.CERT_NONE and a value is not provided for ssl_ca_certs PyMongo will attempt to load system provided CA certificates. If the python version in use does not support loading system CA certificates then the ssl_ca_certs parameter must point to a file of CA certificates. Implies ssl=True. Defaults to ssl.CERT_REQUIRED if not provided and ssl=True.

bigchaindb/backend/mongodb/connection.py
```diff
@@ -129,6 +206,10 @@ def initialize_replica_set(host, port, connection_timeout, dbname, ssl, login, p
     else:
         _wait_for_replica_set_initialization(conn)
         logger.info('Initialized replica set')
+    finally:
+        if conn is not None:
+            logger.info('Closing initial connection to MongoDB')
+            conn.close()
```
krish7919: I think this was a nascent bug in the module; we weren't closing the connection here.

.ci/travis-before-script.sh
```diff
@@ -4,9 +4,24 @@ set -e -x
 
 if [[ "${TOXENV}" == *-rdb ]]; then
     rethinkdb --daemon
-elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb ]]; then
-    wget http://downloads.mongodb.org/linux/mongodb-linux-x86_64-3.4.1.tgz -O /tmp/mongodb.tgz
+elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
+        "${SSL_ENABLED}" == false ]]; then
```
r-marques: Can't we reuse the `BIGCHAINDB_DATABASE_SSL` env var for this?

krish7919: Fixed.

.ci/travis-before-script.sh
```diff
@@ -4,9 +4,24 @@ set -e -x
 
 if [[ "${TOXENV}" == *-rdb ]]; then
     rethinkdb --daemon
-elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb ]]; then
-    wget http://downloads.mongodb.org/linux/mongodb-linux-x86_64-3.4.1.tgz -O /tmp/mongodb.tgz
+elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
+        -z "${BIGCHAINDB_DATABASE_SSL}" ]]; then
```
r-marques: This is becoming bigger and a bit more complex. Lets put in some comments explaining the separation of these two cases.

.ci/travis_script.sh
```diff
@@ -4,8 +4,12 @@ set -e -x
...
 if [[ -n ${TOXENV} ]]; then
   tox -e ${TOXENV}
-elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb ]]; then
-  pytest -v --database-backend=mongodb --cov=bigchaindb
+elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
+    "${BIGCHAINDB_DATABASE_SSL}" == false ]]; then
+  pytest -sv --database-backend=mongodb --cov=bigchaindb
+elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
+    "${BIGCHAINDB_DATABASE_SSL}" == true ]]; then
+  pytest -sv --database-backend=mongodb-ssl --cov=bigchaindb -m bdb_ssl
```
r-marques: Here also some comments since we are using -m here for the first time.

r-marques:
We should also update the `README.md` inside the tests folder to explain how to run the tests for SSL.
We don't need to worry about the docker section for now. Just how to to start mongodb with SSL enabled for the tests and how to run the ssl related tests.

bigchaindb/__init__.py
```diff
@@ -46,6 +45,12 @@
     'backend': os.environ.get('BIGCHAINDB_DATABASE_BACKEND', 'mongodb'),
     'connection_timeout': 5000,
     'max_tries': 3,
+    'ssl': bool(os.environ.get('BIGCHAINDB_DATABASE_SSL', False)),
+    'ca_cert': os.environ.get('BIGCHAINDB_DATABASE_CA_CERT'),
+    'certfile': os.environ.get('BIGCHIANDB_DATABASE_CERTFILE'),
```
r-marques: Typo `BIGCHIANDB -> BIGCHAINDB`

bigchaindb/backend/mongodb/connection.py
```diff
@@ -12,12 +12,17 @@
                                            AuthenticationError)
 from bigchaindb.backend.connection import Connection
 
+from ssl import CERT_REQUIRED
```
r-marques: Lets move this to the beginning of the file. The way we try to organize the imports is:
```python
# standard lib
import time
import logging
from ssl import CERT_REQUIRED

# external dependencies
import pymongo

# internal dependencies
import bigchaindb
from bigchaindb.utils import Lazy
from bigchaindb.common.exceptions import ConfigurationError
from bigchaindb.backend.exceptions import (DuplicateKeyError,
                                           OperationError,
                                           ConnectionError,
                                           AuthenticationError)
from bigchaindb.backend.connection import Connection
```

bigchaindb/backend/mongodb/connection.py
```diff
@@ -63,8 +73,35 @@ def run(self, query):
...
+                                     ssl_crlfile=self.crlfile,
+                                     ssl_cert_reqs=CERT_REQUIRED)
+        return client
+
     def _connect(self):
         """Try to connect to the database.
+        This method reads the configuration options and calls
+        _secure_connect(), if required
 
         Raises:
```
r-marques: The `AuthenticationError` was not introduced in this pr but could you add it here also?

bigchaindb/backend/mongodb/connection.py
```diff
@@ -94,22 +149,44 @@ def _connect(self):
         # `initialize_replica_set` might raise `ConnectionFailure` or `OperationFailure`.
         except (pymongo.errors.ConnectionFailure,
```
r-marques: Should we catch  `pymongo.errors.ConfigurationError` here?

bigchaindb/backend/mongodb/connection.py
```diff
@@ -75,17 +112,35 @@ def _connect(self):
...
+                                   self.keyfile,
+                                   self.keyfile_passphrase,
+                                   self.crlfile)
+
+            # The presence of ca_cert, certfile, keyfile, crlfile implies the
+            # use of certificates for TLS connectivity.
+            # FYI: the connection process might raise a
+            # `ServerSelectionTimeoutError`, that is a subclass of
+            # `ConnectionFailure`.
+            if self.ca_cert is None or self.certfile is None or \
```
r-marques: Wouldn't it be enough to check whether `self.ssl` is true or false. What happens in this case if all these parameters are `None` but `self.ssl` is set to `True`?

bigchaindb/backend/mongodb/connection.py
```diff
@@ -94,22 +149,44 @@ def _connect(self):
...
+                                       ssl=ssl,
+                                       ssl_ca_certs=ca_cert,
+                                       ssl_certfile=certfile,
+                                       ssl_keyfile=keyfile,
+                                       ssl_pem_passphrase=keyfile_passphrase,
+                                       ssl_crlfile=crlfile,
+                                       ssl_cert_reqs=CERT_REQUIRED)
+
+    except (pymongo.errors.ConnectionFailure,
+            pymongo.errors.OperationFailure) as exc:
```
r-marques: Should we also catch `pymongo.errors.ConfigurationError`?

bigchaindb/backend/mongodb/connection.py
r-marques: We should probably refactor https://github.com/bigchaindb/bigchaindb/pull/1510/files#diff-9e7a185b21f57863613b664d55961cc6R204 or else the `finally` will not be reached if the replicaset is already initialized

bigchaindb/backend/mongodb/connection.py
krish7919: That means the MongoDB instance will use the certificates from the OS to connect to the database.
The `self.ca_cert`, etc. kick in only for the self-signed certificates.
Source: [link](http://api.mongodb.com/python/current/examples/tls.html#specifying-a-ca-file)

bigchaindb/backend/mongodb/connection.py
krish7919: Wouldn't the `finally` block be always executed even if there is a `return`?

krish7919: From the [docs](https://docs.python.org/3.6/tutorial/errors.html):

> A finally clause is always executed before leaving the try statement, whether an exception has occurred or not. When an exception has occurred in the try clause and has not been handled by an except clause (or it has occurred in an except or else clause), it is re-raised after the finally clause has been executed. The finally clause is also executed “on the way out” when any other clause of the try statement is left via a break, continue or return statement. A more complicated example:

tests/backend/mongodb-ssl/test_ssl_connection.py
```diff
@@ -0,0 +1,306 @@
...
+    with pytest.raises(ConnectionError):
+        conn = connect()
+        conn.db
+
+    assert mock_client.call_count == 3
+
+
+@mock.patch('bigchaindb.backend.mongodb.connection.initialize_replica_set')
+@mock.patch('pymongo.MongoClient')
+def test_ssl_connection_run_errors(mock_client, mock_init_repl_set):
```
r-marques: There are a some tests in this file that don't look related to ssl and are already tested under `tests/backend/mongodb/test_connection.py`. Wouldn't it be enough to check that we are able to connect using ssl and that everything is correctly handled if there are configuration errors?

tests/conftest.py
```diff
@@ -110,7 +114,7 @@ def _restore_dbs(request):
 
 
 @pytest.fixture(scope='session')
-def _configure_bigchaindb(request):
+def _configure_bigchaindb(request, certs_dir):
```
r-marques: Is `certs_dir` needed here?

tests/conftest.py
```diff
@@ -120,6 +124,10 @@ def _configure_bigchaindb(request):
         test_db_name = '{}_{}'.format(TEST_DB_NAME, xdist_suffix)
 
     backend = request.config.getoption('--database-backend')
+
+    if backend == 'mongodb-ssl':
+        backend = 'mongodb'
+
     config = {
```
r-marques: Do we have all the information we need here to run the ssl related tests or do we still rely on env variables to be set?

tests/backend/mongodb-ssl/test_ssl_connection.py
krish7919: Don't all the tests in `test_connection.py` relate to the connection?
I replicated all the tests to `test_ssl_connection.py`.

tests/conftest.py
krish7919: As discussed offline, removing all dependence on env variables.
We will still need `BIGCHAINDB_DATABASE_BACKEND` and `BIGCHAINDB_DATABASE_SSL`.
For ex:
```
BIGCHAINDB_DATABASE_BACKEND=mongodb \
BIGCHAINDB_DATABASE_SSL= \
pytest -sv --database-backend=mongodb -x --ff -vv
```
or 
```
BIGCHAINDB_DATABASE_BACKEND=mongodb \
BIGCHAINDB_DATABASE_SSL=true \
pytest -sv --database-backend=mongodb-ssl -x --ff -vv
```

.ci/travis-before-script.sh
krish7919: Fixed.

bigchaindb/backend/mongodb/connection.py
krish7919: Done deal.

tests/conftest.py
krish7919: It is now!

bigchaindb/backend/mongodb/connection.py
```diff
@@ -69,47 +78,107 @@ def _connect(self):
         Raises:
             :exc:`~ConnectionError`: If the connection to the database
                 fails.
+            :exc:`~AuthenticationError`: If there is a OperationFailure due to
+                Authentication failure after connecting to the database.
```
r-marques: Also `ConfigurationError`

tests/backend/mongodb-ssl/test_ssl_connection.py
r-marques: Sure. I just don't think that testing again things like the driver tries to connect 3 times before failing is really relevant since it has nothing to do with ssl

tests/conftest.py
r-marques: Now its enough to just do:
```bash
pytest --database-backend=mongodb-ssl -m bdb_ssl
```
without having to specify any env variables

krish7919: Yup.

PR #1511, opened 2017-05-31
Benchmark for CREATE transactions

libscott:
Benchmark to test throughput of CREATE transactions via the web interface.

Check out the readme: https://github.com/bigchaindb/bigchaindb/pull/1511/files#diff-3c57b0f1abdc1c79ed986908a370ae06

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -64,11 +64,22 @@ services:
       - ./setup.cfg:/usr/src/app/setup.cfg
       - ./pytest.ini:/usr/src/app/pytest.ini
       - ./tox.ini:/usr/src/app/tox.ini
+      - ./scripts:/usr/src/app/scripts
     environment:
       BIGCHAINDB_DATABASE_BACKEND: mongodb
       BIGCHAINDB_DATABASE_HOST: mdb
       BIGCHAINDB_DATABASE_PORT: 27017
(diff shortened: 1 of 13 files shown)


libscott:
throughput

docker-compose.yml
```diff
@@ -64,11 +64,22 @@ services:
...
     environment:
       BIGCHAINDB_DATABASE_BACKEND: mongodb
       BIGCHAINDB_DATABASE_HOST: mdb
       BIGCHAINDB_DATABASE_PORT: 27017
       BIGCHAINDB_SERVER_BIND: 0.0.0.0:9984
+      BIGCHAINDB_GRAPHITE_HOST: graphite
     ports:
       - "9984"
-    command: bigchaindb start
+    command: bash -c 'bigchaindb -y configure mongodb && bigchaindb start'
```
sbellem: `bigchaindb start` should be sufficient since the dockerfile build step takes care of the configuration step

sbellem: **Note**: PR #1517 improves the Dockerfile build step so that it is not only mongodb specific, but also supports rethinkdb.

libscott: It should be, but for some reason it's not. We can look at it tomorrow if you want.

PR #1518, opened 2017-06-07
Require asset.data for CREATE and asset.id for TRANSFER in the schemas

libscott:
Troy asked me about this, turns out we were not validating the presence of the asset properties in the schemas. This PR fixes.

### DIFF
--- a/bigchaindb/common/schema/transaction_transfer.yaml
+++ b/bigchaindb/common/schema/transaction_transfer.yaml
@@ -12,6 +12,8 @@ properties:
         "$ref": "#/definitions/sha3_hexdigest"
         description: |
           ID of the transaction that created the asset.
+    required:
+    - id
   inputs:
     type: array
     title: "Transaction inputs"
--- a/tests/validation/test_transaction_structure.py
+++ b/tests/validation/test_transaction_structure.py
@@ -101,6 +101,15 @@ def test_create_tx_asset_type(create_tx):
     validate_raises(create_tx)
 
 
+def test_create_tx_no_asset_data(create_tx):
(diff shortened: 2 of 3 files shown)


tests/validation/test_transaction_structure.py
```diff
@@ -101,6 +101,15 @@ def test_create_tx_asset_type(create_tx):
     validate_raises(create_tx)
 
 
+def test_create_tx_no_asset_data(create_tx):
```
ttmc: Is there a similar test for TRANSFER transactions with no `asset.id`?

libscott: No... But we dont have a test for every schematic violation. Also, a transaction with no ID will get an asset id mismatch error.

bigchaindb/common/schema/transaction_transfer.yaml
```diff
@@ -12,6 +12,8 @@ properties:
         "$ref": "#/definitions/sha3_hexdigest"
         description: |
           ID of the transaction that created the asset.
+    required:
+    - id
```
ttmc: At the top level, I suppose the `inputs` key is also required? Maybe that's already handled by requiring it (an array) to have at least one item?

tests/validation/test_transaction_structure.py
ttmc: :+1:

bigchaindb/common/schema/transaction_transfer.yaml
ttmc: :+1:

PR #1519, opened 2017-06-07
Updated some docs about the asset model

ttmc:
* Changed the wording from "digital asset model" to just "asset model" because BigchainDB can be used to track non-digital assets.
* Clarified what keys are _required_ inside the `"asset"` and their allowed values (see PR #1518).
* Gave some concrete examples of "asset" contents for the CREATE and TRANSFER cases.
* Added a link to the page about the asset model, from the page about the transaction model.
* Did some general tidying up.


libscott:
Looks good

Issue #1169, opened 2017-02-08
[http api] Include block_id and transaction status in response to GET /transactions


sbellem:
A transaction has a status (`IN_BACKLOG`, `IN_UNDECIDED_BLOCK`, `IN_VALID_BLOCK`, `IN_INVALID_BLOCK`) and potentially belongs to a block. This information is important and could be returned as part of the response to a `GET /transactions/id` for instance.

This could also save clients from making subsequent calls to get the transaction status and block id, via `GET /statuses` and `GET /blocks` respectively.

We may also wish to consider adding a timestamp, if that makes any sense, and some information about the votes.

TimDaub:
Closed according to decision in #1522

For everyone wondering about this functionality.
We already have a very similar function implemented [here](https://docs.bigchaindb.com/projects/server/en/v0.10.2/http-client-server-api.html#get--api-v1-blocks?tx_id=tx_id&status=UNDECIDED|VALID|INVALID)

Issue #1039, opened 2017-01-12
GET /statuses?tx_id also needs to be able to return `INVALID`


diminator:
Currently when a transaction is in an `invalid` block, it returns a `404`

This is because it relies on `get_transaction(self, txid, include_status=True)`

A separate query if the above returns `(None, None)` we want to query whether that transaction is found in an `invalid` block

also affects the docs #830

sbellem:
> also affects the docs #830

addressed in 22806907a8ef46eddf0d83d22557836f6ac4e3c9

r-marques:
This is not easy to implement because we don't correctly keep track of invalid transactions. The reason for this is that we add validity to blocks and not transactions.

There are several cases that may happen here:

1. The transaction is invalid and is silently dropped at block creation time.
In this case we don't store the transaction anywhere so the transaction actually does not exist anywhere in BigchainDB. If we wanted to keep track of this we would need to store invalid transactions somewhere (another table or collection).
2. A transaction is invalid and was included in a block (which should not happen in the first place). Although tricky we could probably conclude that a particular transaction is invalid because it is in a block that was voted invalid and was never included in a valid block.
3. A transaction is valid but was included in an invalid block. In this case consensus was not actually reached yet about the transaction. During the query I may get a response that the transaction was invalid but this would be a misleading response because the transaction would go back into the backlog and eventually be included in a valid block.

If we really want this functionality the best way would be to store all invalid transactions in a separate table, collection for a limited period of time so that we don't store invalid transactions forever.

Our current behaviour is similar to how other ledges behave. An invalid transaction may be accepted by the network and will be marked as unconfirmed and once its actually processed and found invalid it will be silently discarded.

TimDaub:
Closing according to discussion in #1522.

General issue is now under #1530.

PR #1522, opened 2017-06-07
Breaking changes proposal for HTTP API

TimDaub:
See document.

Will not merge into master. This is just for reviewing.
Let's have a presentation on Thursday or Friday on these proposals.

### DIFF
--- a/bdb-ip.md
+++ b/bdb-ip.md
@@ -38,17 +38,18 @@ All issues are ordered in a dependency graph to avoid backtracking in decision
 making and merge conflicts in implementation (root nodes do not have
 dependencies):
 
-- 0. What does "breaking change" mean
-    - 1. Inconsistent naming of short-hand ids
-    - 2. Inconsistency in using relative links
-        - 3. Consolidate / and /api/v1
-            - 4. Remove host and port from URLs in API and root endpoint
-        - 5. /outputs?unspents= returns unexpected results
-            - 6. A new /outputs endpoint
-- 7. /statuses?tx_id needs to return status invalid
-- 8. /transaction/ID needs status flag
-- 9. /transaction/id and /transaction?asset_id?operation=CREATE return same content
-- 10. Include block_id and transaction status in /transactions/id
+0. What does "breaking change" mean
+    0.5 tx_id or transaction_id
+    1. Inconsistent naming of short-hand ids
+    2. Inconsistency in using relative links
+        3. Consolidate / and /api/v1
+            4. Remove host and port from URLs in API and root endpoint
+        5. /outputs?unspents= returns unexpected results
+            6. A new /outputs endpoint
+7. /statuses?tx_id needs to return status invalid
+8. /transaction/ID needs status flag
+9. /transaction/id and /transaction?asset_id?operation=CREATE return same content
+10. Include block_id and transaction status in /transactions/id
 
 
 ## Second task
@@ -105,6 +106,56 @@ Proposal 1, as we only have two official driver and several users are in the
 process of building drivers.
 
 
+### 0.5 tx_id/txid or transaction_id
+
+General inconsistency: `block_id`, `asset_id` and `tx_id` instead of
+`transaction_id`.
+
+
+#### Proposal 1
+
+- Leave everything as is and tolerate inconsistency
+
+Pro:
+
+- no work
+
+Contra:
+
+- inconsistency
+
+
+#### Proposal 2
+
+- Rename all occurrences of `txid` and `tx_id` to `transaction_id`
+
+Pro:
+
+- remove inconsistency
+
+Contra:
+
+- work
+- long word to write
+
+
+#### Proposal 3
+
+- Rename all occurrences to `tx_id`
+
+Pro:
+
+- practical solution to inconsistency
+
+
+Contra:
+
+- inconsistency between `tx_id` and `api/v1/transactions`
+
+
+#### Favorite Proposal: 3
+
+
 ### [1. Inconsistent naming of short-hand ids](https://github.com/bigchaindb/bigchaindb/issues/1134)
 
 Current naming of ids according to [HTTP API documentation
@@ -117,7 +168,6 @@ and the [Websocket Event Stream API
     - `txid` in `transaction.inputs.fulfills`
     - `id` in transaction payload
     - `id` in `/assets?search={text_search}` (this is also the asset's id!)
-    - `ID` in prose text
 - Asset id:
     - `asset_id` in query parameters and Event Stream API
     - `id` in `transaction.asset` payload
@@ -126,20 +176,11 @@ and the [Websocket Event Stream API
     - `block_id` in query parameters and Event Stream API
     - `id` in the block payload
 
-General inconsistency: `block_id`, `asset_id` and `tx_id` instead of
-`transaction_id`.
 
 [As pointed out by
 @r-marques](https://github.com/bigchaindb/bigchaindb/pull/1522#discussion_r120815083),
 we'll ignore `type_id` in a Crypto-condition fulfillment for the scope of this
 discussion.
-
-
-Even though, this shouldn't be an argument (as we see the HTTP API descoped
-from the underlying implementation):
-
-- [`txid` usage in bdb-server](https://github.com/bigchaindb/bigchaindb/search?utf8=%E2%9C%93&q=txid&type=)
-- [`tx_id` usage in bdb-server](https://github.com/bigchaindb/bigchaindb/search?utf8=%E2%9C%93&q=tx_id&type=)
 
 
 #### Proposal 1
@@ -236,7 +277,7 @@ notice `_links` being used here (both blocks and transactions).
 This pattern is not repeated in `/assets`, the events API, `/blocks?tx_id`
 
 
-#### Proposal 1
+#### Proposal 1 (with `_links`)
 
 - Remove all relative links and usage of `_links` when linking to specific
   resources (doesn't count for `_links` in `/` and `/api/v1` endpoint.
@@ -248,7 +289,7 @@ This pattern is not repeated in `/assets`, the events API, `/blocks?tx_id`
   "abc", "output": 0, ...}`
 
 
-#### Proposal 2
+#### Proposal 2 (without `_links`)
 
 - Remove *ALL* `_links` and just point (links would still stay, but extra
   nesting via `_links` would be removed)
@@ -258,37 +299,6 @@ This pattern is not repeated in `/assets`, the events API, `/blocks?tx_id`
 - In `/outputs` instead of relatively linking to `../transactions/...`, return
   list of objects like (syntax not for debate here, see issue 8.): `{"tx_id":
   "abc", "output": 0, ...}`
-
-An example for `/` (assuming we don't do anything with issue 6.):
-
-```
-{
-    "api_v1": "http://example.com:9984/api/v1/",
-    "docs": "https://docs.bigchaindb.com/projects/server/en/v0.11.0.dev/"
-    "keyring": [
-        "6qHyZew94NMmUTYyHnkZsB8cxJYuRNEiEpXHe1ih9QX3",
-        "AdDuyrTyjrDt935YnFu4VBCVDhHtY2Y6rcy7x2TFeiRi"
-    ],
-    "public_key": "NC8c8rYcAhyKVpx1PCV65CBmyq4YUbLysy3Rqrg8L8mz",
-    "software": "BigchainDB",
-    "version": "0.11.0.dev"
-}
-```
-
-An example for `/api/v1` (assuming we don't do anything with issue 6.):
-
-
-```
-{
-    "docs": "https://docs.bigchaindb.com/projects/server/en/v0.11.0.dev/http-client-server-api.html",
-    "self": "http://example.com:9984/api/v1/",
-    "statuses": "http://example.com:9984/api/v1/statuses/",
-    "streams_v1": "ws://example.com:9985/api/v1/streams/valid_tx",
-    "transactions": "http://example.com:9984/api/v1/transactions/"
-}
-```
-
-You get the idea.
 
 
 Pro:
@@ -320,7 +330,7 @@ Contra:
   with JSON-LD on the rise
 
 
-#### Favorite proposal: Proposal 2
+#### Favorite proposal: Proposal 1
 
 
 ### [3. Consolidate / and /api/v1](https://github.com/bigchaindb/bigchaindb/issues/1147)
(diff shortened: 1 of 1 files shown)


bdb-ip.md
```diff
@@ -0,0 +1,437 @@
+# Improvement proposals for HTTP API
+
+
+Currently, there is two milestones grouping issues for upcoming HTTP API
+endpoints:
+
+- [Changes that break the existing HTTP API of BDB
+  v0.11](https://github.com/bigchaindb/bigchaindb/milestone/11)
+- [Changes that are additions to the exiting HTTP API of BDB
```
r-marques: issue #1169 proposes a change on the data that is returned by the endpoint. Should this be considered a breaking change?

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+    - `tx_id` in query parameters and Event Stream API
+    - `txid` in `transaction.inputs.fulfills`
+    - `id` in transaction payload
+    - `id` in `/assets?search={text_search}` (this is also the asset's id!)
+    - `ID` in prose text
+- Asset id:
+    - `asset_id` in query parameters and Event Stream API
+    - `id` in `transaction.asset` payload
+    - `id` in `/assets?search={text_search}` (this is also the transaction's id!)
+- Fulfillment id:
```
r-marques: The details of a condition are handled internally by cryptoconditions. Does the spec care about this? Should this be consistent with BigchainDB or with the other cryptoconditions implementations?

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+
+General inconsistency: `block_id`, `asset_id` and `tx_id` instead of
+`transaction_id`.
+
+
+Even though, this shouldn't be an argument (as we see the HTTP API descoped
+from the underlying implementation):
+
+- [`txid` usage in bdb-server](https://docs.bigchaindb.com/projects/server/en/master/websocket-event-stream-api.html)
+- [`txid` usage in bdb-server](https://docs.bigchaindb.com/projects/server/en/master/websocket-event-stream-api.html)
```
r-marques: These two links are the same.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+
+
+##### Proposal 1
+
+- Call any identifier that points to an object externally is a `object_id`
+    - e.g. `transaction.inputs.fulfills.tx_id` points to an external tx, hence
```
r-marques: If an identifier within the data is `object.id` and an identifier that points to an external object is `object_id` shouldn't `tx_id` actually be `transaction_id`?

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+      name
+
+Unfortunately, these rules still leave inconsistencies that are hard to
+resolve. Assuming we could agree on `asset_id` and `tx_id` easily, they could
+be a nice way forward, as it would mean that we don't have to change a lot of
+code:
+- `transaction.inputs.fulfills.txid` ==> `transaction.inputs.fulfills.tx_id`
+- `id` in /assets?search={text_search} ==> `tx_id` or `asset_id`, as it's the
+  /asset endpoint, `asset_id` seems more logical
+- In a TRANSFER transaction `transaction.asset.id` would become
```
r-marques: Here the main problem is that we are using the same `id` to identify 2 different objects

bdb-ip.md
TimDaub: Absolutely. Thanks

bdb-ip.md
TimDaub: I thought about this as well.

Ultimately, we'll have to go with whatever json structure Interledger proposes, meaning that we might just want to ignore everything within a condition or fulfillment.

bdb-ip.md
TimDaub: Thanks, I'll make the change.

bdb-ip.md
TimDaub: Yup, I mentioned that here: https://github.com/bigchaindb/bigchaindb/pull/1522/files#diff-2aeaddbadbcc8fab3ff37cb441498a43R59

I'm not sure how anal we have to be about consistency.
Of course, if we're all on board we should call it `transaction_id`.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+Currently, there is two milestones grouping issues for upcoming HTTP API
+endpoints:
+
+- [Changes that break the existing HTTP API of BDB
+  v0.11](https://github.com/bigchaindb/bigchaindb/milestone/11)
+- [Changes that are additions to the exiting HTTP API of BDB
+  v0.11](https://github.com/bigchaindb/bigchaindb/milestone/9)
+
+After releasing BDB v1.0, breaking changes to the external APIs are not
+permitted anymore. Additions to the APIs are however always allowed.
```
krish7919: Should we add an exclusion here - permitted for bugfixes and security considerations?

One more point to note might be that:
Suppose we change something in BigchainDB API, and make the corresponding changes to the (officially) supported drivers, can we claim that as long as the driver API (with the user programs) remain the same, the change is permissible. That might give us more leeway going forward if we come up with something unexpected.

TimDaub: > Should we add an exclusion here - permitted for bugfixes and security considerations?

Yes.

> can we claim that as long as the driver API (with the user programs) remain the same, the change is permissible

Good point. Something we should discuss. I'll add it to the agenda.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+- Call any identifier within the data (describing the data) an `id`
+    - `transaction.id`, points to itself, hence no need for redundancy in it's
+      name
+
+Unfortunately, these rules still leave inconsistencies that are hard to
+resolve. Assuming we could agree on `asset_id` and `tx_id` easily, they could
+be a nice way forward, as it would mean that we don't have to change a lot of
+code:
+- `transaction.inputs.fulfills.txid` ==> `transaction.inputs.fulfills.tx_id`
+- `id` in /assets?search={text_search} ==> `tx_id` or `asset_id`, as it's the
```
r-marques: The `id` in the asset is both the asset_id and the transaction_id that create the asset (same problem as in the comment below).

Another thing to consider is how the data is stored in the underlying database. We sometimes let the database itself enforce that a key is unique and rethinkdb only enforces uniqueness of the primary key. That is why we use `id` so much.

This is not necessarily a problem since we can specify the primary for a table in rethinkdb and mongodb allows us to enforce uniqueness in any index. This is something to be aware if we decide to change the name of some keys.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+- Everything becomes `object_id` *or* everything becomes `objectid`
+
+Contra:
+
+- There is huge redundancy in naming e.g. `transaction.tx_id`, etc.
+- Lots of changes to be applied
+- Doesn't resolve `tx_id` and `transaction_id` (but can potentially)
+
+
+##### Personal favorite: Proposal 1 then Proposal 2
```
r-marques: 👍

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+I can't think of any proposal. If anyone has a proposal that makes sense,
+please let me know.
+
+
+#### [Consolidate / and /api/v1](https://github.com/bigchaindb/bigchaindb/issues/1147)
+
+
+#### Proposal 1
+
+- In `/`, move `/api/v1` inside of `_links`. Remove `docs` destinction.
```
ttmc: I'd be okay with this but sending a request to `/api/v1` should return something other than 404 Not Found. Maybe something like 301 Redirect (permanent) to `/`?

bdb-ip.md
TimDaub: Fully agreed. Any ideas on how to resolve this?

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+Contra:
+
+- Doesn't resolve issue properly as the endpoint should:
+    - Return an array with both spent and unspent endpoint
+    - Return an array with only spent outputs
+    - Return an array with only unspent outputs
+- Doesn't resolves mind-stretch exercise
+
+
+##### Personal favorite: Proposal 1
```
r-marques: I also agree. Proposal 2 always makes me think twice.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+    - Event stream API for rejected transactions?
+
+##### Proposal 1
+
+- Generalize use case (e.g. As a BigchainDB user, I want to know when my
+  transactions get rejected by a BigchainDB node) and open new ticket
+- Close #1039
+
+
+##### Personal favorite: Proposal 1 lol
```
r-marques: This is a hard problem indeed and #1039 has a good discussion about the issue.
I agree that if it is a feature that we want we can create a new issue for it (and add it to the non-breaking changes milestone?)

In the meantime the api validates the transaction when it is pushed to a node and if invalid it sends a message back to client with the reason.

If the transaction is accepted by a node and ends up being invalid (double spend attempt?) the status endpoint will return a `404`. The client can always resend the transaction and get back a message with why the transaction is invalid.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+  and why should I care?")
+
+
+##### Personal favorite: Proposal 2
+
+
+#### [/transaction/id and /transaction?asset_id?operation=CREATE return same content](https://github.com/bigchaindb/bigchaindb/issues/1129)
+
+I can't think of any proposal. If anyone has a proposal that makes sense,
+please let me know.
```
ttmc: I think it would be okay to leave the longer form implemented so that there's no missing analogue to the TRANSFER version, but add a sentence or two to the documentation noting that they can get the same transaction (except not in a list) by using the `/transaction/id` version.

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+  states](https://github.com/bigchaindb/bigchaindb/issues/1038#issuecomment-274049790))
+  Safes us and the user lots of time
+- Might be used as a substitute endpoint instead of /status?tx_id
+- Ideally, the inner workings of BDB do not need to be fully understood in
+  order to work with BDB (compare to databases: Many database users aren't
+  experts on the details of the database's inner workings: "What is a BACKLOG,
+  and why should I care?")
+
+
+##### Personal favorite: Proposal 2
```
r-marques: There is one problem with Proposal 2.

1. The user does a POST and gets a `202`
2. The user does a GET for that transaction id and gets a `404` (the transaction is still in the BACKLOG or UNDECIDED but a this point the user has no idea what happened)
3. The user does another GET for that transaction id and now it gets a `200`

Another proposal would be mix of proposal 1 and proposal 4 where instead of using a status flag the transaction document always contains the status:

```json
{
    "transaction": {"..."},
    "status": "VALID | UNDECIDED | BACKLOG"
}
```
but this still suffers from all the contras of Proposal 4.

Or make use of HTTP status codes:
```python
response = get('transactions/04c00267af82c161b4bf2ad4a47d1ddbfeb47eef1a14b8d51f37d6ee00ea5cdd')
if response.status_code == 202:
    # transaction still being processed
elif response.status_code == 201:
    # transaction is valid and was persisted
elif response.status == 404:
    # transaction was not persisted
```

bdb-ip.md
```diff
@@ -0,0 +1,437 @@
...
+
+
+#### [/transaction/id and /transaction?asset_id?operation=CREATE return same content](https://github.com/bigchaindb/bigchaindb/issues/1129)
+
+I can't think of any proposal. If anyone has a proposal that makes sense,
```
r-marques: This is related to https://github.com/bigchaindb/bigchaindb/pull/1522/files#r120816238

bdb-ip.md
TimDaub: Added

bdb-ip.md
TimDaub: I'll remove the Fulfillment id from the discussion, as it's out of the scope of this document.

bdb-ip.md
TimDaub: Changed. Thx

bdb-ip.md
TimDaub: This is now point 0. on the agenda.

bdb-ip.md
TimDaub: This was added as a requirement to proposal 1.

bdb-ip.md
TimDaub: Let me try to argue the problem away:

According to [w3c's status code definition](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), a 202 has the following properties:

```
The request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon [...].
The 202 response is intentionally non-committal. Its purpose is to allow a server to accept a request for some other process (perhaps a batch-oriented process that is only run once per day) without requiring that the user agent's connection to the server persist until the process is completed. The entity returned with this response SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled.
```

to highlight:

```
SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled.
```

For a user knowing HTTP status code, there is no unexpected behavior in this case.
We send 202 back and out HTTP API clearly has a **status monitor**.

bdb-ip.md
```diff
@@ -0,0 +1,1000 @@
...
+
+@TimDaub: "It's not clear if this is a confirmed use case."
+
+
+#### Use case 6: [Outputs filtered by specific asset(s)](https://github.com/bigchaindb/bigchaindb/issues/1473)
+
+- Add `?asset_id` to query parameters and filter by
+
+
+#### Proposed /outputs endpoint design
```
TimDaub: @jmduque: 

We're currently having an ongoing design process to update the /outputs endpoint to something more useful. I tried to model the endpoint heavily to the proposal you made on gitter a while ago. I'd love to hear your feedback on it.

bdb-ip.md
```diff
@@ -0,0 +1,1000 @@
...
+
+- official Python driver
+- official Javascript driver
+
+Notice that compatibility is achieved here through officially supported
+drivers that update according to the breaking changes in the HTTP and Events
+API.
+
+
+#### Favorite Proposal: Proposal 1
```
r-marques: I also don't think Proposal 2 is enough. For instance if the data model returned from an endpoint changes the driver may not care or may be modified to work with the new data model but the user application consuming this data will break.

bdb-ip.md
```diff
@@ -0,0 +1,1000 @@
...
+
+
+#### Use case 2: Sha256Threshold outputs
+
+As a BigchainDB user
+I want to retrieve all Sha256Threshold outputs from multiple transactions for given public keys
+so that I can create a follow-up TRANSFER transaction.
+
+
+##### Scenario 2.1: Query for single Sha256Threshold output with single public key
```
r-marques: Does this mean a treshold condition with multiple ed25519 subconditions for the same public key?

TimDaub: As outlined in the scenario description below:

> Given a valid transaction with a single spent or unspent Sha256Threshold output composed from **one or more Ed25519 outputs** exists

Simply didn't put it in the title as I wanted the titles to be short.

TimDaub:
We were able to reach consensus on the base of this document.
Next steps are to implement the changes agreed on (for core: largely done already) and write up a list of all the breaking changes.

In favor of the list of breaking changes (#1547), I'm closing this PR.

PR #1536, opened 2017-06-14
Update docs with SSL environment variables

krish7919:
Documenting the environment variables required for SSL connectivity to

### DIFF
--- a/docs/server/source/server-reference/configuration.md
+++ b/docs/server/source/server-reference/configuration.md
@@ -36,6 +36,9 @@ For convenience, here's a list of all the relevant environment variables (docume
 `BIGCHAINDB_LOG_FMT_CONSOLE`<br>
 `BIGCHAINDB_LOG_FMT_LOGFILE`<br>
 `BIGCHAINDB_LOG_GRANULAR_LEVELS`<br>
+`BIGCHAINDB_DATABASE_SSL`<br>
+`BIGCHIANDB_DATABASE_LOGIN`<br>
+`BIGCHAINDB_DATABASE_PASSWORD`<br>
 `BIGCHAINDB_DATABASE_CA_CERT`<br>
 `BIGCHAINDB_DATABASE_CERTFILE`<br>
 `BIGCHAINDB_DATABASE_KEYFILE`<br>
@@ -107,8 +110,18 @@ The settings with names of the form `database.*` are for the database backend
 * `database.replicaset` is only relevant if using MongoDB; it's the name of the MongoDB replica set, e.g. `bigchain-rs`.
 * `database.connection_timeout` is the maximum number of milliseconds that BigchainDB will wait before giving up on one attempt to connect to the database backend.
 * `database.max_tries` is the maximum number of times that BigchainDB will try to establish a connection with the database backend. If 0, then it will try forever.
+* `database.ssl` is a flag that determines if BigchainDB connects to the
+  backend database over SSL or not. This can be set to any value (the presence of a value
+  qualifies as set or true in this case) to enable SSL connections. Do not
+  specify this parameter to connect over plaintext connection.
+  Note: This parameter is only supported for the MongoDB backend currently.
+* `database.login` and `database.password` are the login and password used to
+  authenticate to the database before performing any operations, specified in
+  plaintext.
 * `database.ca_cert`, `database.certfile`, `database.keyfile` and `database.crlfile` are the paths to the CA, signed certificate, private key and certificate revocation list files respectively.
+  Note: These parameters are only supported for the MongoDB backend currently.
 * `database.keyfile_passphrase` is the private key decryption passphrase, specified in plaintext.
+  Note: This parameter is only supported for the MongoDB backend currently.
 
 **Example using environment variables**
 ```text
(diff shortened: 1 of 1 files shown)


docs/server/source/server-reference/configuration.md
```diff
@@ -36,6 +36,11 @@ For convenience, here's a list of all the relevant environment variables (docume
 `BIGCHAINDB_LOG_FMT_CONSOLE`<br>
 `BIGCHAINDB_LOG_FMT_LOGFILE`<br>
 `BIGCHAINDB_LOG_GRANULAR_LEVELS`<br>
+`BIGCHAINDB_DATABASE_CA_CERT`<br>
+`BIGCHAINDB_DATABASE_CERTFILE`<br>
+`BIGCHAINDB_DATABASE_KEYFILE`<br>
+`BIGCHAINDB_DATABASE_KEYFILE_PASSPHRASE`<br>
+`BIGCHAINDB_DATABASE_CRLFILE`<br>
```
r-marques: There is also `BIGCHAINDB_DATABASE_SSL`

krish7919: `BIGCHAINDB_DATABASE_SSL`, `BIGCHIANDB_DATABASE_LOGIN` and `BIGCHAINDB_DATABASE_PASSWORD` are undocumented. I will add them.

docs/server/source/server-reference/configuration.md
```diff
@@ -102,6 +110,18 @@ The settings with names of the form `database.*` are for the database backend
...
+  qualifies as set or true in this case) to enable SSL connections. Do not
+  specify this parameter to connect over plaintext connection.
+  Note: This parameter is only supported for the MongoDB backend currently.
+* `database.login` and `database.password` are the login and password used to
+  authenticate to the database before performing any operations, specified in
+  plaintext.
+* `database.ca_cert`, `database.certfile`, `database.keyfile` and `database.crlfile` are the paths to the CA, signed certificate, private key and certificate revocation list files respectively.
+  Note: These parameters are only supported for the MongoDB backend currently.
+* `database.keyfile_passphrase` is the private key decryption passphrase, specified in plaintext.
+  Note: This parameter is only supported for the MongoDB backend currently.
```
krish7919: I have added a note that some params are only supported for the MongoDB backend.
We can modify docs when we begin to support both MongoDB and RethinkDB.

docs/server/source/server-reference/configuration.md
```diff
@@ -102,6 +110,18 @@ The settings with names of the form `database.*` are for the database backend
 * `database.replicaset` is only relevant if using MongoDB; it's the name of the MongoDB replica set, e.g. `bigchain-rs`.
 * `database.connection_timeout` is the maximum number of milliseconds that BigchainDB will wait before giving up on one attempt to connect to the database backend.
 * `database.max_tries` is the maximum number of times that BigchainDB will try to establish a connection with the database backend. If 0, then it will try forever.
+* `database.ssl` is a flag that determines if BigchainDB connects to the
+  backend database over SSL or not. This can be set to any value (the presence of a value
+  qualifies as set or true in this case) to enable SSL connections. Do not
+  specify this parameter to connect over plaintext connection.
+  Note: This parameter is only supported for the MongoDB backend currently.
+* `database.login` and `database.password` are the login and password used to
```
ttmc: I noticed that the default values of `database.login` and `database.password` are both `null`. Please document that that is possible and what it means.

docs/server/source/server-reference/configuration.md
```diff
@@ -102,6 +110,18 @@ The settings with names of the form `database.*` are for the database backend
 * `database.replicaset` is only relevant if using MongoDB; it's the name of the MongoDB replica set, e.g. `bigchain-rs`.
 * `database.connection_timeout` is the maximum number of milliseconds that BigchainDB will wait before giving up on one attempt to connect to the database backend.
 * `database.max_tries` is the maximum number of times that BigchainDB will try to establish a connection with the database backend. If 0, then it will try forever.
+* `database.ssl` is a flag that determines if BigchainDB connects to the
+  backend database over SSL or not. This can be set to any value (the presence of a value
+  qualifies as set or true in this case) to enable SSL connections. Do not
+  specify this parameter to connect over plaintext connection.
```
ttmc: I noticed that the default value of `ssl` is `false`. According to the above, "the presence of a value qualifies as set or true in this case" which just leaves me very confused; it implies that setting it to anything means TRUE (including setting it to `true`). I think what is needed here is some concrete examples of possible real-world values for this setting.

ttmc:
I just did `bigchaindb -y configure mongodb` on the `master` branch and got the following default values for the `database.*` configuration settings:
```js
    "database": {
        "ssl": false,
        "ca_cert": null,
        "replicaset": "bigchain-rs",
        "crlfile": null,
        "connection_timeout": 5000,
        "keyfile_passphrase": null,
        "backend": "mongodb",
        "port": 27017,
        "max_tries": 3,
        "host": "localhost",
        "keyfile": null,
        "name": "bigchain",
        "certfile": null,
        "login": null,
        "password": null
    }
```

Please add the new / missing ones to the **Default values** section (for MongoDB; the RethinkDB defaults aren't changed [I checked]).

I try to keep the parameters in the same order throughout, so please re-order them to be in the same order as the documentation.

docs/server/source/server-reference/configuration.md
krish7919: I was going through the code now, I was wrong about my understanding of the parsing process.
The `ssl` parameter can take the value `true` or `false`.
My statement is applicable when travis is run (or test are run locally).
I found a bug in my travis changes and have opened an issue for it. Will close it and then update the docs.

PR #1538, opened 2017-06-14
Feat/1441/remove host and port

### LINKED ISSUE #1465: hostname of websocket events api (streams_v1) link is incorrect
For instance, in the case below,   `'streams_v1'` should be : `'ws://bdb-mdb:9985/api/v1/streams/valid_tx'`. 

```python
>>> import requests

>>> requests.get('http://bdb-mdb:9984/api/v1/').json()
{'_links': {'docs': 'https://docs.bigchaindb.com/projects/server/en/v0.10.1/http-client-server-api.html',
  'self': 'http://bdb-mdb:9984/api/v1/',
  'statuses': 'http://bdb-mdb:9984/api/v1/statuses/',
  'streams_v1': 'ws://0.0.0.0:9985/api/v1/streams/valid_tx',
  'transactions': 'http://bdb-mdb:9984/api/v1/transactions/'}}
```

TimDaub:
Fixes #1441 
Maybe also fixes #1465 (@sbellem confirmed)

Docs:
docs


bigchaindb/web/views/base.py
```diff
@@ -21,14 +21,11 @@ def make_error(status_code, message=None):
...
-
-
 def base_ws_uri():
     """Base websocket uri."""
     # TODO Revisit as this is a workaround to address issue
     # https://github.com/bigchaindb/bigchaindb/issues/1465.
+    # NOTE: This is now only for the WS API. Env-variable should be renamed to
+    # e.g. WS_HOST or WEBSOCKET_HOST
     host = request.environ['HTTP_HOST'].split(':')[0]
     return 'ws://{}:{}'.format(host, config['wsserver']['port'])
```
krish7919: We shouldn't add the host and the port for websockets either.
We can document that users need to access WS services using the `ws://` prefix.

TimDaub: OK. Here is what I'd do (pls correct me @krish7919, @sbellem, @r-marques):

- rename `HTTP_HOST` to `WS_HOST`
- and

```
def base_ws_uri():
    return request.environ['WS_HOST']
```
(or just remove that function - in any case, base the streams base_uri on the env-variables.)
Makes sense?

r-marques: I agree with renaming the env and keeping the host and port for the events api.

krish7919: I would prefer that we remove `base_ws_uri()` completely and only give the relative that should be appended to `ws://host:post/`.

bigchaindb/web/views/base.py
```diff
@@ -25,7 +25,7 @@ def base_ws_uri():
     """Base websocket uri."""
     # TODO Revisit as this is a workaround to address issue
```
TimDaub: @sbellem: Can you confirm that #1535 appropriately addresses #1465, thx :)

TimDaub:
Codecov is failing b/c I removed a comment in 555e823, lol.

PR #1541, opened 2017-06-14
Remove relative links

### LINKED ISSUE #1525: Inconsistency in using relative links
See [/outputs endpoint](https://docs.bigchaindb.com/projects/server/en/master/http-client-server-api.html#get--api-v1-outputs?public_key=public_key)

```
[
  "../transactions/2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e/outputs/0",
"../transactions/2d431073e1477f3073a4693ac7ff9be5634751de1b8abaa1f4e19548ef0b4b0e/outputs/1"
]
```

[/status endpoint](https://docs.bigchaindb.com/projects/server/en/master/http-client-server-api.html#get--api-v1-statuses?tx_id=tx_id)

```
{
  "status": "valid",
  "_links": {
    "tx": "/transactions/38100137cea87fb9bd751e2372abb2c73e7d5bcf39d940a5516a324d9c7fb88d"
  }
}
```

notice `_links` being used here (both blocks and transactions).

This pattern is not repeated in `/assets`, the events API, `/blocks?tx_id`

r-marques:
resolves #1525 

- Outputs now return and object with `{'txid': 'aaa', 'output': 0}` instead of links
- Statuses endpoints only returns the status
- Remove the documented Location HTTP headers since they are not implemented

### DIFF
--- a/bigchaindb/web/views/outputs.py
+++ b/bigchaindb/web/views/outputs.py
@@ -24,5 +24,5 @@ class OutputListApi(Resource):
         with pool() as bigchain:
             outputs = bigchain.get_outputs_filtered(args['public_key'],
                                                     include_spent)
-            # NOTE: We pass '..' as a path to create a valid relative URI
-            return [u.to_uri('..') for u in outputs]
+            return [{'transaction_id': output.txid, 'output': output.output}
+                    for output in outputs]
--- a/docs/server/source/http-client-server-api.rst
+++ b/docs/server/source/http-client-server-api.rst
@@ -176,7 +176,7 @@ not already been spent.
    a base58 encoded ed25519 public key associated with transaction output
    ownership.
 
-   Returns a list of links to transaction outputs.
+   Returns a list of transaction outputs.
 
    :param public_key: Base58 encoded public key associated with output ownership. This parameter is mandatory and without it the endpoint will return a ``400`` response code.
    :param unspent: Boolean value ("true" or "false") indicating if the result set should be limited to outputs that are available to spend. Defaults to "false".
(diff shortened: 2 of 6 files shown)


bigchaindb/web/views/outputs.py
```diff
@@ -24,5 +24,5 @@ def get(self):
         with pool() as bigchain:
             outputs = bigchain.get_outputs_filtered(args['public_key'],
                                                     include_spent)
-            # NOTE: We pass '..' as a path to create a valid relative URI
-            return [u.to_uri('..') for u in outputs]
```
TimDaub: The [`to_uri` method](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/transaction.py#L196) might have only been used for this API endpoint.

As we're not using it here anymore, we might want to check if we cannot simply remove it now.

docs/server/source/http-client-server-api.rst
```diff
@@ -265,16 +267,10 @@ Statuses
 
    **Example response**:
 
-   .. literalinclude:: http-samples/get-statuses-block-invalid-response.http
```
TimDaub: Why remove?

bigchaindb/web/views/outputs.py
r-marques: It is still being used for instance to check that all the inputs in a transaction are unique. I guess the same could be achieve without it since the `TransactionLink` implements `__hash__` but that requires some refactor that was not in the scope of this pr

docs/server/source/http-client-server-api.rst
r-marques: Before if the block was invalid it only returned the `status`. If it was valid it would return the `status` and `_links`.
Now it always returns only the `status` key so I though the two example responses were redundant.

r-marques:
Remove the docs build folder and rebuild the docs. There should be no more `_links` in the documentation.

The test is being skipped so I did not change it.

PR #1543, opened 2017-06-14
Fix /transactions/id endpoint

TimDaub:
Fixes: #1038

As was summarized [here](https://github.com/bigchaindb/bigchaindb/issues/1038#issuecomment-307831678):

- /transactions/id only returns valid transactions
- Location header included in response
- Add strategy for polling or listening to notes

Docs:

- GET /transactions/id
tx_docs1

- POST /transactions response
tx_docs2

### DIFF
--- a/docs/server/source/http-client-server-api.rst
+++ b/docs/server/source/http-client-server-api.rst
@@ -46,12 +46,12 @@ Transactions
 
    Get the transaction with the ID ``tx_id``.
 
-   This endpoint returns a transaction if it was included in a ``VALID`` block,
-   if it is still waiting to be processed (``BACKLOG``) or is still in an
-   undecided block (``UNDECIDED``). All instances of a transaction in invalid
-   blocks are ignored and treated as if they don't exist. If a request is made
-   for a transaction and instances of that transaction are found only in
-   invalid blocks, then the response will be ``404 Not Found``.
+   This endpoint returns a transaction if it was included in a ``VALID`` block.
+   All instances of a transaction in invalid/undecided blocks or the backlog
+   are ignored and treated as if they don't exist. If a request is made for a
+   transaction and instances of that transaction are found only in
+   invalid/undecided blocks or the backlog, then the response will be ``404 Not
+   Found``.
 
    :param tx_id: transaction ID
    :type tx_id: hex string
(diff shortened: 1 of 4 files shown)


docs/server/source/http-client-server-api.rst
```diff
@@ -147,7 +147,17 @@ Transactions
    .. literalinclude:: http-samples/post-tx-response.http
       :language: http
 
+   .. note::
+       Is the server returning a ``202`` HTTP status code, then the transaction
```
ttmc: I think you meant to write, "If the server is returning a 202..."

TimDaub: thx, fixed

docs/server/source/http-client-server-api.rst
```diff
@@ -147,7 +147,17 @@ Transactions
...
+       has been accepted for processing. To check the status of the
+       transaction, poll the link to the `status monitor
+       <https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-statuses?tx_id=tx_id>`_
+       provided in the ``Location`` header or listen to server's `WebSocket
+       Event Stream API
+       <https://docs.bigchaindb.com/projects/server/en/latest/websocket-event-stream-api.html>`_.
```
ttmc: Let's try to avoid hard-coded URLs like this, when possible. Instead you can use something like:
```text
or listen to the server's :ref:`WebSocket Event Stream API <The WebSocket Event Stream API>`.
```

TimDaub: thx fixed

docs/server/source/http-client-server-api.rst
```diff
@@ -147,7 +147,16 @@ Transactions
    .. literalinclude:: http-samples/post-tx-response.http
       :language: http
 
+   .. note::
+       If the server is returning a ``202`` HTTP status code, then the
+       transaction has been accepted for processing. To check the status of the
+       transaction, poll the link to the `status monitor
+       <https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-statuses?tx_id=tx_id>`_
```
ttmc: I will push a commit to fix this hard-coded URL. It's probably not a good idea to use the title of that section as its reference, so I'll create a proper proper reference label instead.

TimDaub: thanks :)

PR #1552, opened 2017-06-16
Enable Auth over TLS connections

### LINKED ISSUE #1549: Version bigchiandb/mongodb container independent of the official mongodb container
Currently, I have versioned the bigchaindb/mongodb container as per the mongodb version, which is 3.4.4.
This is not a feasible versioning scheme if we are not following the mongodb release cycle.

Today, I had to update the containers after enabling auth. And I had to bump up the version number too.
But I cannot set it to 3.4.5 (as it hasn't been released yet).

In the future, we can look at the bigchaindb/mongodb Dockerfile and find out which version of MongoDB are we running.

Thoughts, @ttmc ?

krish7919:
Fixes #1549


bigchaindb/backend/mongodb/connection.py
```diff
@@ -138,7 +139,7 @@ def _connect(self):
         except (pymongo.errors.ConnectionFailure,
                 pymongo.errors.OperationFailure) as exc:
             logger.info('Exception in _connect(): {}'.format(exc))
-            if "Authentication fail" in str(exc):
```
r-marques: Did this changed or mongodb produces a different message for `SCRAM` and `X509`?
On my machine I do get `OperationFailure: Authentication failed.` when using `SCRAM` authentication

bigchaindb/backend/mongodb/connection.py
```diff
@@ -126,10 +128,9 @@ def _connect(self):
                                              ssl_pem_passphrase=self.keyfile_passphrase,
                                              ssl_crlfile=self.crlfile,
                                              ssl_cert_reqs=CERT_REQUIRED)
-
-            # authenticate with the specified user if the connection succeeds
-            if self.login is not None and self.password is not None:
-                client[self.dbname].authenticate(self.login, self.password)
+                if self.login is not None:
+                    client[self.dbname].authenticate(self.login,
+                                                     mechanism="MONGODB-X509")
```
r-marques: single quotes

bigchaindb/backend/mongodb/connection.py
```diff
@@ -175,16 +178,19 @@ def initialize_replica_set(host, port, connection_timeout, dbname, ssl, login,
                                        ssl_pem_passphrase=keyfile_passphrase,
                                        ssl_crlfile=crlfile,
                                        ssl_cert_reqs=CERT_REQUIRED)
+            if login is not None:
+                logger.info('Authenticating to the database...')
+                conn[dbname].authenticate(login, mechanism="MONGODB-X509")
```
r-marques: single quotes

bigchaindb/backend/mongodb/connection.py
```diff
@@ -175,16 +178,19 @@ def initialize_replica_set(host, port, connection_timeout, dbname, ssl, login,
...
 
     except (pymongo.errors.ConnectionFailure,
             pymongo.errors.OperationFailure) as exc:
+        logger.info('Exception in _connect(): {}'.format(exc))
+        if 'auth failed' in str(exc):
```
r-marques: same as https://github.com/bigchaindb/bigchaindb/pull/1552/files#r123485502

bigchaindb/backend/mongodb/connection.py
```diff
@@ -113,6 +113,8 @@ def _connect(self):
                                              replicaset=self.replicaset,
                                              serverselectiontimeoutms=self.connection_timeout,
                                              ssl=self.ssl)
+                if self.login is not None and self.password is not None:
```
ttmc: I wonder if we should raise an exception if they set one of `self.login` or `self.password`, but not the other?

docs/server/source/production-deployment-template/easy-rsa.rst
```diff
@@ -69,14 +69,17 @@ The comments in the file explain what each of the variables mean.
    echo 'set_var EASYRSA_REQ_OU "IT"' >> vars
    echo 'set_var EASYRSA_REQ_EMAIL "dev@bigchaindb.com"' >> vars
 
+We follow the convention of modifying the OU to ``ROOT-CA``,
+``MongoDB-Instance``, ``BigchainDB-Instance``, ``MongoDB-Mon-Instance`` and
+``MongoDB-Backup-Instance`` while issuing certificates.
```
ttmc: Why? I always thought OU meant the Organizational Unit within the company, such as "marketing" or "IT". Do we need to include information like this (e.g. "MongoDB-Instance") in the certificate? If so, is there another setting/value where we could store that?

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -487,7 +487,119 @@ Step 11: Start a Kubernetes StatefulSet for MongoDB
...
 
-Step 12: Start a Kubernetes Deployment for MongoDB Monitoring Agent
+Step 12: Configure Users and Access Control for MongoDB
+-------------------------------------------------------
+
+  * Create a user on MongoDB with authorization to create more users and assign
+    roles to them.
+    Note: You need to do this only when setting up the first MongoDB node of
+    the cluster.
+    Log in to the MongoDB instance and open a mongo shell using the certificates
```
ttmc: Maybe add a blank line before this sentence so it doesn't seem like it's part of the "Note:" (unless it is part of the "Note:").

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -487,7 +487,119 @@ Step 11: Start a Kubernetes StatefulSet for MongoDB
...
+           } ]
+         } )
+
+    The ``hostname`` in this case will be the value set in
+    ``mdb-instance-name`` in the ConfigMap.
+    For example, if the value set in the ``mdb-instance-name`` is
+    ``mdb-instance-0``, set the ``hostname`` above to the value ``mdb-instance-0``.
+  
+  * The instance should be voted as the ``PRIMARY`` in the replica set as this
+    is the only instance in the replica set. This can bbe observed by observing
```
ttmc: bbe

bigchaindb/backend/mongodb/connection.py
krish7919: Good point. I have only observed this message with X509 auth; I will test with SCRAM and compare.

krish7919: This is just for follow-up:
We get `Authentication failed` with plain SCRAM auth, and `Could not find user ...` if there is no user in the database as specified in `database.login` parameter, and `auth failed` if the presented certificate has a different name than the one in the database, and the one specified in `database.login`.

@r-marques and I consulted and have decided that we will remove the `AuthenticationError` in BigchainDB and just log the specific error, along with raising it again.

bigchaindb/backend/mongodb/connection.py
krish7919: It will still result in an error over at this stage and we will write the same checks that MongoDB driver has.

krish7919: Only `login` is required for X509, and both `login` and `password` are required for SCRAM.

docs/server/source/production-deployment-template/easy-rsa.rst
krish7919: The OUs need to be different because all the cluster members are in one OU.
This is how MongoDB can differentiate if one of the connections to it is from another cluster member or another client.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
krish7919: :+1:

PR #1555, opened 2017-06-16
Bug while running travis tests

krish7919:
This got left behind in the `travis_script.sh` file when working on the comments on PR #1510

### DIFF
--- a/.ci/travis_script.sh
+++ b/.ci/travis_script.sh
@@ -5,7 +5,7 @@ set -e -x
 if [[ -n ${TOXENV} ]]; then
   tox -e ${TOXENV}
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
-    "${BIGCHAINDB_DATABASE_SSL}" == false ]]; then
+    -z "${BIGCHAINDB_DATABASE_SSL}" ]]; then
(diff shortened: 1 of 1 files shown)


.ci/travis_script.sh
```diff
@@ -5,7 +5,7 @@ set -e -x
 if [[ -n ${TOXENV} ]]; then
   tox -e ${TOXENV}
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
-    "${BIGCHAINDB_DATABASE_SSL}" == false ]]; then
+    -z "${BIGCHAINDB_DATABASE_SSL}" ]]; then
```
ttmc: So `-z "${BIGCHAINDB_DATABASE_SSL}"` will be true if `BIGCHAINDB_DATABASE_SSL` is not set?

Bash is so strange. (This is called Bash bashing.)

r-marques: No. It will be true if `BIGCHAINDB_DATABASE_SSL` is an empty string

krish7919: Yes, link [here](http://tldp.org/LDP/abs/html/comparison-ops.html).

r-marques: Maybe just checking that the variable is not set would be a better option.

PR #1571, opened 2017-06-21
Updated crypto-conditions-related root docs

ttmc:
This PR updates the root docs related to crypto-conditions.
- No more hashlock/preimage condition
- No more timeout condition or escrow

The page about transaction concepts was also revised.
- Wrote more details about what is in the inputs and outputs (but it's a "concepts" page so the deep technical details were avoided). Linked to the docs page about **Inputs and Outputs**.
- Added links to some example transactions elsewhere in the docs.

### DIFF
--- a/docs/root/source/transaction-concepts.md
+++ b/docs/root/source/transaction-concepts.md
@@ -42,7 +42,7 @@ See BigchainDB Server [issue #626](https://github.com/bigchaindb/bigchaindb/issu
 
 A CREATE transaction must be signed by all the owners.
 (If you're looking for that signature,
-it's in the one "fulfillment" of the one input.)
+it's in the one "fulfillment" of the one input, albeit encoded.)
 
 
 ## TRANSFER Transactions
--- a/docs/root/source/transaction-concepts.md
+++ b/docs/root/source/transaction-concepts.md
@@ -37,7 +37,7 @@ Each output also has a list of all the public keys associated
 with the conditions on that output.
 Loosely speaking, that list might be interpreted as the list of "owners."
 A more accurate word might be fulfillers, signers, controllers,
-or tranfer-enablers.
+or transfer-enablers.
 See BigchainDB Server [issue #626](https://github.com/bigchaindb/bigchaindb/issues/626).
 
 A CREATE transaction must be signed by all the owners.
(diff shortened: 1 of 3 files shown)


docs/root/source/transaction-concepts.md
```diff
@@ -15,30 +15,57 @@ one might register an identity or a creative work. The things are often called
...
+BigchainDB supports a variety of conditions,
+a subset of the [Interledger Protocol (ILP)](https://interledger.org/)
+crypto-conditions. For details, see
+[the documentation about Inputs and Outputs](https://docs.bigchaindb.com/projects/server/en/latest/data-models/inputs-outputs.html).
+
+Each output also has a list of all the public keys associated
+with the conditions on that output.
+Loosely speaking, that list might be interpreted as the list of "owners."
+A more accurate word might be fulfillers, signers, controllers,
+or tranfer-enablers.
```
sbellem: tranfer --> transfer

docs/root/source/transaction-concepts.md
```diff
@@ -15,30 +15,57 @@ one might register an identity or a creative work. The things are often called
...
+See BigchainDB Server [issue #626](https://github.com/bigchaindb/bigchaindb/issues/626).
+
+A CREATE transaction must be signed by all the owners.
+(If you're looking for that signature,
+it's in the one "fulfillment" of the one input.)
```
sbellem: Perhaps we could somehow mention that the fulfillment is encoded, and hence the signature will not be visible? For instance, something along the lines:

Note that the fulfillment URI is encoded. One could to retrieve the signature, using a crypto-conditions library.

As an aside the fulfillment URI is encoded using the ASN.1 Distinguished Encoding Rules (DER), and the resulting bytes are encoded using `BASE64URL` as per [RFC 4648, Section 5](https://tools.ietf.org/html/rfc4648#section-5). More details at https://tools.ietf.org/html/draft-thomas-crypto-conditions-02#section-9

I am not sure how much details this section wishes to expose though.

docs/root/source/transaction-concepts.md
ttmc: Ooh, good catch! There are two other instances of that as well. I'll fix all three.

docs/root/source/transaction-concepts.md
ttmc: I'll mention that the signature is encoded (here in the root docs), but the technical details could be put somewhere else (e.g. in the server docs). I'll make a note of your comment here and look for a place to add that elsewhere.

PR #1572, opened 2017-06-21
Rabbitmq for events api

emotional-engineering:


ttmc:
Hi @emotional-engineering 

Before we can merge this pull request, we need you or your organization to agree to one of our contributor agreements. I could explain more about why, but I suspect that you're contributing on behalf of a company which has already agreed to one of our contributor agreements. If so, this code would already be covered by that agreement, but I need to confirm that my suspicion is correct.

Please email me at troy@bigchaindb.com to say if you're just contributing this code on your own, or if you're contributing on behalf of some company, and if so, the name of that company. (I will then have to confirm with that company.)

vrde:
I'll close this in favor or #1707.

Issue #1558, opened 2017-06-19
Using/ignoring the value of "version" in transactions for BDB Server version 1.0.0+


ttmc:
[The BigchainDB transaction model](https://docs.bigchaindb.com/projects/server/en/latest/data-models/transaction-model.html) contains a top-level key-value pair:
```json
    "version": "<version number of the transaction model>",
```

What values are allowed for "version"? The schema file at `bigchaindb/bigchaindb/common/schema/transaction.yaml` contains:
```yaml
  version:
    type: string
    pattern: "^0\\."
    description: |
        BigchainDB transaction schema version.
```

What will happen if someone puts "1.0" as the value for "version"?

Is the value of the "version" used by BigchainDB Server anywhere today, or is it just ignored?

If someone puts "2.0" as the value of "version" today (or with BigchainDB Server 1.0.0), will that value be ignored, so that their transaction might be considered valid today, but in the future, it won't be considered valid, once there is an actual version 2 of the transaction schema/model?

Maybe BigchainDB Server should explicitly require that the value of "version" be "1.0" starting with BigchainDB Server version 1.0.0?

ttmc:
Note: It looks like we'll be calling the version 1.0.0 release candidate version `1.0.0rc1` to comply with [the Python way of doing semantic versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme) (which agrees with [regular Semantic Versioning](http://semver.org/) except there's no hyphen between the 1.0.0 and the rc1, i.e. not `1.0.0-rc1`).

We need to make sure that won't break something, because we've never had a version name that looked like that before. For example, we've had `0.11.0.dev` not `0.11.0dev`.

libscott:
> Maybe BigchainDB Server should explicitly require that the value of "version" be "1.0" starting with BigchainDB Server version 1.0.0?

Yes, that's the idea. After 1.0, transaction model version will break step with server version.

TimDaub:
A quick task check list:

- [ ] Adjust javascript driver
- [ ] Adjust python driver
- [ ] Adjust CLI
- [x] Make change in BDB server #1574
- [ ] Add to changelog in #1561
- [ ] Add to upgrade-guide in #1547

ttmc:
[...]

PR #1574, opened 2017-06-22
tx version break step with server version

libscott:
Transaction now has its own version and it must be `1.0` to validate.

Part of the resolution of #1558

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -414,7 +414,7 @@ class Transaction(object):
                 ``id`` property.
             metadata (dict):
                 Metadata to be stored along with the Transaction.
-            version (int): Defines the version number of a Transaction.
+            version (string): Defines the version number of a Transaction.
     """
     CREATE = 'CREATE'
     TRANSFER = 'TRANSFER'
@@ -440,6 +440,7 @@ class Transaction(object):
                     lock.
                 metadata (dict): Metadata to be stored along with the
                     Transaction.
+                version (string): Defines the version number of a Transaction.
         """
         if operation not in Transaction.ALLOWED_OPERATIONS:
             allowed_ops = ', '.join(self.__class__.ALLOWED_OPERATIONS)
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -10,7 +10,6 @@ from bigchaindb.common.exceptions import (KeypairMismatchException,
                                           InvalidHash, InvalidSignature,
                                           AmountError, AssetIdMismatch)
 from bigchaindb.common.utils import serialize, gen_timestamp
-import bigchaindb.version
 
 
 class Input(object):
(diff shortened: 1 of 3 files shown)


tests/validation/test_transaction_structure.py
```diff
@@ -165,14 +165,7 @@ def test_high_amounts(create_tx):
...
-    short_ver = bigchaindb.version.__short_version__
-    assert create_tx.version == re.match(r'^(.*\d)', short_ver).group(1)
-
-    validate(create_tx)
-
-    # At version 1, transaction version will break step with server version.
-    create_tx.version = '1.0.0'
+    create_tx.version = '0.10'
+    validate_raises(create_tx)
+    create_tx.version = '110'
```
ttmc: What about a test to see if `'1.0'` is okay?

bigchaindb/common/transaction.py
```diff
@@ -421,7 +420,7 @@ class Transaction(object):
     TRANSFER = 'TRANSFER'
     GENESIS = 'GENESIS'
     ALLOWED_OPERATIONS = (CREATE, TRANSFER, GENESIS)
-    VERSION = '.'.join(bigchaindb.version.__short_version__.split('.')[:2])
+    VERSION = '1.0'
```
r-marques: Why do we need to change this. Wouldn't be enough to update `version.py`?

bigchaindb/common/transaction.py
```diff
@@ -441,7 +440,6 @@ def __init__(self, operation, asset, inputs=None, outputs=None,
                     lock.
                 metadata (dict): Metadata to be stored along with the
                     Transaction.
-                version (int): Defines the version number of a Transaction.
```
r-marques: This should not be removed but we could change the type of version to `str`

tests/validation/test_transaction_structure.py
libscott: implied by all the other tests

bigchaindb/common/transaction.py
libscott: Transaction version is breaking step with server version, so it makes sense to hardcode it in the code that generates the transactions. When the server version changes, the transaction version should not neccesarily change.

bigchaindb/common/transaction.py
libscott: Oops

tests/validation/test_transaction_structure.py
libscott: Added by popular demand

bigchaindb/common/transaction.py
r-marques: Up until now the transactions had the same version as the server.
This is saying that the transaction version and the server version should be two different things.
I don't necessarily disagree but this is a decision that affects all bigchaindb drivers

ttmc: Tim made a TODO list of all the things that must be done, including updating all the drivers: https://github.com/bigchaindb/bigchaindb/issues/1558#issuecomment-310317072

PR #1581, opened 2017-06-23
1.0

sbellem:


bigchaindb/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.11.0.dev'
-__short_version__ = '0.11.dev'
+__version__ = '1.0.0rc1'
+__short_version__ = '1.0rc1'
```
r-marques: Will we have any problems with readthedocs because of the `rc1`?

sbellem: Good question!

k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: bigchaindb
-        image: bigchaindb/bigchaindb:0.10.2
+        image: bigchaindb/bigchaindb:1.0.0rc1
```
krish7919: :+1:

PR #1547, opened 2017-06-15
[WIP] Upgrade guide BDB v0.10 -> v1.0

TimDaub:
WIP, @r-marques and me will add notes on all breaking changes.

PR #1589, opened 2017-06-26
Remove condition details signature, rename subfulfillments to subconditions

### LINKED ISSUE #1583: Remove "output.condition.details.signature" from transaction
Currently our transaction outputs have a `signature` field in the case of an ed25519 condition (see image). This is because it's a fulfillment type that's encoded to get the information neccesary to sign a payload, and the fulfillment may contain a signature.

The proposal is to remove that from the transaction model, since the signature field should never be populated.

@r-marques @diminator @tim @sbellem 

screenshot from 2017-06-23 11-40-49

libscott:
Also implementing condition details dict conversion, we no longer need Fulfillment.[to|from]_dict in cryptoconditions.

Fixes #1583 

TODO:
- [x] document in upgrade guide #1547 
- [x] Fix condition URI in Multiple New Owners docs section

### DIFF
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -160,7 +160,7 @@ def _fulfillment_from_details(data):
     """
     if data['type'] == 'ed25519-sha-256':
         public_key = base58.b58decode(data['public_key'])
-        return Ed25519Sha256(public_key)
+        return Ed25519Sha256(public_key=public_key)
 
     if data['type'] == 'threshold-sha-256':
         try:
--- a/bigchaindb/common/transaction.py
+++ b/bigchaindb/common/transaction.py
@@ -10,7 +10,7 @@ from bigchaindb.common.crypto import PrivateKey, hash_data
 from bigchaindb.common.exceptions import (KeypairMismatchException,
                                           InvalidHash, InvalidSignature,
                                           AmountError, AssetIdMismatch,
-                                          ThresholdDepthOverflow)
+                                          ThresholdTooDeep)
 from bigchaindb.common.utils import serialize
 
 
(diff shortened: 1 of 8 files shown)


TimDaub:
As always, this tickets requires changes on various components before it can be closed:

A quick task check list:

- [ ] Adjust javascript driver
- [ ] Adjust python driver
- [ ] Adjust CLI
- [x] Make change in BDB server (this PR here)
- [ ] Add to changelog in
- [ ] Add to upgrade-guide in

bigchaindb/common/transaction.py
```diff
@@ -125,11 +118,66 @@ def from_dict(cls, data):
...
+        return {
+            'type': 'threshold-sha-256',
+            'threshold': fulfillment.threshold,
+            'subfulfillments': subfulfillments,
+        }
+
+    raise UnsupportedTypeError(fulfillment.type_name)
+
+
+def _fulfillment_from_details(data, limit=10, depth=0):
```
TimDaub: @r-marques The current default limit for threshold is defined at depth 10 here.
Any take on if we should limit this? And if so, to what depth?

TimDaub: I'd rename this method to something that empathizes even more that we're taking a jsonized condition and loading it into Python memory as a "**condition** object"

libscott: @r-marques noted that this was an implementation detail and not neccesary, im going to give a non recursive implementation a shot

TimDaub: ok

bigchaindb/common/transaction.py
```diff
@@ -125,11 +118,63 @@ def from_dict(cls, data):
...
+def _fulfillment_from_details(data):
+    """
+    Load a fulfillment for a signing spec dictionary
+
+    Args:
+        data: tx.output[].condition.details dictionary
+    """
+    if data['type'] == 'ed25519-sha-256':
+        public_key = base58.b58decode(data['public_key'])
+        return Ed25519Sha256(public_key)
```
sbellem: `public_key` is required to be passed as a keyword argument

libscott: Danke!

libscott:
Good to go on merging?

tests/common/test_transaction.py
```diff
@@ -94,13 +98,17 @@ def test_output_serialization(user_Ed25519, user_pub):
 
 
 def test_output_deserialization(user_Ed25519, user_pub):
+    import base58
```
sbellem: This could be imported at the top since `base58` is not under test, and is a third party library that we assume to be stable.

sbellem:
> Good to go on merging?

Good to go for me. The `base58` imports in the tests could be imported at the top, as per PEP 8.

PR #1594, opened 2017-06-28
Document graphite configuration parameter

### LINKED ISSUE #1537: Document undocumented graphite.host config setting?
I used `bigchaindb -y configure mongodb` to generate a default configuration file (in `~/.bigchaindb`) and when I looked at the generated file, it contained:
```json
    "graphite": {
        "host": "localhost"
    }
```

This setting is not documented in [the BigchainDB Configuration Settings docs page](https://docs.bigchaindb.com/projects/server/en/latest/server-reference/configuration.html). Should it be?

Should that be a configuration setting at all? Can we hard-code the value as `localhost` and leave it up to benchmarkers to edit the BigchainDB Server code if they want to change it to something else? (I'm not sure. I'm just asking.)

I'm assigning this issue to @libscott because I believe he added this setting.

libscott:
Fixes #1537

### DIFF
--- a/docs/server/source/server-reference/configuration.md
+++ b/docs/server/source/server-reference/configuration.md
@@ -45,6 +45,7 @@ For convenience, here's a list of all the relevant environment variables (docume
 `BIGCHAINDB_DATABASE_KEYFILE`<br>
 `BIGCHAINDB_DATABASE_KEYFILE_PASSPHRASE`<br>
 `BIGCHAINDB_DATABASE_CRLFILE`<br>
+`BIGCHAINDB_GRAPHITE_HOST`<br>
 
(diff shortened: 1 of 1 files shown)


docs/server/source/server-reference/configuration.md
```diff
@@ -492,3 +493,28 @@ logging of the `core.py` module to be more verbose, you would set the
 ```
 
 **Defaults to**: `"{}"`
+
+
+## Graphite
```
ttmc: Can you change the structure of this subsection to be parallel to the subsection about wsserver.host and wsserver.port ?

e.g.
* It should all be under one heading of the form `## graphite.host`
* Start the JSON blocks with three backticks followed by `js`
* Have an example using the environment variable
* ... Basically follow the exact same structure as the subsection about wsserver.host and wsserver.port. You could copy and paste that subsection here and then just edit it.

libscott:
screenshot from 2017-06-28 13-29-25

PR #1598, opened 2017-06-28
More prod depl template copyedits

ttmc:
Fixes:

* Added missing `\` to the end of a line in the `az acs create` command
* Removed an unnecessary trailing comma from `secret.yaml`

Also:

* Added a subsection about where to find the 3scale credentials
* In `secret.yaml` comments: Clarified that the entire certificate chain (i.e. all intermediate certs in addition to the actual cert) is needed for `https-certs.cert.pem`
* Other minor clarifications.

### DIFF
--- a/docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
+++ b/docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
@@ -29,8 +29,8 @@ where all data values must be base64-encoded.
 This is true of all Kubernetes ConfigMaps and Secrets.)
 
 
-vars
-~~~~
+vars.mdb-instance-name and Similar
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Your BigchainDB cluster organization should have a standard way
 of naming instances, so the instances in your BigchainDB node
@@ -59,11 +59,14 @@ of all *other* nodes in your BigchainDB cluster
 * If you're deploying the first node in the cluster,
   the value should be ``""`` (an empty string).
 * If you're deploying the second node in the cluster,
-  the value should be one public key inside double quotes.
+  the value should be the BigchainDB public key of the first/original
+  node in the cluster, inside double quotes.
   For example,
   ``"EPQk5i5yYpoUwGVM8VKZRjM8CYxB6j8Lu8i8SG7kGGce"``
 * If there are two or more other nodes already in the cluster,
-  the value should be a colon-separated list of public keys
+  the value should be a colon-separated list
+  of the BigchainDB public keys
+  of those other nodes,
   inside double quotes.
   For example,
   ``"DPjpKbmbPYPKVAuf6VSkqGCf5jzrEh69Ldef6TrLwsEQ:EPQk5i5yYpoUwGVM8VKZRjM8CYxB6j8Lu8i8SG7kGGce"``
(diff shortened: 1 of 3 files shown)


docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
```diff
@@ -59,11 +59,14 @@ of all *other* nodes in your BigchainDB cluster
 * If you're deploying the first node in the cluster,
   the value should be ``""`` (an empty string).
 * If you're deploying the second node in the cluster,
-  the value should be one public key inside double quotes.
+  the value should be the BigchainDB public key of the first/original
+  node in the cluster, inside double quotes.
```
krish7919: Yaml doesn't care about double quotes. Maybe Kubernetes does.

docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
```diff
@@ -119,6 +122,31 @@ field (``bdb-certs.bdb-user``), i.e.
    emailAddress=dev@bigchaindb.com,CN=test-bdb-ssl,OU=BigchainDB-Instance,O=BigchainDB GmbH,L=Berlin,ST=Berlin,C=DE
 
 
+threescale-credentials.*
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete the ``threescale-credentials`` Secret if you're not using 3scale.
```
krish7919: `or leave it blank`, which is much simpler than asking to delete a formatted file as yaml.

k8s/configuration/secret.yaml
```diff
@@ -77,7 +77,7 @@ data:
   # Base64-encoded CA certificate (ca.crt)
   ca.pem: "<b64 encoded CA certificate>"
   # Base64-encoded CRL file
-  crlfile: "<b64 encoded CRL>",
+  crlfile: "<b64 encoded CRL>"
```
krish7919: :+1:

k8s/configuration/secret.yaml
```diff
@@ -94,8 +94,9 @@ type: Opaque
 data:
   # Base64-encoded HTTPS private key
   cert.key: "<b64 encoded HTTPS private key>"
-  # Base64-encoded HTTPS Signed Certificate or Certificate Chain
-  cert.pem: "<b64 encoded HTTPS Signed Certificate or Certificate Chain>"
+  # Base64-encoded HTTPS certificate chain
+  # including all intermediate certificates and the actual certificate
+  cert.pem: "<b64 encoded HTTPS certificate chain"
```
krish7919: It depends on the certificate provider whether you need the complete chain or not.
One of our certificate providers is not the root and hence we need a chain.
More info here: https://support.dnsimple.com/articles/what-is-ssl-certificate-chain/

krish7919: And since we are specifying that, we can also specify the order in which the chain can be formed.
Info here: https://www.digicert.com/ssl-support/pem-ssl-creation.htm under `Creating a .pem with the Entire SSL Certificate Trust Chain`

docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
ttmc: Okay I will remove the "inside double quotes" prepositional phrase (both instances).

docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
ttmc: Okay I added a note about that option.

k8s/configuration/secret.yaml
ttmc: Oh interesting! I guess the order of the certs does matter... I'll add some docs about that.

PR #1610, opened 2017-06-29
Add documentation on exposed ports in k8s/nginx

krish7919:
While fixing this, I also realized that the parameter names (or port names in the *-svc for both nginx and nginx-3scale) are not consistent. I will open an issue for consistent naming and fix it once we are done with your testing, as we need to change names in quite a few places.

Opened #1611.

### DIFF
--- a/docs/server/source/production-deployment-template/node-on-kubernetes.rst
+++ b/docs/server/source/production-deployment-template/node-on-kubernetes.rst
@@ -128,7 +128,16 @@ Step 4.1: Vanilla NGINX
      the ConfigMap followed by ``-dep``. For example, if the value set in the
      ``ngx-instance-name`` is ``ngx-instance-0``, set  the
      ``spec.selector.app`` to ``ngx-instance-0-dep``.
-   
+     
+   * Set ``ngx-public-mdb-port`` to 27017, or the port number on which you
+     want to expose MongoDB service.
+
+   * Set ``ngx-public-api-port`` to 80, or the port number on which you want to
+     expose BigchainDB API service.
+
+   * Set ``ngx-public-ws-port`` to 81, or the port number on which you want to
+     expose BigchainDB Websocket service.
+     
    * Start the Kubernetes Service:
 
      .. code:: bash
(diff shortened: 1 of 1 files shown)


docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -128,7 +128,16 @@ Step 4.1: Vanilla NGINX
      the ConfigMap followed by ``-dep``. For example, if the value set in the
      ``ngx-instance-name`` is ``ngx-instance-0``, set  the
      ``spec.selector.app`` to ``ngx-instance-0-dep``.
-   
+     
+   * Set ``ngx-public-mdb-port`` to 27017, or the port number on which you
+     want to expose MongoDB service.
```
ttmc: Should they be changing `port` or `targetPort`? Maybe add a clarifying note.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -152,6 +161,18 @@ Step 4.2: OpenResty NGINX + 3scale
      ``ngx-instance-name`` is ``ngx-instance-0``, set  the
      ``spec.selector.app`` to ``ngx-instance-0-dep``.
    
+   * Set ``ngx-public-mdb-port`` to 27017, or the port number on which you
+     want to expose MongoDB service.
```
ttmc: Maybe add a clarifying note here as well, just like in the vanilla NGINX case.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
ttmc: It's not technically correct to write it as `ngx-public-mdb-port.port` because the YAML file hierarchy looks like this:

```yaml
spec:
  selector:
    app: ngx-instance-0-dep
  ports:
  - port: 27017
    targetPort: 27017
    name: ngx-public-mdb-port
    protocol: TCP
```

However, maybe it's clear what you mean, even though it's not 100% accurate. What do you think?

krish7919: Even I couldn't find a better way to write that. And that made sense intuitively.

ttmc:
Feel free to merge.

PR #1612, opened 2017-06-29
Do not allow duplicate GENESIS block creation - #1556

### LINKED ISSUE #1556: Handle case when a user posts a transaction with 'operation'=='GENESIS'
I discussed this with Rodolphe and apparently if an external user posts a CREATE transaction with 'operation'=='GENESIS', it will pass through all the validations and be accepted as valid. He suggested that the validation could add a check like the following:  

if operation is GENESIS, then check if a GENESIS block already exists, and if one does, then the transaction is invalid.

Note: A GENESIS block is just a block containing a single GENESIS transaction, created by building a CREATE transaction (with one owner, the node creating the transaction) but with operation changed to 'GENESIS' etc. See [the code for `prepare_genesis_block()` in core.py](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/core.py).

krish7919:
Resolves #1556

### DIFF
--- a/bigchaindb/pipelines/block.py
+++ b/bigchaindb/pipelines/block.py
@@ -85,7 +85,7 @@ class BlockPipeline:
             tx.validate(self.bigchain)
             return tx
         except ValidationError as e:
-            logger.error('Invalid tx: %s', e)
+            logger.warning('Invalid tx: %s', e)
             self.bigchain.delete_transaction(tx.id)
             return None
 
--- a/bigchaindb/pipelines/block.py
+++ b/bigchaindb/pipelines/block.py
@@ -74,10 +74,13 @@ class BlockPipeline:
 
         # If transaction is not valid it should not be included
         try:
+            # Do not allow an externally submitted GENESIS transaction.
+            # A simple check is enough as a pipeline is started only after the
+            # creation of GENESIS block, or after the verification of a GENESIS
+            # block. Voting will fail at a later stage if the GENESIS block is
+            # absent.
             if tx.operation == Transaction.GENESIS:
-                blocks_count = backend.query.count_blocks(self.bigchain.connection)
-                if blocks_count:
-                    raise GenesisBlockAlreadyExistsError('Duplicate GENESIS transaction')
+                raise GenesisBlockAlreadyExistsError('Duplicate GENESIS transaction')
 
             tx.validate(self.bigchain)
             return tx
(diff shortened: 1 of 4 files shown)


libscott:
LGTM

bigchaindb/pipelines/block.py
```diff
@@ -73,10 +74,15 @@ def validate_tx(self, tx):
 
         # If transaction is not valid it should not be included
         try:
+            if tx.operation == Transaction.GENESIS:
+                blocks_count = backend.query.count_blocks(self.bigchain.connection)
```
r-marques: I don't think there is a need for this check. When the pipelines are started they expect that bigchaindb was already initialized or else it will fail.

krish7919: So you mean do not go to the db to check for a genesis block?

r-marques: Yes. When the pipeline is started the genesis block already needs to be present.
Things like this https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/pipelines/vote.py#L175 would fail if the genesis block is not there.

ttmc: I would replace the trip to the database with a line or two of comments based on what Rodolphe just wrote, so it's clear to the reader why the trip isn't necessary.

krish7919: @r-marques @ttmc Changed in latest commit.

bigchaindb/pipelines/block.py
```diff
@@ -73,10 +74,18 @@ def validate_tx(self, tx):
...
+            # block. Voting will fail at a later stage if the GENESIS block is
+            # absent.
+            if tx.operation == Transaction.GENESIS:
+                raise GenesisBlockAlreadyExistsError('Duplicate GENESIS transaction')
+
             tx.validate(self.bigchain)
             return tx
         except ValidationError as e:
-            logger.warning('Invalid tx: %s', e)
+            logger.error('Invalid tx: %s', e)
```
r-marques: Maybe I am just being picky here but I don't think this should be treated as an error since no error as actually occurred. An invalid transaction was submitted and was correctly handled by the pipeline.

krish7919: :+1: 
This change came in while I was testing log levels! This has to be undone.

PR #1619, opened 2017-06-30
Secure WebSocket Support

krish7919:
Since we are dependent on both websockets and Openresty (for HTTPS and integration with 3scale), and since Openresty isn't updated at a higher velocity, we have decided to move ahead with the plan of running 2 NGINX instances/containers:
1. Openresty service to handle the BigchainDB HTTP API with authentication (using 3scale)
2. NGINX service to support secure websockets (`wss://`).


k8s/nginx-ws/container/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+LABEL maintainer "dev@bigchaindb.com"
+WORKDIR /
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get autoremove \
+    && apt-get clean
+COPY nginx_ws.conf.template /etc/nginx/nginx.conf
+COPY nginx_ws_entrypoint.bash /
+EXPOSE 81 444 9985
+ENTRYPOINT ["/nginx_ws_entrypoint.bash"]
```
libscott: shouldnt that be `.sh`?

krish7919: @libscott It's opinionated, and differs from person to person, and team to team. A filename is just a name under *nix.
Personally, I use it so that a user/engineer can easily find at a glance (using `ls` or something) that it's a bash script, rather than open the file to read the first line. Truly portable scripts can use `.sh`, but I would still want to show what a file contains from a name.

Also, that is the convention followed in all the other modules under `k8s/` now, and I would prefer to stick with it. :)

k8s/dev-setup/bigchaindb-dep.yaml
```diff
@@ -0,0 +1,73 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: bdb-dep
+    spec:
+      terminationGracePeriodSeconds: 10
+      containers:
+      - name: bigchaindb
+        image: bigchaindb/bigchaindb:1.0.0rc1
```
ttmc: I guess this can be upgraded to 1.0.0 now.

PR #1637, opened 2017-07-05
Changed to version 1.0.1.dev in version.py

ttmc:


ttmc:
Never mind, I will create a different branch and PR that uses `1.1.0.dev` and `1.1.dev` instead.

PR #1638, opened 2017-07-05
MongoDB socket timeout to handle master re-election

### LINKED ISSUE #1627: If the primary node of mongodb replica set is power off, the run_changefeed(conn, table, last_ts) function is hung on
When I power off the primary node of mongodb (not command "reboot" or "shutdown"), the run_changefeed function is hung on the line "record = cursor.next()" because the socket is not set to timeout mode. Default read mode is to read from primary node. If the bug occurs, the remaining nodes can not vote for new blocks because they cannot read from the mongodb by run_changefeed function.

The solution is:
```python
            client = pymongo.MongoClient(self.host,
                                         self.port,
                                         replicaset=self.replicaset,
                                         serverselectiontimeoutms=self.connection_timeout,
                                         ssl=self.ssl,
                                         socketTimeoutMS=20000, #add this
                                         readPreference='nearest') #add this
```
if socketTimeoutMS is set, when the primary node is powered off, after socketTimeoutMS time, the run_changefeed function is going on again.
readPreference is set to improve the performance.

libscott:
Fixes #1627

### DIFF
--- a/bigchaindb/backend/mongodb/connection.py
+++ b/bigchaindb/backend/mongodb/connection.py
@@ -111,7 +111,8 @@ class MongoDBConnection(Connection):
                                              self.port,
                                              replicaset=self.replicaset,
                                              serverselectiontimeoutms=self.connection_timeout,
-                                             ssl=self.ssl)
+                                             ssl=self.ssl,
+                                             **MONGO_OPTS)
                 if self.login is not None and self.password is not None:
                     client[self.dbname].authenticate(self.login, self.password)
             else:
@@ -126,7 +127,8 @@ class MongoDBConnection(Connection):
                                              ssl_keyfile=self.keyfile,
                                              ssl_pem_passphrase=self.keyfile_passphrase,
                                              ssl_crlfile=self.crlfile,
-                                             ssl_cert_reqs=CERT_REQUIRED)
+                                             ssl_cert_reqs=CERT_REQUIRED,
+                                             **MONGO_OPTS)
                 if self.login is not None:
                     client[self.dbname].authenticate(self.login,
                                                      mechanism='MONGODB-X509')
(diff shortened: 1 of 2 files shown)


libscott:
It's worth noting that I observed a strange stack overflow while running the test suite, but could not reproduce it a second time.

[stack overflow.txt](https://github.com/bigchaindb/bigchaindb/files/1124771/stack.overflow.txt)

bigchaindb/backend/mongodb/connection.py
```diff
@@ -143,6 +145,11 @@ def _connect(self):
             raise ConfigurationError from exc
 
 
+MONGO_OPTS = {
+    'socketTimeoutMS': 20000,
+}
```
krish7919: Would it help to add `readPreference='nearest'` as the user suggested.
This might let the writes go to the primary and reads to the local MongoDB instance in the replica set.
Helps with decentralization and performance too. Thoughts, @libscott @r-marques ?

r-marques: I'm afraid of possible side effects since the secondary replicas are always lagging behind the primary.

libscott:
I've also added a fix for the stack overflow error I encountered, I did not track down the reason (I think the sys.excinfo got cleared or something) but it had something to do with hitting the max stack depth so I removed this possibility.

bigchaindb/common/transaction.py
```diff
@@ -155,26 +155,26 @@ def _fulfillment_to_details(fulfillment):
     raise UnsupportedTypeError(fulfillment.type_name)
 
 
-def _fulfillment_from_details(data):
+def _fulfillment_from_details(data, _depth=0):
```
r-marques: Shouldn't the same be done to `_fulfillment_to_details`?

libscott: There isn't an instance where converting a fulfillment to details would crash the system, since all transactions enter in a serialized form and are deserialised. It's only the other way around when building transactions, ie in the driver, and I'm not very worried about the use case of making a 100 deep threshold condition.

bigchaindb/backend/mongodb/connection.py
libscott: Ignore until specific needs come up.

PR #1654, opened 2017-07-11
Wrote more about how to calculate a fulfillment string

### LINKED ISSUE #1615: Document some details about how "fulfillment" is encoded?
See the comment by @sbellem in https://github.com/bigchaindb/bigchaindb/pull/1571#pullrequestreview-47158159

ttmc:
Resolves #1615 

@sbellem I am putting you as a reviewer because this is based on a comment you made on a previous PR:  https://github.com/bigchaindb/bigchaindb/pull/1571#discussion_r124830052

### DIFF
--- a/docs/server/source/data-models/inputs-outputs.rst
+++ b/docs/server/source/data-models/inputs-outputs.rst
@@ -17,7 +17,7 @@ An input has the following structure:
 
    {
        "owners_before": ["<The public_keys list in the output being spent>"],
-       "fulfillment": "<Fulfillment URI fulfilling the condition of the output being spent>",
+       "fulfillment": "<String that fulfills the condition in the output being spent>",
        "fulfills": {
            "output_index": "<Index of the output being spent (an integer)>",
            "transaction_id": "<ID of the transaction containing the output being spent>"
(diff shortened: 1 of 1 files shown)


docs/server/source/data-models/inputs-outputs.rst
```diff
@@ -28,10 +28,22 @@ You can think of the ``fulfills`` object as a pointer to an output on another tr
 A CREATE transaction should have exactly one input. That input can contain one or more ``owners_before``, a ``fulfillment`` (with one signature from each of the owners-before), and the value of ``fulfills`` should be ``null``). A TRANSFER transaction should have at least one input, and the value of ``fulfills`` should not be ``null``.
 See the reference on :ref:`inputs <Input>` for more description about the meaning of each field.
 
-To calculate a fulfillment URI, you can use one of the
+The ``fulfillment`` string fulfills the condition in the output that is being spent (transferred).
+To calculate it:
+
+1. Determine the fulfillment as per the `Crypto-Conditions spec (version 02) <https://tools.ietf.org/html/draft-thomas-crypto-conditions-02>`_.
```
sbellem: We should point to version 03 (https://tools.ietf.org/html/draft-thomas-crypto-conditions-03) now since version 02 has expired.

If you prefer we can treat this in a separate ticket as they may be other places where the change is needed.

The changes in version 03 are minor and we will support/enforce them, so there is no concern there and we can safely point to version 03.

The changes are mostly concerned with a stricter requirement for the ordering of subtypes in the condition URI.

Issue #1655, opened 2017-07-11
[k8s] MongoDB StatefulSet hitting memory limit, so k8s restarts it. What to do?


ttmc:
We noticed that the memory limit was being reached sometimes and so Kubernetes was restarting the MongoDB StatefulSet. The file where the memory limit is set is:

`bigchaindb/k8s/mongodb/mongo-ss.yaml`

The current memory limit is 768 Mi.

krish7919:
@muawiakh wanna take a stab at this?

krish7919:
Logs to support this conclusion:

```text

33956s random time.
Jul 10 16:16:10 k8s-agent-B14C0819-0 systemd[1]: apt-daily.timer: Adding 10h 27min 22.534970s random time.
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel: conn5 invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=-998
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel: conn5 cpuset=99e74377fb675240bd7a062aade2f910d43f1cfda2636f3634d655e52e707a26 mems_allow
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel: CPU: 1 PID: 42897 Comm: conn5 Not tainted 4.4.0-81-generic #104-Ubuntu
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel: Hardware name: Microsoft Corporation Virtual Machine/Virtual Machine, BIOS 090006  05/23
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel:  0000000000000286 000000008eaa634e ffff8801334dfc88 ffffffff813f9683
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel:  ffff8801334dfd68 ffff880139532a00 ffff8801334dfcf8 ffffffff8120b7ae
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel:  ffffffff810ac1a9 ffff8801334dfcc8 ffffffff8119251b ffff880293d50000
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel: Call Trace:
Jul 10 16:16:12 k8s-agent-B14C0819-0 kernel:  [<ffffffff813f9683>] dump_stack+0x63/0x90
[...]

PR #1663, opened 2017-07-12
Fix for MongoDB hitting memory limit on k8s

muawiakh:
- MongoDB StatefulSet hitting memory limit, so k8s restarts it.
We have had multiple instances of restarts lately.
-  Changing it to 3.5 GB, data and reasoning to back it up
is mentioned in the ticket #1655

### DIFF
--- a/k8s/mongodb/mongo-ss.yaml
+++ b/k8s/mongodb/mongo-ss.yaml
@@ -72,7 +72,7 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: 768Mi
+            memory: 3.5G
         livenessProbe:
           tcpSocket:
             port: mdb-port


k8s/mongodb/mongo-ss.yaml
```diff
@@ -72,7 +72,7 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: 768Mi
+            memory: 3.5G
```
krish7919: Two points:
1. I referred to the links given below, and none of them specify memory as `X.XG`. I think the only unit supported is Mi (maybe Gi) in Kubernetes. Please verify this.
https://kubernetes.io/docs/tasks/administer-cluster/cpu-memory-limit/#enforcing-limits-at-point-of-creation
https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run
https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/

2. The section on what happens if one does not specify memory or cpu limits is that the pod can use as much resource as possible. So, I am thinking if it wouldn't be better to just remove this configuration, and then open a tocket to ensure that we put in scheduling constraints so that Mongo is not co-located with Nginx or BigchainDB.
https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/#if-you-dont-specify-limits-or-requests

ttmc: This Kubernetes docs page says that G and Gi are both okay:

https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/#cpu-and-ram-units

muawiakh: Exactly, it accepts both. This is the guide I referred to before pushing the change.

ttmc: Suggestion 2 from @krish7919 sounds reasonable. I've never used k8s scheduling constraints but if they can in fact keep Mongo by itself on an agent, that would be cool.

muawiakh: @krish7919 I am bumping up the RAM in this change so things don't break in the near future. We discussed that we can create/track another ticket where we can evaluate the resource consumption, minimum requirements for all the services i.e. mongodb, bigchaindb, nginx instances etc and update the pods and their configuration/scheduling accordingly.

PR #1676, opened 2017-07-17
Remove 3scale port number from configs

krish7919:
As we no longer use 3scale GET requests to test the setup, we can safely remove this configuration from our setup.

In case we need to check if 3scale has been configured correctly, the best way today would be to send some `POST` requests to our HTTP API and check the error message, if any.

We can think about including this code in case we decide to meter `GET`s in the future.

### DIFF
--- a/docs/server/source/production-deployment-template/node-on-kubernetes.rst
+++ b/docs/server/source/production-deployment-template/node-on-kubernetes.rst
@@ -173,18 +173,6 @@ Step 4.2: NGINX with HTTPS + 3scale
      ``public-mdb-port`` in the file which specifies where MongoDB is
      available.
 
-   * Set ``ports[2].port`` and ``ports[2].targetPort`` to the value set in the
-     ``threescale-api-port`` in the ConfigMap above. This is the
-     ``public-threescale-port`` in the file which specifies where OpenResty is
-     available.
-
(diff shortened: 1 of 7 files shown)


docs/server/source/production-deployment-template/node-config-map-and-secrets.rst
```diff
@@ -29,6 +29,39 @@ where all data values must be base64-encoded.
 This is true of all Kubernetes ConfigMaps and Secrets.)
 
 
+
+vars.cluster-fqdn
+~~~~~~~~~~~~~~~~~
+
+The ``cluster-fqdn`` field specifies the domain you would have
+:ref:`registered before <2. Register a Domain and Get an SSL Certificate for It>`.
```
muawiakh: nit: 'it' instead of 'It'

muawiakh: Can be addressed later.

ttmc:
If we want to go back to metering GETs in the future, we can always go back in the Git history, since we had that before.

krish7919:
> If we want to go back to metering GETs in the future, we can always go back in the Git history, since we had that before.

That's precisely why this is a separate PR and not clubbed with other PRs.

krish7919:
@ttmc Yup. We are good to go wrt grep.

PR #1681, opened 2017-07-17
K8s Service and Deployment for ipdb-proxy

krish7919:


krish7919:
As per discussion with @kremalicious and @vrde, we will stick to a non-kubernetes deployment for now, and think about moving to a Kubernetes deployment in the future.

PR #1699, opened 2017-07-25
Remove an invalid message in the k8s production deployment guide

muawiakh:
- The guide is outdated and expects and error message when the following command is run on the toolbox container in the nginx with https+3scale integration section:
$ curl -X GET https://<cluster-fqdn>
The above command should not return an error since this workflow is supported

### DIFF
--- a/docs/server/source/production-deployment-template/node-on-kubernetes.rst
+++ b/docs/server/source/production-deployment-template/node-on-kubernetes.rst
@@ -866,7 +866,7 @@ To test the NGINX instance with HTTPS and 3scale integration:
 
    $ wsc -er wss://<cluster-fqdn>/api/v1/streams/valid_transactions
 
-   $ curl -X GET https://<cluster-fqdn>
+   $ curl -X GET http://<cluster-fqdn>:27017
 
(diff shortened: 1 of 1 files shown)


muawiakh:
fixes #1699

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -866,7 +866,7 @@ To test the NGINX instance with HTTPS and 3scale integration:
 
    $ wsc -er wss://<cluster-fqdn>/api/v1/streams/valid_transactions
 
-   $ curl -X GET https://<cluster-fqdn>
+   $ curl -X GET http://<cluster-fqdn>:27017
```
ttmc: Now that we're using MongoDB authentication for all communications with MongoDB, does MongoDB still respond with the "It looks like you are trying to access MongoDB over HTTP..." message, or does it say something about authentication being required?

krish7919: It just gives the same error message.

PR #1700, opened 2017-07-26
Update SemVer policy

vrde:
*My editor removes trailing whitespaces, so this PR contains other changes. If you prefer I can disable that feature from my editor and commit again.*

This PR is related to the (relatively long) discussion in #1644. IMO we are having this discussion because we didn't properly define what our **Public API** is.

The following paragraph is taken from the [SemVer 2 introduction](http://semver.org/spec/v2.0.0.html#introduction) (emphasis mine)
> As a solution to this problem, I propose a simple set of rules and requirements that dictate how version numbers are assigned and incremented. These rules are based on but not necessarily limited to pre-existing widespread common practices in use in both closed and open-source software. **For this system to work, you first need to declare a public API. This may consist of documentation or be enforced by the code itself. Regardless, it is important that this API be clear and precise. Once you identify your public API, you communicate changes to it with specific increments to your version number.** Consider a version format of X.Y.Z (Major.Minor.Patch). Bug fixes not affecting the API increment the patch version, backwards compatible API additions/changes increment the minor version, and backwards incompatible API changes increment the major version.

We never defined this in a **clear and precise** way. This PR is to fix it. Of course this is **my** view on what we should include and exclude form our **public API**. I'd :heart: to have feedback from as many people as possible: those few lines of **human readable text** will define the public interface of BigchainDB.

Please note that I've removed the reference to the *Python form of Semantic Versioning*. We are not developing a Python package, but a server. (We might want to reference that anyway, since the server package is stored in pypi).

Waiting for your comments!
giphy


CHANGELOG.md
```diff
@@ -1,7 +1,20 @@
...
 
 All _notable_ changes to this project will be documented in this file (`CHANGELOG.md`).
-This project adheres to [the Python form of Semantic Versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme) (or at least we try).
+
+This project adheres to [Semantic Versioning 2](http://semver.org/spec/v2.0.0.html), the BigchainDB **public API**:
+* Includes:
+  * This definition of **public API**.
+  * The Command Line Interface.
+  * The transaction schema.
+  * All the endpoints exposed by the HTTP API (including payloads sent to and received from them).
```
r-marques: What is the difference between this and **public API**

CHANGELOG.md
```diff
@@ -1,7 +1,20 @@
...
+  * The docker setup.
+* Excludes:
```
r-marques: I agree with this. In the future we may move some of these things to the `Includes` section.

I like the notion of `experimental` since we can have our users testing some features before we make them permanent. We should however have a better definition of  what an `experimental` feature is:
- How to we mark a feature as experimental
- We need to make sure that an experimental feature if in a stable version does not break the server

CHANGELOG.md
vrde: It's recursive, the idea is that the definition of **public API** in `CHANGELOG.md` is part of the **public API** itself. If we don't have this rule, then someone can change **what we mean** as **public API** and mess everything up.

r-marques:
I love the idea of being more explicit in to what the semver includes and doesn't include

CHANGELOG.md
vrde: Good point! I guess we should define what `experimental` means somewhere else in the docs. We can have a keyword (maybe `experimental` itself?) we use as a prefix for API endpoints (like `/experimental/whatever`), for CLI flags (like `--experimental-whatever`), configuration (like `{"experimental": {"key": "value"}`, or `{"experimental-key": "value"}`).

muawiakh: +1, but I think we should keep experimental features as part of this document too. e.g. adding experimental support for feature-X. Under the release tag.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All _notable_ changes to this project will be documented in this file (`CHANGELO
 This project adheres to [Semantic Versioning 2](http://semver.org/spec/v2.0.0.html), the BigchainDB **public API**:
 * Includes:
   * This definition of **public API**.
+  * The configuration schema, and the path where the configuration is stored.
```
r-marques: 👍

CHANGELOG.md
```diff
@@ -1,7 +1,21 @@
...
 All _notable_ changes to this project will be documented in this file (`CHANGELOG.md`).
-This project adheres to [the Python form of Semantic Versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme) (or at least we try).
+
+This project adheres to [Semantic Versioning 2](http://semver.org/spec/v2.0.0.html), the BigchainDB **public API**:
+* Includes:
+  * This definition of **public API**.
+  * The configuration defaults, schemas, and paths where configuration files are stored.
+  * The Command Line Interface.
+  * The transaction schema.
+  * All the endpoints exposed by the HTTP API (including payloads sent to and received from them).
```
TimDaub: Might want to mention the WebSocket Event Stream API here explicitly as well.

We defined it here as well: https://github.com/bigchaindb/bigchaindb/pull/1547/files#diff-57fcbb9dfad73b9008ffed4bbbb799ddR10

CHANGELOG.md
ttmc: On a different note, there's a slight difference between [Python's version of semantic versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme) and [the standard at semver.org](http://semver.org/spec/v2.0.0.html). In particular, the version labels used for release candidates are slightly different. This matters for things like pip/PyPI, which expect the Python conventions.

TimDaub: I think there is even a distinction between `experimental` and `external`.

We might want to never provide backwards compatibility to some of our components, but these could be fairly stable (the opposite of `experimental`).

E.g. As we see transactions and assets as the main interaction paradigm of BigchainDB, we could view blocks and votes as `external`, hence not give backwards compatibility guarantees regarding semver.
We can't really make that change now anymore however, as we've already promised backwards-compatibility to our data models in general: https://github.com/bigchaindb/bigchaindb/pull/1547/files#diff-57fcbb9dfad73b9008ffed4bbbb799ddR10

vrde:
@TimDaub thanks for pointing out previous work around this issue :clap:, I'll add the WebSocket Event Stream API to this PR.

CHANGELOG.md
vrde: > As we see transactions and assets as the main interaction paradigm of BigchainDB, we could view blocks and votes as external, hence not give backwards compatibility guarantees regarding semver.

IMO changing those models breaks compatibility:
* If someone is building a *bigchaindb explorer* they will use those models.
* Changing those models means to migrate the DB and—I guess—all the nodes should "vote again".

If this is what we want, then we might explicitly add those models to this PR, even if they are included in the sentence:
> All the endpoints exposed by the HTTP API (including payloads sent to and received from them)

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
 # Change Log (Release Notes)
 
 All _notable_ changes to this project will be documented in this file (`CHANGELOG.md`).
-This project adheres to [the Python form of Semantic Versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme) (or at least we try).
+
+This project adheres to [Semantic Versioning 2](http://semver.org/spec/v2.0.0.html), the BigchainDB **public API**:
```
ttmc: There are some small differences between [Semantic Versioning 2](http://semver.org/spec/v2.0.0.html) and [Python's form of Semantic Versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme). The main difference is in the version label string (point 9 in Semantic Versioning). This matters to pip and PyPI.

Note: I thought I made this comment already yesterday but for some reason it didn't get saved.

CHANGELOG.md
TimDaub: > IMO changing those models breaks compatibility:

Fully agreed. Let's not change them any time soon. We've promised to keep the data models backwards compatible already. The granularity we described above however can be subject to a discussion once we release 2.0

CHANGELOG.md
ttmc: My proposal is to change it back to saying that this project adheres to the Python form of Semantic Versioning.

PR #1703, opened 2017-07-28
Support for advertised host, port, scheme for Websockets

### LINKED ISSUE #1626: WSSERVER_EXTERNAL_HOST setting
Currently we have this:
```json
{
"api": {
"v1": {
"assets": "/api/v1/assets/",
"docs": "https://docs.bigchaindb.com/projects/server/en/v1.0.0rc1/http-client-server-api.html",
"outputs": "/api/v1/outputs/",
"statuses": "/api/v1/statuses/",
"streams": "ws://0.0.0.0:9985/api/v1/streams/valid_transactions",
"transactions": "/api/v1/transactions/"
}
},
"docs": "https://docs.bigchaindb.com/projects/server/en/v1.0.0rc1/",
"keyring": [],
"public_key": "983tiwtPKXrcWSkDTMiGCNoYAMpSgW5PV2KHqvu12Zsh",
"software": "BigchainDB",
"version": "1.0.0rc1"
}
```

Since we have decided to include a `ws://host:port` section for the `streams` key, we need a way to overwrite the default value of `0.0.0.0` in the configuration.
Thoughts, @TimDaub ?

krish7919:
Fixes #1626.

I took the liberty of not naming the configs as `WSSERVER_EXTERNAL_HOST`, etc. (as mentioned in the issue) and named them `BIGCHAINDB_WSSERVER_ADVERTISED_SCHEME`, `BIGCHAINDB_WSSERVER_ADVERTISED_HOST` and `BIGCHAINDB_WSSERVER_ADVERTISED_PORT` as they map directly to what setting is to be advertised.


Dockerfile
```diff
@@ -14,5 +14,9 @@ ENV BIGCHAINDB_CONFIG_PATH /data/.bigchaindb
 ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 ENV BIGCHAINDB_WSSERVER_HOST 0.0.0.0
 ENV BIGCHAINDB_WSSERVER_SCHEME ws
+
+ENV BIGCHAINDB_WSSERVER_ADVERTISED_SCHEME ws
+ENV BIGCHAINDB_WSSERVER_ADVERTISED_HOST 0.0.0.0
```
ttmc: I'm pretty sure users won't be able to find the WebSocket API at ws://0.0.0.0:9985 but I'm not sure what to put instead of 0.0.0.0

bigchaindb/web/views/base.py
```diff
@@ -23,7 +23,7 @@ def make_error(status_code, message=None):
 
 def base_ws_uri():
     """Base websocket uri."""
-    scheme = config['wsserver']['scheme']
-    host = config['wsserver']['host']
-    port = config['wsserver']['port']
+    scheme = config['wsserver']['advertised_scheme']
```
ttmc: Maybe you could add to the docstring of this function to say how it's supposed to be used (and not used).

docs/server/source/server-reference/configuration.md
```diff
@@ -248,6 +251,38 @@ export BIGCHAINDB_WSSERVER_PORT=9985
 }
 ```
 
+## wsserver.advertised.scheme, wsserver.advertised.host and wsserver.advertised.port
+
+These settings are for the advertising the Websocket URL to external clients in
+the root api endpoint. These configurations might be useful if your deployment
```
ttmc: Please write `API` in all-capital letters when writing it in prose, because it's an acronym.

docs/server/source/server-reference/configuration.md
```diff
@@ -248,6 +251,38 @@ export BIGCHAINDB_WSSERVER_PORT=9985
...
+is hosted behind a firewall, NAT, etc. where the exposed public IP or domain is
+different from where BigchainDB is running.
+
+**Example using environment variables**
+```text
+export BIGCHAINDB_WSSERVER_ADVERTISED_SCHEME=wss
+export BIGCHAINDB_WSSERVER_HOST=mybigchaindb.com
+export BIGCHAINDB_WSSERVER_PORT=443
```
ttmc: You need to add `ADVERTISED` in the above two lines.

docs/server/source/server-reference/configuration.md
```diff
@@ -248,6 +251,38 @@ export BIGCHAINDB_WSSERVER_PORT=9985
...
+```
+
+**Example config file snippet**
+```js
+"wsserver": {
+    "advertised_scheme": "wss",
+    "advertised_host": "mybigchaindb.com",
+    "port": 443
```
ttmc: advertised_port

docs/server/source/server-reference/configuration.md
```diff
@@ -248,6 +251,38 @@ export BIGCHAINDB_WSSERVER_PORT=9985
 }
 ```
 
+## wsserver.advertised.scheme, wsserver.advertised.host and wsserver.advertised.port
```
ttmc: Should be `wss.advertised_scheme` etc. (three times).

Dockerfile
krish7919: Switch it to the default value of `localhost` or unset it here so it defaults to `localhost`?

bigchaindb/web/views/base.py
krish7919: I didn't quite understand this. Use of the `advertised_*` variables are in the docs.
Do you mean replicate that over here?

ttmc: I meant that it would be nice to put some more details in the comments at the top of this function definition, so users of the function know what it is for, i.e. when it is supposed to be used.

krish7919: Cool. Will do.

Dockerfile
ttmc: I don't understand why all these BIGCHAINDB_WSSERVER_* environment variables are being set in the Dockerfile. If the environment variable is set, it overrides the default value, and any value in a local config file.

Why isn't BIGCHAINDB_WSSERVER_PORT set?

krish7919: This was as it was in the first version of Dockerfile that I dealt with; and it has remained that way from then.
And extending that, I have added the `WSSERVER_ADVERTISED_*` settings accordingly.
Maybe you can file a separate issue for it and we can handle the Dockerfile separately?

ttmc: Okay I created https://github.com/bigchaindb/bigchaindb/issues/1704

ttmc:
I guess you should also set the same environment variables in `Dockerfile-dev`?

Dockerfile-dev
```diff
@@ -12,6 +12,9 @@ ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
 ENV BIGCHAINDB_WSSERVER_HOST 0.0.0.0
 ENV BIGCHAINDB_WSSERVER_SCHEME ws
 
+ENV BIGCHAINDB_WSSERVER_ADVERTISED_HOST 0.0.0.0
+ENV BIGCHAINDB_WSSERVER_ADVERTISED_SCHEME ws
+
```
ttmc: Why no `ENV BIGCHAINDB_WSSERVER_ADVERTISED_PORT 9985` like in the Dockerfile?

krish7919: Because there is not `BIGCHAINDB_WSSERVER_PORT` in the Dockerfile-dev.
I believe we have #1704  which will help @sbellem and me to answer what can and cannot get included in the Dockerfiles. I guess we can follow that one to resolve the case with the Dockerfiles.

r-marques:
Don't know about the docker files but the rest looks good to me.

docs/server/source/server-reference/configuration.md
```diff
@@ -248,6 +251,38 @@ export BIGCHAINDB_WSSERVER_PORT=9985
...
+export BIGCHAINDB_WSSERVER_ADVERTISED_HOST=mybigchaindb.com
+export BIGCHAINDB_WSSERVER_ADVERTISED_PORT=443
+```
+
+**Example config file snippet**
+```js
+"wsserver": {
+    "advertised_scheme": "wss",
+    "advertised_host": "mybigchaindb.com",
+    "advertised_port": 443
```
TimDaub: Instead of name spacing the variables, we could simply nest one level further:

```json
"wsserver": {
  "advertised": {
    "scheme": "wss",
   ...
}
```

ttmc: In BigchainDB Server version 1.0.0, we already had `wsserver.scheme`, `wsserver.host` and `wsserver.port` as configuration settings. This PR adds three more: `wsserver.advertised_scheme`, `wsserver.advertised_host` and `wsserver.advertised_port`. I like the fact that they're at the same level in the tree/hierarchy as the existing settings. Flat is better than nested, etc.

It also makes sense to keep the words "advertised" and "scheme" together in a compound word-phrase. It's like New Orleans, which isn't New.Orleans (i.e. New doesn't have an Orleans). Similarly, it's not like wsserver has an advertised which has a scheme. The wsserver has an advertised_scheme.

krish7919: @TimDaub I kept it at the same level as it isn't something like:
```
"wsserver": {
    "internal" : {
        "port": ...
    },
    "advertised": {
        "port": ...
    }
}
```

but goes hand in hand with what is specified right next to each other, or at the same indent level.

PR #1707, opened 2017-07-31
Add plugin support for events

### LINKED ISSUE #1701: Add plugin support to tap into the blocks stream
BigchainDB WebSocket Event API emits new valid transactions as soon as a block reached quorum.
While this lightweight approach is suitable for many applications, other use cases require integration with AMPQ services like RabbitMQ (see #1572).

To support those use cases (without bloating the codebase), we can add support to tap into the **blocks** stream using a plugin architecture. In this way custom use cases can easily be implemented with *few lines of code™*.

vrde:
Closes #1701.

This PR adds support for plugins that can tap into the BigchainDB events. Please read the [submitted documentation](https://github.com/bigchaindb/bigchaindb/blob/0f139a1999ea1aafa5c45086944d52cc1eb8e971/docs/server/source/events/event-plugin-api.rst) to understand how it works.

I've also created a sample plugin called [console_print](https://github.com/bigchaindb/bigchaindb_events_plugin_example).


bigchaindb/config_utils.py
```diff
@@ -287,3 +287,16 @@ def load_consensus_plugin(name=None):
                         'consensus.BaseConsensusRules`'.format(type(plugin)))
 
     return plugin
+
+
+def load_block_publisher_plugins(names):
+    plugins = []
+
+    if names is None:
```
r-marques: We need to decide what the no plugins default should be (`[]` or `None`).
https://github.com/bigchaindb/bigchaindb/pull/1707/files#diff-38ba73c9ead7452abd625b636f098e99R64 is using `[]` as the default if the config is not set.

bigchaindb/processes.py
```diff
@@ -54,8 +55,16 @@ def start():
...
     p_websocket_server = mp.Process(name='ws',
                                     target=websocket_server.start,
-                                    args=(events_queue,))
+                                    args=(pubsub.get_subscriber_queue(),))
     p_websocket_server.start()
 
     # start message
     logger.info(BANNER.format(bigchaindb.config['server']['bind']))
+
+    for name, plugin in load_block_publisher_plugins(bigchaindb.config.get('block_publishers', [])):
```
r-marques: Should we use a more generic name instead of `block_publishers`. In the feature we may update this to emit other types of events.

r-marques:
Looking really good to me

krish7919:
@vrde : I see code to add queues. Is there any mechanism devoted to remove entries from the queue?

A client might connect to get a websocket event and then disconnect and go away forever, or connect with a new public IP or something else.

vrde:
> If addition or removal of plugins need a restart, then we should be good here.

Yup, adding or removing plugins need a restart, no hot loading here :fire_engine:.

docs/server/source/events/event-plugin-api.rst
```diff
@@ -0,0 +1,67 @@
...
+-------------------------------
+
+Creating your own plugin is really easy, and can be summarized in few steps:
+
+1. Create a new Python package that defines the entry point ``bigchaindb.events`` in its ``setup.py``.
+2. In your entry point, define two properties:
+
+   - ``event_types``: a variable to tell BigchainDB to which events your plugin is interested.
+   - ``run``: a function that will process the events coming from BigchainDB.
+3. Install the newly created Python in the current environment.
```
ttmc: Python _package_?

docs/server/source/events/event-plugin-api.rst
```diff
@@ -0,0 +1,67 @@
...
+4. Add the plugin name to your BigchainDB configuration.
+5. (Re)start BigchainDB
+
+If the installation was successful, the plugin will be run in a different
+process. Your plugin will receive events through a ``multiprocessing.Queue``
+object. Remember: it's your plugin responsibility to consume that queue.
```
ttmc: "it's your plugin's responsibility"

(apostrophe s)

docs/server/source/events/event-plugin-api.rst
```diff
@@ -0,0 +1,67 @@
...
+We decided to make this internal event system public, to allow developers to
+integrate BigchainDB with other applications, such as AMPQ systems.
+
+
+Available events
+----------------
+
+The event types are listed in the source file ``bigchaindb/events.py``.
+
+.. list-table:: Event Types
```
vrde: It's much better than writing tables in the classic way :)

bigchaindb/events.py
```diff
@@ -1,33 +1,91 @@
...
+POISON_PILL = 'POISON_PILL'
+
+
+class EventTypes:
+    """Container class that holds all the possible
+    events BigchainDB manages."""
+
+    # If you add a new Event Type, make sure to add it
+    # to the docs in docs/server/source/event-plugin-api.rst
+    ALL = ~0
```
vrde: This is a *bit* (eh eh eh) tricky. It's the bitwise negation of `0`, that is a number with all `1`s. It's always `True` as long as you `&` with something `True`:

```
>>> bin(~0 & 0b0010000)
'0b10000'
>>> bin(~0 & 0b0)
'0b0'
>>> bin(~0 & 0b0000001)
```

vrde: It is used [here](https://github.com/bigchaindb/bigchaindb/pull/1707/files#diff-60d798ce60a59a778f9a841f17bef565R79) to dispatch an event to the correct queues.

An event it is a number that is a power of two. If you represent this number you have just a `1` followed by zero or more `0`s.

A subscriber can listen to a combination of events just by `or`ing the ids of those events, the name for this is *bitmask*.

When the dispatcher gets a new event, it `and`s the event's id with all the available *bitmasks*. If the result is greater than `0`, then the event belongs to that bitmask, and the dispatcher forwards the event to the subscribers.

Issue #1432, opened 2017-04-21
Use a proper persistentVolumeReclaim policy in k8s deployment


krish7919:
While creating a [PersistentVolumeClaim](https://github.com/bigchaindb/bigchaindb/blob/master/k8s/mongodb/mongo-pvc.yaml), we do not use a persistentVolumeReclaimPolicy.

Currently it is marked as:
```
  # FIXME(Uncomment when ACS supports this!)
  # persistentVolumeReclaimPolicy: Retain
```

This is because ACS does not support it yet.
Update the file when ACS supports this feature.

muawiakh:
`persistentVolumeReclaimPolicy` can only be specified when we are creating a `persistentVolume` and not `persistentVolumeClaim` . In our current deployment model we use `persistentVolumeClaim` to dynamically provision a volume for our claim. Marking this ticket as invalid.

On the other hand we can provide some documentation on how to provision a `persistentVolume` and then bind a `persistentVolumeClaim` to it. @krish7919 Let me know if there is a requirement for it. Otherwise we can close this ticket and track it as part of another one.

krish7919:
@muawiakh Ideally, we do need a PV.

I referred to the [1.6 api reference](https://kubernetes.io/docs/api-reference/v1.6/#persistentvolumeclaimspec-v1-core) and the [1.7 api reference](https://kubernetes.io/docs/api-reference/v1.7/#persistentvolumespec-v1-core), and do not find support of PV in 1.6.

Can you please clarify if PV with Azure is supported in v1.6?

shahbazn:
Creating PV is supported with Azure on v1.6 but you need to create a vhd disk file and upload to your blob storage container in your storage account and then use its UUID and disk URI in PV templates. Not sure how kubernetes handles creating VHDs dynamically while creating PVs but for now we can change the Reclaim policy for the PVs once these are created that way we can retain the Azure Storage disk blob even if we decide to disintegrate the cluster and can use the same disk with a different cluster without loosing any data.

PR #1720, opened 2017-08-15
Remove persistentVolumeReclaimPolicy from the pvc request

muawiakh:
- Reclaim policy can only be specified when we explicitly create
a persistent Volume. Removing this from a persistent volume claim
request yaml.


muawiakh:
Parially fixes #1432

Issue #1424, opened 2017-04-20
Run mongodb monitoring and backup agents as non-root user


krish7919:
`USER mongodb-mms-agent ` in Dockerfile for both backup and monitoring agents.
Will need dir and file permission changes in entrypoint script and Dockerfile.

muawiakh:
We need to modify the read permissions on the secrets and allow
the `mongodb-mms-agent` to read the credentials. Alternative approach is to change the `readOnly` permissions from mounted volumes `/etc/mongod/cloud` and `/etc/mongod/ssl` and make `mongodb-mms-agent` the owner of the above mentioned directories in the `entrypoint` script but I went with the earlier approach because changing readOnly on the volumes might be such a good idea.

ttmc:
@muawiakh Can this issue be close now?

muawiakh:
Yes.

PR #1723, opened 2017-08-15
Run mongodb monitoring and backup agents as non-root user

muawiakh:
- We need to modify the read permissions on the secrets and allow
the mongodb-mms-agent to read the credentials.


muawiakh:
fixes #1424

PR #1737, opened 2017-08-31
Added link to Ruby Gem by @nileshtrivedi

ttmc:
I added a link to the **bigchaindb** Ruby Gem by @nileshtrivedi 
on the **Drivers & Tools** page.


diminator:
Maybe also add it to the getting started page cc @kremalicious

kremalicious:
@diminator yes, adding a new tool or driver to Get Started page should be your first thought instead of adding to docs. Way more traffic on Get started page then on docs, so much more impact:

Last 30 days, visits to Drivers & Tools docs page: **267**
Last 30 days, visits to Get Started page: **3,936**

kremalicious:
https://github.com/ascribe/bigchain-website/pull/171

ttmc:
The number of page views should not be used as a measure of "impact."

Even if the number of page views is 1, if that one person is the right person, that one page view might have enormous consequences.

In short, not all page views are equal impact.

Issue #1224, opened 2017-02-27
Validate block voters list


libscott:
Validate that block voters list matches server keyring (possibly up to some threshold) when voting

@ttmc from #1168:

> I had some more thoughts about the problem of a malicious node writing a baloney voters list into a block it creates. There are at least two bad things a malicious node can do to the voters list in blocks it creates:
> 
> * remove legit public keys of the other federation nodes, preventing them from voting on the block
> * add fake "sybil" public keys of imaginary federation nodes, for which it has all the corresponding private keys, so it can produce valid votes
> 
> While checking the voters list against the "federation keyring" is one way to prevent those problems, there are other options. In particular, each node could have a rule something like:
> 
> "One condition for this block to get a vote-of-valid from me is that its voters list must differ from mine by at most one public key."
> 
> (The list lengths can differ, but so can the list values.)
> 
> It can't be 100% because we need to allow for nodes to come and go from the federation from time to time, but it will take time for all federation nodes to update their local keyrings.
> 
> If we take this approach, we need to re-evaluate how many votes are needed on a block to switch it from undecided to valid/invalid. 50% is probably too low. 66.7% seems like it would suffice. It's not like block validity is a matter of opinion: the usual outcome is that all nodes will agree that a block is valid/invalid. Any disagreement from any node should be a red flag to investigate further.

ttmc:
BigchainDB 2.0 is coming out soon and it doesn't care about the "voter list" (list of current validators). It now leaves those concerns to separate consensus software. Therefore this is no longer an issue for BigchainDB, and I am closing it.

PR #1740, opened 2017-09-01
Test incorrect voting behavior

TimDaub:
@CsterKuroi wrote on gitter:

> There's a logical problem in the line https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/voting.py#L33 .
If a bad node nodeA set block['block']['voters'] =['`<public key of nodeA >`'] ,
all nodes would check nodeA's vote only as the eligible_voters =set(block['block']['voters']) & set(keyring)=set(['`<public key of nodeA >`']) .
So whether the block is valid depends only on the nodeA's vote.
Am I right?

This PR is to confirm their intuition. Seems right to me.

Don't have time to look into this further now, but it seems like a critical bug. Happy if someone wants to take this over.

Thx to @CsterKuroi

### DIFF
--- a/tests/test_voting.py
+++ b/tests/test_voting.py
@@ -235,14 +235,41 @@ def test_block_election(b):
         'other_previous_block': {},
     }
 
+# TODO: Check all other tests in this class for similar mistakes
+
 
 @patch('bigchaindb.voting.Voting.verify_vote_signature', return_value=True)
 def test_duplicate_vote_throws_critical_error(b):
     keyring = 'abc'
+
+    # block_election calls count_votes that calls `cls.verify_vote_schema`.
     block = {'id': 'xyz', 'block': {'voters': 'ab'}}
+    # As `vote` has a required property `signature`, the vote here should
+    # technically be counted as `malformed` (or invalid), rendering this test
+    # incorrect as `CriticalDuplicateVote` could be trigger elsewhere.
     votes = [{
+        # according to the vote's schema a public key needs to have up to 43
+        # chars, making the keyring in this test to not pass
+        # `verify_vote_schema`.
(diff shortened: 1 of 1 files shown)


tests/test_voting.py
```diff
@@ -235,14 +235,41 @@ def verify_vote_schema(cls, vote):
...
         'vote': {'is_block_valid': True, 'previous_block': 'a'}
     } for c in 'aabc']
     with pytest.raises(CriticalDuplicateVote):
         Voting.block_election(block, votes, keyring)
+    # TODO: Fix this test
+    assert False
+
+
+@patch('bigchaindb.voting.Voting.verify_vote_signature', return_value=True)
+@patch('bigchaindb.voting.Voting.verify_vote_schema', return_value=True)
```
TimDaub: This is not idea. Ideally we setup the test such that we can remove patching this method. @vrde and I just used it to save time.

ttmc:
This isn't a new issue, it's a special case of an old one. The basic problem is that a node can stuff whatever list of node public keys its wants into the blocks that it builds, including a list of its friends, or just its own public key (as this issue highlights).

Previous related issues / proposals include: 

#187 (closed) - An old proposal to have some federation-agreed list of current-node public keys, a "federation keyring"

#1224 (open) - Validate block voters list

tests/test_voting.py
vrde: I don't mind patching `verify_vote_schema` since we are testing for a malicious, valid vote. We also have `test_verify_vote_schema` that should cover the schema validation.

vrde:
We are making architectural changes that will make this issue invalid. Hold tight.

PR #1747, opened 2017-09-12
Added link to new Java driver in the docs

ttmc:
This PR adds a link to the Java driver written by Authenteq.

I kept the link to the old and not-working Java driver, but noted those characteristics in its description. Some folks might find it interesting. Anyone wanting a Java driver will go for the one labelled "newer, working" (I hope!).


kremalicious:
Now that the new Java driver is marked as "official" I feel we should remove the non-working one since community-efforts should rather be spent on that official one then. What do you think?

ttmc:
@kremalicious I have changed this PR so that if it gets merged, the docs will only list the new Java driver.

kremalicious:
ok, then I'm also going to remove the non-working one from the Get Started page

PR #1749, opened 2017-09-14
Automation for single node deployment for quickstart

muawiakh:
- Change consists of two deployment models:
  - Using Vagrant(single node, with/without docker)
  - Using Ansible(single node, with/without docker)
- Updated quickstart documentation.
- Some WIP comments, which will be addressed later. Depending on the
  requirements.

### DIFF
--- a/docs/server/source/appendices/run-with-vagrant.md
+++ b/docs/server/source/appendices/run-with-vagrant.md
@@ -1,0 +1,137 @@
+# Run BigchainDB with Vagrant
+
+**NOT for Production Use**
+
+You can use the following instructions to deploy a BigchainDB node
+for dev/test using Vagrant. Vagrant will setup a BigchainDB node with
+all the dependencies along with MongoDB, BigchainDB Python driver. You
+can also tweak the following configurations for the BigchainDB node.
+- Vagrant Box
+  - Currently, we support the following boxes:
+    - `ubuntu/xenial64 # >=16.04`
+    - `centos/7 # >=7`
+    - `fedora/24 # >=24`
+  - **NOTE** : You can choose any other vagrant box of your choice but these are
+  the minimum versioning requirements.
+- Resources and specs for your box.
+  - RAM
+  - VCPUs
+  - Network Type
+    - Currently, only `private_network` is supported.
+  - IP Address
+- Setup type
+  - `quickstart`
+- Deploy node with Docker
+  - Deploy all the services in Docker containers or as processes.
+- Upstart Script
+- Vagrant Provider
+  - Virtualbox
+  - VMware
+
+## Install dependencies | Vagrant
+1. [VirtualBox](https://www.virtualbox.org/wiki/Downloads) >= 5.0.0
+2. [Vagrant](https://www.vagrantup.com/downloads.html) >= 1.16.0
+
+## Clone the BigchainDB repository | Vagrant
+```text
+$ git clone https://github.com/bigchaindb/bigchaindb.git
+```
+
+## Configuration | Vagrant
+Navigate to `bigchaindb/pkg/config/` inside the repository.
+```text
+$ cd bigchaindb/pkg/config/
+```
+
+Edit the `bdb-config.yaml` as per your requirements. Sample `bdb-config.yaml`:
+
+```text
+---
+- name: "bdb-node-01"
+  box:
+    name: "ubuntu/xenial64"
+  ram: "2048"
+  vcpus: "2"
+  setup_type: "quickstart"
+  deploy_docker: false
+  network:
+    ip: "10.20.30.40"
+    type: "private_network"
+  upstart: "/bigchaindb/scripts/bootstrap.sh"
+```
+
+**Note**: You can spawn multiple instances as well using `bdb-config.yaml`. Here is a sample `bdb-config.yaml`:
+```text
+---
+- name: "bdb-node-01"
+  box:
+    name: "ubuntu/xenial64"
+  ram: "2048"
+  vcpus: "2"
+  setup_type: "quickstart"
+  deploy_docker: false
+  network:
+    ip: "10.20.30.40"
+    type: "private_network"
+  upstart: "/bigchaindb/scripts/bootstrap.sh"
+- name: "bdb-node-02"
+  box:
+    name: "ubuntu/xenial64"
+  ram: "4096"
+  vcpus: "3"
+  setup_type: "quickstart"
+  deploy_docker: false
+  network:
+    ip: "10.20.30.50"
+    type: "private_network"
+  upstart: "/bigchaindb/scripts/bootstrap.sh"
+```
+
+
+## Local Setup | Vagrant
+To bring up the BigchainDB node, run the following command:
+
+```text
+$ vagrant up
+```
+
+*Note*: There are some vagrant plugins required for the installation, user will be prompted to install them if they are not present. Instructions to install the plugins can be extracted from the message.
+
+```text
+$ vagrant plugin install <plugin-name>
+```
+
+After successfull execution of Vagrant, you can log in to your fresh BigchainDB node.
+
+```text
+$ vagrant ssh <instance-name>
+```
+
+## Make your first transaction
+Once you are inside the BigchainDB node, you can verify that BigchainDB docker/process is running.
+
+Verify BigchainDB process:
+```text
+$ ps -ef | grep bigchaindb
+```
+
+OR
+
+Verify BigchainDB Docker:
+```text
+$ docker ps | grep bigchaindb
+```
+
+BigchainDB python driver is pre-installed in the instance, so you can instantly make transactions and verify the functionality. The `bdb_root_url` can be one of the following:
+```text
+# BigchainDB is running as a process
+bdb_root_url = http://<HOST-IP>:9984
+
+OR
+
+# BigchainDB is running inside a docker container
+bdb_root_url = http://<HOST-IP>:<DOCKER-PUBLISHED-HOST-PORT>
+```
+For more details on *how to make a transaction?* Please refer to [Basic Usage Examples](https://docs.bigchaindb.com/projects/py-driver/en/latest/connect.html).
+
+*Note*: If you want to make transactions remotely, you need to install the `bigchaindb-driver`. For detailed instructions on how to install the driver and make your first transaction. Please refer to [Quickstart/Installation](https://docs.bigchaindb.com/projects/py-driver/en/latest/quickstart.html) of BigchainDB driver.
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -1,12 +1,6 @@
 # Quickstart
 
-This page has instructions to set up a single stand-alone BigchainDB node manually for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). You might try one of the following deployment methods
-as well, depending on your choice:
-- [BigchainDB with Docker](appendices/run-with-docker.html).
-- [BigchainDB with Vagrant](appendices/run-with-vagrant.html).
-- [BigchainDB with Ansible](appendices/run-with-ansible.html).
-
-For manual installation, We will assume you're using Ubuntu 16.04 or similar.
+This page has instructions to set up a single stand-alone BigchainDB node for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). We will assume you're using Ubuntu 16.04 or similar. If you're not using Linux, then you might try [running BigchainDB with Docker](appendices/run-with-docker.html).
 
 A. Install MongoDB as the database backend. (There are other options but you can ignore them for now.)
 
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -1,6 +1,12 @@
 # Quickstart
 
-This page has instructions to set up a single stand-alone BigchainDB node for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). We will assume you're using Ubuntu 16.04 or similar. If you're not using Linux, then you might try [running BigchainDB with Docker](appendices/run-with-docker.html).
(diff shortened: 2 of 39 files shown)


docs/server/source/quickstart.md
```diff
@@ -1,6 +1,12 @@
 # Quickstart
 
-This page has instructions to set up a single stand-alone BigchainDB node for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). We will assume you're using Ubuntu 16.04 or similar. If you're not using Linux, then you might try [running BigchainDB with Docker](appendices/run-with-docker.html).
+This page has instructions to set up a single stand-alone BigchainDB node manually for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). You might try one of the following deployment methods
+as well, depending on your choice:
+- [BigchainDB with Docker](appendices/run-with-docker.html).
```
ttmc: I think a "Quickstart" page should have no steps where the reader must make a decision. A Quickstart page should be "on rails." This PR adds a four-option decision first thing on the Quickstart page. Let's remove that.

The sentence "Instructions for other cases are elsewhere." already links to the Introduction page, and that page has "Setup Instructions for Various Cases". Let's put links to the various other setup options there. We could maybe organize those links differently.

What are all these different options for? Help the reader decide which one is for them. Which ones are just for trying BigchainDB? Which ones are for being a developer of BigchainDB Server software? How would you categorize these options:

* Run BigchainDB with Docker
* Run BigchainDB with Docker on Mac
* Run BigchainDB with Vagrant
* Run BigchainDB with Ansible

ttmc: While you're editing the Introduction page:

I noticed that one of the links on the Introduction page is broken (because a previous PR deleted the page that it linked to). That link can be removed. It's the one with text, "Deploy a bare-bones RethinkDB-based node on any Ubuntu machine with Ansible". That whole line can be deleted.

docs/server/source/quickstart.md
```diff
@@ -1,6 +1,12 @@
...
+- [BigchainDB with Vagrant](appendices/run-with-vagrant.html).
+- [BigchainDB with Ansible](appendices/run-with-ansible.html).
+
+For manual installation, We will assume you're using Ubuntu 16.04 or similar.
```
ttmc: Maybe remove this line and embed the "Ubuntu 16.04" in the opening line, e.g.

This page has instructions to set up a single stand-alone BigchainDB node on Ubuntu 16.04, for learning or experimenting.

I removed the word "manually" because even this method uses helper tools such as `apt` and `pip` (but not Ansible or Vagrant). Basically, it's not clear what the word "manually" means in the context of installing software, so let's just avoid it.

muawiakh: If I don't specify manually both the deployment methods will be the same and manually implies the opposite of `automat(ic/ed)`. All deployments need some manual intervention or pre-req installations etc.

docs/server/source/quickstart.md
muawiakh: Removed the broken link. For the other comment ,I can think of two options.
- We can change the `quickstart` in a way that we specify all these ways to install and the user can pick whatever they want to choose i.e. the new `quickstart` page can look like this:
```text
  - Run BigchainDB on Host(Ubuntu 16.04)
  - Run BigchainDB with Docker
  - Run BigchainDB with Vagrant
  - Run BigchainDB with Ansible
```

OR

- We don't change the instructions on the `quickstart` page and change the `introduction` page to something like this.
```text
....
  - Set up and run a local dev/test node for developing and testing BigchainDB Server.
    - Run BigchainDB on Host/Local Dev.
    - Run BigchainDB with Docker.
    - Run BigchainDB with Vagrant.
    - Run BigchainDB with Ansible.
...
```
Instead of `setup-run-node` having OptionA,B..., we can just redirect from the introduction page to the relevant installation. Comments? But option 2 will need some refactoring in terms of files and updating the `setup-run-node` which I think might be broken. @ttmc

ttmc: Maybe it would be better to move the list of all the installation options out of the "Introduction" page and into a new page titled "Install BigchainDB". That would be better for search engines as well.

I definitely prefer to keep the "Quickstart" page in its own little world, not asking the reader to make any decisions.

I know it would take time, but it would be a good idea to refactor `setup-run-node`. Maybe into separate pages? In fact, all the various installation methods (pages) could be moved from the Appendices into their own section.

docs/server/source/quickstart.md
ttmc: If you want to keep the word "manually" then that is fine.

docs/server/source/quickstart.md
muawiakh: This makes sense. How about we merge this PR as is, if you don't have any comments on the  rest of the components and I will issue another PR for the docs refactor.

ttmc: On second thought, refactoring the installation-related docs should be done in a separate PR. This PR is already big enough.

muawiakh: So should I take the docs changes out of this PR? And make it part of another one?

docs/server/source/appendices/run-with-vagrant.md
```diff
@@ -0,0 +1,137 @@
...
+
+BigchainDB python driver is pre-installed in the instance, so you can instantly make transactions and verify the functionality. The `bdb_root_url` can be one of the following:
+```text
+# BigchainDB is running as a process
+bdb_root_url = http://<HOST-IP>:9984
+
+OR
+
+# BigchainDB is running inside a docker container
+bdb_root_url = http://<HOST-IP>:<DOCKER-PUBLISHED-HOST-PORT>
```
krish7919: Is the BigchainDB process inside the vagrant box run in a container?
The <OR> over here makes it confusing to know if it is a stand-alone process or a process inside a docker container.

muawiakh: Yes, in `run-with-vagrant` guide it is running as a docker container inside the box.

docs/server/source/quickstart.md
krish7919: "We" => "we"

krish7919: We => we.

pkg/Vagrantfile
```diff
@@ -0,0 +1,56 @@
...
+        vmwf.vmx['numvcpus'] = instance['vcpus']
+      end
+
+      bdb.vm.provider 'virtualbox' do |vb, override|
+        vb.memory = instance["ram"]
+        vb.cpus = instance['vcpus']
+      end
+    end
+  end
+end
```
krish7919: Question: Can a used deploy a BDB vagrant box locally and then use the driver from the host to send transactions?
From what I understand looking at the networking, this only supports a private network which means that the user has to send transactions from within the VM. This can make user not use the local deployment for testing at all, as s/he will have to copy scripts to the box and back.
This can also hamper our plan to run a test setup using vagrant, as we will have to bundle all the tests in the vagrant box.

Can we change the networking to bridged or host-only mode to allow this to happen?
Or should we add this constraint to the deployment and do the testing in a different way?

pkg/ansible/roles/bigchaindb-driver/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+dependencies_deb:
+  - python3-dev
+  - libffi-dev
+  - libssl-dev
+  - python3-pip
+
+dependencies_yum:
+  - gcc-c++
+  - "@Development Tools"
+  - python34-devel
```
krish7919: I think we only support python 3.5 and 3.6 now. Refer to https://github.com/bigchaindb/bigchaindb/blob/master/.travis.yml

pkg/ansible/roles/bigchaindb-driver/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  - libffi-devel
+  - openssl-devel
+  - python34-setuptools
```
krish7919: Same as above comment.

pkg/Vagrantfile
muawiakh: You can still run the tests/send transactions from the host, we don't need to copy them inside the `vagrant box`, I send transactions from my host by specifying the `bdb_root_url` of the bigchaindb process running inside the box.

pkg/ansible/roles/bigchaindb-driver/tasks/centos.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: Install dependencies
+  yum:
+    name: "{{ item }}"
+    state: present
+    update_cache: yes
+  with_items: "{{ dependencies_yum }}"
+  tags: [bigchaindb-driver]
+
+- name: Install pip
+  shell: "easy_install-3.4 pip"
```
krish7919: Same comment as the above.

pkg/ansible/roles/bigchaindb/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  - python3-dev
+  - libffi-dev
+  - build-essential
+  - libssl-dev
+  - python3-pip
+
+dependencies_yum:
+  - gcc-c++
+  - "@Development Tools"
+  - python34-devel
```
krish7919: Same comment as the above.

pkg/ansible/roles/bigchaindb/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  - libffi-devel
+  - openssl-devel
+  - python34-setuptools
```
krish7919: Same comment as the above.

docs/server/source/quickstart.md
muawiakh: I will address this in the new PR which will refactor some documents regarding installation.

muawiakh: Meanwhile, `quickstart` will stay unchanged for this PR, I will revert this diff.

pkg/ansible/roles/mongodb/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+distribution_major: "{{ ansible_distribution_major_version }}"
+server_arch: "amd64,arm64"
+
+# MongoDB Repos
+mongodb_apt_repo: "deb [arch={{ server_arch }}] http://repo.mongodb.org/apt/{{ distribution_name }} {{ distribution_codename }}/{{ mongodb_package }}/{{ mongodb_version }} {{'main' if ansible_distribution == 'debian' else 'multiverse'}}"
+mongodb_yum_base_url: "https://repo.mongodb.org/yum/{{ ansible_os_family|lower }}/$releasever/{{ mongodb_package }}/{{ mongodb_version }}/{{ ansible_architecture }}"
+mongodb_dnf_base_url: "https://repo.mongodb.org/yum/{{ ansible_os_family|lower }}/7/{{ mongodb_package }}/{{ mongodb_version }}/{{ ansible_architecture }}"
+
+# MongoDB running config
+mongodb_storage_path: /data/db
```
krish7919: This has been changed to `/data/db/main` in the container recently.

pkg/ansible/roles/mongodb/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+mongodb_config_path: /data/configdb
+se_linux: "TODO: (muawiakh)"
+directories:
+  - "{{ mongodb_storage_path }}"
+  - "{{ mongodb_log_path }}"
+  - "{{ mongodb_config_path }}"
+
+# Docker configuration
+mongodb_default_port: 27017
+mongodb_docker_image: "mongo:3.4.1"
```
krish7919: mongo:3.4.4 from [here](https://github.com/bigchaindb/bigchaindb/blob/master/k8s/mongodb/container/Dockerfile).

pkg/ansible/roles/mongodb/defaults/main.yml
muawiakh: Done.

pkg/ansible/roles/mongodb/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+mongodb_docker_name: "mongodb"
+mongodb_docker_published_ports:
+  - 172.17.0.1:27017:27017
+mongodb_docker_volumes:
+  - "/tmp/mongodb_docker/db:{{ mongodb_storage_path }}"
+  - "/tmp/mongodb_docker/configdb:{{ mongodb_config_path }}"
```
krish7919: A user recently changed this to `$HOME/...` instead of `/tmp/...` as he/she didn't want to lose data on system restarts.
We might wanna consider that in case a user does a `vagrant halt` and a `vagrant up` repeatedly.

pkg/ansible/roles/mongodb/defaults/main.yml
muawiakh: Done.

pkg/config/bdb-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+- name: "bdb-node-01" # Instance name
+  box:
+    name: "ubuntu/xenial64" # Box name
+  ram: "2048"
+  vcpus: "2"
+  setup_type: "quickstart" # Currently, only quickstart is supported.
+  deploy_docker: true # [true, false]
+  network:
+    ip: "10.20.30.50"
+    type: "private_network" # Currently, only private network is supported.
```
krish7919: Comment about `private_network` above. Cross referencing here for clarity.

pkg/Vagrantfile
krish7919: So, you mean vagrant takes care of resolving the BDB vagrant box name?

pkg/config/bdb-config.yaml
muawiakh: I have added support for `public_network` as well. So this will not be an issue anymore.

pkg/Vagrantfile
muawiakh: On second thought, I have added support for `public network` as well, so this should not be a problem.

PR #1757, opened 2017-09-25
Adjust resources for MongoDB Stateful Set

muawiakh:
- Currently, MongoDB container crashed because of resource constaints i.e.
out of memory exception. This change updates the resources and provides
data on how the configure/calculate them(if not following the guide).
- Also, add the ability to specify the storage engine(WiredTiger) cache
size for MongoDB, this configuration also helps with keeping the resources
constrained for MongoDB containers.
- Minor changes in some other documents as well.

### DIFF
--- a/docs/server/source/production-deployment-template/client-tls-certificate.rst
+++ b/docs/server/source/production-deployment-template/client-tls-certificate.rst
@@ -92,7 +92,7 @@ consolidated file containing both the public and private keys.
 
 .. code:: bash
 
-   cat /path/to/mdb-instance-0.crt /path/to/mdb-instance-0.key > mdb-instance-0.pem
+   cat /path/to/bdb-instance-0.crt /path/to/bdb-instance-0.key > bdb-instance-0.pem
 
     OR
 
--- a/docs/server/source/production-deployment-template/node-on-kubernetes.rst
+++ b/docs/server/source/production-deployment-template/node-on-kubernetes.rst
@@ -542,6 +542,19 @@ Step 12: Start a Kubernetes StatefulSet for MongoDB
     - ``mdb-certs``
     - ``ca-auth``
 
+  * **Optional**: You can also change the value for ``STORAGE_ENGINE_CACHE_SIZE``, for more information
+    regarding this configuration, please consult the `MongoDB Official
+    Documentation <https://docs.mongodb.com/manual/reference/configuration-options/#storage.wiredTiger.engineConfig.cacheSizeGB>`_.
+
+  * **Optional**: If you are not using the **Standard_D2_v2** virtual machines for Kubernetes agents as per the guide,
+    please update the ``resources`` for ``mongo-ss``. We suggest allocating ``memory`` using the following scheme
+    for a MongoDB StatefulSet:
+
+    .. code:: bash
+
+      memory = (Total_Memory_Agent_VM_GB - 2GB)
+      STORAGE_ENGINE_CACHE_SIZE = memory / 2
+
   * Create the MongoDB StatefulSet using:
 
     .. code:: bash
--- a/docs/server/source/production-deployment-template/node-on-kubernetes.rst
+++ b/docs/server/source/production-deployment-template/node-on-kubernetes.rst
@@ -542,7 +542,7 @@ Step 12: Start a Kubernetes StatefulSet for MongoDB
     - ``mdb-certs``
     - ``ca-auth``
 
-  * **Optional**: You can also change the value for ``STORAGE_ENGINE_CACHE_SIZE``, for more information
+  * **Optional**: You can change the value for ``STORAGE_ENGINE_CACHE_SIZE`` in the ConfigMap ``storage-engine-cache-size``, for more information
     regarding this configuration, please consult the `MongoDB Official
     Documentation <https://docs.mongodb.com/manual/reference/configuration-options/#storage.wiredTiger.engineConfig.cacheSizeGB>`_.
 
--- a/k8s/mongodb/container/docker_build_and_push.bash
+++ b/k8s/mongodb/container/docker_build_and_push.bash
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-docker build -t bigchaindb/mongodb:3.1 .
+docker build -t bigchaindb/mongodb:3.2 .
 
-docker push bigchaindb/mongodb:3.1
+docker push bigchaindb/mongodb:3.2
--- a/k8s/mongodb/container/mongod_entrypoint.bash
+++ b/k8s/mongodb/container/mongod_entrypoint.bash
@@ -46,6 +46,10 @@ while [[ $# -gt 1 ]]; do
           MONGODB_IP="$2"
           shift
           ;;
+      --storage-engine-cache-size)
+          STORAGE_ENGINE_CACHE_SIZE="$2"
+          shift
+          ;;
       *)
           echo "Unknown option: $1"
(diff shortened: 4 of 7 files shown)


docs/server/source/production-deployment-template/client-tls-certificate.rst
```diff
@@ -92,7 +92,7 @@ consolidated file containing both the public and private keys.
 
 .. code:: bash
 
-   cat /path/to/mdb-instance-0.crt /path/to/mdb-instance-0.key > mdb-instance-0.pem
+   cat /path/to/bdb-instance-0.crt /path/to/bdb-instance-0.key > bdb-instance-0.pem
```
krish7919: Why?
We never need a `bdb-instance-X` consolidated file.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -500,6 +500,19 @@ Step 12: Start a Kubernetes StatefulSet for MongoDB
...
+    Documentation <https://docs.mongodb.com/manual/reference/configuration-options/#storage.wiredTiger.engineConfig.cacheSizeGB>`_.
+
+  * **Optional**: If you are not using the **Standard_D2_v2** virtual machines for Kubernetes agents as per the guide,
+    please update the ``resources`` for ``mongo-ss``. We suggest allocating ``memory`` using the following scheme
+    for a MongoDB StatefulSet:
+
+    .. code:: bash
+
+      memory = (Total_Memory_Agent_VM_GB - 2GB)
+      STORAGE_ENGINE_CACHE_SIZE = memory / 2
```
krish7919: I would suggest that we link this from our doc to MongoDB docs, so that in case they change it tomorrow, we are not left with outdated info here.

k8s/mongodb/container/mongod_entrypoint.bash
```diff
@@ -61,7 +65,8 @@ if [[ -z "${REPLICA_SET_NAME:?REPLICA_SET_NAME not specified. Exiting!}" || \
     -z "${MONGODB_IP:?MONGODB_IP not specified. Exiting!}" || \
     -z "${MONGODB_KEY_FILE_PATH:?MONGODB_KEY_FILE_PATH not specified. Exiting!}" || \
     -z "${MONGODB_CA_FILE_PATH:?MONGODB_CA_FILE_PATH not specified. Exiting!}" || \
-    -z "${MONGODB_CRL_FILE_PATH:?MONGODB_CRL_FILE_PATH not specified. Exiting!}" ]] ; then
+    -z "${MONGODB_CRL_FILE_PATH:?MONGODB_CRL_FILE_PATH not specified. Exiting!}" ||
```
ttmc: Why doesn't this line end with a `\` like the ones above it?

k8s/mongodb/container/mongod_entrypoint.bash
```diff
@@ -61,7 +65,8 @@ if [[ -z "${REPLICA_SET_NAME:?REPLICA_SET_NAME not specified. Exiting!}" || \
...
+    -z "${STORAGE_ENGINE_CACHE_SIZE:=''}" ]] ; then
```
ttmc: The stuff inside the `${}` is different in this line, compared to the previous lines. What does it mean?

muawiakh:
Let me update the patch. Seems like a mistake.

k8s/mongodb/container/docker_build_and_push.bash
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-docker build -t bigchaindb/mongodb:3.1 .
+docker build -t bigchaindb/mongodb:3.2 .
```
shahbazn: are we changing version number for each new build?

k8s/mongodb/container/mongod_entrypoint.bash
shahbazn: +1

docs/server/source/production-deployment-template/client-tls-certificate.rst
muawiakh: I wanted to update the `mdb` in the `client-tls-certificate` page because that is not relevant here. I know that is not correct, but in the `secret.yaml` file the secret corresponding to bdb certificate is named `bdb-instance.pem`, either we change that name to `bdb-instance.crt`, or keep this consolidated file.  There is no impact on the functionality. What do you suggest?

docs/server/source/production-deployment-template/node-on-kubernetes.rst
muawiakh: This is not from an official MongoDB `docs`. It is a from a use case/presentation on MongoDB website.

k8s/mongodb/container/mongod_entrypoint.bash
muawiakh: Done, I committed the wrong branch, by mistake.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
```diff
@@ -500,6 +500,19 @@ Step 12: Start a Kubernetes StatefulSet for MongoDB
     backend port. Set it to the value specified for ``mongodb-backend-port``
     in the ConfigMap.
 
+  * **Optional**: You can also change the value for ``STORAGE_ENGINE_CACHE_SIZE``, for more information
+    regarding this configuration, please consult the `MongoDB Official
+    Documentation <https://docs.mongodb.com/manual/reference/configuration-options/#storage.wiredTiger.engineConfig.cacheSizeGB>`_.
+
+  * **Optional**: If you are not using the **Standard_D2_v2** virtual machines for Kubernetes agents as per the guide,
```
shahbazn: generally recommended specs would be better instead of azure specific instance type

k8s/mongodb/container/mongod_entrypoint.bash
muawiakh: So in the previous lines, everything is mandatory e.g. `MONGODB_PORT`, `MONGODB_FQDN` etc and we exit if anything is missing but this is an optional argument. If this is not specified we don't need to kill the container.

docs/server/source/production-deployment-template/client-tls-certificate.rst
krish7919: :+1:

k8s/mongodb/container/docker_build_and_push.bash
krish7919: @shahbazn Yes! That's the whole idea.

muawiakh: Yes, we can try to automate in future.

shahbazn: @krish7919 we should come up with better versioning strategy. with increased development and bigger team sizes its gonna get messy really fast if we keep bumping this every time a PR gets merged. may be add build number and update the general availability image for each named release with the release cycle. but we can have further discussion on this once we move towards automating the build process.

muawiakh: Yes, that should be the approach, but we have to adjust to this for now.

docs/server/source/production-deployment-template/node-on-kubernetes.rst
muawiakh: The general specs are down below #L513 and #L514. Referring to Azure specific size here because the template uses that.

PR #1759, opened 2017-09-26
Update BDB Docker image version for k8s

ttmc:
Part of the BigchainDB release process for version 1.1.

@krish7919 Is there anywhere else we should be updating the version number of the bigchaindb/bigchaindb image to 1.1.0? (That's its identity on Docker Hub.)

### DIFF
--- a/k8s/bigchaindb/bigchaindb-dep.yaml
+++ b/k8s/bigchaindb/bigchaindb-dep.yaml
@@ -12,7 +12,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: bigchaindb
-        image: bigchaindb/bigchaindb:1.0.1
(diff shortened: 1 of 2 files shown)


krish7919:
There are 2 files which use the docker image currently:
1. k8s/dev-setup/bigchaindb.yaml
2. k8s/bigchaindb/bigchaindb-dep.yaml

But I can take care of the dev-setup config files in a separate PR.
What would you prefer?

ttmc:
I would prefer to take care of both in a single PR, so I'll add another commit to this PR. Thanks!

I'll make a note about updating `k8s/dev-setup/bigchaindb.yaml` in our official `RELEASE_PROCESS.md` document (after doing this release).

krish7919:
@ttmc Thanks.

PR #1762, opened 2017-09-26
Update nginx config for well being

muawiakh:
- turned off server tokens so the server does not leak nginx information
  on errors and header
- Added header to turn off cross site scripting
- use stable release of nginx instead of mainline
- limit available methods
- update response code

### DIFF
--- a/k8s/nginx-http/container/Dockerfile
+++ b/k8s/nginx-http/container/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.13.1
+FROM nginx:stable
 LABEL maintainer "dev@bigchaindb.com"
 WORKDIR /
 RUN apt-get update \
--- a/k8s/nginx-http/container/nginx.conf.template
+++ b/k8s/nginx-http/container/nginx.conf.template
@@ -80,10 +80,6 @@ http {
       # max client request body size: avg transaction size.
       client_max_body_size 15k;
 
-      if ($request_method !~ ^(GET|OPTIONS|POST)$) {
-        return 444;
-      }
-
       # No auth for GETs, forward directly to BDB.
       if ($request_method = GET) {
         proxy_pass http://$bdb_backend:BIGCHAINDB_API_PORT;
@@ -91,6 +87,7 @@ http {
 
       # POST requests get forwarded to OpenResty instance. Enable CORS too.
       if ($request_method = POST ) {
+        add_header 'Access-Control-Allow-Origin' '*';
         add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
         add_header 'Access-Control-Allow-Headers' 'DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
         add_header 'Access-Control-Expose-Headers' 'DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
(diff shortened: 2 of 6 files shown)


k8s/nginx-http/container/nginx.conf.template
```diff
@@ -45,6 +45,12 @@ http {
 
   keepalive_timeout 60s;
 
+  # Do not expose nginx data/version number in error response and header
+  server_tokens off;
```
krish7919: :+1:

k8s/nginx-http/container/nginx.conf.template
```diff
@@ -45,6 +45,12 @@ http {
...
+
+  # To prevent cross-site scripting
+  add_header X-XSS-Protection "1; mode=block";
```
krish7919: I am skeptical that this setting would break usage of JavaScript driver, where other websites can call `test.ipdb.io` API.

k8s/nginx-http/container/nginx.conf.template
```diff
@@ -75,18 +80,21 @@ http {
       # max client request body size: avg transaction size.
       client_max_body_size 15k;
 
+      if ($request_method !~ ^(GET|OPTIONS|POST)$) {
+        return 444;
+      }
```
krish7919: Minor thing: We can move the complete block to the bottom of all the other 3 `if`s and reduce one if...else lookup in nginx.

muawiakh: +1, will take are of it.

k8s/nginx-http/container/nginx.conf.template
```diff
@@ -75,18 +80,21 @@ http {
...
+
       # No auth for GETs, forward directly to BDB.
       if ($request_method = GET) {
         proxy_pass http://$bdb_backend:BIGCHAINDB_API_PORT;
       }
 
       # POST requests get forwarded to OpenResty instance. Enable CORS too.
       if ($request_method = POST ) {
-        add_header 'Access-Control-Allow-Origin' '*';
```
krish7919: Can we test this using the Javascript driver from another browser please?

k8s/nginx-http/container/nginx.conf.template
muawiakh: I have tested the config with Python driver and direct REST calls and everything seems to be in place.

k8s/nginx-https/container/nginx.conf.template
```diff
@@ -135,7 +147,7 @@ http {
     location / {
         add_header Upgrade "TLS/1.2, HTTP/1.1" always;
         default_type text/plain;
-        return 426 'Consider using the HTTPS protocol next time!';
+        return 301 'Consider using the HTTPS protocol next time!';
```
krish7919: From [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/426):

> The HTTP 426 Upgrade Required client error response code indicates that the server refuses to perform the request using the current protocol but might be willing to do so after the client upgrades to a different protocol.
> The server sends an Upgrade header with this response to indicate the required protocol(s).

krish7919: If you want to use `301`, then maybe add something like:
`return 301 https://$host$request_uri;`

k8s/nginx-http/container/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nginx:1.13.1
+FROM nginx:stable
```
krish7919: :+1:

k8s/nginx-https/container/nginx.conf.template
```diff
@@ -59,11 +65,13 @@ http {
   # Frontend server for the external clients; acts as HTTPS termination point.
   server {
     listen CLUSTER_FRONTEND_PORT ssl;
+    add_header X-XSS-Protection "1; mode=block";
```
krish7919: Do we need this twice?

PR #1772, opened 2017-09-28
Change Group ID to Project ID for Mongo DB Cloud Manager configs

shahbazn:
MongoDB cloud manager UI has been updated and they have changed **Group ID** to **Project ID** and merged **Settings -> Group Settings** into one consolidated panel **Settings**

- This PR updates the k8s docs accordingly

### DIFF
--- a/k8s/configuration/secret.yaml
+++ b/k8s/configuration/secret.yaml
@@ -14,9 +14,9 @@ metadata:
   namespace: default
 type: Opaque
 data:
-  # Base64-encoded Group ID
-  # Group ID used by MongoDB deployment
-  group-id: "<b64 encoded Group ID>"
+  # Base64-encoded Project ID
+  # Project ID used by MongoDB deployment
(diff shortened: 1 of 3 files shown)


k8s/configuration/secret.yaml
```diff
@@ -14,9 +14,9 @@ metadata:
   namespace: default
 type: Opaque
 data:
-  # Base64-encoded Group ID
-  # Group ID used by MongoDB deployment
-  group-id: "<b64 encoded Group ID>"
+  # Base64-encoded Project ID
+  # Project ID used by MongoDB deployment
+  group-id: "<b64 encoded Project ID>"
```
muawiakh: why not change this key to `project-id` as well?

krish7919: Because internally it is still `group-id` in the configuration file.
So, I would maybe change it later when MongoDB starts using `project id` as a parameter in the configuration file.

PR #1781, opened 2017-10-11
Add JS Driver button on home page of root docs

ttmc:
I just noticed that there's no big blue button for the JavaScript Driver on the home page of the root docs.

This PR adds one.


ttmc:
It looks like somebody already went ahead and pushed this change straight to the `master` branch, therefore I will close this pull request.

PR #1796, opened 2017-10-23
Provide log server port as config parameter

### LINKED ISSUE #1660: bigchaindb listens on port 9020 by default, cannot be disabled and port cannot be changed
As far as I can see from the python source, bigchaindb 1.0.0 always opens a logging server on TCP port 127.0.0.1:9020. It seems like this cannot be disabled and also the port number cannot be changed.

This effectively prohibits me from running multiple bigchaindb instances on the same host (say for the purpose of setting up a development system). I'm aware that I could have multiple docker containers, but I prefer to keep things as simple as possible.

kansi:
Converted log server's binding port as a parameter in `.bigchaindb` config file
Fixes #1660

### DIFF
--- a/bigchaindb/log/configs.py
+++ b/bigchaindb/log/configs.py
@@ -63,6 +63,6 @@ SUBSCRIBER_LOGGING_CONFIG = {
     'root': {
         'level': logging.DEBUG,
         'handlers': ['console', 'file', 'errors'],
-        'port': 9020
+        'port': DEFAULT_SOCKET_LOGGING_PORT
     },
 }
--- a/bigchaindb/log/configs.py
+++ b/bigchaindb/log/configs.py
@@ -62,6 +62,7 @@ SUBSCRIBER_LOGGING_CONFIG = {
     'loggers': {},
     'root': {
         'level': logging.DEBUG,
-        'handlers': ['console', 'file', 'errors']
+        'handlers': ['console', 'file', 'errors'],
+        'port': 9020
     },
 }
--- a/bigchaindb/log/setup.py
+++ b/bigchaindb/log/setup.py
@@ -25,17 +25,29 @@ def _normalize_log_level(level):
         raise ConfigurationError('Log level must be a string!') from exc
 
 
-def setup_pub_logger():
+def setup_pub_logger(logging_port=None):
+    if logging_port is None:
+        logging_port = DEFAULT_SOCKET_LOGGING_PORT
+
     dictConfig(PUBLISHER_LOGGING_CONFIG)
     socket_handler = logging.handlers.SocketHandler(
-        DEFAULT_SOCKET_LOGGING_HOST, DEFAULT_SOCKET_LOGGING_PORT)
+        DEFAULT_SOCKET_LOGGING_HOST, logging_port)
     socket_handler.setLevel(logging.DEBUG)
     logger = logging.getLogger()
     logger.addHandler(socket_handler)
 
 
 def setup_sub_logger(*, user_log_config=None):
-    server = LogRecordSocketServer()
+    kwargs = {}
+    logging_port = None
+
+    if user_log_config is not None:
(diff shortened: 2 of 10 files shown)


bigchaindb/log/configs.py
```diff
@@ -62,6 +62,7 @@
     'loggers': {},
     'root': {
         'level': logging.DEBUG,
-        'handlers': ['console', 'file', 'errors']
+        'handlers': ['console', 'file', 'errors'],
+        'port': 9020
```
r-marques: Should we use DEFAULT_SOCKET_LOGGING_PORT instead?

bigchaindb/log/setup.py
```diff
@@ -25,17 +25,29 @@ def _normalize_log_level(level):
         raise ConfigurationError('Log level must be a string!') from exc
 
 
-def setup_pub_logger():
+def setup_pub_logger(logging_port=None):
+    if logging_port is None:
```
r-marques: Maybe we can abbreviate this if statement with

```python
logging_port = logging_port or DEFAULT_SOCKET_LOGGING_PORT
```

bigchaindb/log/configs.py
kansi: Yes, that would be better.

bigchaindb/log/setup.py
```diff
@@ -25,17 +25,29 @@ def _normalize_log_level(level):
...
+        DEFAULT_SOCKET_LOGGING_HOST, logging_port)
     socket_handler.setLevel(logging.DEBUG)
     logger = logging.getLogger()
     logger.addHandler(socket_handler)
 
 
 def setup_sub_logger(*, user_log_config=None):
-    server = LogRecordSocketServer()
+    kwargs = {}
+    logging_port = None
```
r-marques: `logging_port` could be initialized the same way you did in https://github.com/bigchaindb/bigchaindb/pull/1796/files#diff-30313dc6455cdf917cde64a230c4ec99R60 this way the code would be more consistent.

r-marques:
I left some suggestions but other than that looks good 👍

docs/server/source/server-reference/configuration.md
```diff
@@ -319,7 +319,8 @@ holding the logging configuration.
         "granular_levels": {
             "bichaindb.backend": "info",
             "bichaindb.core": "info"
-        }
+        },
+        "port": 7070
```
ttmc: Please add the name of the associated environment variable to the big list at the top of this page as well. I guess it would be `BIGCHAINDB_LOG_PORT` 

Add it right after `BIGCHAINDB_LOG_GRANULAR_LEVELS`

docs/server/source/server-reference/configuration.md
```diff
@@ -533,6 +535,22 @@ logging of the `core.py` module to be more verbose, you would set the
 **Defaults to**: `"{}"`
 
 
+### log.port
```
ttmc: I know this isn't something you did but I just noticed that three lines above this, it says log.granular_levels defaults to `"{}"` but that's not quite right. It's not a string. It defaults to `{}`. We may as well change it in this PR, since it's such a minor cosmetic fix.

kansi: sure

kansi:
@r-marques the build is failing because of `flake8` issue which I fixed [here](https://github.com/bigchaindb/bigchaindb/pull/1797/commits/50eb8571332bfbfa4288952e7cfce99cc695a252). If the [pull request](https://github.com/bigchaindb/bigchaindb/pull/1797) could be merged into the `master` then I could merge the same here and pass the build.

r-marques:
@kansi go ahead and merge the other pr.

kansi:
@r-marques merged the other pr to `master` and updated this branch. I think this is good to go.

PR #1798, opened 2017-10-24
Move implicit database initialization in "bigchaindb start" to command line flag

### LINKED ISSUE #1661: bigchaindb should not create tables automatically
Creating the tables automatically is a bad idea. Assume a cluster of rethinkdb servers which do not have the tables yet. Starting bigchaindb simultaneously on two servers may result in the tables being created on 2 rethinkdb nodes. This will eventually lead to conflicts. Rethinkdb then complains loudly about multiple conflicting tables with the same name.

The tables must be created manually on one node of the rethinkdb cluster. All bigchaindb instances must wait until the tables become available.

I would have hoped that there is at least an option to disable automatic table creation.

kansi:
Migrate database setup during `bigchaindb start` to command line flag `--init`
Fixes #1661

### DIFF
--- a/bigchaindb/commands/bigchaindb.py
+++ b/bigchaindb/commands/bigchaindb.py
@@ -196,7 +196,9 @@ def run_start(args):
         logger.info('RethinkDB started with PID %s' % proc.pid)
 
     try:
-        _run_init()
+        if args.initialize_database:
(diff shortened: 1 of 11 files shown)


r-marques:
I think this ready to merge 👍

PR #1803, opened 2017-10-27
Add secondary index for "id" in bigchain collection.

### LINKED ISSUE #1656: Possible mongodb query not using secondary index.
From @ferOnti in gitter https://gitter.im/bigchaindb/bigchaindb?at=595ad1e86ae41d5d33b052db

> Hi, I was stressing the bigchaindb, and after 20k assets, I found this "slow query" in mongodb.log:

```bash
2017-07-03T23:21:26.490+0000 I COMMAND [conn4274] command bigchain.bigchain command: mapReduce { mapreduce: "bigchain", map: "function () {
        this.block.transactions.forEach(function (tx) {
            var key = tx.operation
            var value = 1
            emit(key, value);  
        })
    }", reduce: "function (key, values) {
        //var reduceVal = {count:0, sum: 0}
        var reduceVal = 0
        values.forEach(function(value) {   
            reduceVal +...", query: {}, out: { inline: 1 } } planSummary: COLLSCAN keysExamined:0 docsExamined:3362 numYields:26 reslen:245 locks:{ Global: { acquireCount: { r: 130 } }, Database: { acquireCount: { r: 4, R: 61 } }, Collection: { acquireCount: { r: 4 } } } protocol:op_query 178ms
```
> I think this query is related to get the last transaction

Did this happened only when inserting assets? Or was there some reads also?

From the `planSummary` we can see that this query is doing a collection scan instead of an index scan.

kansi:
fixes #1656

### DIFF
--- a/bigchaindb/backend/mongodb/schema.py
+++ b/bigchaindb/backend/mongodb/schema.py
@@ -50,6 +50,11 @@ def drop_database(conn, dbname):
 def create_bigchain_secondary_index(conn, dbname):
     logger.info('Create `bigchain` secondary index.')
 
+    # secondary index on block id which is should be unique
+    conn.conn[dbname]['bigchain'].create_index('id',
+                                               name='block_id',
+                                               unique=True)
+
     # to order blocks by timestamp
     conn.conn[dbname]['bigchain'].create_index([('block.timestamp',
                                                  ASCENDING)],
--- a/bigchaindb/backend/mongodb/schema.py
+++ b/bigchaindb/backend/mongodb/schema.py
@@ -52,8 +52,7 @@ def create_bigchain_secondary_index(conn, dbname):
 
     # secondary index on block id which is should be unique
     conn.conn[dbname]['bigchain'].create_index('id',
-                                               name='block_id',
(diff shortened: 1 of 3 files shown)


bigchaindb/backend/mongodb/schema.py
```diff
@@ -50,6 +50,10 @@ def drop_database(conn, dbname):
 def create_bigchain_secondary_index(conn, dbname):
     logger.info('Create `bigchain` secondary index.')
 
+    # secondary index on block id which is should be unique
```
r-marques: small typo

bigchaindb/backend/mongodb/schema.py
```diff
@@ -50,6 +50,10 @@ def drop_database(conn, dbname):
...
+    conn.conn[dbname]['bigchain'].create_index('id',
```
r-marques: If the index needs to be unique we need to specify it with `unique=True`.

kansi: Yeah, missed that, my bad.

PR #1805, opened 2017-10-30
Validation for "language" key in asset.data object

### LINKED ISSUE #1670: MongoDB text search does not support Star Wars languages
When creating a text index with MongoDB the index is created with a default language (usually en).
MongoDB also allow us to override the language used to create the text index for a particular document by adding the field `language` inside the document being written. From the [documentation](https://docs.mongodb.com/manual/tutorial/specify-language-for-text-index/):

> If a collection contains documents or embedded documents that are in different languages, include a field named language in the documents or embedded documents and specify as its value the language for that document or embedded document.
MongoDB will use the specified language for that document or embedded document when building the text index

If you specify a language that is not supported by MongoDB the insert operation will fail. This is a problem for us because we don't know how BigchainDB clients will use the asset field.

For example I was testing with star wars data from [swapi.co](http://swapi.co/):
```json
{
  "average_height": "300",
  "language": "Huttese",
  "url": "http://swapi.co/api/species/5/",
  "eye_colors": "yellow, red",
  "name": "Hutt",
  "hair_colors": "n/a",
  "people": [
    "http://swapi.co/api/people/16/"
  ],
  "edited": "2014-12-20T21:36:42.146000Z",
  "homeworld": "http://swapi.co/api/planets/24/",
  "classification": "gastropod",
  "films": [
    "http://swapi.co/api/films/3/",
    "http://swapi.co/api/films/1/"
  ],
  "designation": "sentient",
  "average_lifespan": "1000",
  "skin_colors": "green, brown, tan",
  "created": "2014-12-10T17:12:50.410000Z"
}
```

When trying to insert this asset MongoDB fails with `WriteError: language override unsupported: Huttese`.

We need to somehow either disable language override so that `language` fields in documents do not affect how the text index is built, or default to some language when there is an error.

kansi:
Fixes #1670 for BigchainDB v2.0 by changing the schema of the transaction model (i.e. by restricting the allowed values of `"language"` in assets).

### DIFF
--- a/bigchaindb/common/schema/transaction.yaml
+++ b/bigchaindb/common/schema/transaction.yaml
@@ -117,6 +117,25 @@ definitions:
         anyOf:
         - type: object
           additionalProperties: true
+          patternProperties:
+            language:
+              type: string
+              enum:
+                - danish
+                - dutch
+                - english
+                - finnish
+                - french
+                - german
+                - hungarian
+                - italian
+                - norwegian
+                - portuguese
+                - romanian
+                - russian
+                - spanish
+                - swedish
+                - turkish
         - type: 'null'
   output:
     type: object
--- a/tests/web/test_transactions.py
+++ b/tests/web/test_transactions.py
@@ -47,6 +47,42 @@ def test_post_create_transaction_endpoint(b, client):
     assert res.json['outputs'][0]['public_keys'][0] == user_pub
 
 
+@pytest.mark.parametrize("language,expected_status_code", [
+    ('danish', 202),
+    ('dutch', 202),
+    ('english', 202),
+    ('finnish', 202),
+    ('french', 202),
+    ('german', 202),
+    ('hungarian', 202),
+    ('italian', 202),
+    ('norwegian', 202),
+    ('portuguese', 202),
+    ('romanian', 202),
+    ('russian', 202),
+    ('spanish', 202),
+    ('swedish', 202),
+    ('turkish', 202),
+    ('any', 400),
+])
+@pytest.mark.bdb
+def test_post_create_transaction_with_language(b, client, language, expected_status_code):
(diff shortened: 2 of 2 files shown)


ttmc:
We'll have to update the docs about the asset model as well. Those docs will be moving under the [IPDB Protocol docs](https://github.com/ipdb/ipdb-protocol).

bigchaindb/common/schema/transaction.yaml
```diff
@@ -117,6 +117,25 @@ definitions:
         anyOf:
         - type: object
           additionalProperties: true
+          patternProperties:
+            language:
```
r-marques: Will this be applied to any `language` key within any nested document?
Also should we support two and three letter codes https://docs.mongodb.com/manual/reference/text-search-languages/?

kansi: No. I think we can only handle this at the API level in #1806 . Also, I think it would be sufficient to just have the full names @ttmc What do you think?

ttmc: I'd prefer to allow both the long and short (two-letter) names for the languages, just like MongoDB.

kansi: Sure, then we will keep the list as it is in #1806

r-marques: Maybe I am missing something but I don't understand the reason for this change.
This will not handle the case where the `language` is in a nested document and https://github.com/bigchaindb/bigchaindb/pull/1806 already covers all cases.

kansi: @r-marques sorry for confusion, this pull request is not required anymore. We can close it.

kansi:
Closed as #1806 is the correct way to fix this.

PR #1806, opened 2017-10-31
Api level fix for asset language

### LINKED ISSUE #1670: MongoDB text search does not support Star Wars languages

kansi:
API-level fix for the MongoDB `language` issue targeting the v1.x series.

It doesn't change the schema of the transaction model. It just makes an aspect of the schema explicit which until now has been implicit, i.e. if the backend database is MongoDB, then there is a list of allowed values for "language".

Fixes #1670

### DIFF
--- a/docs/root/source/bft.md
+++ b/docs/root/source/bft.md
@@ -1,6 +1,8 @@
 # BigchainDB and Byzantine Fault Tolerance
 
-While BigchainDB is not currently [Byzantine fault tolerant (BFT)](https://en.wikipedia.org/wiki/Byzantine_fault_tolerance), we plan to offer it as an option.
-We anticipate that turning it on will cause a severe dropoff in performance. See [Issue #293](https://github.com/bigchaindb/bigchaindb/issues/293).
+While BigchainDB is not currently [Byzantine fault tolerant (BFT)](https://en.wikipedia.org/wiki/Byzantine_fault_tolerance), we plan to offer it as an option. 
+Update Nov 2017: we're actively working on this, the next release or two will likely have support. More details to come in blog form and github issues
+
+Related issue: [Issue #293](https://github.com/bigchaindb/bigchaindb/issues/293). We anticipate that turning on BFT will cause a dropoff in performance (for a gain in security).
 
 In the meantime, there are practical things that one can do to increase security (e.g. firewalls, key management, and access controls).
--- a/tests/README.md
+++ b/tests/README.md
@@ -105,17 +105,17 @@ $ docker-compose build
 First, start `RethinkDB` in the background:
 
 ```text
-$ docker-compose up -d rdb
+$ docker-compose -f docker-compose.rdb.yml up -d rdb
 ```
 
 then run the tests using:
 
 ```text
-$ docker-compose run --rm bdb-rdb py.test -v
+$ docker-compose -f docker-compose.rdb.yml run --rm bdb-rdb py.test -v
 ```
 
 to rebuild all the images (usually you only need to rebuild the `bdb` and
- `bdb-rdb` images).
+ `bdb-rdb` images). If that fails, then do `make clean-pyc` and try again.
 
 ## Automated Testing of All Pull Requests
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,6 +25,15 @@ USER_PRIVATE_KEY = '8eJ8q9ZQpReWyQT5aFCiwtZ5wDZC4eDnCen88p3tQ6ie'
 USER_PUBLIC_KEY = 'JEAkEJqLbbgDRAtMm8YAjGp759Aq2qTn9eaEHUj2XePE'
 
 
+def pytest_runtest_setup(item):
+    if isinstance(item, item.Function):
+        if item.get_marker('skip_travis_rdb'):
+            if (os.getenv('TRAVIS_CI') == 'true' and
+                    os.getenv('BIGCHAINDB_DATABASE_BACKEND') == 'rethinkdb'):
+                pytest.skip(
+                    'Skip test during Travis CI build when using rethinkdb')
+
+
 def pytest_addoption(parser):
     from bigchaindb.backend.connection import BACKENDS
 
--- a/tests/integration/test_federation.py
+++ b/tests/integration/test_federation.py
@@ -97,6 +97,7 @@ def process_vote(steps, result=None):
 
 @pytest.mark.bdb
 @pytest.mark.genesis
+@pytest.mark.skip_travis_rdb
 def test_elect_valid(federation_3):
     [bx, (s0, s1, s2)] = federation_3
     tx = input_single_create(bx[0])
@@ -115,6 +116,7 @@ def test_elect_valid(federation_3):
 
 
 @pytest.mark.bdb
+@pytest.mark.skip_travis_rdb
 @pytest.mark.genesis
 def test_elect_invalid(federation_3):
     [bx, (s0, s1, s2)] = federation_3
@@ -135,6 +137,7 @@ def test_elect_invalid(federation_3):
 
 @pytest.mark.bdb
 @pytest.mark.genesis
+@pytest.mark.skip_travis_rdb
 def test_elect_sybill(federation_3):
     [bx, (s0, s1, s2)] = federation_3
     tx = input_single_create(bx[0])
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -5,6 +5,7 @@ import pytest
 pytestmark = [pytest.mark.bdb, pytest.mark.usefixtures('processes')]
 
 
+@pytest.mark.skip_travis_rdb
 def test_double_create(b, user_pk):
     from bigchaindb.models import Transaction
     from bigchaindb.backend.query import count_blocks


bigchaindb/backend/schema.py
```diff
@@ -16,10 +16,14 @@
...
 import bigchaindb
 from bigchaindb.backend.connection import connect
+from bigchaindb.common.exceptions import ValidationError
 
 logger = logging.getLogger(__name__)
 
 TABLES = ('bigchain', 'backlog', 'votes', 'assets')
+VALID_LANGUAGES = ('danish' 'dutch' 'english' 'finnish' 'french' 'german'
+                   'hungarian' 'italian' 'norwegian' 'portuguese' 'romanian'
+                   'russian' 'spanish' 'swedish' 'turkish')
```
ttmc: Why are there no commas between these strings, similar to `TABLES`?

Also, according to https://docs.mongodb.com/manual/reference/text-search-languages/ , MongoDB text search also supports some two- and three-letter language codes such as `en`, `ara` and `zhs`.

bigchaindb/backend/schema.py
```diff
@@ -99,3 +103,21 @@ def init_database(connection=None, dbname=None):
     create_database(connection, dbname)
     create_tables(connection, dbname)
     create_indexes(connection, dbname)
+
+
+def validate_if_exists_asset_language(tx_body):
+    data = tx_body['asset'].get('data', {})
+
+    if data and 'language' in data:
```
ttmc: Maybe I shouldn't, but I always get concerned about relational operator precedence, such as whether `and` or `in` takes precedence. You could assuage that concern by putting `'language' in data` inside parentheses.

bigchaindb/backend/schema.py
```diff
@@ -99,3 +103,21 @@ def init_database(connection=None, dbname=None):
...
+
+        language = data.get('language')
+        backend = bigchaindb.config['database']['backend']
+
+        if backend == 'mongodb' and language not in VALID_LANGUAGES:
```
ttmc: Similar to my previous remark, I'd have much more peace of mind if this was written
```python
if (backend == 'mongodb') and (language not in VALID_LANGUAGES):
```

bigchaindb/backend/schema.py
```diff
@@ -21,9 +21,11 @@
...
 
 TABLES = ('bigchain', 'backlog', 'votes', 'assets')
-VALID_LANGUAGES = ('danish' 'dutch' 'english' 'finnish' 'french' 'german'
-                   'hungarian' 'italian' 'norwegian' 'portuguese' 'romanian'
-                   'russian' 'spanish' 'swedish' 'turkish')
+VALID_LANGUAGES = ('danish', 'dutch', 'english', 'finnish', 'french', 'german',
+                   'hungarian', 'italian', 'norwegian', 'portuguese', 'romanian',
+                   'russian', 'spanish', 'swedish', 'turkish',
+                   'da', 'nl', 'en', 'fi', 'fr', 'de', 'hu', 'it', 'nb', 'pt',
+                   'ro', 'ru', 'es', 'sv', 'tr')
```
ttmc: There are still some languages missing here: `ara`, `prs`, `pes`, `urd`, `zhs`, and `zht` (from https://docs.mongodb.com/manual/reference/text-search-languages/ ).

ttmc: `none` is also a valid value for language.

ttmc:
@kansi Do you think it would make sense to say something about the special "language" key/field in the docs about the asset model?

bigchaindb/backend/schema.py
ttmc: I wonder if the long-form versions of those other languages also works, e.g. `traditional chinese`

kansi:
@ttmc added the suggested changes. Regarding docs, I think we should update the asset model docs with a note which describes the benifits and issues of using `language` key in the asset data model.

ttmc:
@kansi I reloaded the page in GitHub but I'm not seeing support for the other "language" values I mentioned above:  https://github.com/bigchaindb/bigchaindb/pull/1806#pullrequestreview-73133273

Also, please make the suggested changes to the asset model docs _in this pull request_.

ttmc:
I have another question:  Will the "language" issue also be a problem for the value of "metadata" (which can also contain arbitrary JSON)? Or is it an issue that only arises when there's full-text search for the thing in question?

bigchaindb/backend/schema.py
```diff
@@ -99,3 +105,21 @@ def init_database(connection=None, dbname=None):
...
     create_tables(connection, dbname)
     create_indexes(connection, dbname)
+
+
+def validate_if_exists_asset_language(tx_body):
+    data = tx_body['asset'].get('data', {})
+
+    if data and ('language' in data):
+
+        language = data.get('language')
```
r-marques: This only checks the `language` is set in the top level document. With MongoDB the `language` can be overridden in embedded documents.

Something like:
```python
asset={'msg': {'language': 'hello'}}
```
will pass the check but will still fail when writing to mongodb.

kansi:
@ttmc @r-marques  I think I misunderstood when you pointed this issue earlier. I thought you were suggesting to include the 2 letter codes for languages. 
Nevertheless, before we add the remaining languages we have to consider the fact that some of the languages (`Arabic`, `Farsi`, `Urdu`, `Simplified Chinese`, and `Traditional Chinese`) seem to be only supported by the enterprise version. So, if we include all these in our set of valid languages and user doesn't use the enterpirse version of MongoDB but still specifies on langauages only supported in the enterprise version then we are again left with the same issue of invalid langauge. 

One way to solve this would be to just add a note in the docs regarding community and enterprise languages and warn the user.

Other way to solve this would be to make supported languages a part of the config and validate the `language` field from the config. Thereafter if the user decides to use a language from the enterprise version they can simply added it to the list of languages. But here  again we risk the user mispelling the language name which would again lead to the same issue.

A third way would be to only allow languages supported in the community version. This will limit the languages users can user but ensure that the language issue will never arise.

r-marques:
I think the easiest would be to only support languages that are in the community version.

ttmc:
Oh wow, I didn't notice the bit about some languages only being supported in the enterprise version of MongoDB. What a mess...

I agree with @r-marques --- let's just support the community version for now.

I think `"none"` is still a valid value for `"language"`, even in the community version.

kansi:
@r-marques We consulted with a member from the MongoDB team regarding whether the db has an API to list the supported lanaguages. They responsded that it may be introduced in the newer versions. So, till that time we would just support the community lanaguages?

kansi:
@ttmc The `metadata` object should not have any issue as it is not indexed.

kansi:
@ttmc Ok, So the community supported languages are the ones that you see in this commit. If will add `none` to this list and update the docs.

bigchaindb/backend/schema.py
ttmc: @kansi This comment remains unresolved.

bigchaindb/backend/schema.py
ttmc: Okay we decided to allow only those languages supported by MongoDB community edition (so not `ara` etc.).

`'none'` still needs to be added to the above.

kansi: Sure, I will add `'none'` to the list and update asset docs.

bigchaindb/backend/schema.py
kansi: @ttmc   @r-marques is right. we will need to interate over all the nested keys.

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
...
+
+
+def validate_if_exists_language(obj, key):
+    """Validate all nested "language" key in `obj`.
+
+       Args:
+           obj (dict): dictonary whose "language" key is to be validated.
+
+       Returns:
+           None: validation successfull
```
vrde: successfull → successful

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
...
+
+def validate_language(value):
+    """Check if `value` is a valid language
+       https://docs.mongodb.com/manual/reference/text-search-languages/
+
+        Args:
+            value (str): language to validated
+
+        Returns:
+            None: validation successfull
```
vrde: successfull → successful

bigchaindb/common/utils.py
```diff
@@ -97,6 +97,30 @@ def validate_all_keys(obj_name, obj, validation_fun):
...
+       using `validation_fun`
+
+        Args:
+            obj (dict): dictonary object.
+            key (str): key whose value is to be validated.
+            validation_fun (function): function used to validate the value
+            of `key`.
+
+        Returns:
+            None: indicates validation successfull
```
vrde: successfull → successful

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
...
+
+        Raises:
+            ValidationError: raises execption incase language is not valid.
+    """
+    if value not in VALID_LANGUAGES:
+        error_str = ('MongoDB does not support text search for the '
+                     'language "{}". If you do not understand this error '
+                     'message then please rename key/field "language" to '
+                     'something else like "lang".').format(value)
+        raise ValidationError(error_str) from ValueError()
```
vrde: the `raise ... from` idiom is used to raise a new exception keeping the context of another exception that already happened. I'm not sure the usage here is correct, I guess the correct line should be just a simple `raise ValidationError(error_str)`.

bigchaindb/common/utils.py
```diff
@@ -97,6 +97,30 @@ def validate_all_keys(obj_name, obj, validation_fun):
...
+
+        Raises:
+            ValidationError: `validation_fun` will raise this error on failure
+    """
+    for vkey, value in obj.items():
+        if vkey == key:
+            validation_fun(value)
+        elif type(value) is dict:
+            validate_all_value_for_key(value, key, validation_fun)
+    return
```
vrde: If you agree with the previous comment, then you can omit the `return` here.

bigchaindb/common/utils.py
vrde: This function doesn't provide to the caller any useful return value, what do you think about removing the `Returns` part? As an alternative idea, you can return `obj` on success.

bigchaindb/common/utils.py
```diff
@@ -97,6 +97,30 @@ def validate_all_keys(obj_name, obj, validation_fun):
...
+        Returns:
+            None: indicates validation successfull
+
+        Raises:
+            ValidationError: `validation_fun` will raise this error on failure
+    """
+    for vkey, value in obj.items():
+        if vkey == key:
+            validation_fun(value)
+        elif type(value) is dict:
```
vrde: I'd [suggest](https://stackoverflow.com/a/1549854/597097) using `isinstance(value, dict)`.

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
...
+       Returns:
+           None: validation successfull
+
+        Raises:
+            ValidationError: raises execption incase language is not valid.
+    """
+    backend = bigchaindb.config['database']['backend']
+
+    if backend == 'mongodb':
+        data = obj.get(key, {}) or {}
```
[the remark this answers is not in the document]
kansi: Incase the value of `key` is `None` i.e. `{key: None}` then the expression `obj.get(key, {})` will resolve to `None` so we need `or {}` to return `{}`

bigchaindb/common/utils.py
kansi: We can remove the `Returns` as its exactly correct. Furthermore, I think we could return `True` instead of `obj` because intutively a validation function should return `True` or `False`?

bigchaindb/common/utils.py
kansi: @vrde I think I will omit `return` as it makes the code a little more homogenous to rest of the code base.

bigchaindb/common/utils.py
```diff
@@ -92,9 +92,28 @@ def validate_all_keys(obj_name, obj, validation_fun):
...
             validate_all_keys(obj_name, value, validation_fun)
-    return
+
+
+def validate_all_value_for_key(obj, key, validation_fun):
+    """Validates value for all (nested) occurences of `key` in `obj`
+       using `validation_fun`
+
+        Args:
+            obj (dict): dictonary object.
```
vrde: dictonary → dictionary

bigchaindb/common/utils.py
```diff
@@ -92,9 +92,28 @@ def validate_all_keys(obj_name, obj, validation_fun):
     """
     for key, value in obj.items():
         validation_fun(obj_name, key)
-        if type(value) is dict:
+        if isinstance(value, dict):
             validate_all_keys(obj_name, value, validation_fun)
-    return
+
+
+def validate_all_value_for_key(obj, key, validation_fun):
```
vrde: `validate_all_value_for_key` → `validate_all_values_for_key` ?

bigchaindb/common/utils.py
```diff
@@ -92,9 +92,28 @@ def validate_all_keys(obj_name, obj, validation_fun):
...
+    """Validates value for all (nested) occurences of `key` in `obj`
```
vrde: Should be "Validate", I'm not sure we documented this but we prefer imperative sentences. See also [Good Python docstrings by example](http://blog.dolphm.com/pep257-good-python-docstrings-by-example/).

bigchaindb/common/utils.py
vrde: `validate_all_value_for_key` → `validate_all_values_for_key`?

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
     create_database(connection, dbname)
     create_tables(connection, dbname)
     create_indexes(connection, dbname)
+
+
+def validate_language_key(obj, key):
+    """Validate all nested "language" key in `obj`.
+
+       Args:
+           obj (dict): dictonary whose "language" key is to be validated.
```
vrde: dictonary → dictionary

bigchaindb/backend/schema.py
```diff
@@ -99,3 +106,43 @@ def init_database(connection=None, dbname=None):
...
+
+       Returns:
+           None: validation successful
+
+        Raises:
+            ValidationError: raises execption incase language is not valid.
```
vrde: execption → exception
incase → in case

bigchaindb/common/utils.py
kansi: sad to see myself making these mistakes !

bigchaindb/backend/schema.py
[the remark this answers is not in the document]
vrde: Good point. What if key points to something that is not `None` and it's not a `dict`, like `{key: 'foobar'}`?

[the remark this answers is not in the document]
ttmc: @kansi Since both @vrde and I asked about this, please add a comment line in the code to explain why the ` or {}` is there.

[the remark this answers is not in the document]
kansi: @vrde that is a good point. @ttmc @vrde I think we should do the following which will help us void the confusion,
```python
        data = obj.get(key, {})
        if isinstance(data, dict):
            validate_all_values_for_key(data, 'language', validate_language)
```
I think the above is much clearer and correctly handles cases @vrde pointed out.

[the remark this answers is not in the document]
ttmc: This does seem clearer.

vrde:
:clap: good job @kansi!

ttmc:
@sbellem already merged #1821 so this PR should use the same code for skipping those tests.

PR #1811, opened 2017-11-06
Updated metadata and asset model docs

### LINKED ISSUE #1810: Update the docs about metadata and asset.data
Because of recent pull requests https://github.com/bigchaindb/bigchaindb/pull/1806 and https://github.com/bigchaindb/bigchaindb/pull/1807 we need to update the docs:

1. The keys (strings) in `metadata` and `asset.data` can't contain the `.` character or the null character (Unicode code point 0000). They also can't begin with `$`.
2. The value of `language` in `asset.data` (at any level in the hierarchy) must be one of the allowed values. See https://docs.mongodb.com/manual/reference/text-search-languages/ and note that only the languages supported by MongoDB community edition are allowed.

kansi:
Fixes #1810

### DIFF
--- a/bigchaindb/backend/schema.py
+++ b/bigchaindb/backend/schema.py
@@ -16,10 +16,17 @@ import logging
 
 import bigchaindb
 from bigchaindb.backend.connection import connect
+from bigchaindb.common.exceptions import ValidationError
+from bigchaindb.common.utils import validate_all_values_for_key
 
 logger = logging.getLogger(__name__)
 
 TABLES = ('bigchain', 'backlog', 'votes', 'assets')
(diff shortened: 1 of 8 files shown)


docs/server/source/data-models/asset-model.md
```diff
@@ -18,3 +18,11 @@ In a `TRANSFER` transaction, the `"asset"` must contain exactly one key-value pa
     "id": "38100137cea87fb9bd751e2372abb2c73e7d5bcf39d940a5516a324d9c7fb88d"
 }
 ```
+
+
+.. note::
```
ttmc: This is a Markdown file so the `.. note::` thing isn't going to work. (That's a Sphinx reStructuredText thing.) I will make the edit since there are some other minor things I'd like to change.

PR #1812, opened 2017-11-07
Metadata text search api

### LINKED ISSUE #1592: Text search on transactions metadata
Similar to the text search provided via `/assets?search=` for assets, we'd like to have the same functionality for a transactions meta data.

Internally for BigchainDB this would mean:

- splitting meta data and transaction apart in the underlying data base
- content linking between the two
- implement an endpoint similar to `/asset?search=` for meta data

This use case was put forth by a high priority customer.

kansi:
Fixes #1592

### DIFF
--- a/bigchaindb/backend/mongodb/query.py
+++ b/bigchaindb/backend/mongodb/query.py
@@ -366,10 +366,7 @@ def get_new_blocks_feed(conn, start_block_id):
 
 @register_query(MongoDBConnection)
 def text_search(conn, search, *, language='english', case_sensitive=False,
-                diacritic_sensitive=False, text_score=False, limit=0, table=None):
-    if table is None:
-        table = 'assets'
-
+                diacritic_sensitive=False, text_score=False, limit=0, table='assets'):
     cursor = conn.run(
         conn.collection(table)
         .find({'$text': {
--- a/bigchaindb/backend/mongodb/query.py
+++ b/bigchaindb/backend/mongodb/query.py
@@ -366,9 +366,12 @@ def get_new_blocks_feed(conn, start_block_id):
 
 @register_query(MongoDBConnection)
 def text_search(conn, search, *, language='english', case_sensitive=False,
-                diacritic_sensitive=False, text_score=False, limit=0):
+                diacritic_sensitive=False, text_score=False, limit=0, table=None):
+    if table is None:
+        table = 'assets'
+
     cursor = conn.run(
-        conn.collection('assets')
+        conn.collection(table)
         .find({'$text': {
                 '$search': search,
                 '$language': language,
@@ -381,7 +384,7 @@ def text_search(conn, search, *, language='english', case_sensitive=False,
     if text_score:
         return cursor
 
-    return (_remove_text_score(asset) for asset in cursor)
+    return (_remove_text_score(obj) for obj in cursor)
 
 
 def _remove_text_score(asset):
--- a/bigchaindb/backend/mongodb/query.py
+++ b/bigchaindb/backend/mongodb/query.py
@@ -266,10 +266,28 @@ def write_assets(conn, assets):
 
 
 @register_query(MongoDBConnection)
+def write_metadata(conn, metadata):
+    try:
+        return conn.run(
+            conn.collection('metadata')
+            .insert_many(metadata, ordered=False))
+    except OperationError:
+        return
+
+
+@register_query(MongoDBConnection)
 def get_assets(conn, asset_ids):
     return conn.run(
         conn.collection('assets')
         .find({'id': {'$in': asset_ids}},
+              projection={'_id': False}))
+
+
+@register_query(MongoDBConnection)
+def get_metadata(conn, txn_ids):
+    return conn.run(
+        conn.collection('metadata')
+        .find({'id': {'$in': txn_ids}},
               projection={'_id': False}))
 
 
--- a/bigchaindb/core.py
+++ b/bigchaindb/core.py
@@ -672,7 +672,7 @@ class Bigchain(object):
         """
         return backend.query.write_metadata(self.connection, metadata)
 
-    def text_search(self, search, *, limit=0):
+    def text_search(self, search, *, limit=0, table=None):
         """
         Return an iterator of assets that match the text search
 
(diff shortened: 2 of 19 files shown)


bigchaindb/backend/mongodb/query.py
```diff
@@ -348,9 +366,12 @@ def get_new_blocks_feed(conn, start_block_id):
 
 @register_query(MongoDBConnection)
 def text_search(conn, search, *, language='english', case_sensitive=False,
-                diacritic_sensitive=False, text_score=False, limit=0):
+                diacritic_sensitive=False, text_score=False, limit=0, table=None):
+    if table is None:
```
r-marques: I think it would be cleaner and less error prone to make table a required parameter or at least change the default value of table from `None` to `assets` to eliminate the need for this `if` statement.

bigchaindb/core.py
```diff
@@ -390,6 +395,13 @@ def get_spent(self, txid, output):
                     ' with the chain'.format(txid))
             # if its not and invalid transaction
             if status is not None:
+                if 'metadata' not in transaction:
```
r-marques: Isn't this already handled by `Transaction.from_db` in `Bigchain.get_transaction`?

kansi: @r-marques `transactions` here are fetched using `backend.query.get_spent`. So, they don't have metadata preloaded. We could assign transaction to a variable here `_, status = self.get_transaction ...` rather than ignoring which would help us avoid the `if`?

bigchaindb/web/views/metadata.py
```diff
@@ -0,0 +1,50 @@
...
+
+from bigchaindb.backend.exceptions import OperationError
+from bigchaindb.web.views.base import make_error
+
+logger = logging.getLogger(__name__)
+
+
+class MetadataApi(Resource):
+    def get(self):
+        """API endpoint to perform a text search on transaction metadata.
```
ttmc: This new HTTP API endpoint must be documented in the page about the HTTP API, similar to the documentation about the asset text search endpoint: https://docs.bigchaindb.com/projects/server/en/master/http-client-server-api.html#assets

bigchaindb/models.py
```diff
@@ -352,11 +361,15 @@ def from_db(cls, bigchain, block_dict, from_dict_kwargs=None):
...
+        txn_ids = cls.get_txn_ids(block_dict)
+        metadata = bigchain.get_metadata(txn_ids)
+        # reconstruct block
         block_dict = cls.couple_assets(block_dict, assets)
+        block_dict = cls.couple_metadata(block_dict, metadata)
         kwargs = from_dict_kwargs or {}
         return cls.from_dict(block_dict, **kwargs)
 
-    def decouple_assets(self):
+    def decouple_assets(self, block_dict=None):
```
r-marques: Why do we need that additional optional parameter?

bigchaindb/models.py
```diff
@@ -365,6 +378,9 @@ def decouple_assets(self):
             the block being the dict of the block with no assets in the CREATE
             transactions.
         """
+        if block_dict is None:
```
r-marques: This seems irrelevant. `block_dict = deepcopy(self.to_dict())` is always called regardless.

bigchaindb/models.py
```diff
@@ -376,6 +392,28 @@ def decouple_assets(self):
...
+            tuple: (metadatas, block) with the metadatas being a list of dict/null and
+            the block being the dict of the block with no metadata in any transaction.
+        """
+        if block_dict is None:
+            block_dict = deepcopy(self.to_dict())
+
+        metadatas = []
+        for transaction in block_dict['block']['transactions']:
+            metadata = transaction.pop('metadata')
+            if isinstance(metadata, dict):
```
r-marques: `metadata` is either `None` or a `dict`. This is enforced by the schema validation. So a check to see if it's not None should be enough, right?

bigchaindb/models.py
```diff
@@ -376,6 +392,28 @@ def decouple_assets(self):
...
+                metadata.update({'id': transaction['id']})
```
r-marques: We should probably nest the metadata under a new key or else if the metadata contains a key named `id` this will override it. We can probably have a data model similar to the asset. E.g.:
```python
{
     'id': '<transaction id>'
    'metadata': '<the metadata dict>'
}
```

bigchaindb/models.py
```diff
@@ -402,6 +440,39 @@ def couple_assets(block_dict, assets):
...
+                a database call.
+
+        Returns:
+            dict: The dict of the reconstructed block.
+        """
+        # create a dict with {'<txid>': metadata}
+        metadatal = {m.pop('id'): m for m in metadatal}
+        # add the metadata to their corresponding transactions
+        for transaction in block_dict['block']['transactions']:
+            if 'metadata' not in transaction:
```
r-marques: `metadata` should not be in the transaction at this point.

bigchaindb/models.py
kansi: agreed.

bigchaindb/models.py
kansi: this is done in order to make the interface same as `decouple_metadata`.

bigchaindb/core.py
r-marques: I think it would be better to use the transaction returned by `get_transaction`. This way we reduce the number of database calls and we eliminate the need to handle the metadata.

bigchaindb/models.py
kansi: That is true as we have only one place where `decouple_assets` is called but I tired to generalize and make the interface for `decouple_assets` and `decouple_metadata` similar. We can do it either way

bigchaindb/web/views/metadata.py
kansi: Sure.

bigchaindb/models.py
r-marques: That is ok but the logic here is still wrong. `block_dict = deepcopy(self.to_dict())` is always called  and the `if` statement is irrelevant.

kansi: @r-marques Its my bad, I didn't realize you were talking about [this](https://github.com/bigchaindb/bigchaindb/pull/1812/files#diff-9f1ab5b5daf1b3072e2123b6939bfb07R383)?

ttmc:
@kansi I don't know why codecov is upset about a supposed 0.06% change in coverage. I wouldn't worry about it.

kansi:
@ttmc sure, just need a final review from @r-marques then I will merge it.

PR #1820, opened 2017-11-08
Do not use xdist for RethinkDB tests

vrde:
Tests should be independent, running them in parallel using `xdist` should not be a problem. @kansi found that changing the order of few tests under `tests/integration/test_integration.py` changes the outcome of the test suite, so my guess is that disabling `xdist` for now should make the test suite pass.

This is *not* a nice way to fix the issue.


vrde:
Closing because @kansi and I are thinking of another approach.

PR #1821, opened 2017-11-08
Bypass CI issues for rethinkdb

kansi:


tests/integration/test_federation.py
```diff
@@ -7,6 +7,7 @@
 from contextlib import contextmanager
 from bigchaindb.common.crypto import generate_key_pair
 from tests.pipelines.stepping import create_stepper
+from tests.conftest import skip_travis_rdb
```
sbellem: Is this import really needed?

kansi: Is there another way I can access `skip_travis_rdb` in this module?

kansi: @sbellem I think I can directly do this as `@pytest.mark.usefixtures("skip_travis_rdb")`

sbellem:
**NOTE**: The approach used was based on https://docs.pytest.org/en/latest/example/markers.html#marking-platform-specific-tests-with-pytest

PR #1822, opened 2017-11-09
Flat UTXO collection and first integration with Tendermint™

vrde:


bigchaindb/models.py
```diff
@@ -30,7 +30,11 @@ def validate(self, bigchain):
         """
         input_conditions = []
 
-        if self.operation == Transaction.TRANSFER:
+        if self.operation == Transaction.CREATE:
+            if bigchain.get_transaction(self.to_dict()['id']):
+                raise DoubleSpend('transaction `{}` already exists'
```
ttmc: The problem here isn't a "DoubleSpend", it's more like a DuplicateTransaction. Bonus: we already have that exception:

https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/exceptions.py#L103

In fact, you could even remove the `if self.operation == Transaction.CREATE:` and check _all_ transactions to see if they are duplicates, since a duplicate TRANSFER is also not allowed (but caught by other, more expensive logic later).

vrde: Oh great! Thanks for pointing this out :dancer:

Issue #1808, opened 2017-11-01
Update quickstart documentation for Docker for Mac


muawiakh:
* BigchainDB version: `N/A`
* Operating System: `MacOS`
* Deployment Type: `Docker`
* BigchainDB driver: `no`

### Description
Currently, the [Quickstart](https://docs.bigchaindb.com/projects/server/en/latest/appendices/run-with-docker.html) documentation for to run BigchainDB with Docker containers, for MacOS ,assumes Docker is installed and running with Docker machine and doesn't cater for the latest `Docker for Mac` installations and the guide is currently broken for that. 

### Steps to Reproduce
- Download [Docker for Mac](https://docs.docker.com/docker-for-mac/install/) on your Mac.
- Follow the [Quickstart](https://docs.bigchaindb.com/projects/server/en/latest/appendices/run-with-docker.html) guide for dockers

muawiakh:
Fixed.

PR #1825, opened 2017-11-09
Update quickstart guide for 'docker for mac' users

muawiakh:


docs/server/source/appendices/run-with-docker.md
```diff
@@ -30,6 +30,17 @@ docker run \
   -y configure \
   [mongodb|rethinkdb]
 
+# For Docker for Mac users
+docker run \
+  --interactive \
+  --rm \
+  --tty \
+  --volume $HOME/bigchaindb_docker:/data \
+  --env BIGCHAINDB_DATABASE_HOST=172.17.0.1 \
```
shahbazn: why do we need a hard coded value here?

docs/server/source/appendices/run-with-docker.md
```diff
@@ -78,6 +89,16 @@ docker run \
   --restart=always \
   --volume $HOME/bigchaindb_docker:/data \
   rethinkdb:2.3
+
+# For Docker for Mac users
```
shahbazn: can you please add the docker version pre requisit since newer version dont use docker-machine while older needed that

muawiakh: The older version with docker-machine not called `Docker for Mac`. `Docker for Mac` is specifically introduced to run docker directly on MacOS not having a proxy(default docker-machine.)

docs/server/source/appendices/run-with-docker.md
muawiakh: Because it is the default docker bridge gateway.

ttmc: Maybe add a comment about that in the text, otherwise it seems rather arbitrary.

muawiakh:
fixes #1808

PR #1840, opened 2017-11-13
Updated the changelog for the v1.2 release

ttmc:


ttmc:
I'm not going to wait for the Travis CI tests to pass because they'll probably fail (because they're all failing most of the time lately) and all I did was change the changelog.

PR #1847, opened 2017-11-13
Removed auto-generated tx & vote schema documentation

ttmc:
Most of the documentation about transactions is moving over to the IPDB Protocol docs, so we don't need it here (in the BigchainDB Server docs). This PR is a step towards realizing that goal.

This PR removes the auto-generated pages **Transaction Schema** and **Vote Schema** from the docs (and all the associated code for doing the auto-generation of those docs).

The YAML files for encoding those schemas were kept since they're still used (indirectly) for transaction and vote schema validation. I will remove the un-needed documentation-related stuff from those YAML files in a separate PR. I may also include links to the actual _final_ YAML schema files from the docs, since those are useful for some people. (The current auto-generated docs don't actually tell you all the schema constraints, so that approach would be more thorough.) Maybe those YAML files could be embedded nicely in the Sphinx-generated docs somehow.


ttmc:
Whoa, all the tests passed on Travis CI. Merging now!

PR #1851, opened 2017-11-14
Return commit hash to tendermint

### LINKED ISSUE #1829: Send the hash to tendermint
_description will be added soon_

kansi:
Fixes #1829

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -50,7 +50,7 @@ def get_latest_block(conn):
 
 
 @register_query(LocalMongoDBConnection)
-def create_new_block(conn, block):
+def store_block(conn, block):
     try:
         return conn.run(
             conn.collection('blocks')
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -4,6 +4,7 @@ from bigchaindb import backend
 from bigchaindb.backend.exceptions import DuplicateKeyError
 from bigchaindb.backend.utils import module_dispatch_registrar
 from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
+from pymongo import DESCENDING
 
 
 register_query = module_dispatch_registrar(backend.query)
@@ -38,4 +39,21 @@ def get_spent(conn, transaction_id, output):
                        'inputs.fulfills.output_index': output},
                       {'_id': 0}))
     except IndexError:
+        pass
+
+
+@register_query(LocalMongoDBConnection)
+def get_latest_block(conn):
+    return conn.run(
+        conn.collection('blocks')
+        .find_one(sort=[('height', DESCENDING)]))
+
+
+@register_query(LocalMongoDBConnection)
+def create_new_block(conn, block):
+    try:
+        return conn.run(
+            conn.collection('blocks')
+            .insert_one(block))
+    except DuplicateKeyError:
         pass
--- a/bigchaindb/tendermint/core.py
+++ b/bigchaindb/tendermint/core.py
@@ -27,9 +27,9 @@ class App(BaseApplication):
 
     def init_chain(self, validators):
         """Initialize chain with block of height 0"""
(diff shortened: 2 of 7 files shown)


bigchaindb/tendermint/core.py
```diff
@@ -46,4 +50,11 @@ def deliver_tx(self, raw_transaction):
             return Result.error(log='Invalid transaction')
         else:
             self.bigchaindb.store_transaction(transaction)
+            self.block_txn_ids.append(transaction.to_dict()['id'])
             return Result.ok()
+
+    def commit(self):
+        """ Return merkle root of the transactions included in the
+        block"""
+        merkle_root = calculate_hash(self.block_txn_ids)
```
kansi: Should move this to `end_block` because `commit` is blocking

bigchaindb/tendermint/utils.py
```diff
@@ -12,3 +13,11 @@ def decode_transaction(raw):
     """Decode a transaction from Base64 to a dict."""
 
     return json.loads(raw.decode('utf8'))
+
+
+def calculate_hash(key_list):
```
ttmc: Does Tendermint expect a SHA-256 hash (i.e. a SHA-2 hash), or can we give it a SHA3-256 hash instead? BigchainDB usually uses SHA3-256 hashes. For example, the ID of a BigchainDB transaction is a SHA3-256 hash (represented as a 64-character hex string. 256 bits = 32 bytes and each byte can be represented by two hex characters).

kansi: Tendermint doesn't seem to specify any such contraint, so I have moved to SHA3-256.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -39,3 +40,20 @@ def get_spent(conn, transaction_id, output):
...
+
+@register_query(LocalMongoDBConnection)
+def get_latest_block(conn):
+    return conn.run(
+        conn.collection('blocks')
+        .find_one(sort=[('height', DESCENDING)]))
+
+
+@register_query(LocalMongoDBConnection)
+def create_new_block(conn, block):
```
vrde: We have other methods called `store_<obj>`, like `store_transaction` or `store_asset`. Following this pattern, we can call this method `store_block`, what do you think?

bigchaindb/tendermint/core.py
```diff
@@ -1,11 +1,12 @@
 """This module contains all the goodness to integrate BigchainDB
 with Tendermint."""
 
-
 from abci import BaseApplication, Result
+from abci.types_pb2 import ResponseEndBlock, ResponseInfo
```
vrde: You can import everything from `abci`. As you can see from their [`abci/__init__.py`](https://github.com/davebryson/py-abci/blob/master/abci/__init__.py) they export all the useful classes.

bigchaindb/tendermint/core.py
```diff
@@ -19,6 +20,29 @@ def __init__(self, bigchaindb=None):
...
             bigchaindb = BigchainDB()
         self.bigchaindb = bigchaindb
+        self.block_txn_ids = []
+        self.block_txn_hash = ''
+        self.validators = None
+        self.new_height = None
+
+    def init_chain(self, validators):
+        """Initialize chain with block of height 0"""
+        print("\ninit_chain::\n")
```
vrde: I'd use `log`. Or you can even remove the `print` for now, and we address logging later.

bigchaindb/backend/localmongodb/query.py
kansi: sure

bigchaindb/tendermint/core.py
kansi: Sure, will remove it. Was using it for testing locally.

bigchaindb/tendermint/core.py
kansi: I can see`ResponseInfo` is exported but `ResponseEndBlock` is not exported there.

PR #1852, opened 2017-11-14
[Workaround]: Travis failures

muawiakh:
- Marking test_double_create as serial
- If this works, we don't need to skip the tests
- Finger crossed.

### DIFF
--- a/.ci/travis_script.sh
+++ b/.ci/travis_script.sh
@@ -7,11 +7,14 @@ if [[ -n ${TOXENV} ]]; then
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
     -z "${BIGCHAINDB_DATABASE_SSL}" ]]; then
     # Run the full suite of tests for MongoDB over an unsecure connection
-  pytest -sv --database-backend=mongodb --cov=bigchaindb
+  pytest -sv --database-backend=mongodb -m "serial"
+  pytest -sv --database-backend=mongodb --cov=bigchaindb -m "not serial"
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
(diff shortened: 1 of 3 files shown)


ttmc:
This PR doesn't make any changes on the test named test_websocket_block_event (but the description says it marks it as serial).

muawiakh:
@ttmc yes I reverted that, Let me change the description.

.ci/travis_script.sh
```diff
@@ -7,11 +7,13 @@ if [[ -n ${TOXENV} ]]; then
...
+  pytest -sv --database-backend=mongodb -m "serial"
+  pytest -sv --database-backend=mongodb --cov=bigchaindb -m "not serial"
 elif [[ "${BIGCHAINDB_DATABASE_BACKEND}" == mongodb && \
     "${BIGCHAINDB_DATABASE_SSL}" == true ]]; then
     # Run a sub-set of tests over SSL; those marked as 'pytest.mark.bdb_ssl'.
   pytest -sv --database-backend=mongodb-ssl --cov=bigchaindb -m bdb_ssl
 else
-  pytest -sv -n auto --cov=bigchaindb
+  pytest -sv -m "serial"
+  pytest -sv --cov=bigchaindb -m "not serial"
```
ttmc: Why only have `--cov=bigchaindb` for the "not serial" tests?

This block might be easier to understand if you added a comment like:
```text
# Run the full suite of tests for RethinkDB (the default backend when testing)
```

muawiakh: I will add the comment and code coverage needs to be checked once, that is my understanding, please correct me if I am wrong. Otherwise I can add `--cov=bigchaindb` in the `serial` marked tests as well.

muawiakh:
@ttmc you don't need to run code coverage twice.  Adding `--cov=bigchaindb` once takes care of everything.

PR #1853, opened 2017-11-14
Removed all 'description' keys & values from the YAML files

ttmc:
This is a small follow-up to pull request #1847 

- It removes the `description` keys and values from all the YAML files (which are used to define the schemas of transactions and votes).
- It also removes the function `drop_schema_descriptions()` because it's not needed any more. The descriptions are gone, so there are no descriptions to drop.

In a future pull request, I will either link to the appropriate YAML files from the documentation, or I will embed them right in the documentation, as a reference for what the _actual_ schema definitions are.

Note:  There were no `description` keys to remove in `transaction_create.yaml`.


ttmc:
Note:  Pull request #1855 creates a new page that shows the full contents of the `vote.yaml` file (in the Appendices, with a link to it from **The Vote Model** page). A similar thing will be done with the transaction-related YAML files.

Issue #1828, opened 2017-11-09
Revisit HTTP API for tendermint integration


sbellem:
- [x] Identify tendermint endpoints and their purpose
- [x] Bind bigchaindb endpoints to tendermint ones, when relevant.

kansi:
Usefull links:
- [List of tendermint endpoints](http://tendermint.readthedocs.io/en/master/specification/rpc.html#endpoints) 
- [Events API](https://docs.bigchaindb.com/projects/server/en/latest/events/websocket-event-stream-api.html) could be built using `/subscribe?event=` endpoint from tendermint.
- [Get block](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-blocks-block_id) could be suplimented via `/block?height=_`

kansi:
#### [Transaction status](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-statuses?transaction_id=transaction_id)
According to comments [here](https://github.com/tendermint/tendermint/blob/master/rpc/core/tx.go#L11) it seems we cannot figure out if a transaction is `undecided` or `invalid`. This would imply that inorder to keep track of the transaction status i.e. `undecided`, `valid`, `invalid` we would need to maintain the state it BigchainDB?

vrde:
@kansi great work here! :nail_care:

Ignore for now the last two endpoints `blocks` and `votes`. The implementation depends on how far we are with #1845. All other endpoints depend on the `transactions` and `assets` collection, so you should be able to integrate them """easily""".

ttmc:
Some questions about the `votes` endpoint:

- It seems Tendermint has no "votes." It has "pre-votes", "pre-commits" and "commits"? Is that right?
- Are those things available via some Tendermint API?
- Would any end-users care about those things (e.g. auditors) enough to _want_ endpoints to get them?

ttmc:
@sbellem and @kansi I think this issue can be closed because issues #2036 and #2037 should take care off all remaining things-to-do.

PR #1864, opened 2017-11-16
Integrate api GET "/transactions?asset_id=&operation="

kansi:
#1828

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -1,10 +1,12 @@
 """Query implementation for MongoDB"""
 
+from pymongo import DESCENDING
+
 from bigchaindb import backend
 from bigchaindb.backend.exceptions import DuplicateKeyError
 from bigchaindb.backend.utils import module_dispatch_registrar
 from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
-from pymongo import DESCENDING
+from bigchaindb.common.transaction import Transaction
 
 
 register_query = module_dispatch_registrar(backend.query)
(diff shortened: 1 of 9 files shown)


tests/tendermint/test_core.py
```diff
@@ -1,10 +1,13 @@
 import json
 
+import pytest
```
vrde: Instead of marking each function, you can apply the marker to the [whole module](https://docs.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules):

```python
pytestmark = pytest.mark.tendermint
```

tests/web/test_transactions.py
```diff
@@ -276,6 +276,7 @@ def test_post_invalid_transfer_transaction_returns_400(b, client, user_pk):
     assert res.json['message'] == expected_error_message
 
 
+@pytest.mark.tendermint
```
vrde: That's amazing! :100: 
We can now enable tests without changing them because we maintain backward compatibility.

vrde: That's amazing! :100:
We can now enable tests without changing them because we maintain backward compatibility.

tests/tendermint/test_core.py
vrde: Instead of marking each function, you can apply the marker to the [whole module](https://docs.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules):

```python
pytestmark = pytest.mark.tendermint
```

kansi: yes, that should be better for all tests under `tests/tendermint/`, for others we individually whitelist.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -77,3 +79,30 @@ def store_block(conn, block):
             .insert_one(block))
     except DuplicateKeyError:
         pass
+
+
+@register_query(LocalMongoDBConnection)
+def get_txids_filtered(conn, asset_id, operation=None):
+    match_create = {
+        'operation': 'CREATE',
```
vrde: I just noticed we don't have any index on `operation` (this is true also for previous versions of BigchainDB), would it make sense to create one? If so, we can create an issue and solve it later.

kansi: Good point #1893

PR #1871, opened 2017-11-20
add tests to improve codecov

codegeschrei:


tests/tendermint/test_lib.py
```diff
@@ -27,3 +28,40 @@ def test_asset_is_separated_from_transaciton(b):
...
+
+def test_validation_error(b):
+    from bigchaindb.models import Transaction
+    from bigchaindb.common.crypto import generate_key_pair
+
+    alice = generate_key_pair()
+    asset = {'': ''}
+    tx = Transaction.create([alice.public_key],
+                            [([alice.public_key], 1)],
+                            asset=asset)\
```
vrde: Asset can be `None`, so no need to have `{'': ''}`.

tests/tendermint/test_lib.py
```diff
@@ -27,3 +28,40 @@ def test_asset_is_separated_from_transaciton(b):
...
+                    .sign([alice.private_key]).to_dict()
+
+    tx = b.validate_transaction(tx)
+    b.write_transaction(tx)
+
+    assert mock_post.called
+    assert 'broadcast_tx_async' in str(mock_post.call_args)
```
vrde: Check out [assert_called_with](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with) it should be more suitable for this test case :dancer:

PR #1872, opened 2017-11-20
pre-commit proposal 

codegeschrei:
We already style our code with flake8 but it's not locally enforced.
That's why I want to propose [pre-commit](http://pre-commit.com/)
If people have pre-commit installed it runs with every `git commit` through the given checks.

With pre-commit specific guidelines and style guides can easily be maintained.
An overview of possible hooks you can use [here](http://pre-commit.com/hooks.html)
Note that there are hooks for many different languages!
if that's not enough it's also easy to write new hooks yourself (I did it once)

Every hook I used is a proposal itself and there are many more we can use.

The [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) offer multiple hooks to set, for example, a flake8, json and yml check, trim whitespaces, etc.
I added the following for now:
- trailing whitespaces
- check-merge-conflict (checks for merge conflicts in files)
- debug-statements (checks for debug statements in files)

The [pre-commit-python-sorter](https://github.com/FalconSocial/pre-commit-python-sorter) is a nice hook to sort your imports
I currently set it to `--silent-overwrite` which is convenient since the user does not need to change everything manually

 The [pre-commit-hooks-safety](https://github.com/Lucas-C/pre-commit-hooks-safety/tree/v1.1.0) checks python dependencies. Currently, it only looks for requirement files.

The [pre-commit-mirrors-pydocstyle](https://github.com/chewse/pre-commit-mirrors-pydocstyle). I think with a proper configuration and selection of [error codes](http://www.pydocstyle.org/en/2.1.1/error_codes.html) this can be a good feature. In my settings, I check only for the error code 'D300' which checks if triple double quotes are used. 

The [mirrors-yapf](https://github.com/pre-commit/mirrors-yapf) is a hook for [YAPF](https://github.com/google/yapf), a python formatter with a lot of settings. 
The above pre-commit-hooks offer flake8 but you need to take care of all the errors. Yapf instead just changes the code according to the settings. Some details might not be as nicely formatted as you would do it but at least no one has to take care of it.

If you want to test it and see the changes made or errors thrown by the hooks just get the code, install pre-commit via pip and run `pre-commit run -a` (this might take some time).

[info: the pre-commit accidentally ran on setup.py I'll fix this later]

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+repos:
+-   repo: git://github.com/pre-commit/pre-commit-hooks
+    sha: v1.1.1
+    hooks:
+    -   id: trailing-whitespace
+        args: ['--no-markdown-linebreak-ext']
+    -   id: check-merge-conflict
+    -   id: debug-statements
+    -   id: check-added-large-files
+    -   id: flake8
+
+-   repo: git://github.com/chewse/pre-commit-mirrors-pydocstyle
+    sha: v2.1.1
+    hooks:
+    -   id: pydocstyle
+        args: ['--select=D204,D201,D209,D210,D212,D300,D403']
+
+exclude: '^(?!bigchaindb/)(?!tests/)'
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ dev_require = [
     'ipython',
     'watchdog',
     'logging_tree',
+    'pre-commit'
 ]
 
 docs_require = [


vrde:
I really like this proposal. I was not aware of those nice pre–commit hooks, and they can save us much time in reviewing PRs.

I don't get one thing, how do you hook to the actual `git commit` command? Is it something that happens when you do `pip install pre-commit` (that is in our `setup.py`)?

codegeschrei:
> I don't get one thing, how do you hook to the actual git commit command? Is it something that happens when you do pip install pre-commit (that is in our setup.py)?

That's right. The pre-commit is running whenever you make a commit. There is also a way to avoid it by adding `--no-verify` after the commit message.

codegeschrei:
> Did you base some of the settings on our Python style guide?

I used some already, like the max line length. All of my current settings can also be changed. For some, I just chose what I thought was best. Maybe it's a good idea to upload some examples how the code will look like after yapf formatting so that everyone can make suggestions.

> Maybe you could add some notes about these hooks to that Python style guide document, and to the CONTRIBUTING.md file (i.e. as part of this pull request), so it doesn't surprise anyone.

When this is finished I can make an overview of our styling. We also need a documentation about it because contributors should use the pre-commit

sbellem:
Nice!

We do have a flake8 check on travis (https://github.com/bigchaindb/bigchaindb/blob/master/.travis.yml#L16), however, and we may want to reconcile the two.

The travis job uses tox to run the flake8 check: https://github.com/bigchaindb/bigchaindb/blob/master/tox.ini#L27

The current flake8 config is specified in https://github.com/bigchaindb/bigchaindb/blob/master/setup.cfg#L8
Defaults are used except for the max line length.

Perhaps we could start with a minimal set of requirements, and refine these requirements as time progresses?

codegeschrei:
@sbellem thanks for the links! I'll check it and try to adjust my settings to it.
If they're the same the tox can still run with Travis and will never fail (best case). 
We still need a validation besides the pre-commit. When people just don't install it there will be no checks.

codegeschrei:
I updated the pre-commit settings and yapf styling

- added a check for large files to the pre-commit-hooks
- removed the hook safety check for now, can be added later if we need it but I think it's not part of a minimal 'base' setup
- removed the import sorter because it conflicted with yapf and yapf is more important
- added more error codes to pydocstyle
   - the error codes are all set according to Google-style docstring
- used the global `exclude` statement to run checks only in folders `bigchaindb` and `tests` for now
- the yapf styling file is set according to pep8 styling

If we agree on these settings I'll make another PR with the changes made by the hooks and also change the Travis settings! We shouldn't use flake8 in tox anymore instead we should use pre-commit there too, I already tested it and it works fine.

vrde:
It took me a while to go through the `yapf` configuration, and I'm still not finished :sweat_smile:. For me a better approach is to keep everything you have here but `yapf`, and add a `flake8` hook.

I really like `yapf`, but I'd prefer to see it in a separate pull request, why? Because we need to adapt it more to our needs before letting it change the source code, and I don't want it to block merging this PR.

codegeschrei:
I'll make a second branch with all the changes of the pre-commit, please don't merge this until I'm done. Both branches should be merged at the same time.

PR #1886, opened 2017-11-22
Integrate asset search

kansi:
#1828 
Following APIs are integrated:
- `GET /api/v1/assets?search={text_search}`
- `GET /api/v1/assets?search={text_search}&limit={n_documents}`

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -7,7 +7,7 @@ from bigchaindb.backend.exceptions import DuplicateKeyError
 from bigchaindb.backend.utils import module_dispatch_registrar
 from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
 from bigchaindb.common.transaction import Transaction
-
+from bigchaindb.backend import mongodb
 
 register_query = module_dispatch_registrar(backend.query)
 
@@ -106,3 +106,8 @@ def get_txids_filtered(conn, asset_id, operation=None):
         conn.collection('transactions')
         .aggregate(pipeline))
     return (elem['id'] for elem in cursor)
+
+
+@register_query(LocalMongoDBConnection)
+def text_search(*args, **kwargs):
+    return mongodb.query.text_search(*args, **kwargs)
--- a/tests/web/test_assets.py
+++ b/tests/web/test_assets.py
@@ -87,58 +87,56 @@ def test_get_assets_limit(client, b):
 
 @pytest.mark.bdb
 @pytest.mark.tendermint
-def test_get_assets_tmint(client, tb):
+@pytest.mark.localmongodb
+def test_get_assets_tendermint(client, tb):
     from bigchaindb.models import Transaction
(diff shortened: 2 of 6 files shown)


bigchaindb/backend/localmongodb/query.py
```diff
@@ -7,7 +7,7 @@
 from bigchaindb.backend.utils import module_dispatch_registrar
 from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
 from bigchaindb.common.transaction import Transaction
-
+from bigchaindb.backend import mongodb
```
vrde: :joy: :joy: :joy:
that's what I call backward compatibility.

tests/web/test_assets.py
```diff
@@ -81,3 +83,62 @@ def test_get_assets_limit(client, b):
         res = client.get(ASSETS_ENDPOINT + '?search=abc&limit=1')
         assert res.status_code == 200
         assert len(res.json) == 1
+
+
+@pytest.mark.bdb
+@pytest.mark.tendermint
+def test_get_assets_tmint(client, tb):
```
vrde: Minor comment: I'd call it `test_get_asset_tendermint`, why? Less things to process for your brain :) (I asked myself "why tmint and not tendermint?")

tests/web/test_assets.py
```diff
@@ -81,3 +83,62 @@ def test_get_assets_limit(client, b):
...
+    from bigchaindb.models import Transaction
+    from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
+
+    if isinstance(tb.connection, LocalMongoDBConnection):
```
vrde: I'm a bit confused, what's the reason to mark the test for `mdb` if we don't test it actually against the `mongodb` backend? I'd probably just remove `pytest.mark.bdb` **or** (it's even better, probably) I'd be totally fine in duplicating the test (yep, copypasta) so we have one for `mdb` and one for `tendermint`. Repetition is fine in tests :)

(same applies for the other test below)

tests/web/test_assets.py
```diff
@@ -81,3 +83,62 @@ def test_get_assets_limit(client, b):
...
+        assert len(res.json) == 1
+        assert res.json[0] == {
+            'data': {'msg': 'abc'},
+            'id': tx.id
+        }
+
+
+@pytest.mark.bdb
+@pytest.mark.tendermint
+def test_get_assets_limit_tmint(client, tb):
```
vrde: Same comments I made for `test_get_assets_tmint`.

tests/web/test_assets.py
kansi: In case we merge these changes back into `master`, the tendermint tests won't affect anything when tests are run using `mongodb` or `rethinkdb`

vrde: Oh I see, what if we use [`pytest.mark.skipif`](https://docs.pytest.org/en/latest/skipping.html#id1)?

PR #1916, opened 2017-11-30
Updated & deleted some root docs in the tendermint branch

ttmc:
I updated:

- The index / home page
- Terminology
- How BigchainDB is Decentralized
- Kinds of Node Diversity
- BigchainDB and BFT
- How BigchainDB is Good for Asset Registrations & Transfers
- BigchainDB and Smart Contracts
- Transaction Concepts
- Permissions in BigchainDB

These pages were removed for now, but we may add them back in the future once it's more clear what we should say:

- **Production-Ready?** (because that page has caused a lot of hand-wringing and hand-wringing is a waste of time)
- **Timestamps in BigchainDB** (because transactions don't have timestamps and it's not clear to me what other timestamps are available from Tendermint + our APIs, or what they mean)


docs/root/source/decentralized.md
```diff
@@ -14,8 +14,6 @@ A consortium can increase its decentralization (and its resilience) by increasin
 
 There’s no node that has a long-term special position in the cluster. All nodes run the same software and perform the same duties.
 
-MongoDB and RethinkDB have an “admin” user which can’t be deleted and which can make big changes to the database, such as dropping a table. Right now, that’s a big security vulnerability, but we have plans to mitigate it by:
-1. Locking down the admin user as much as possible.
-2. Having all nodes inspect admin-type requests before acting on them. Requests can be checked against an evolving whitelist of allowed actions. Nodes requesting non-allowed requests can be removed from the list of cluster nodes.
+If someone has (or gets) admin access to a node, they can mess with that node (e.g. change or delete data stored on that node), but those changes should remain isolated to that node. The BigchainDB cluster can only be compromised if more than one third of the nodes get compromised. See the [Tendermint documentation](https://tendermint.readthedocs.io/projects/tools/en/master/introduction.html) for more details.
```
kansi: Even though an attacker cannot manipulate the blockchain but they could possibly throttle the network (DoS) by sending invalid transactions?

vrde: It's a valid point but it's not related to decentralization IMO.

PR #1918, opened 2017-12-01
Upgrade py-abci

kansi:
Upgrade py-abci from v0.2.0 to v0.3.0

### DIFF
--- a/bigchaindb/tendermint/core.py
+++ b/bigchaindb/tendermint/core.py
@@ -62,8 +62,12 @@ class App(BaseApplication):
             logger.debug('check_tx: INVALID')
             return Result.error()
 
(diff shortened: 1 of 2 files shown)


bigchaindb/tendermint/core.py
```diff
@@ -62,7 +62,7 @@ def check_tx(self, raw_transaction):
             logger.debug('check_tx: INVALID')
             return Result.error()
 
-    def begin_block(self, block_hash, header):
+    def begin_block(self, reqBeginBlock):
```
vrde: please use `snake_case`

ttmc: Maybe also write something about "reqBeginBlock" in the docstring.

PR #1928, opened 2017-12-04
Integrate event stream api

kansi:
#1828

### DIFF
--- a/bigchaindb/tendermint/commands.py
+++ b/bigchaindb/tendermint/commands.py
@@ -57,7 +57,7 @@ def start():
     p_websocket_server.start()
 
     # connect to tendermint event stream
-    p_websocket_client = mp.Process(name='ws_client',
+    p_websocket_client = mp.Process(name='ws_to_tendermint',
                                     target=event_stream.start,
                                     args=(exchange.get_publisher_queue(),))
     p_websocket_client.start()
--- a/bigchaindb/tendermint/commands.py
+++ b/bigchaindb/tendermint/commands.py
@@ -6,7 +6,9 @@ from os import getenv
 import bigchaindb
 from bigchaindb.tendermint.lib import BigchainDB
 from bigchaindb.tendermint.core import App
-from bigchaindb.web import server
+from bigchaindb.web import server, websocket_server
+from bigchaindb.tendermint import event_stream
+from bigchaindb.events import Exchange, EventTypes
 
 
 logger = logging.getLogger(__name__)
@@ -26,6 +28,10 @@ BANNER = """
 
 
 def start():
+
+    # Exchange object for event stream api
+    exchange = Exchange()
+
     # start the web api
     app_server = server.create_server(
         settings=bigchaindb.config['server'],
@@ -43,6 +49,18 @@ def start():
             'node',
             '--consensus.create_empty_blocks=false',
         ])
+
+    # start websocket server
+    p_websocket_server = mp.Process(name='ws',
+                                    target=websocket_server.start,
+                                    args=(exchange.get_subscriber_queue(EventTypes.BLOCK_VALID),))
+    p_websocket_server.start()
+
+    # connect to tendermint event stream
+    p_websocket_client = mp.Process(name='ws_client',
+                                    target=event_stream.start,
+                                    args=(exchange.get_publisher_queue(),))
+    p_websocket_client.start()
 
     # We need to import this after spawning the web server
     # because import ABCIServer will monkeypatch all sockets
--- a/bigchaindb/tendermint/event_stream.py
+++ b/bigchaindb/tendermint/event_stream.py
@@ -7,39 +7,34 @@ import aiohttp
 
 from bigchaindb.common.utils import gen_timestamp
 from bigchaindb.events import EventTypes, Event
-from bigchaindb.tendermint.utils import decode_txn_base64
+from bigchaindb.tendermint.utils import decode_transaction_base64
 
 
 HOST = 'localhost'
 PORT = 46657
 URL = f'ws://{HOST}:{PORT}/websocket'
 
-PAYLOAD = {
-    "method": "subscribe",
-    "jsonrpc": "2.0",
-    "params": ["NewBlock"],
-    "id": "bigchaindb_stream"
-}
-
-
 logger = logging.getLogger(__name__)
 
 
-async def connect_and_recv(event_queue):
+@asyncio.coroutine
+def connect_and_recv(event_queue):
     session = aiohttp.ClientSession()
-    async with session.ws_connect(URL) as ws:
-        logger.info('Connected to tendermint ws server')
+    ws = yield from session.ws_connect(URL)
 
-        stream_id = "bigchaindb_stream_{}".format(gen_timestamp())
-        await subscribe_events(ws, stream_id)
+    logger.info('Connected to tendermint ws server')
 
-        async for msg in ws:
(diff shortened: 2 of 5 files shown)


bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+        async for msg in ws:
+            process_event(event_queue, msg.data, stream_id)
+
+            if msg.type in (aiohttp.WSMsgType.CLOSED,
+                            aiohttp.WSMsgType.ERROR):
+                session.close()
+                raise aiohttp.ClientConnectionError()
+
+
+def process_event(event_queue, event, stream_id):
```
kansi: write test for this

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+    "jsonrpc": "2.0",
+    "params": ["NewBlock"],
+    "id": "bigchaindb_stream"
+}
+
+
+logger = logging.getLogger(__name__)
+
+
+async def connect_and_recv(event_queue):
```
vrde: `async` and `await` are not compatible with Python 3.5, that we still support. I'd be happy to drop support for Python 3.5, but it must be something planned.

In case of function definitions and function calls, we can easily use `@asyncio.coroutine` and `yield from`, for `with` and `for` is more difficult.

I've created #1956 to talk about this issue.

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+            event = Event(EventTypes.BLOCK_VALID, new_block)
+            event_queue.put(event)
+
+
+async def subscribe_events(ws, stream_id):
+    PAYLOAD["id"] = stream_id
+    await ws.send_str(json.dumps(PAYLOAD))
+
+
+async def try_connect_and_recv(event_queue, gas):
```
vrde: what is `gas`?

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        await connect_and_recv(event_queue)
+
+    except Exception as e:
+        logger.error('WebSocket connection failed with exception: {}'.format(e))
+        if gas:
+            time.sleep(2)
+            await try_connect_and_recv(event_queue, gas-1)
```
vrde: Oh I see, in `bigchaindb/backend/connection.py` we call it `max_tries`.

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+    PAYLOAD["id"] = stream_id
+    await ws.send_str(json.dumps(PAYLOAD))
+
+
+async def try_connect_and_recv(event_queue, gas):
+    try:
+        await connect_and_recv(event_queue)
+
+    except Exception as e:
+        logger.error('WebSocket connection failed with exception: {}'.format(e))
```
vrde: To me, this should be a `logger.warning`, it's not an error **yet**. It's an error when there are no more retries. As an inspiration, check [`Connection::connect`](https://github.com/bigchaindb/bigchaindb/blob/f35f01f1cb1f794b1336049003c3bd70bc4bc72b/bigchaindb/backend/connection.py#L141).

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,84 @@
...
+            time.sleep(2)
+            await try_connect_and_recv(event_queue, gas-1)
+
+
+def start(event_queue):
+    loop = asyncio.get_event_loop()
+    try:
+        loop.run_until_complete(try_connect_and_recv(event_queue, 5))
+    except (KeyboardInterrupt, SystemExit):
+        logger.info('Shutting down tendermint event stream connection')
```
vrde: capital `T`, `Tendermint`

bigchaindb/tendermint/utils.py
```diff
@@ -15,6 +15,12 @@ def decode_transaction(raw):
     return json.loads(raw.decode('utf8'))
 
 
+def decode_txn_base64(value):
```
vrde: In general, we try to keep the names of functions and variable explicit, so what about a simple `decode_transaction_base64`?

Moreover, would it make sense to also do a `json.loads` inside the function?

bigchaindb/tendermint/commands.py
```diff
@@ -44,6 +50,18 @@ def start():
...
         ])
 
+    # start websocket server
+    p_websocket_server = mp.Process(name='ws',
+                                    target=websocket_server.start,
+                                    args=(exchange.get_subscriber_queue(EventTypes.BLOCK_VALID),))
+    p_websocket_server.start()
+
+    # connect to tendermint event stream
+    p_websocket_client = mp.Process(name='ws_client',
```
vrde: `ws_to_tendermint`? This would make easier to know what the process is doing when doing a `top` from the terminal.

bigchaindb/tendermint/event_stream.py
kansi: Counter for the number of connection attempts to Tendermint stream api.

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,91 @@
...
+        if block_txs:
+            block_txs = [decode_transaction_base64(txn) for txn in block_txs]
+            new_block = {'id': str(block_id), 'transactions': block_txs}
+            event = Event(EventTypes.BLOCK_VALID, new_block)
+            event_queue.put(event)
+
+
+@asyncio.coroutine
+def subscribe_events(ws, stream_id):
+    PAYLOAD["id"] = stream_id
```
vrde: This code is async, and it runs in an event loop in a single thread. Anyway, it's safer to create a local payload in my opinion, and it's more straightforward.

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@asyncio.coroutine
+def try_connect_and_recv(event_queue, max_tries):
+    try:
+        yield from connect_and_recv(event_queue)
+
+    except Exception as e:
+        if max_tries:
+            logger.warning('WebSocket connection failed with exception: {}'.format(e))
```
vrde: The best practice for string formatting and logging is to use lazy evaluation:
```python
logger.warning('WebSocket connection failed with exception %s', e)
```

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,88 @@
...
+            time.sleep(2)
+            yield from try_connect_and_recv(event_queue, max_tries-1)
+        else:
+            logger.error('WebSocket connection failed with exception: {}'.format(e))
```
vrde: In this case you can use [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception)

```python
logger.exception('WebSocket connection failed')
```

bigchaindb/tendermint/event_stream.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def start(event_queue):
+    loop = asyncio.get_event_loop()
+    try:
+        loop.run_until_complete(try_connect_and_recv(event_queue, 5))
+    except (KeyboardInterrupt, SystemExit):
+        logger.info('Shutting down Tendermint event stream connection')
+        return
```
vrde: You can drop the `return` here.

Issue #1895, opened 2017-11-24
Update the docs


vrde:
# Description
We need to update the documentation with the new changes we are introducing by integrating with Tendermint.

This issue is a work in progress and feedback is welcome.

Note: All the paths in this issue are relative to `/bigchaindb/docs/`.

Assumptions:
- We won't support RethinkDB for now (but we are open for contributions there)

Changes:
- [ ] `root/source/bft.md` we now inherit BFT from Tendermint
- [ ] `root/source/decentralized.md` remove paragraph related to MongoDB/RethinkDB *admin user*
- [ ] `root/source/production-ready.md` uhm... I guess we will be closer to be production ready, I think we should wait to change that :)
- [ ] `root/source/terminology.md` update the definition for **BigchainDB Cluster**
- [ ] `root/source/appendices/backend.rst` add `automodule`s for `localmongodb`, remove docs for old backends
- [ ] `root/source/appendices/docker-on-mac.md` add info on how to run docker with `localmongodb` backend, remove docs for old backends
- [ ] `root/source/appendices/firewall-notes.md` add Tendermint specific ports, remove ports for RethinkDB/MongoDB
- [ ] `root/source/appendices/install-with-lxd.md` update with instructions to install Tendermint (or remove that part of the documentation)
- [ ] `root/source/appendices/pipelines.rst` remove file
- [ ] `root/source/appendices/rethinkdb-*.md` remove files
- [ ] `root/source/appendices/run-with-*.md` remove files? We are working on #1887 that covers **Vagrant**
- [ ] `root/source/appendices/the-Bigchain-class.rst` rename `Bigchain` with `BigchainDB` (right now the `BigchainDB` class is under `bigchaindb/tendermint/core.py`, but I guess we will move it to `bigchaindb/core.py`)
- [ ] `root/source/data-models/block-model.rst` TBD. We still have to explore how to expose the `block` model since it will depend on Tendermint
- [ ] `root/source/data-models/vote-model.rst` TBD (but I guess we have to remove it)
- [ ] `root/source/dev-and-test/setup-bdb-docker.md` change backend to `localmongodb`, remove references to **RethinkDB**
- [ ] `root/source/dev-and-test/setup-bdb-host.md` change backend to `localmongodb`, remove references to **RethinkDB**
- [ ] `root/source/http-samples/*.md` TBD because of the `block` model (see before). We might need to remove files related to the `vote` model
[...]

PR #1887, opened 2017-11-22
Scripts to bootstrap devstack setup for bigchaindb with tendermint

shahbazn:
## Description
Adds a single command workflow to bring up development setup with bigchaindb and tendermint

## Issues This PR Fixes
Fixes #1850 

## Todos
- [x] Tested and working on development environment

## How to QA
System requirements: vagrant and virtualbox (install latest versions and you'll be fine)
Install required vagrant plugins
- vagrant plugin install vagrant-cachier
- vagrant plugin install vagrant-vbguest
Set git branch to use: export GIT_BRANCH=shahbaz-dev
down load script: curl -fOL https://raw.githubusercontent.com/bigchaindb/bigchaindb/${GIT_BRANCH}/pkg/scripts/install_stack.sh
run script: bash install_stack.sh

Optionally you can set MONGO_VERSION (current default 3.4) and TM_VERSION (current default 0.12.1) if you wish to configure versions mongodb or tendermint other than defaults

PR #1935, opened 2017-12-05
Updated Quickstart page for Tendermint

ttmc:
This is one of the pull requests to tackle issue #1895. It updates the **Quickstart** page in the docs for BigchainDB Server. It rearranges the order a bit, consolidates some things, removes the letter labels on the steps.

What's new? The installation, init, and running of Tendermint.

I based it on pull request #1887 by @shahbazn 

I tested all the installation, config and run instructions on a brand new Ubuntu 16.04 Server (on AWS), up to just before `bigchaindb start` (which isn't going to work correctly yet because the current `bigchaindb` package on PyPI is the wrong one). The tested instructions worked fine.

### Questions

* Do we still need to specify the replica set when we start MongoDB? i.e.
```text
sudo mongod --replSet=bigchain-rs
```
* Do we need to set all those environment variables? i.e.
```text
$ export BIGCHAINDB_START_TENDERMINT=0
$ export TENDERMINT_HOST=localhost
$ export TENDERMINT_PORT=46657
```
* Do we need to do `tendermint unsafe_reset_all` before `tendermint node`?
* Is it really correct to start Tendermint last? I guess so, since Tendermint sees itself as a client and BigchainDB Server as a Server.

### DIFF
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -1,69 +1,87 @@
 # Quickstart
 
-This page has instructions to set up a single stand-alone BigchainDB node for learning or experimenting. Instructions for other cases are [elsewhere](introduction.html). We will assume you're using Ubuntu 16.04 or similar. You can also try, [running BigchainDB with Docker](appendices/run-with-docker.html).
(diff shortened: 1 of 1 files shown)


ttmc:
@shahbazn  As discussed today, I'm going to change the Quickstart page to be mostly just a redirect to other options (mostly the IPDB Test Net), so I'm closing this pull request (which updated the Quickstart page assuming it was just a variation on the old one). There will be a new pull request changing the Quickstart page in the new way.

PR #1948, opened 2017-12-07
Fixed failing test and removed pytest-catchlog on tendermint branch

ttmc:
The test `test_get_spending_transactions` in `tests/backend/localmongodb/test_queries.py` was failing so I fixed it by just making it more like the same test in `tests/backend/mongodb/test_queries.py`.

I also removed `pytest-catchlog` from the requirements in `setup.py` to make that warning go away. The same was already done on the `master` branch in PR #1936.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,6 @@ tests_require = [
     'hypothesis-regex',
     'pylint',
     'pytest>=3.0.0',
-    'pytest-catchlog>=1.2.2',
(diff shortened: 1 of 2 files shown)


sbellem:
> I also removed pytest-catchlog from the requirements in setup.py to make that warning go away. The same was already done on the master branch in PR #1936.

In general, we can either `cherry-pick` or `merge master` into the `tendermint` branch.

PR #1951, opened 2017-12-08
Proposal for extending post transaction api

kansi:
Extend `POST` transaction api to include query parameter `mode` which would enable the follows,
- asynchoronously post the transaction
- post the transaction and return after its validated
- post the transaction and return after its commited

### DIFF
--- a/proposals/extend-post-txn.md
+++ b/proposals/extend-post-txn.md
@@ -1,0 +1,68 @@
+## Feature : Add query parameter "mode" to `POST` transaction
+Add new query paramter `mode` to the [post transaction api](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#post--api-v1-transactions) such that client can,
+- asynchoronously post the transaction
+- post the transaction and return after its validated
+- post the transaction and return after its commited
+
+## Problem Description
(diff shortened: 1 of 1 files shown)


ttmc:
I think this is a nice proposal and a good way to expose that Tendermint API.

I wonder what the failure modes are, and how we'll handle them.

For example, I suppose it's possible someone to POST a transaction in the sync mode but then the CheckTx takes a long time to happen (or maybe never happens). Is there some timeout (implicit or explicit)? If so, what happens when it's reached?

I wonder about other failure modes.

kansi:
@ttmc that is valid concern. The [tendermint documentation](http://tendermint.readthedocs.io/projects/tools/en/master/using-tendermint.html#broadcast-api) doesn't provide much details regarding this. This aspect needs to properly explored before we decide to implement this feature.

proposals/extend-post-txn.md
```diff
@@ -0,0 +1,68 @@
...
+### Data model impact
+N/A
+
+### API impact
+The query parameter `mode` will be introduced to [`POST /api/v1/transaction`](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#post--api-v1-transactions) and will have the following possible values
+- `async`
+- `sync`
+- `commit`
+
+To preserve compability with the existing behavour of the api the default value of `mode` would be `async`.
```
vrde: Just to be 100% clear, I'd say:

To preserve compability with the existing behavour of the API, the query parameter `mode` is optional. In this case the default value would be `async`.

PR #1957, opened 2017-12-12
Simplify the "Develop & Test BigchainDB Server" section of the docs

ttmc:
This is one of the pull requests to tackle issue #1895.

This pull request makes a massive simplification the the **Develop & Test BigchainDB Server** section of the docs. It replaces the entire section with one page saying:

> If you’d like to help develop and test BigchainDB Server, then see the CONTRIBUTING.md file.

with a link to the `CONTRIBUTING.md` file in the `master` branch on GitHub.

@shahbazn If you want to mention the Vagrant stuff anywhere (and you might not), then you can mention it in the `CONTRIBUTING.md` file. (Same with Ansible if we keep that.) The `CONTRIBUTING.md` file links to the page about how to run all the unit tests (i.e. `tests/README.md`), and it should continue to do so.

@muawiakh I'm putting you as a reviewer mainly so that you know this is what I propose to do.


sbellem:
I have strong doubts about this simplification. Contributing docs need to address multiple details and to try to cram them all up in one page seems to me to be over-optimistic and not realistic. One may look at [Contributing to Django](https://docs.djangoproject.com/en/dev/internals/contributing/) as an example.

ttmc:
@sbellem  Last week, you told me you think I should convert the IPDB Transaction Spec (which is currently 20+ pages) into one page. Your justification was one example: the Interledger spec docs.

Now you think that one page is bad. Your justification was one example: the Django docs.

Anyway, this PR is merged. If you want to do something else, then make your own PR.

sbellem:
This is just feedback, and the concern I expressed is legit. I gave the example of Django as an example of how an active open source project handles their documentation for contributors. This was not meant to be a justification but simply an example. The case at hand here is not about a protocol specification, but about the details of contributing code to an open source project.

I simply fear that we will be struggling to add new documentation when needed, because one `md` file appears to me as too minimalistic.

PR #1976, opened 2017-12-21
Proposal: Migrate cli for Tendermint

kansi:
## Description
With the tendermint integration some of the sub-commands have become irrelevant. This proposal discusses how to migrate the same.

### DIFF
--- a/proposals/migrate-cli.md
+++ b/proposals/migrate-cli.md
@@ -54,7 +54,7 @@ optional arguments:
 
 Commands:
   {configure,show-config,export-my-pubkey,init,drop,start,set-shards,set-replicas,add-replicas,remove-replicas}
-    configure           Prepare the config file and create the node keypair
+    configure           Prepare the config file
     show-config         Show the current configuration
     init                Init the database
     drop                Drop the database
--- a/proposals/migrate-cli.md
+++ b/proposals/migrate-cli.md
@@ -147,6 +147,7 @@ Following test cases should be added
 
 ### Assignee(s)
 Primary assignee(s): @muawiakh
+
 Secondary assignee(s): @kansi, @ttmc
 
 ### Targeted Release
--- a/proposals/migrate-cli.md
+++ b/proposals/migrate-cli.md
@@ -1,0 +1,68 @@
+# Migrate Bigchaindb cli for Tendermint
+
+## Problem Description
+With Tendermint integration some of the cli sub-commands have been rendered obsolete. It would be only appropriate to white list sub-commands depending on the backend.
+
+### Use cases
+- Avoid confusing the user by not displaying irrelevant sub-commands.
+
+
+## Proposed Change
+Following sub-commands should be updated
+
+- `bigchaindb --help`: list the relevant sub-commands depending on the configured backend. In case the backend is not configured then the default backend `mongodb` should be assumed.
+
+Following sub-commands should be depreciated for `localmongodb` backend.
+
+- `bigchaindb export-my-pubkey`
+- `bigchaindb set-shards`
+- `bigchaindb set-replicas`
+
+**NOTE**: In case the user attempts to execute the above depreciated sub-commands an error message stating their in-compatibility with `localmongodb` should be displayed.
+
+### Usage example
+N/A
+
+### Data model impact
+N/A
+
+### API impact
+N/A
+
+### Security impact
+N/A
+
+### Performance impact
+N/A
+
+### End user impact
+N/A
+
+### Deployment impact
+N/A
+
+### Documentation impact
+The documentation for depreciated sub-commands should indicate that they are not compatible with `localmongodb` backend.
+
+
+### Testing impact
+Following test cases should be added
+- Set `localmongodb` as backend, then executing the depreciated sub-commands should give an error.
+- Set `localmongodb` as backend, then executing `bigchaindb --help`, `bigchaindb -h` should not list the deprecated sub-commands.
+
+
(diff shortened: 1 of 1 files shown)


proposals/migrate-cli.md
```diff
@@ -0,0 +1,68 @@
...
+With Tendermint integration some of the cli sub-commands have been rendered obsolete. It would be only appropriate to white list sub-commands depending on the backend.
+
+### Use cases
+- Avoid confusing the user by not displaying irrelevant sub-commands.
+
+
+## Proposed Change
+Following sub-commands should be updated
+
+- `bigchaindb --help`: list the relevant sub-commands depending on the configured backend. In case the backend is not configured then the default backend `mongodb` should be assumed.
```
ttmc: In BigchainDB-with-Tendermint, I thought the only supported backend was `localmongodb`.

proposals/migrate-cli.md
```diff
@@ -0,0 +1,68 @@
+# Migrate Bigchaindb cli for Tendermint
```
ttmc: There's a bigger question here, and that is, will the two products share the same Python code for handling the `bigchaindb` command-line command, or will they have different code for that? I guess your thinking is that they _will_ share the same code. I don't know which is best without attempting implementation.

proposals/migrate-cli.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Problem Description
+With Tendermint integration some of the cli sub-commands have been rendered obsolete. It would be only appropriate to white list sub-commands depending on the backend.
+
+### Use cases
+- Avoid confusing the user by not displaying irrelevant sub-commands.
+
+
+## Proposed Change
```
ttmc: Some other sub-commands might require modification. For example, `bigchaindb configure` shouldn't allow `rethinkdb` or `mongodb` if this is BigchainDB-with-Tendermint.

proposals/migrate-cli.md
```diff
@@ -0,0 +1,68 @@
...
+Following sub-commands should be updated
+
+- `bigchaindb --help`: list the relevant sub-commands depending on the configured backend. In case the backend is not configured then the default backend `mongodb` should be assumed.
+
+Following sub-commands should be depreciated for `localmongodb` backend.
+
+- `bigchaindb export-my-pubkey`
+- `bigchaindb set-shards`
+- `bigchaindb set-replicas`
```
ttmc: What about `bigchaindb add-replicas` and `bigchaindb remove-replicas`?

proposals/migrate-cli.md
kansi: I concur with the idea that it is difficult to comtemplate the effort entailed in either of the approaches. But I guess we could start off with the current code and replace bits and pieces as we move ahead?

proposals/migrate-cli.md
kansi: In case we have both 1.x and 2.0 in the master then we would have to allow the user to configure the database with either of options i.e. `rethinkdb`, `mongodb`, `localmongodb`. Once configured the sub-commands will be available depending on the configured backend.

proposals/migrate-cli.md
kansi: The [comment](https://github.com/bigchaindb/bigchaindb/pull/1976#discussion_r158327626) should clarify this.

proposals/migrate-cli.md
kansi: I was a little unsure about these as they could different meaning for BDB 2.0 i.e. `add-replicas` could allow the user to setup redundant nodes. The same could be achieved by directly configuring mongodb which in my opinon is a better way.

This is up for discussion.

proposals/migrate-cli.md
ttmc: :+1:

proposals/migrate-cli.md
ttmc: :+1:

proposals/migrate-cli.md
ttmc: Indeed, "local MongoDB" (with Tendermint) might, in fact, be an entire sharded cluster (for use only by that one BigchainDB node)...

sbellem:
I would prefer that we first clarify the bigger picture with respect to the future of the current BigchainDB, version versus the one integrating with Tendermint.

kansi:
The crux of this PR is to enforce that the user should configure bigchaindb before they can see the list of avaiable commands using `help` subcommand.

ttmc:
The thinking has changed since this proposal was written. We've decided not to worry about maintaining potential support for a future "product 1" in the `tendermint` branch.

- `bigchaindb configure` should only allow `localmongodb` (not `rethinkdb` or `mongodb`)
- Remove `bigchaindb export-my-pubkey`? (A BigchainDB node still _has_ a public key, but it's a Tendermint thing, not a BigchainDB thing.)
- Remove `bigchaindb set-shards` (was for RethinkDB only)
- Remove `bigchaindb set-replicas` (was for RethinkDB only)
- Remove `bigchaindb add-replicas` (was for MongoDB only)
- Remove `bigchaindb remove-replicas` (was for MongoDB only)

Maybe other changes?

kansi:
@ttmc Those changes seems about right. Furthermore, I think it would better to create a new PR and discuss the changes there. The context of discussion here is quite old and might confuse someone.

muawiakh:
Partially fixes #2087

muawiakh:
@ttmc +1

PR #1964, opened 2017-12-13
add test cases for integration testing

codegeschrei:
Adding possible integration test scenarios and give a template for docstrings.

PR #1977, opened 2017-12-22
Proposal: Network driver for integration testing

kansi:
## Description
Spec for network driver which would expose required api for integration testing.

### DIFF
--- a/proposals/integration-infrastructure.md
+++ b/proposals/integration-infrastructure.md
@@ -1,0 +1,100 @@
+# Integration testing simulation
+
+## Problem Description
+The integration testing proposal introduced new set of requirements which would require that the infrastructure be controlled at the code level in order to write the tests.
+
+
+## Proposed Change
+The following components shall be required
+
+### Network driver `ND`
+The network driver is module which would implement the basic infrastructure management operations which would then be used in test cases to simulate different scenarios.
+
+The `Node` class shall have the following methods:
+- `url`
+- `start()`
+- `stop()`
+- `start_tendermint()`
+- `stop_tendermint()`
+- `configure_tendermint()`
+- `reset_tendermint()`
+- `start_db()`
+- `stop_db()`
+- `reset_db()`
+- `configure_bigchaindb()`
+- `configure_clock()`
+- `configure_QoS()` (discussed later)
+
+
+`configure_clock()`: This method should allow to change the clock on a given node. This could facilitate simulation of nodes located in different time zones.
+
+`configure_QoS()`: This methods allows configuring different scenarios like network latency, packet loss etc. The underlying implementation could be facilitated via [tc](https://wiki.linuxfoundation.org/networking/netem). 
+
+
+The network driver implementation could be achieved using the following libraries
+- https://github.com/kubernetes-incubator/client-python
+- https://github.com/docker/docker-py
+
+
+### Usage example
+
+```python
+def test_valid_transaction_is_synced_nodes(network):
+    alice, bob = crypto.generate_key_pair(), crypto.generate_key_pair()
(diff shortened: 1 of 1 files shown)


muawiakh:
The changes look good to me and +1 for putting them in the document and I really like the reference that you have given to the python clients, but I think this sort document should have more details, to make the execution smoother e.g.
- Where will be run this testing suite? Per commit? Nightly? On Travis? Setup a local/cloud based CI system?
- What will be the network size? Is it configurable?
- Can we add and remove nodes?
- What will we use, simply the Docker API/Docker compose or a Kubernetes environment? Because depending on what we choose, there will be a considerable amount of development time.
  - Automation of the infrastructure deployment.
- How do we clean up the system? After every test? Or after running the whole suite? 
  - How much time would it take, approximately?
- Is there a test plan? Can you link to it?

The usage example you have provided is very nice and simple and a developer writing a test should only be exposed to this minimal API. +1 for that. 

Let me know what you think and we can sync up on this.

kansi:
@muawiakh Those are some important queries, Following are tentative responses.

> -  Where will be run this testing suite? Per commit? Nightly? On Travis? Setup a local/cloud based CI system?

I am not really sure on how frequently integration tests should be run but I think it will depend on resource and time consumption. Furthermore, since we plan to use docker compose/kubernetes we should be able to simulate a test net on a single machine so I think Travis should prove sufficient?   (**suggestions regarding this are welcome**)

> - What will be the network size? Is it configurable?

The network size would depend on the tests since we wish to simulate scenarios where nodes join and leave the network.

> - Can we add and remove nodes?

Yes. This answer is purely from an implementation perspective i.e. the network driver should expose api which should allow adding and removing nodes. I acknowledge that achieving this with tendermint is not easy though (considering the current stable release). But this is expected to improve in future.

> - What will we use, simply the Docker API/Docker compose or a Kubernetes environment? Because depending on what we choose, there will be a considerable amount of development time.
        - Automation of the infrastructure deployment.

I am inclined towards using Kubernetes since we use it for production but I have kept this open in case someone has a strong case for using docker API.

> - How do we clean up the system? After every test? Or after running the whole suite?

In my view the system should be cleaned after the whole test suite. The primary reason is that it is only more realistic to have a some transactions lying around in the blockchain. Nevertheless, I think a `reset` api should available in the network driver.

> - How much time would it take, approximately?

You mean writing the whole integration test suite? I think it would depend on the time it takes to write the network driver because it is going to do the heavy lifiting. 

> - Is there a test plan? Can you link to it?

Is [this](https://github.com/bigchaindb/bigchaindb/pull/1964) what you are looking for?

Let's try to sync and discuss this in more detail 

Another point of this PR was to get the discussion on these things started. Thanks for these important queries/inputs.

muawiakh:
This makes sense, maybe you can integrate all these changes/comments in the document and I can review again. I would really like to help you out and work together on this. Let me know.

proposals/integration-infrastructure.md
```diff
@@ -0,0 +1,108 @@
+# Integration testing simulation
```
muawiakh: Changes requested in the discussion.

kansi: Sure, I am on it

kansi:
@ttmc The network driver itself will not setup the network, rather it would expose api which faciliate network setup. The test suite can have fixutres which setup the network (per session, module or per test case) or the tests can themselves create a network. The network driver will implement a class `Node` which would translate to a pod in kubernetes that runs `BigchainDB`, `MongoDB` and `Tendermint`.

The network driver has  a performace testing api i.e. `configure_QoS` which could simulate some network scenarios on a given node. Currently the api uses `tc` command to simulate network conditions but this suggestion is tentative as I couldn't find anything more suitable.

ttmc:
Maybe make this into a BEP?

PR #1988, opened 2018-01-08
Change "IPDB Testnet" to "BigchainDB Testnet" in docs

ttmc:
This PR is for the `tendermint` branch. It changes the "IPDB Testnet" to the "BigchainDB Testnet" in the docs (root and server docs).

The "IPDB Transaction Spec" will continue to be called that for now, so that wasn't changed.

The image in `docs/server/source/_static/arch.jpg` should also be updated but it needs more updates than just the "IPDB Node" label (e.g. it needs to add the Tendermint stuff), so I'll leave that for another PR.


ttmc:
@shahbazn I merged this (small) PR because I wanted to do a follow-up PR to update all links to the IPDB Transaction Spec. Feel free to take a look at this PR and give comments. It's small so that shouldn't take too long.

PR #1991, opened 2018-01-10
Network driver for integration tests

kansi:
Network driver for writing integration tests #1977 

#### `Node` Class
- [x] `ip` : Get the ip of the node.
- [x] `start()`: Start a node i.e. it will start `mongodb`, `tendermint`, `bigchaindb`
- [x] `stop()`: Stop a node.
- [x] `reset()`: Reset the state of an existing node i.e. data stored in Tendermint and MongoDB will be purged. **NOTE**: It is only realistic to reset the nodes after all the tests have been executed rather than resetting the network after each test.
- [x] `start_tendermint()`: Run Tendermint process on a given node.
- [x] `stop_tendermint()`: Stop Tendermint process on a given node.
- [x] `reset_tendermint()`: Reset Tendermint i.e. `tendermint unsafe_reset_all`.
- [x] `dial_seeds()`: Add Tendermint peers.
- [ ] `configure_tendermint()`: Alter Tendermint configuration on a given node.
- [x] `tendermint_uri`: Return the URI (along with port) where Tendermint is running.
- [x] `start_db()`: Start MongoDB daemon on a given node.
- [x] `stop_db()`: Stop MongoDB daemon on a given node.
- [x] `start_bigchaindb()`: Start bigchaindb server
- [x] `stop_bigchaindb()`: Stop bigchaindb server
- [x] `reset_bigchaindb()`: Flush bigchaindb data
- [ ] `configure_bigchaindb()`: Update BigchainDB configuration.
- [ ] `configure_clock()`: The method should allow to change the clock on a given node. This could facilitate simulation of nodes located in different time zones.
- [ ] `configure_QoS()`: Configure different scenarios like network latency, packet loss etc. The underlying implementation could be facilitated via [tc](https://wiki.linuxfoundation.org/networking/netem). 

#### `Network` Class
- [x] `ensure_started()`: ensures that `n` nodes are running in the network.
- [x] `ensure_connected()`: ensures that the running nodes in a network are connected.
- [x] `stop()`: ensure that `n` nodes are stopped in a network.


network_driver/Dockerfile.tendermint
```diff
@@ -0,0 +1,7 @@
+FROM tendermint/tendermint:develop
+WORKDIR /
+COPY ./tnode /tendermint
+COPY tendermint_entrypoint.sh tendermint_entrypoint.sh
+VOLUME /tmp/bigchain/tendermint /tendermint_node_data
```
kansi: @muawiakh @shahbazn The volume mount is same for all nodes. This will be problematic right?

muawiakh: these are the mount points for the container, you can attach it to different persistent volumes/storage classes. Shouldn't be a problem.

network_driver/build_images.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# build mongodb image
+docker build -t mongodb:bdb-itest -f Dockerfile.mongo .
+
+# build mongodb image
```
codegeschrei: I guess you meant tendermint

PR #1996, opened 2018-01-11
add mode parameter to transaction endpoint, fixes #1995

### LINKED ISSUE #1995: implement 'extending post transaction api'
implement the proposal #1951

codegeschrei:
## Description
Implementation of the proposal #1951 

## Issues This PR Fixes
Fixes #1995 

## Related PRs
#1951 

## Todos
- [x] Tested and working on development environment
- [x] Unit tests (if appropriate)
- [x] Added/Updated all related documentation. 

## Deployment Notes
None

## How to QA
None

None

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -41,10 +41,9 @@ class BigchainDB(Bigchain):
         # TODO: handle connection errors!
         requests.post(ENDPOINT, json=payload)
 
-    def write_transaction(self, transaction, **kwargs):
+    def write_transaction(self, transaction, **mode):
         # This method offers backward compatibility with the Web API.
         """Submit a valid transaction to the mempool."""
-        mode = kwargs
         self.post_transaction(transaction, mode)
 
     def store_transaction(self, transaction):
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -19,21 +19,21 @@ logger = logging.getLogger(__name__)
 TENDERMINT_HOST = getenv('TENDERMINT_HOST', 'localhost')
 TENDERMINT_PORT = getenv('TENDERMINT_PORT', '46657')
 ENDPOINT = 'http://{}:{}/'.format(TENDERMINT_HOST, TENDERMINT_PORT)
+MODE_LIST = ('broadcast_tx_async',
+             'broadcast_tx_sync',
+             'broadcast_tx_commit')
 
 
 class BigchainDB(Bigchain):
 
     def post_transaction(self, transaction, mode):
         """Submit a valid transaction to the mempool."""
-        mode_list = ('broadcast_tx_async',
-                     'broadcast_tx_sync',
-                     'broadcast_tx_commit')
-        if not mode or mode['mode'] not in mode_list:
+        if not mode or mode not in MODE_LIST:
             raise ValidationError(('Mode must be one of the following {}.')
-                                  .format(', '.join(mode_list)))
+                                  .format(', '.join(MODE_LIST)))
 
         payload = {
-            'method': mode['mode'],
+            'method': mode,
             'jsonrpc': '2.0',
             'params': [encode_transaction(transaction.to_dict())],
             'id': str(uuid4())
@@ -41,7 +41,7 @@ class BigchainDB(Bigchain):
         # TODO: handle connection errors!
         requests.post(ENDPOINT, json=payload)
 
-    def write_transaction(self, transaction, **mode):
+    def write_transaction(self, transaction, mode):
         # This method offers backward compatibility with the Web API.
         """Submit a valid transaction to the mempool."""
         self.post_transaction(transaction, mode)
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -23,11 +23,17 @@ ENDPOINT = 'http://{}:{}/'.format(TENDERMINT_HOST, TENDERMINT_PORT)
 
 class BigchainDB(Bigchain):
 
-    def post_transaction(self, transaction):
+    def post_transaction(self, transaction, mode):
         """Submit a valid transaction to the mempool."""
+        mode_list = ('broadcast_tx_async',
+                     'broadcast_tx_sync',
+                     'broadcast_tx_commit')
+        if not mode or mode['mode'] not in mode_list:
+            raise ValidationError(('Mode must be one of the following {}.')
+                                  .format(', '.join(mode_list)))
 
         payload = {
-            'method': 'broadcast_tx_async',
+            'method': mode['mode'],
             'jsonrpc': '2.0',
             'params': [encode_transaction(transaction.to_dict())],
             'id': str(uuid4())
@@ -35,11 +41,11 @@ class BigchainDB(Bigchain):
         # TODO: handle connection errors!
         requests.post(ENDPOINT, json=payload)
 
-    def write_transaction(self, transaction):
+    def write_transaction(self, transaction, **kwargs):
         # This method offers backward compatibility with the Web API.
         """Submit a valid transaction to the mempool."""
-
-        self.post_transaction(transaction)
+        mode = kwargs
+        self.post_transaction(transaction, mode)
 
     def store_transaction(self, transaction):
         """Store a valid transaction to the transactions collection."""
--- a/bigchaindb/web/views/parameters.py
+++ b/bigchaindb/web/views/parameters.py
@@ -29,4 +29,14 @@ def valid_operation(op):
         return 'CREATE'
     if op == 'TRANSFER':
         return 'TRANSFER'
-    raise ValueError('Operation must be "CREATE" or "TRANSFER')
+    raise ValueError('Operation must be "CREATE" or "TRANSFER"')
+
+
+def valid_mode(mode):
+    if mode == 'async':
+        return 'broadcast_tx_async'
+    if mode == 'sync':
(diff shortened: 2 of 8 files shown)


tests/web/test_transactions.py
```diff
@@ -410,3 +410,39 @@ def inner(self, tx_id, include_status):
...
+    from bigchaindb.models import Transaction
+    from bigchaindb.common.crypto import generate_key_pair
+    alice = generate_key_pair()
+    tx = Transaction.create([alice.public_key],
+                            [([alice.public_key], 1)],
+                            asset=None) \
+        .sign([alice.private_key])
+    mode_endpoint = TX_ENDPOINT + '?mode=nope'
+    response = client.post(mode_endpoint, data=json.dumps(tx.to_dict()))
+    assert '400 BAD REQUEST' in response.status
```
kansi: I think error response could also be `assert`ed here

bigchaindb/tendermint/lib.py
```diff
@@ -23,23 +23,29 @@
...
+            'method': mode['mode'],
             'jsonrpc': '2.0',
             'params': [encode_transaction(transaction.to_dict())],
             'id': str(uuid4())
         }
         # TODO: handle connection errors!
         requests.post(ENDPOINT, json=payload)
 
-    def write_transaction(self, transaction):
+    def write_transaction(self, transaction, **kwargs):
```
kansi: The function is being called by unpacking the args, so `mode` be directly extracted here right?

bigchaindb/tendermint/lib.py
```diff
@@ -23,23 +23,29 @@
 
 class BigchainDB(Bigchain):
 
-    def post_transaction(self, transaction):
+    def post_transaction(self, transaction, mode):
         """Submit a valid transaction to the mempool."""
+        mode_list = ('broadcast_tx_async',
+                     'broadcast_tx_sync',
+                     'broadcast_tx_commit')
+        if not mode or mode['mode'] not in mode_list:
```
kansi: I think adding parantheses around the appropriate sub-conditions would add more clarity.

bigchaindb/web/views/transactions.py
```diff
@@ -55,6 +55,12 @@ def post(self):
         Return:
             A ``dict`` containing the data about the transaction.
         """
+        parser = reqparse.RequestParser()
+        parser.add_argument('mode', type=parameters.valid_mode)
+        args = parser.parse_args()
+        if not str(args['mode']) or str(args['mode']) == 'None':
```
kansi: Small query, could these conditions be handled in `valid_mode` function?

codegeschrei: Actually no. 'None' is the default value, so the first case (empty mode) should actually never exist. The second check is for the default value which is set by flask. Since it is set by flask there is no need for validation. But I'll change it. I didn't think of it but I can just change the default value and we don't need this line at all.

bigchaindb/tendermint/lib.py
codegeschrei: I decided to not add them since it is still readable. But I'm also still open to change it if more people think we should have parentheses in this case.

tests/web/test_transactions.py
codegeschrei: Done

bigchaindb/web/views/transactions.py
kansi: I couldn't understand you, what I was asking is if following is possible
```python
def valid_mode(mode):
    if mode == 'async' or mode == None or mode == '':
        return 'broadcast_tx_async'
    if mode == 'sync':
        return 'broadcast_tx_sync'
    if mode == 'commit':
        return 'broadcast_tx_commit'
    raise ValueError('Mode must be "async", "sync" or "commit"')
```

codegeschrei: No. As far as I understand it when we don't set a mode the mode will be set to a default. In this case, it will never be `None`. Furthermore, I changed it now, so the default is always `'broadcast_tx_async'` instead of `'None'`. 
If I revert the changes and the default will be `'None'` it then is set by Flask and therefore doesn't enter the validation. 
I hope that explains it.

kansi: Yes, I think I understand it now.

docs/server/source/http-client-server-api.rst
```diff
@@ -161,6 +165,25 @@ Transactions
...
+   .. note::
+
+       This option is only available when using BigchainDB with Tendermint.
+
+   Tendermint offers a `broadcast API
+   <http://tendermint.readthedocs.io/projects/tools/en/master/using-tendermint.html#broadcast-api>`_ with three different modes to send transactions.
+   By setting the mode, a new transaction can be pushed with a different mode than the default. The default mode is ``broadcast_tx_async``, which
+   will return immediately and not wait to see if the transaction is valid.
+
+   :query string mode: (Optional) One of the three supported modes to send a transaction: ``broadcast_tx_async``, ``broadcast_tx_sync``, ``broadcast_tx_commit``.
```
kansi: Shouldn't the possible mode values be `async`, `sync` and `commit` here?

codegeschrei: oh! You're right! I'll change it.

bigchaindb/tendermint/lib.py
```diff
@@ -23,23 +23,28 @@
 
 class BigchainDB(Bigchain):
 
-    def post_transaction(self, transaction):
+    def post_transaction(self, transaction, mode):
         """Submit a valid transaction to the mempool."""
+        mode_list = ('broadcast_tx_async',
```
vrde: You can move it outside the method, or outside the class, and make it an all capital "constant".

bigchaindb/web/views/parameters.py
```diff
@@ -29,4 +29,14 @@ def valid_operation(op):
         return 'CREATE'
     if op == 'TRANSFER':
         return 'TRANSFER'
-    raise ValueError('Operation must be "CREATE" or "TRANSFER')
+    raise ValueError('Operation must be "CREATE" or "TRANSFER"')
```
ttmc: Haha nice catch. I did a double-take before I could see what you fixed.

docs/server/source/http-client-server-api.rst
```diff
@@ -137,6 +137,10 @@ Transactions
        <https://docs.bigchaindb.com/projects/py-driver/en/latest/index.html>`_
        to build a valid transaction.
 
+   A generalization of the parameter follows:
+
+   :param mode: (Optional) Set the broadcast method.
```
ttmc: It wasn't clear to me what these two lines meant until later. At first, I thought you'd somehow forgotten to include the `?mode={mode}` in the subsection header (which seemed very unlikely because that's what this whole PR is about).

Maybe after "Push a new transaction." you could have a sentence or two noting that there's a variation of this endpoint that includes a `mode` parameter, documented below (with a link to it). Also note that this version (with no mode specified) is the same as using `mode=async`.

docs/server/source/http-client-server-api.rst
```diff
@@ -161,6 +165,25 @@ Transactions
    :statuscode 202: The pushed transaction was accepted in the ``BACKLOG``, but the processing has not been completed.
    :statuscode 400: The transaction was malformed and not accepted in the ``BACKLOG``.
 
+.. http:post:: /api/v1/transactions?mode={mode}
+
+   .. note::
+
+       This option is only available when using BigchainDB with Tendermint.
```
ttmc: Since the naming of things isn't settled yet, this should be clear enough for now.

docs/server/source/http-client-server-api.rst
ttmc: It would be helpful to the reader to summarize what each mode means here, e.g. 

- `async` means post the transaction and return
- `sync` means post the transaction and return after `checkTx` is executed
- `commit` means post the transaction and return after the transaction has been committed

docs/server/source/http-client-server-api.rst
```diff
@@ -161,6 +165,25 @@ Transactions
...
+   By setting the mode, a new transaction can be pushed with a different mode than the default. The default mode is ``broadcast_tx_async``, which
+   will return immediately and not wait to see if the transaction is valid.
+
+   :query string mode: (Optional) One of the three supported modes to send a transaction: ``broadcast_tx_async``, ``broadcast_tx_sync``, ``broadcast_tx_commit``.
+
+   The only change for a request, compared to the one above, is to set the mode
+
+    .. sourcecode:: http
+
+        GET /api/v1/transactions?mode={mode} HTTP/1.1
```
ttmc: Maybe make this an example rather than this generic line, which is the same as the subsection heading, e.g.

...is to set the mode. For example:

    GET /api/v1/transactions?mode=commit HTTP/1.1

ttmc: Actually, are the expected response status codes the same, i.e. `202 Accepted` and `400 Bad Request`?

codegeschrei: The response status codes don't change.

docs/server/source/http-client-server-api.rst
ttmc: I see you changed the order of these two versions of the POST /api/v1/transactions endpoint to have the one with the `mode` parameter first. That's more clear.

docs/server/source/http-client-server-api.rst
ttmc: ok

codegeschrei:
@ttmc yes of course :) Thanks!

Issue #1946, opened 2017-12-07
HTTP api integration with tendermint


kansi:
Following from #1828 most of the APIs have been integrated. Below are some remaining apis which would need some thinking or they are no longer much useful because of tendermint.

<hr>

#### [Transaction status](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-statuses?transaction_id=transaction_id) = **GET /api/v1/statuses?transaction_id={transaction_id}** `DEPRICATED`

Done?
- [x] pull request #1996 : add mode parameter to transaction endpoint, resolved #1951 and #1995

EDIT: The above PR allows the user to post a transaction in different modes. So, the user doesn't need to continously check the status of a transaction i.e. they can `POST` the transaction in `async` or `commit` mode.

The transaction status is crucial API for the client to figure out if and when a transaction has been successfully accepted by the network. The older api provided three `status` values for a posted transaction i.e. `backlog`, `undecided` and `valid`. To replicate the same we can take the following approach:
- create a `statuses` collection.
- when a txn in posted from bigcahindb to tendermint, add the txn to `statuses` collection with `status` as `backlog`
- when a txn passes `checkTx` the status is updated from `backlog` to `undecided`
- after `deliverTx` is executed successfully the txn in removed from the `statuses` collection as the txn is now present in the `transactions` collection which implies that its `valid`.

In order to check the status of a txn, the `transactions` collection is check first and if the txn in not found then it is searched in `statuses` collection else HTTP `404` status is returned.

**NOTE**: the `backlog` status of a txn is only available from the node to which the txn was sent by the client.

**NOTE**: an alternate way to achieve this is suggested here #1951 

<hr>

#### [Get Block](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-blocks-block_id) = **GET /api/v1/blocks/{block_id}**

- [x] Done? See pull request #1970 

There are two ways to port this API
- Stored the block `id` a given transaction belongs to in the transaction itself. **NOTE** this would change the txn schema
- Store a list of txn `ids` in each document of `blocks` collection and then load the txns accordingly.

**NOTE**: Fields `"node_pubkey"`, `"voters"`, `"signature"` might not be avaiable because of integration with Tendermint. Although fields like `"node_pubkey"` and `"signature"` could be computed when abci calls `commit` to commit the block. 

<hr>

[...]

PR #2021, opened 2018-02-01
Integrate get block with transaction id api

kansi:
### Description
Following from issue #1946 this PR implements `GET /api/v1/blocks?transaction_id={transaction_id}&status={UNDECIDED|VALID|INVALID}` api. In order to maintain compatiability with 1.x there is no change to the query parameters i.e. the client still specifies `transaction_id` and `status` but in the current scenrio specifiying `status` other than `VALID` will always return empty list in response.

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -191,7 +191,7 @@ class BigchainDB(Bigchain):
         else:
             return block
 
-    def get_blocks_status_containing_tx(self, txid):
+    def get_block_containing_tx(self, txid):
         """Retrieve the list of blocks (block ids) containing a
            transaction with transaction id `txid`
 
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -153,7 +153,8 @@ class BigchainDB(Bigchain):
             return block
 
     def get_blocks_status_containing_tx(self, txid):
-        """Retrieve block id containing transaction id `txid`
+        """Retrieve the list of blocks (block ids) containing a
+           transaction with transaction id `txid`
 
         Args:
             txid (str): transaction id of the transaction to query
@@ -162,6 +163,9 @@ class BigchainDB(Bigchain):
             Block id list (list(int))
         """
         blocks = list(backend.query.get_block_with_transaction(self.connection, txid))
+        if len(blocks) > 1:
+            logger.critical('Transaction id %s exists in multiple blocks', txid)
+
         return [block['height'] for block in blocks]
 
     def validate_transaction(self, tx):
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -152,6 +152,18 @@ class BigchainDB(Bigchain):
         else:
             return block
 
+    def get_blocks_status_containing_tx(self, txid):
+        """Retrieve block id containing transaction id `txid`
+
+        Args:
+            txid (str): transaction id of the transaction to query
+
+        Returns:
+            Block id list (list(int))
+        """
+        blocks = list(backend.query.get_block_with_transaction(self.connection, txid))
+        return [block['height'] for block in blocks]
+
     def validate_transaction(self, tx):
         """Validate a transaction against the current status of the database."""
 
--- a/docs/server/source/http-client-server-api.rst
+++ b/docs/server/source/http-client-server-api.rst
@@ -645,19 +645,22 @@ Blocks
 
    :statuscode 400: The request wasn't understood by the server, e.g. just requesting ``/blocks`` without the ``block_id``.
 
-.. http:get:: /api/v1/blocks?transaction_id={transaction_id}&status={UNDECIDED|VALID|INVALID}
+.. http:get:: /api/v1/blocks?transaction_id={transaction_id}
 
    Retrieve a list of ``block_id`` with their corresponding status that contain a transaction with the ID ``transaction_id``.
 
-   Any blocks, be they ``UNDECIDED``, ``VALID`` or ``INVALID`` will be
-   returned if no status filter is provided.
+   .. note::
+       The query parameter ``status`` has been depricated. The parameter allowed
+       users to filter blocks based on their status i.e. only blocks with the specified
+       status were included in the response. Since then this behavior has changed
+       and now block are created only after the transactions are accepted by the
+       network i.e. blocks have only one status ``VALID``
 
    .. note::
        In case no block was found, an empty list and an HTTP status code
        ``200 OK`` is returned, as the request was still successful.
 
    :query string transaction_id: transaction ID *(required)*
-   :query string status: Filter blocks by their status. One of ``VALID``, ``UNDECIDED`` or ``INVALID``.
 
    **Example request**:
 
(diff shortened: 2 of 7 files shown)


ttmc:
@kansi I am having a hard time understanding what all this pull request is doing. I guess it's some code related to the

**GET /api/v1/blocks?transaction_id={transaction_id}&status={UNDECIDED|VALID|INVALID}**

endpoint, but there doesn't seem to be any change to the Flask code, nor any changes noted in the HTTP API documentation. Maybe you could add some description in the opening comment of the PR.

kansi:
@ttmc I updated the issue description, updating the docs.

vrde:
For me, we can drop the `status` parameter.

kansi:
@vrde @ttmc The `status` query paramter is removed. A note has been added in the documentation to reflect the same.

bigchaindb/tendermint/lib.py
```diff
@@ -152,6 +152,18 @@ def get_block(self, block_id, include_status=False):
         else:
             return block
 
+    def get_blocks_status_containing_tx(self, txid):
+        """Retrieve block id containing transaction id `txid`
```
ttmc: Maybe:

Retrieve the list of blocks (block ids) containing a transaction with transaction id `txid`

docs/server/source/http-client-server-api.rst
```diff
@@ -645,19 +645,22 @@ Blocks
...
 
-.. http:get:: /api/v1/blocks?transaction_id={transaction_id}&status={UNDECIDED|VALID|INVALID}
+.. http:get:: /api/v1/blocks?transaction_id={transaction_id}
 
    Retrieve a list of ``block_id`` with their corresponding status that contain a transaction with the ID ``transaction_id``.
 
-   Any blocks, be they ``UNDECIDED``, ``VALID`` or ``INVALID`` will be
-   returned if no status filter is provided.
+   .. note::
+       The query parameter ``status`` has been depricated. The parameter allowed
```
ttmc: deprecated

ttmc: Also, maybe change "The parameter allowed" to "It allowed"

docs/server/source/http-client-server-api.rst
```diff
@@ -671,7 +674,7 @@ Blocks
 
    :resheader Content-Type: ``application/json``
 
-   :statuscode 200: A list of blocks containing a transaction with ID ``transaction_id`` was found and returned.
+   :statuscode 200: A list containing atmost one block height where ``transaction_id`` was found and returned.
```
ttmc: atmost --> at most

docs/server/source/http-client-server-api.rst
```diff
@@ -645,19 +645,22 @@ Blocks
 
    :statuscode 400: The request wasn't understood by the server, e.g. just requesting ``/blocks`` without the ``block_id``.
 
-.. http:get:: /api/v1/blocks?transaction_id={transaction_id}&status={UNDECIDED|VALID|INVALID}
+.. http:get:: /api/v1/blocks?transaction_id={transaction_id}
 
    Retrieve a list of ``block_id`` with their corresponding status that contain a transaction with the ID ``transaction_id``.
```
ttmc: This line should also be edited. How about:
```text
Retrieve a list of block IDs (block heights), such that the blocks with those IDs
contain a transaction with the ID ``transaction_id``.
```

docs/server/source/http-client-server-api.rst
ttmc: I wonder what happens if two or more block heights are found? Would it just happily return the list and a 200 status code.

It seems to me that something is seriously wrong if a list of more than one block height comes back. Maybe a CRITICAL logging message should be written to the logs, and the HTTP response could ask them to report the incident to the network operators?

kansi: Under normal operation that should not happen but if it does its an indication of some malicious activity on the node or something is critically wrong. Either way it would be good to generate a log message.

ttmc: I see that you added the CRITICAL log message if the response list has more than one block ID.

I guess what will happen is that the HTTP response will contain a list with more than one block ID, and the response status code will be 200. Maybe the description of that 200 status code could be changed a bit, to reflect that possibility.

The status code 200 can also happen if there are no blocks containing a transaction with that transaction ID.

Maybe

```text
   :statuscode 200: The request was properly formed and zero or more blocks were found containing the specified ``transaction_id``.
```

Maybe the sentence describing this endpoint ("Retrieve a list of block IDS...") could be followed by a second sentence noting that the list should be either empty or contain only one block ID.

bigchaindb/tendermint/lib.py
ttmc: :+1:

docs/server/source/http-client-server-api.rst
ttmc: :+1:

docs/server/source/http-client-server-api.rst
ttmc: :+1:

docs/server/source/http-client-server-api.rst
```diff
@@ -671,7 +674,7 @@ Blocks
 
    :resheader Content-Type: ``application/json``
 
-   :statuscode 200: A list of blocks containing a transaction with ID ``transaction_id`` was found and returned.
+   :statuscode 200: A list containing at most one block height where ``transaction_id`` was found and returned.
```
ttmc: I see that you added the CRITICAL log message if the response list has more than one block ID. Good!

If there's more than one block with the given transaction_id, I guess what will happen is that the HTTP response will contain a list with more than one block ID, and the response status code will be 200. Maybe the description of that 200 status code could be changed a bit, to reflect that possibility.

I think the status code 200 can also happen if there are no blocks containing a transaction with that transaction ID.

Maybe:

```text
   :statuscode 200: The request was properly formed and zero or more blocks were found containing the specified ``transaction_id``.
```

Also, maybe the sentence describing this endpoint ("Retrieve a list of block IDS...") could be followed by a second sentence noting that the list should be either empty or contain only one block ID.

bigchaindb/tendermint/lib.py
```diff
@@ -191,6 +191,22 @@ def get_block(self, block_id, include_status=False):
         else:
             return block
 
+    def get_blocks_status_containing_tx(self, txid):
```
vrde: Since we have instant finality, I'd keep the `status` out of the tendermint module. It's OK to change the code in ` bigchaindb/web/views/blocks.py` to call another method.

bigchaindb/tendermint/lib.py
```diff
@@ -191,6 +191,22 @@ def get_block(self, block_id, include_status=False):
...
+           transaction with transaction id `txid`
+
+        Args:
+            txid (str): transaction id of the transaction to query
+
+        Returns:
+            Block id list (list(int))
+        """
+        blocks = list(backend.query.get_block_with_transaction(self.connection, txid))
+        if len(blocks) > 1:
```
vrde: I wonder if we really need this check. This is Tendermint's responsability.

kansi: I agree this is Tendermint's responsibility but aim here to generate a meaningful log message in case state is critically inconsistant.

PR #2045, opened 2018-02-13
Crash recovery mechanism

### LINKED ISSUE #2018: Design recovery mechanisms for node crash
### Description
As of now the transactions are written to the database during `deliver_tx`, with #2017 the transaction will be written during block `commit`. Since we are using mongodb as the underlying database, `insert_many` is not transactional. So, in case `insert_many` crashes midway the database will be left in an inconsistent state. This crash might be because of improper bigchaindb shutdown, system failure etc.

### Solutions
This error could be handeled in various ways,

#### Approach 1
Catch when `insert_many` throws error and delete the documents which were successfully inserted so as to return to a valid state (two-phase-commit). This approach may work in some cases but will fall short in scenarios where the bigchain process is killed for some reason or the vm shuts down abruptly etc.

#### Approach 2
Devise a recovery mechanism i.e. during `commit` the last db operation is to store the block (which contains a list of transaction ids for that block and app hash) after which the db is in a consistent state. After this the new app hash is returned to tendermint. So, when a crash occurs bigchaindb could be doing the following:
 - failed while writing transactions with `insert_many`
 - Wrote transactions but failed to write the block
 - Wrote transaction and block but failed while returning the new app hash

In general what we might have is a bunch of transactions which don't belong to any block or an app hash which is not registered back to tendermint.  To get to a consistent state we can create a command `bigchaindb recover` which would find and delete "dangling" transactions i.e. transactions which are not included in any block. Furthermore, we delete the latest committed block along with its transactions which would handle the case where transactions and block are written successfully but failure occurred while returning the new app hash to tendermint.

To summarize: "dangling" transactions and the latest block are deleted so when Bighchaindb is started back again it will sync up and start adding new transactions.

NOTE: We could implement both approach 1 & 2

kansi:
Fixes #2018 
The PR implements approach 1 and 2 from issue #2018.
- A new subcommand `recover` has been added. The sub-command checks for any assets or metadata which don't have a corresponding transaction. Furthermore, it also checks for transactions which are not included in any blocks.
- The implemetation also catches any execption during `store_bulk_transactions` and reverts the operation.


vrde:
This thing just popped up in my mind today (unfortunately not during discussing #2018): do we really need **approach 1**? What if instead we just run **approach 2** (aka the `recover` command) automatically every time we start a node, and specifically run only `delete_zombie_transactions` (no need for `delete_latest_block`)?

Our [`commit`](https://github.com/bigchaindb/bigchaindb/blob/a7ed8cf4cbd794d9c66578e305074b2335b1d3e0/bigchaindb/tendermint/core.py#L113) function is:
```python
    def commit(self):
        """Store the new height and along with block hash."""

        # register a new block only when new transactions are received
        if self.block_txn_ids:
            self.bigchaindb.store_bulk_transactions(self.block_transactions)
            block = Block(app_hash=self.block_txn_hash,
                          height=self.new_height,
                          transactions=self.block_txn_ids)

        data = self.block_txn_hash.encode('utf-8')
        return Result.ok(data=data)
```

The main ways `commit ` can create an inconsistent state are:
- during/after `store_bulk_transactions` and before `store_block`. AFAIK this can be detected by `delete_zombie_transactions` on startup.
- after `store_block` but before returning the result to the Tendermint state machine. We will know this only after Tendermint queries [`info`](https://github.com/bigchaindb/bigchaindb/blob/a7ed8cf4cbd794d9c66578e305074b2335b1d3e0/bigchaindb/tendermint/core.py#L38) and we end up in an inconsistent state.

I'm asking those questions because I see an opportunity to simplify the code a bit 👨‍🚀

kansi:
> do we really need approach 1? 

Not really sure if we do. The original thought behind proposing it was to handle scenarios where mongodb connection breaks down while `commit`ting a transaction. 
If the error is propogated it may crash the abci server? If we treat this as a fatal crash and bring down the node then we won't need to use approach 1 because on restart we plan to execute recover by default.

> ... and specifically run only delete_zombie_transactions (no need for delete_latest_block)?

`delete_latest_block` was added with the understanding that after writing the new block and before returning the new app hash if there is a node failure then tendermint will not have the latest app hash. Futhermore, when the node is brought back online bigchaindb would try to return this new (un-registered) apphash to tendermint as the lastest app hash. Since tendermint would have no knowledge of this hash it wouldn't know what to do?

vrde:
> If the error is propogated it may crash the abci server? If we treat this as a fatal crash and bring down the node then we won't need to use approach 1 because on restart we plan to execute recover by default.

If we are talking about the same error, it is already propagated by the `raise` in the `try ... except` block in `store_block` and `store_bulk_transactions`. AFAIK the case covered by _approach 1_ is:
1. something bad happens while writing transactions in MongoDB
1. the transactions **not** belonging to any block are deleted
1. the system exits (because of the `raise`)

> `delete_latest_block` was added with the understanding that after writing the new block and before returning the new app hash if there is a node failure then tendermint will not have the latest app hash. Futhermore, when the node is brought back online bigchaindb would try to return this new (un-registered) apphash to tendermint as the lastest app hash. Since tendermint would have no knowledge of this hash it wouldn't know what to do?

In the Tendermint code there is a [check in the ReplayBlocks function](https://github.com/tendermint/tendermint/blob/1d6f00859dd46beae036146782f0f088b2e50e63/consensus/replay.go#L261) that should answer the question. I don't know if Tendermint communicates the error back to BigchainDB. Can you check that? I can do it if you don't have time.

About this PR, it makes sense to me, but I'd discard _approach 1_, what do you think?

kansi:
> In the Tendermint code there is a check in the ReplayBlocks function that should answer the question. I don't know if Tendermint communicates the error back to BigchainDB. Can you check that? I can do it if you don't have time

@vrde Tendermint node crashes with the following error:

```bash
ERROR: Failed to create node: Error starting proxy app connections: Error on replay: App block height (10) is higher than core (1)
```

Furthermore, instead of always calling `delete_latest_block` we can check the latest block height that tendermint recorded and act accordintly i.e. we call `http://localhost:46657/status` and check the lastest block height witnessed by Tendermint and based on which we decided to delete the latest block.

Regarding **approach 1** I think we can skip it because we could have to crash the node anyway to recover from it so no point handle all these errors and still crashing the node.

vrde:
> Furthermore, instead of always calling delete_latest_block we can check the latest block height that tendermint recorded and act accordintly i.e. we call http://localhost:46657/status and check the lastest block height witnessed by Tendermint and based on which we decided to delete the latest block.

Oh yes! That's correct. 

> Regarding approach 1 I think we can skip it because we have to crash the node anyway to recover from it so no point handle all these errors and still crashing the node.

Related to that, would you mind updating this PR to reflect those changes? I can help if you want.

kansi:
@vrde I will update this PR with the aforementioned changes before tomorrow.

bigchaindb/commands/bigchaindb.py
```diff
@@ -154,6 +156,23 @@ def run_init(args):
...
+
+def _run_recover(b):
+    query.delete_zombie_transactions(b.connection)
+
+    tendermint_height = b.get_latest_block_height_from_tendermint()
+    block = b.get_latest_block()
+
+    if block:
+        while block['height'] > tendermint_height:
+            query.delete_latest_block(b.connection)
```
vrde: Would you mind adding a `log.info` here, saying something like `bigchaindb is ahead of tendermint, removing block %s`

bigchaindb/commands/bigchaindb.py
```diff
@@ -292,6 +313,9 @@ def create_parser():
     subparsers.add_parser('drop',
                           help='Drop the database')
 
+    subparsers.add_parser('recover',
```
vrde: Do we still need to expose the `recover` command?

ttmc:
@kansi @vrde Now that this crash recovery mechanism is merged, and it's a bit different from what was proposed in issue #2018, it would be good to document what actually happens. Fortunately, @shahbazn created a nice project where you could do that: a document that tells the story of a transaction from the time it's first posted to a BigchainDB instance: see pull request #2074

PR #2048, opened 2018-02-13
Changed constructed/allowed transaction 'version' from '1.0' to '2.0'

### LINKED ISSUE #1907: update tx model to v2.0
**after issue #1891/#1899**

- [x] update tx model to v2.0
- [x] add error message for tx model v1.0

ttmc:
Resolves #1907 

- Created three new JSON Schema files for the v2.0 transaction schema, based on the v1.0 files. The only change is the allowed value of `version` (from `"^1\\.0$"` to `"^2\\.0$"`).
- I didn't delete the v1.0 JSON Schema files because we might want those some day (to validate old transactions).
- Updated the `__init__.py` in the directory with the JSON Schema files so that it now uses the version 2.0 JSON Schema files.
- Updated all relevant tests. I only found one, in `tests/validation/test_transaction_structure.py`
- Updated `VERSION` in `common/transaction.py`
- Checked to make sure the example HTTP API docs show "version" with value "2.0". They do.
- Updated the docs page about "The Transaction Model". It just points to the IPDB Transaction Spec.
- If someone submits a transaction with "version" having value "1.0" then the error message comes from the JSON Schema checker.

### DIFF
--- a/docs/server/source/data-models/transaction-model.rst
+++ b/docs/server/source/data-models/transaction-model.rst
@@ -2,21 +2,3 @@ The Transaction Model
 =====================
 
 See the `IPDB Transaction Spec <https://github.com/ipdb/ipdb-tx-spec>`_.
-
-
-The Transaction Schema
-----------------------
-
-BigchainDB checks all transactions (JSON documents)
-against a formal schema defined
-in some `JSON Schema <http://json-schema.org/>`_ files.
-Those files are part of the 
-`IPDB Transaction Spec <https://github.com/ipdb/ipdb-tx-spec>`_.
-Their official source is the ``tx_schema/`` directory
-in the `ipdb/ipdb-tx-spec repository on GitHub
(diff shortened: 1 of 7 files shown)


docs/server/source/data-models/transaction-model.rst
```diff
@@ -2,21 +2,3 @@ The Transaction Model
 =====================
 
 See the `IPDB Transaction Spec <https://github.com/ipdb/ipdb-tx-spec>`_.
```
vrde: Should it refer to something under the BigchainDB organization?

ttmc: For now, the transaction spec / IPDB Transaction Spec lives under the `ipdb` organization. I think I'll eventually rename it and move it under the `bigchaindb` organization (e.g. maybe under the new /rfc repository), but I have higher-priority things to do first.

PR #2053, opened 2018-02-14
Problem: tests don't pick up env vars

vrde:
Solution: add a call to `config_utils.env_config` in the `_bdb` fixture.


bigchaindb/commands/bigchaindb.py
```diff
@@ -271,7 +271,10 @@ def create_parser():
                                           help='Prepare the config file '
                                                'and create the node keypair')
     config_parser.add_argument('backend',
-                               choices=['rethinkdb', 'mongodb', 'localmongodb'],
+                               choices=['localmongodb'],
```
vrde: Here I'm limiting the backend choice to just `localmongodb`. I can understand if we want to move this to a different PR.

Issue #2064, opened 2018-02-16
(Tendermint Integration) Update the Event Plugin API


ttmc:
The Event Plugin API needs to be updated to account for changes now that we're using Tendermint. For example:

- `bigchaindb/bigchaindb/events.py` contains a list of all the possible events. Today that list includes `BLOCK_VALID` and `BLOCK_INVALID`. That list will have to be updated. Maybe `BLOCK_VALID` becomes `BLOCK_COMMITTED` and `BLOCK_INVALID` is removed and not replaced.
- More code updates, including tests.
- The associated documentation also has to be updated, in `bigchaindb/docs/server/source/events/event-plugin-api.rst` and maybe elsewhere.

ttmc:
Note:  When we set up a two-node BigchainDB cluster last Monday, it all worked, but the Events API was throwing errors and not working.

PR #2065, opened 2018-02-16
Updated (almost) all docs regarding "votes" once Tendermint integrated

ttmc:
I grepped all the docs for "vote" (case-insensitive) and updated all associated mentions of votes to account for how things will work with Tendermint integrated. Two entire pages got deleted:

- Data Models - The Vote Model
- Appendices - The Vote Schema File

(Deleting the actual vote schema file at `bigchaindb/common/schema/vote.yaml` caused many things to break so doing that was deferred until later.)

The only exception was the docs page about the Event Stream API. Updating that will require some code changes first, so I created a new issue to do that: #2064


ttmc:
@kansi - While writing the blog post about the changes to the HTTP API, I noticed several more changes that will have to be made to the HTTP API docs. (For example, there's no longer a concept of "invalid blocks" so all that will go away.) I'll do those changes in another PR that uses this branch as a starting point.

PR #2069, opened 2018-02-19
Remove all code for all /statuses endpoints

ttmc:
The HTTP API docs were showing `"statuses": "/api/v1/statuses/",` in the response to the BigchainDB Root URL and in the response to the API Root Endpoint. That's because the code implementing those endpoints was still around.

This pull request removes all code for all the `/statuses` endpoints, and updates all related tests.

The HTTP API docs now exclude all mention of the `/statuses` endpoints, everywhere.


tests/web/test_transactions.py
```diff
@@ -44,9 +44,6 @@ def test_post_create_transaction_endpoint(b, client):
 
     assert res.status_code == 202
 
-    assert '../statuses?transaction_id={}'.format(tx.id) in \
-        res.headers['Location']
```
kansi: 👍  Just a side note. This module is yet to be cleaned and integrated with tendermint compaitable apis.

PR #2070, opened 2018-02-19
Add 'upsert-validator' command and necessary db queries

kansi:
### Description
Allow node admins to add/update/remove validators.

It implements the spec created in pull request https://github.com/bigchaindb/rfc/pull/3


bigchaindb/commands/bigchaindb.py
```diff
@@ -278,6 +292,16 @@ def create_parser():
                                help='The backend to use. It can be either '
                                     'rethinkdb or mongodb.')
 
+    validator_parser = subparsers.add_parser('upsert-validator',
+                                             help='Add/update/delete a validator')
+
+    validator_parser.add_argument('public_key',
+                                  help='Public key of the new validator.')
```
ttmc: It might not be a _new_ validator. It might be an existing validator that's being updated or deleted. Maybe `'Public key of the validator.'`?

bigchaindb/commands/bigchaindb.py
```diff
@@ -278,6 +292,16 @@ def create_parser():
...
+
+    validator_parser.add_argument('power',
+                                  type=int,
+                                  help='Voting power of the validator.')
```
ttmc: Maybe add a short second sentence about how setting it to zero will delete the validator, e.g.

`'Voting power of the validator. Setting it to 0 will delete the validator.'`

ttmc:
Now that I look at this implementation, I worry that we've created something that's way more complicated than necessary. I'll make a suggestion for how to simplify it over on the original spec pull request, https://github.com/bigchaindb/rfc/pull/3

ttmc:
See https://github.com/bigchaindb/rfc/pull/3#issuecomment-367712822

bigchaindb/tendermint/core.py
```diff
@@ -121,3 +131,12 @@ def commit(self):
...
         return Result.ok(data=data)
+
+
+def cast_validator(v):
+    pub_key = v['pub_key']['data']
+    # NOTE: tendermint expects public to be ecoded in go-wire format
+    # so `01` has to be appended
+    pubKey = bytes.fromhex('01{}'.format(pub_key))
+    return Validator(pubKey=pubKey,
+                     power=v['power'])
```
ttmc: Why does nothing have to be appended before the power? I guess an integer is a basic type and doesn't need a prefix in go-wire encoding.

ttmc:
I don't know how to test this PR, we need more tooling for that. I will create an issue to revisit testing this once we _do_ have the tooling (hopefully soon).

The docs need updating as well: a new HTTP API endpoint, a new CLI subcommand

kansi:
@ttmc WIll create a seperate PR for updating the docs. The tasks are being tracked here #2098

bigchaindb/tendermint/core.py
kansi: Yup, but more specifically `power` is not expected to be encoded in go-wire format i.e. only `pubKey` is expected to be encoded in go-wire format.

bigchaindb/common/utils.py
```diff
@@ -5,6 +5,8 @@
 import bigchaindb
 from bigchaindb.common.exceptions import ValidationError
 
+VALIDATOR_UPDATE_ID = 'validator01'
```
muawiakh: nit: `validator01` is a bit ambiguous in terms of the idea, that this used to block new requests to add/remove validators.

kansi: @muawiakh I just want a string which I can use as a unique id. If you have any suggestions on better naming do let me know.

ttmc: Maybe change `validator01` to `a_unique_id_string` so it becomes super-clear what its purpose is.

ttmc:
I suggested a solution to [the nit raised by @muawiakh ](https://github.com/bigchaindb/bigchaindb/pull/2070#discussion_r173761128).

Once that nit is taken care of, which wouldn't take long, I think we could merge this pull request. (@vrde is also listed as a reviewer but he's away and it's not clear when he will get back.)

bigchaindb/tendermint/lib.py
```diff
@@ -335,5 +335,15 @@ def validate_transaction(self, tx, current_transactions=[]):
     def fastquery(self):
         return fastquery.FastQuery(self.connection, self.me)
 
+    def get_validator_update(self):
+        update = backend.query.get_validator_update(self.connection,
+                                                    VALIDATOR_UPDATE_ID)
+        print(update)
```
vrde: Should be a `log` or removed.

bigchaindb/tendermint/core.py
```diff
@@ -121,3 +131,12 @@ def commit(self):
 
         data = self.block_txn_hash.encode('utf-8')
         return Result.ok(data=data)
+
+
+def cast_validator(v):
+    pub_key = v['pub_key']['data']
+    # NOTE: tendermint expects public to be ecoded in go-wire format
```
vrde: `ecoded` → `encoded`

bigchaindb/tendermint/core.py
```diff
@@ -121,3 +131,12 @@ def commit(self):
 
         data = self.block_txn_hash.encode('utf-8')
         return Result.ok(data=data)
+
+
+def cast_validator(v):
```
vrde: Is it more casting or encoding?

kansi:
> AFAIK I cannot update a validator that has not been processed yet, unless I remove it and add it again, is it correct?

@vrde In general new request cannot be made until the existing request has be processed. This is true for all the operations (add/update/delete). So the remove will not work either.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -4,10 +4,12 @@
 
 from bigchaindb import backend
 from bigchaindb.backend.exceptions import DuplicateKeyError
+from bigchaindb.common.exceptions import MultipleValidatorOperationError
```
vrde: It should be plural `MultipleValidatorsOperationError` :pray:

kansi: I didn't use the plural because the term `ValidatorOperation` is the name for each update request to validator set i.e.
```json
{"validator": {"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
               "power":10},
 "update_id": "validator01"
}
```
The naming is such because each operation can request update to only 1 validator object. So, the plural would be `ValidatorOperations` or `MultipleValidatorOperation`

vrde: Fair enough!

PR #2074, opened 2018-02-20
Adding tutorial about bigchaindb post transaction workflow

shahbazn:
## Description
This PR adds a tutorial to capture the detailed workflow of what happens when someone posts a transaction to the bigchaindb test network.

### DIFF
--- a/docs/what-happens-when-post-tx.md
+++ b/docs/what-happens-when-post-tx.md
@@ -4,15 +4,14 @@ Imagine I have a bigchaindb test network deployed at https://test.bigchaindb.com
 I'd probably type something like this in my terminal:
 
 ```bash
-curl -XPOST https://test.bigchaindb.com -d @tx_data.json
+curl -XPOST https://test.bigchaindb.com/api/v1/transactions -d @tx_data.json
 ```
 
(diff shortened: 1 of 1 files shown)


ttmc:
For now, I think it would be most useful to focus on what happens to a transaction once it arrives "at" a BigchainDB instance, skipping over the journey of the transaction from the client to the BigchainDB instance.

If we get this document in nice shape, it could be added to the ReadTheDocs docs somewhere.

ttmc:
I wonder if this could become an "Informational" BEP, over in the `bigchaindb/BEPs` repository.

ttmc:
Another possibility is that this ("The life of a BigchainDB transaction") goes in the new BigchainDB whitepaper, because it provides a nice narrative structure very similar to parts of the old whitepaper, and once one reads that, they have a good overview of the BigchainDB system and how all the pieces work together.

ttmc:
I've decided to put a section about "The life of a BigchainDB transaction" in the new BigchainDB whitepaper (which should be done in about a week). That new whitepaper isn't done yet, but I'm going to close this pull request just so we can de-clutter the list.

ttmc:
If you're curious, you can read the new whitepaper (a work in progress as I write this) at https://www.overleaf.com/read/spdhxcycjggm

PR #2078, opened 2018-02-21
Problem: processes don't have an intelligible name

vrde:
Solution: when a process starts, update the "process title" as well.

This is how `pstree` looks like with this patch.
image

### DIFF
--- a/bigchaindb/tendermint/commands.py
+++ b/bigchaindb/tendermint/commands.py
@@ -1,7 +1,8 @@
 import logging
 import subprocess
-import multiprocessing as mp
 from os import getenv
+
+import setproctitle
 
 import bigchaindb
 from bigchaindb.tendermint.lib import BigchainDB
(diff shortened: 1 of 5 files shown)


bigchaindb/tendermint/commands.py
```diff
@@ -37,36 +38,38 @@ def start():
...
         bigchaindb_factory=BigchainDB)
-    p_webapi = mp.Process(name='webapi', target=app_server.run)
+    p_webapi = Process(name='webapi', target=app_server.run)
     p_webapi.start()
 
     # start message
     logger.info(BANNER.format(bigchaindb.config['server']['bind']))
 
-    if int(getenv('BIGCHAINDB_START_TENDERMINT', 1)):
+    if int(getenv('BIGCHAINDB_START_TENDERMINT', 0)):
```
kansi: +1

vrde: oh oh oh :santa: :santa: :santa: I forgot this. I think the default setting should be `0`, but I'll put it back to `1` because it doesn't belong here.

PR #2079, opened 2018-02-21
Problem: exchange pubsub doesn't start on boot

vrde:
Solution: start it on boot


ttmc:
Haha, love it. This is my kind of pull request.

PR #2080, opened 2018-02-21
Hardcode static ports for k8s deployment

### LINKED ISSUE #2077: Make container ports static for k8s deployment templates
* BigchainDB version: `tendermint`
* Deployment Type: `k8s`
* BigchainDB driver: `no`

### Description
Currently we manually configure ports in deployment and service files for k8s deployment e.g [here](https://github.com/bigchaindb/bigchaindb/blob/d5f45d4094a3e3671cf55722b9632d083ffccf56/k8s/bigchaindb/bigchaindb-dep.yaml#L123). Instead these ports should be constant since these are always the same for each deployment

shahbazn:
## Description
Currently we manually configure ports in deployment and service files for k8s deployment e.g [here](https://github.com/bigchaindb/bigchaindb/blob/d5f45d4094a3e3671cf55722b9632d083ffccf56/k8s/bigchaindb/bigchaindb-dep.yaml#L123). Instead these ports should be constant since these are always the same for each deployment

## Issues This PR Fixes
Fixes #2077 

## Todos
- [ ] Added/Updated all related documentation #2081 .

### DIFF
--- a/k8s/bigchaindb/bigchaindb-dep.yaml
+++ b/k8s/bigchaindb/bigchaindb-dep.yaml
@@ -120,13 +120,13 @@ spec:
         - |
           bigchaindb -l DEBUG start
         ports:
-        - containerPort: "<bigchaindb-api-port from ConfigMap>"
+        - containerPort: 9984
           protocol: TCP
           name: bdb-port
-        - containerPort: "<bigchaindb-ws-port from ConfigMap>"
+        - containerPort: 9985
           protocol: TCP
           name: bdb-ws-port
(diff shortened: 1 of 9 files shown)


ttmc:
This alone will lead to a great shortening of the docs, and a reduction in work-to-be-done for setup. Nice.

shahbazn:
@muawiakh its part of the todo's being tracked as a separate ticket.
>Looks good to me and this is a good change, but this will require a lot of changes in the documents. Can you take care of those as well? Can you mention that in the PR(other requirements/dependencies)?

PR #2082, opened 2018-02-21
Problem: tendermint is started by default

vrde:
Solution: do not start tendermint. It's up to the user to start tendermint or not for now.

### DIFF
--- a/bigchaindb/tendermint/commands.py
+++ b/bigchaindb/tendermint/commands.py
@@ -1,6 +1,4 @@
 import logging
-import subprocess
-from os import getenv
 
 import setproctitle
 
@@ -43,13 +41,6 @@ def start():
 
     # start message
     logger.info(BANNER.format(bigchaindb.config['server']['bind']))
-
-    if int(getenv('BIGCHAINDB_START_TENDERMINT', 1)):
-        subprocess.Popen([
-            'tendermint',
-            'node',
-            '--consensus.create_empty_blocks=false',
-        ])
 
     # start websocket server
     p_websocket_server = Process(name='ws',


bigchaindb/tendermint/commands.py
```diff
@@ -44,7 +44,7 @@ def start():
     # start message
     logger.info(BANNER.format(bigchaindb.config['server']['bind']))
 
-    if int(getenv('BIGCHAINDB_START_TENDERMINT', 1)):
+    if int(getenv('BIGCHAINDB_START_TENDERMINT', 0)):
```
ttmc: `BIGCHAINDB_START_TENDERMINT` could just have a default value of `False` or  `0`, rather than doing a check like this here (and anywhere else it gets used). That way Tendermint won't get started unless they explicitly set it to `True` or `1` (either using an environment variable or a BigchainDB config file).

`BIGCHAINDB_START_TENDERMINT` should also be documented in the docs page about BigchainDB configuration settings, i.e. in `bigchaindb/docs/server/source/server-reference/configuration.md`

What will happen if `BIGCHAINDB_START_TENDERMINT` is truthy _and_ Tendermint is already running? Maybe there should be a check if Tendermint is already running, before running it again?

vrde: > `BIGCHAINDB_START_TENDERMINT` could just have a default value of `False` or `0`, rather than doing a check like this here (and anywhere else it gets used). That way Tendermint won't get started unless they explicitly set it to `True` or `1` (either using an environment variable or a BigchainDB config file).

Yes, it should be in the config file, but →

> `BIGCHAINDB_START_TENDERMINT` should also be documented in the docs page about BigchainDB configuration settings, i.e. in `bigchaindb/docs/server/source/server-reference/configuration.md`

→ `BIGCHAINDB_START_TENDERMINT` was supposed to be a quick workaround, and →

> What will happen if `BIGCHAINDB_START_TENDERMINT` is truthy and Tendermint is already running? Maybe there should be a check if Tendermint is already running, before running it again?

→ this problem was not taken into account. To me, we can even remove the whole code that is starting tendermint.

ttmc: > To me, we can even remove the whole code that is starting tendermint.

Okay!

kansi: @ttmc I think it will eventually get removed (soon) because production setup will start tendermint in seperate container and dev (by dev I mean users building ontop of BDB) will use some container management utility to setup the system so tendermint will have its own container there too.

vrde: If you don't mind I can change the scope of this PR to remove that part of the code.

vrde:
Oops, I didn't push the latest code. Doing it now.

PR #2084, opened 2018-02-21
Update docker images and Dockerfiles

muawiakh:


k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: bigchaindb
-        image: bigchaindb/bigchaindb:1.3.0
+        image: bigchaindb/bigchaindb:unstable
```
ttmc: I wonder if we should change the `imagePullPolicy` on these "unstable" Docker images to `Always` --- because the content of the `bigchaindb/bigchaindb:unstable` image can change. Either that or we could adopt some naming convention for the "unstable" images such as `unstable1`, `unstable2`, etc.

k8s/mongodb/container/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
+FROM mongo:3.4.13
 LABEL maintainer "dev@bigchaindb.com"
 WORKDIR /
 RUN apt-get update \
     && apt-get -y upgrade \
     && apt-get autoremove \
     && apt-get clean
-COPY mongod.conf.tmt.template /etc/mongod.conf
-COPY mongod_tmt_entrypoint.bash /
+COPY mongod.conf.template /etc/mongod.conf
```
ttmc: Did you do a grep for all occurrences of `tmt`?

k8s/mongodb/mongo-ss.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: mongodb
-        image: bigchaindb/localmongodb:1.0
+        image: bigchaindb/localmongodb:unstable
```
ttmc: See my other comment about the `unstable` bigchaindb/bigchaindb images. Do you think the `bigchaindb/localmongodb:unstable` image will change much?

k8s/nginx-http/nginx-http-dep.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       terminationGracePeriodSeconds: 10
       containers:
       - name: nginx
-        image: bigchaindb/nginx_http:1.1
+        image: bigchaindb/nginx_http:unstable
```
ttmc: See my other comments about the `unstable` images, and whether to change the `imagePullPolicy` or to adopt a naming convention such as `unstable1`, `unstable2`, etc.

k8s/bigchaindb/bigchaindb-dep.yaml
muawiakh: I think changing the `imagePullPolicy` is a better idea, I will take care of it.

k8s/mongodb/container/Dockerfile
muawiakh: yes.

k8s/bigchaindb/bigchaindb-dep.yaml
muawiakh: I think as long as, we keep it `unstable` i.e. till the official release, I will keep the policy as `Always`

k8s/tendermint/tendermint_container/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM tendermint/tendermint:develop
+FROM tendermint/tendermint:0.13
```
shahbazn: lets keep all images to unstable for now until we are ready to tag a realease?

muawiakh: The tendermint we will use for k8s is still unstable, I am changing this because the current integration is based out of `0.13` and they can make breaking changes in `develop` branch as the base image.

shahbazn: okay

k8s/mongodb/container/Dockerfile
ttmc: :+1: Yay grep!

Issue #2037, opened 2018-02-12
Bring back a /votes endpoint in the HTTP API?


ttmc:
There was and HTTP API endpoint:

`GET /api/v1/votes?block_id={block_id}`

but it went away with the Tendermint integration. Can we bring something like that back?

Copied from https://github.com/bigchaindb/bigchaindb/issues/1946 

> The block data provided by tendermint only contains the validators hash. This hash can verified by hashing the current known validator peers. Furthermore, the signature field would need to be dropped. It would be good to re-evaluate this api from the point of view of what tendermint is providing.

Copied from https://github.com/bigchaindb/bigchaindb/issues/1828 

> [List of votes](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-votes?block_id=block_id) could be extracted from block header obtained via`/block?height=_`. [More info](http://tendermint.readthedocs.io/en/master/specification/block-structure.html#header)

PR #2086, opened 2018-02-22
Make votes endpoint return 404 and a helpful message

ttmc:
The votes endpoint in the HTTP API is going away, but it might come back in the future; see issue #2037 

This PR keeps the votes endpoint (and puts existing code + test code in comments), but changes the response to 'The votes endpoint is gone now, but it might return in the future.' with HTTP status code 404 Not Found, which means "The requested resource could not be found but may be available in the future."


ttmc:
Travis CI is failing on some flake8 errors. I'll fix those before merging. Turns out, I had pre-commit installed but not *installed*...

Issue #2081, opened 2018-02-21
Update k8s deployment docs


shahbazn:
### Description
As part of simplifying k8s production deployment, documentation will change significantly and will be updated once simplification work is complete

PR #2088, opened 2018-02-22
Adding secret access token header checks for authorization

### LINKED ISSUE #2035: Restricted-Access Private Network without Using 3scale
### Problem

The current production deployment template uses 3scale to ensure that POST requests to the network (from anyone) only get through if they come from a client with an account (app_id and app_key).

A private network wants to launch so that _all_ HTTP requests (POST and GET) sent to the nodes in the network get be dropped unless they come from a small set of known (and unchanging) clients/sources. They don't need 3scale. They will want a modified version of the production deployment template.

### Potential Solution

- Generate a special HTTP header and share it with all the known clients/sources.
- Have a single NGINX in each node which checks for that HTTP header value. If it's present, let the request pass through to the network. (HTTP headers are encrypted if HTTPS is used.)

Are there other simpler or better options?

shahbazn:
## Description
Currently production deployment templates use threescale apicast for authentication and authorization. This PR adds a second option to restrict POST API access to a bigchaindb node based on a secret access token programmed into nginx conf file.

## Issues This PR Fixes
Fixes #2035 

## Todos
- [ ] Added/Updated all related documentation #2081

### DIFF
--- a/k8s/nginx-https/container/nginx.conf.template
+++ b/k8s/nginx-https/container/nginx.conf.template
@@ -1,6 +1,7 @@
 # Frontend API server that:
 # 1. Acts as the HTTPS termination point.
-# 2. Forwards BDB HTTP requests to OpenResty backend.
+# 2. Authorizes HTTP requests with secret token header
+#    and forwards to BDB backend.
 # 3. Forwards BDB WS requests to BDB backend.
 # 4. Does health check with LB.
 
--- a/k8s/nginx-https/container/nginx.conf.template
+++ b/k8s/nginx-https/container/nginx.conf.template
@@ -57,9 +57,6 @@ http {
   map $remote_addr $bdb_backend {
     default BIGCHAINDB_BACKEND_HOST;
   }
-  map $remote_addr $openresty_backend {
-    default OPENRESTY_BACKEND_HOST;
-  }
 
   # Frontend server for the external clients; acts as HTTPS termination point.
   server {
@@ -116,7 +113,7 @@ http {
         return 403;
       }
 
-      # POST requests get forwarded to OpenResty instance. Enable CORS too.
+      # POST requests get forwarded to BDB.
       if ($request_method = POST ) {}
         proxy_pass http://$bdb_backend:BIGCHAINDB_API_PORT;
       }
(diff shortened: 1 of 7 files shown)


muawiakh:
quick question, the default value for `authorization_mode` is `threescale`?

shahbazn:
@muawiakh yes default is set to threescale in entry point script
>quick question, the default value for `authorization_mode` is `threescale`?

ttmc:
@shahbazn Maybe the default `authorization_mode` should be changed to `secret-header` so that our friends have one less thing to think about?

k8s/nginx-https/container/nginx.conf.template
```diff
@@ -100,27 +100,27 @@ http {
...
         add_header 'Content-Length' 0;
         return 204;
       }
 
+      # Check for security header to authorize POST requests
+      if ( $http_x_secret_access_token != "SECRET_ACCESS_TOKEN" ) {
+        return 403;
+      }
+
+      # POST requests get forwarded to OpenResty instance. Enable CORS too.
```
ttmc: Is this comment correct? It seems the following code is forwarding POST requests to the BigchainDB instance.

shahbazn: no will update this good catch

shahbazn:
@ttmc both are separate parameters `authorization_mode` set the mode of authorization and `secrete-access-token` is used to set the secrete token if `authorization_mode=secrete-access-token`
>@shahbazn Maybe the default authorization_mode should be changed to secret-header so that our friends have one less thing to think about?

ttmc:
What are the possible values of `authorization_mode`?

shahbazn:
@ttmc possible values for `authorization_mode` are `[threescale | secret-token]`

k8s/nginx-https/container/nginx.conf.threescale.template
```diff
@@ -0,0 +1,198 @@
+# Frontend API server that:
+# 1. Acts as the HTTPS termination point.
+# 2. Forwards BDB HTTP requests to OpenResty backend.
```
shahbazn: missed this

Issue #2090, opened 2018-02-22
Problem: deadlock between tendermint and abci server


sbellem:
### Problem
When tendermint is first started it tries to connect to the ABCI server. Until it establishes a connection with the ABCI server, calls to it such as requesting status information may fail.

`run_recover` tries to make a request to tendermint before the ABCI server is started and this causes errors, e.g.

```bash
$ docker-compose -f docker-compose.tendermint.yml up --no-deps bdb                  
Creating bigchaindb_bdb_1 ... done                              
Attaching to bigchaindb_bdb_1   
bdb_1          | INFO:bigchaindb.config_utils:Configuration loaded from `/root/.bigchaindb`                                      
bdb_1          | [2018-02-22 23:22:13] [INFO] (bigchaindb.commands.bigchaindb) BigchainDB Version 1.4.0.dev (MainProcess - pid: 1)
bdb_1          | [2018-02-22 23:22:13] [DEBUG] (bigchaindb.config_utils) System already configured, skipping autoconfiguration (MainProcess - pid: 1)
bdb_1          | [2018-02-22 23:22:13] [DEBUG] (bigchaindb.backend.connection) Connection: <class 'bigchaindb.backend.localmongodb.connection.LocalMongoDBConnection'> (MainProcess - pid: 1)
bdb_1          | [2018-02-22 23:22:13] [DEBUG] (urllib3.connectionpool) Starting new HTTP connection (1): tendermint (MainProcess - pid: 1)
bdb_1          | Traceback (most recent call last):             
bdb_1          |   File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 141, in _new_conn                   
bdb_1          |     (self.host, self.port), self.timeout, **extra_kw)                                                           
bdb_1          |   File "/usr/local/lib/python3.6/site-packages/urllib3/util/connection.py", line 83, in create_connection       
bdb_1          |     raise err  
bdb_1          |   File "/usr/local/lib/python3.6/site-packages/urllib3/util/connection.py", line 73, in create_connection       
bdb_1          |     sock.connect(sa)                           
bdb_1          | ConnectionRefusedError: [Errno 111] Connection refused                                                          
bdb_1          |                
bdb_1          | During handling of the above exception, another exception occurred:                                             
bdb_1          |                
bdb_1          | Traceback (most recent call last):             
[...]

PR #2091, opened 2018-02-23
Rollback crash recovery mechanism

kansi:
Following from #2090 , the crash recovery mechanism is being disabled untill a fix is created.


muawiakh:
I understand that this is merged but I think the reverted changes should also include travis files. @kansi @vrde

kansi:
@muawiakh I think @sbellem will try to address those changes as he is more acquainted with them.

PR #2092, opened 2018-02-23
Migrate cli for tendermint integration

### LINKED ISSUE #2087: Migrate cli for Tendermint
With Tendermint integration some of the cli sub-commands have been rendered obsolete. It would be only appropriate to remove those sub-commands.

- [x] Submit a proposal on how this problem will be handled.
- [x] Implement the accepted proposal.(Core changes)
- [x] Update the documentation.

muawiakh:
## Description
With Tendermint integration some of the cli sub-commands were rendered obsolete. This PR removes those commands and references to those commands.

- `bigchaindb --help` (Updated help message)
- `bigchaindb export-my-pubkey`
- `bigchaindb set-shards`
- `bigchaindb set-replicas`
- `bigchaindb add-replicas`
- `bigchaindb remove-replicas`
- Two `bigchaindb start` options:
  - `--dev-allow-temp-keypair`
  - `--dev-start-rethinkdb`

## Issues This PR Fixes
Partially Fixes #2087 

## Related PRs

Repo/Branch | PR
------ | ------
bigchaindb/bigchaindb/tree/tendermint | [Docs](https://github.com/bigchaindb/bigchaindb/pull/2093)
bigchaindb/bigchaindb/tree/tendermint | [Spec document](https://github.com/bigchaindb/bigchaindb/pull/1976)

## Deployment Notes
N/A

## How to QA
Re-enabled unit tests.

CLI and documentation only.


bigchaindb/tendermint/event_stream.py
```diff
@@ -13,7 +13,7 @@
 
 HOST = getenv('BIGCHAINDB_TENDERMINT_HOST', 'localhost')
 PORT = int(getenv('BIGCHAINDB_TENDERMINT_PORT', 46657))
-URL = f'ws://{HOST}:{PORT}/websocket'
+URL = 'ws://{HOST}:{PORT}/websocket'
```
kansi: I couldn't understand this, why don't we need formating?

bigchaindb/commands/bigchaindb.py
```diff
@@ -10,22 +10,17 @@
...
 from bigchaindb import backend
 from bigchaindb.backend import schema
 from bigchaindb.backend import query
-from bigchaindb.backend.admin import (set_replicas, set_shards, add_replicas,
-                                      remove_replicas)
-from bigchaindb.backend.exceptions import OperationError
 from bigchaindb.commands import utils
 from bigchaindb.commands.messages import (
     CANNOT_START_KEYPAIR_NOT_FOUND,
-    RETHINKDB_STARTUP_ERROR,
```
ttmc: I think _maybe_ you can remove the import of `CANNOT_START_KEYPAIR_NOT_FOUND` as well, including the call (below) that uses it. A BigchainDB instance doesn't have any internal keypair any more, so `KeypairNotFoundException` is probably defunct and can also be deleted everywhere in all the code. Maybe such keypair-related changes could be done in a separate pull request, since it's _a different concern_ that's only peripherally related to the CLI changes (or is it?). There could be a PR to scrub all code related to the BigchainDB node's keypair. A simpler principle to apply might just be: "make lots of small pull requests, each with narrow scope."

export_my_pubkey is definitely _in_ scope since that's a CLI command.

bigchaindb/commands/bigchaindb.py
```diff
@@ -313,63 +237,12 @@ def create_parser():
     start_parser = subparsers.add_parser('start',
                                          help='Start BigchainDB')
 
-    start_parser.add_argument('--dev-allow-temp-keypair',
```
ttmc: `--dev-allow-temp-keypair` and `--dev-start-rethinkdb` were undocumented and quasi-secret, nevertheless they existed as valid CLI arguments. I wonder if you should mention removing them in the proposal, i.e. in PR #1976

compose/travis/Dockerfile
```diff
@@ -12,11 +12,8 @@ ENV PYTHONUNBUFFERED 0
 
 ENV BIGCHAINDB_DATABASE_PORT 27017
 ENV BIGCHAINDB_DATABASE_BACKEND $backend
-ENV BIGCHAINDB_SERVER_BIND 0.0.0.0:9984
```
ttmc: I don't understand why these Dockerfile ENV variables are being removed in this PR. Same for the corresponding settings removed in `docker-compose.travis.yml`.

bigchaindb/tendermint/event_stream.py
muawiakh: This is a syntax error no? flake8 was also failing.

compose/travis/Dockerfile
muawiakh: These variables are not needed by the travis configuration, but they were failing the tests, I enabled because default values for BIGCHAINDB_SERVER_BIND is `localhost:9984`, BIGCHAINDB_WSSERVER_HOST is `localhost` and BIGCHAINDB_WSSERVER_ADVERTISED_HOST is `localhost`. Which made some tests fail that assert the default values. Either I needed to change here since this is not being used OR change the default values in the BigchainDB config. But if we set it to `0.0.0.0`, some use cases will need to be updated as well.

bigchaindb/tendermint/event_stream.py
ttmc: f-strings were added in Python 3.6:

https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498

bigchaindb/commands/bigchaindb.py
ttmc: I added a bullet point about removing these in the comment description of this PR (at the top of this page).

bigchaindb/tendermint/event_stream.py
kansi: @muawiakh My query was, do variables get replaced without using `f`? because I think without `f` value of `URL` will be `'ws://{HOST}:{PORT}/websocket'`

compose/travis/Dockerfile
ttmc: Okay maybe leave them removed then. I just didn't understand why removal of some CLI subcommands would require changing `compose/travis/Dockerfile` in this way, but now I do.

bigchaindb/tendermint/event_stream.py
muawiakh: Makes sense, now. I didn't know about that, I will update again.

bigchaindb/commands/bigchaindb.py
muawiakh: Makes sense, I think addressing this in a separate PR is a better option.

compose/travis/Dockerfile
sbellem: I doubt very strongly that the env vars should be changed or removed. Running the tests in a dev environment with docker-compose fails for some tests now. The tests, or config mechanism needs to be fixed not the docker files. 

If the same modifications are applied to the docker-compose file used for development the following error occurs:

```bash
bdb_1          | Process ws:                                                                             
bdb_1          | Traceback (most recent call last):                                                      
bdb_1          |   File "/usr/local/lib/python3.6/asyncio/base_events.py", line 1044, in create_server   
bdb_1          |     sock.bind(sa)                             
bdb_1          | OSError: [Errno 99] Cannot assign requested address                                                           
bdb_1          |                                                                                         
bdb_1          | During handling of the above exception, another exception occurred:                                           
bdb_1          |                                                                                         
bdb_1          | Traceback (most recent call last):            
bdb_1          |   File "/usr/local/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap   
bdb_1          |     self.run()                                                                          
bdb_1          |   File "/usr/src/app/bigchaindb/utils.py", line 38, in run                                                    
bdb_1          |     super().run()                                                                       
bdb_1          |   File "/usr/local/lib/python3.6/multiprocessing/process.py", line 93, in run                                 
bdb_1          |     self._target(*self._args, **self._kwargs)                                           
bdb_1          |   File "/usr/src/app/bigchaindb/web/websocket_server.py", line 173, in start                                  
bdb_1          |     port=config['wsserver']['port'])          
bdb_1          |   File "/usr/local/lib/python3.6/site-packages/aiohttp/web.py", line 457, in run_app                          
bdb_1          |     asyncio.gather(*server_creations, loop=loop)                                                              
bdb_1          |   File "/usr/local/lib/python3.6/asyncio/base_events.py", line 467, in run_until_complete                     
bdb_1          |     return future.result()                    
bdb_1          |   File "/usr/local/lib/python3.6/asyncio/base_events.py", line 1048, in create_server                         
bdb_1          |     % (sa, err.strerror.lower()))
bdb_1          | OSError: [Errno 99] error while attempting to bind on address ('::1', 9985, 0, 0): cannot assign requested add
ress
```

Some key parts of the application (seems to be `bigchaindb/web/websocket_server.py` in this case) are not tested, and that is most probably why removing the env vars for travis worked, but this can hardly be a path forward.

PR #2093, opened 2018-02-23
Migrate CLI documentation

### LINKED ISSUE #2087: Migrate cli for Tendermint

muawiakh:
## Description
With Tendermint integration some of the cli sub-commands were rendered obsolete. This PR removes those commands and references to those commands in docs.

## Issues This PR Fixes
Partially fixes #2087 

## Related PRs

Repo/Branch | PR
------ | ------
bigchaindb/bigchaindb/tree/tendermint | [Spec Doc](https://github.com/bigchaindb/bigchaindb/pull/1976)
bigchaindb/bigchaindb/tree/tendermint | [Core changes](https://github.com/bigchaindb/bigchaindb/pull/2092)

Docs only!

### DIFF
--- a/docs/server/source/server-reference/bigchaindb-cli.md
+++ b/docs/server/source/server-reference/bigchaindb-cli.md
@@ -15,10 +15,10 @@ Show the version number. `bigchaindb -v` does the same thing.
 
 ## bigchaindb configure
 
-Generate a local configuration file (which can be used to set some or all [BigchainDB node configuration settings](configuration.html)). It will ask you for the values of other configuration settings.
+Generate a local configuration file (which can be used to set some or all [BigchainDB node configuration settings](configuration.html)). It will ask you for the values of some configuration settings.
 If you press Enter for a value, it will use the default value.
 
-At this point only one database backend is supported: `localmongodb`.
+At this point, only one database backend is supported: `localmongodb`.
 
 If you use the `-c` command-line option, it will generate the file at the specified path:
 ```text
@@ -27,7 +27,7 @@ bigchaindb -c path/to/new_config.json configure localmongodb
 
 If you don't use the `-c` command-line option, the file will be written to `$HOME/.bigchaindb` (the default location where BigchainDB looks for a config file, if one isn't specified).
 
-If you use the `-y` command-line option, then there won't be any interactive prompts: it will use the default values for all the other configuration settings.
+If you use the `-y` command-line option, then there won't be any interactive prompts: it will use the default values for all the configuration settings.
 ```text
 bigchaindb -y configure localmongodb
 ```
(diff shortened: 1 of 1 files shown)


docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -39,32 +38,23 @@ bigchaindb -y configure rethinkdb
...
-various backend database indexes,
-and the genesis block.
+Create a backend database (Local MongoDB), all database tables/collections,
+various backend database indexes, and the genesis block.
 
 
 ## bigchaindb drop
 
-Drop (erase) the backend database (a RethinkDB or MongoDB database).
+Drop (erase) the backend database (a Local MongoDB database).
```
kansi: Not sure if the document should include the word 'Local'. It was called `localmongodb` because the name `mongodb` was already there in the code. So, `localmongodb` is just an implementation detail. @ttmc ?

muawiakh: I talked to @ttmc about this exactly, and he suggested that it should be localmongodb.

ttmc: What  I meant is that we should keep using `localmongodb` for now, because the different name helps us distinguish the old `mongodb` option from the new one (`localmongodb`). `localmongodb` is even the name of a subdirectory in the `bigchaindb/bigchaindb/backend/` directory. The different name makes it easy for us to distinguish the old from the new, for now. Maybe in the future, once all the old code is gone, we can shorten it to `mongodb`.

Anyway, in this line, we're not referring to that particular label, which is more of a code/configuration value. We're referring to a MongoDB database, the local one used by this node, so what you wrote is fine, but I'd write "Local" in all-lowercase and add a clarifying phrase:

`Drop (erase) the backend database (the local MongoDB database used by this node).`

(I also changed the article before `local` from `a` to `the` so it agrees with the article before  `backend database`.)

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -39,32 +38,23 @@ bigchaindb -y configure rethinkdb
...
-
-
 ## bigchaindb init
 
-Create a backend database (RethinkDB or MongoDB),
-all database tables/collections,
-various backend database indexes,
-and the genesis block.
+Create a backend database (Local MongoDB), all database tables/collections,
+various backend database indexes, and the genesis block.
```
ttmc: I'm not sure if Tendermint wants us to create a Genesis block. We can leave this text alone for now, but I created a new issue #2094 to figure out what to do with our code to create a Genesis block.

PR #2095, opened 2018-02-26
Automate MongoDB user creation for prod/test deployments

muawiakh:
## Description
- Currently, we had to manually log into the MongoDB container
   and create users, this change will configure the relevant users
   from a single script `configure_mdb.sh`
- Improvements can be done but keeping it minimal for the workshop
- Usage:
  - `bash configure_mdb.sh <mongodb-instance-name>`
    - **Example** `$ bash configure_mdb.sh mdb-instance-1`

### DIFF
--- a/k8s/mongodb/container/configure_mdb_users.template.js
+++ b/k8s/mongodb/container/configure_mdb_users.template.js
@@ -1,4 +1,4 @@
-use admin;
+db = db.getSiblingDB("admin");
 db.createUser({
     user: "MONGODB_ADMIN_USERNAME",
     pwd: "MONGODB_ADMIN_PASSWORD",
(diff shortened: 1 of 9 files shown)


k8s/mongodb/container/configure_mdb_users.template.js
```diff
@@ -0,0 +1,43 @@
+use admin;
```
ttmc: Does this work? I was under the impression that the JavaScript equivalent of `use admin` (a mongo shell command) is

```javascript
db = db.getSiblingDB("admin")
```

muawiakh: Let me try this then I can update.

PR #2106, opened 2018-02-27
Get validator set from the node

kansi:
## Description
Following from bigchaindb/beps#3 this PR introduces the `GET /api/v1/validators` api.

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -253,8 +253,9 @@ class BigchainDB(Bigchain):
 
             return validators
 
-        except Exception:
-            raise Exception('Error while processing data from tendermint.')
+        except requests.exceptions.RequestException as e:
+            logger.error('Error while connecting to Tendermint HTTP API')
+            raise e
 
 
 Block = namedtuple('Block', ('app_hash', 'height', 'transactions'))
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -253,7 +253,7 @@ class BigchainDB(Bigchain):
 
             return validators
 
-        except:
+        except Exception:
             raise Exception('Error while processing data from tendermint.')
 
 
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -243,5 +243,18 @@ class BigchainDB(Bigchain):
     def fastquery(self):
         return fastquery.FastQuery(self.connection, self.me)
 
+    def get_validators(self):
+        try:
+            resp = requests.get(f'{ENDPOINT}validators')
+            validators = resp.json()['result']['validators']
+            for v in validators:
+                v.pop('accum')
+                v.pop('address')
+
+            return validators
+
+        except:
+            raise Exception('Error while processing data from tendermint.')
+
 
 Block = namedtuple('Block', ('app_hash', 'height', 'transactions'))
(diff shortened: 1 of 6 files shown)


bigchaindb/tendermint/lib.py
```diff
@@ -243,5 +243,18 @@ def validate_transaction(self, tx, current_transactions=[]):
     def fastquery(self):
         return fastquery.FastQuery(self.connection, self.me)
 
+    def get_validators(self):
+        try:
+            resp = requests.get(f'{ENDPOINT}validators')
+            validators = resp.json()['result']['validators']
+            for v in validators:
+                v.pop('accum')
```
muawiakh: Just a question: Why are you removing these keys? What is there significance?

tests/web/test_validators.py
```diff
@@ -0,0 +1,47 @@
+import pytest
```
muawiakh: Do we have any tests for an actual MongoDB instance running and querying for `validators` collection?

vrde: We don't store the validators in MongoDB, we [query tendermint directly](https://github.com/bigchaindb/bigchaindb/pull/2106/files#diff-c6511560546a7dc577e7e647b5bfdaceR248).

bigchaindb/tendermint/lib.py
kansi: Since we query Tendermint for the validator set, we just clean up the response a little and serve just the necessary data. The usual response looks something as below,

```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_height": 1,
    "validators": [
      {
        "address": "6FEE524CE1E29425EF369CA63FFB1A9204EB0F12",
        "pub_key": {
          "type": "ed25519",
          "data": "B0E42D2589A455EAD339A035D6CE1C8C3E25863F268120AA0162AD7D003A4014"
        },
        "voting_power": 10,
        "accum": -10
      },
      {
        "address": "F5426F0980E36E03044F74DD414248D29ABCBDB2",
        "pub_key": {
          "type": "ed25519",
          "data": "4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"
        },
        "voting_power": 10,
        "accum": 10
      }
    ]
  }
}
```

ttmc:
Did you update the HTTP API docs to add documentation of this endpoint or was that a separate pull request? (That would be fine, there are just so many pull requests right now.)

ttmc:
Oh, I see you updated the docs in #2123 . Great! Approving.

bigchaindb/web/views/validators.py
```diff
@@ -0,0 +1,24 @@
...
+
+        pool = current_app.config['bigchain_pool']
+
+        try:
+            with pool() as bigchain:
+                validators = bigchain.get_validators()
+
+            return validators
+
+        except Exception:
```
vrde: :warning: catch the exceptions you expect may happen :warning: 

A generic `except Exception` will catch **everything**:
```
>>> issubclass(NameError, Exception)
True

>>> issubclass(SyntaxError, Exception)
True
```

A system that crashes because we didn't catch an exception is easier to debug than one that suppresses **any** exception.

Also, if the exception is not catch I think `flask` is returning a `500` anyway.

kansi: Ideally the only exception I expect is a connection to the Tendermint's HTTP API. In case of such connection error the client should still see `500` which implies that we don't need to handle any exceptions here as `flask` by default is returning `500` anyway. The addendum I made now is to handle the connection exception and create a log message for easier debugging.

ttmc:
It's not clear to me what's blocking the merge of this pull request. Are you waiting for @vrde to approve your changes and answers to his questions?

kansi:
@ttmc Yes

PR #2107, opened 2018-02-27
Problem: Production deployment template was not updated after v1.0 automation

muawiakh:
## Description
- Update the production deployment template for v1.0 automation.
- Add docs in workflow if user does not use 3scale
- Note about accessing UI and adding a ‘/’ at the end
- Make mdb-instance-0 default value for configure mdb users script instead of returning error
- Remove --context from docs in kubectl commands
- Make creating a premium account default instruction(premium storage)

### DIFF
--- a/docs/server/source/production-deployment-template/workflow.rst
+++ b/docs/server/source/production-deployment-template/workflow.rst
@@ -1,3 +1,5 @@
+.. _kubernetes-template-overview:
+
 Overview
 ========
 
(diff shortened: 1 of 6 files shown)


docs/server/source/production-deployment-template/workflow.rst
```diff
@@ -74,112 +76,32 @@ We'll say more about that file below.)
...
-(Each Cloud Manager "Project" has its own ``Project ID``. A ``Project ID`` can
-contain a number of ``Agent API Key`` s. It can be found under
-**Settings**. It was recently added to the Cloud Manager to
-allow easier periodic rotation of the ``Agent API Key`` with a constant
-``Project ID``)
+.. Note::
+    All the operations required to set up a Self-Signed CA can be automatically generated from
+    our :ref:`"How to configure a BigchainDB node" <how-to-configure-a-bigchaindb-node>` guide.
 
+5. Set Up a Self-Signed Certificate Authority
```
ttmc: This step doesn't say _how_ they should "Set Up a Self-Signed Certificate Authority" and it's my understanding that the generate_configs.sh bash script will do that. The instructions to run that are in the next page (linked), so I guess we could remove this step here?

k8s/mongodb/mongo-sc.yaml
```diff
@@ -10,7 +10,7 @@ parameters:
   skuName: Premium_LRS #[Premium_LRS, Standard_LRS]
   location: westeurope
   # If you have created a different storage account e.g. for Premium Storage
-  #storageAccount: <Storage account name>
+  storageAccount: <Storage account name>
```
ttmc: Should this change also be made for the Tendermint StorageClass?

ttmc:
@muawiakh and @shahbazn - Since you're away today, I'm going to merge this pull request and address the issues I raised (and any others that I find) in some follow-up pull requests.

PR #2111, opened 2018-03-03
Edited the "Overview" page in Production Deployment Template for clarity

ttmc:
This pull request mostly edits the **Overview** page in the Production Deployment Template for clarity, since it's the introductory page for anyone using that template.

I moved the "Deploy a Kubernetes cluster" step to later, because they don't have to do that until just before they deploy their node to the cluster. First they can get their SSL certificate, which might take a few hours.

I put a bunch of space at the bottom of the page so they don't click the auto-inserted "Previous" and "Next" buttons, since that would be the wrong thing to do.

In the spirit of an "overview" page, I also let them know what they can expect to do when they "deploy their BigchainDB node."


ttmc:
@shahbazn Did I get the secret token header name right?

Issue #2109, opened 2018-03-01
Problem: tests failing with docker-compose dev base environment


sbellem:
See https://github.com/bigchaindb/bigchaindb/pull/2092#discussion_r171426224

More details will be added as soon as possible. In the meantime the above linked comment should provide a good idea of what the problem is.

## Steps to reproduce the problem
```bash
$ docker-compose -f docker-compose.tendermint.yml up -d bd
$ docker-compose -f docker-compose.tendermint.yml run --rm bdb pytest -vv
```
### Failures
```bash
========================================================== FAILURES ===========================================================                                                                                                                                                                                              
__________________________________________________ test_bigchain_show_config __________________________________________________                                                                                                                                                                                              
                                                                                                                               
capsys = <_pytest.capture.CaptureFixture object at 0x7f69507a5080>                                                             
                                                                                                                               
    @pytest.mark.tendermint                                                                                                    
    @pytest.mark.usefixtures('ignore_local_config_file')                                                                                                                                      
    def test_bigchain_show_config(capsys):                                                                                                                                                    
        from bigchaindb import config                                                                                                                                                                              
        from bigchaindb.commands.bigchaindb import run_show_config                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                             
        args = Namespace(config=None)                                                                                                                                                                              
        _, _ = capsys.readouterr()                                                                                                                                                                                 
        run_show_config(args)                                                                                                  
        output_config = json.loads(capsys.readouterr()[0])                                                                                                                                                         
        del config['CONFIGURED']                                                                                                                                                                                   
        config['keypair']['private'] = 'x' * 45                                                                                                                                                                    
>       assert output_config == config                                                                                                                                                                             
E       AssertionError: assert {'backlog_rea...ing': [], ...} == {'backlog_reas...ing': [], ...}                                                              
E         Omitting 5 identical items, use -vv to show                                                                                   
E         Differing items:                                                                                                                                                                                         
E         {'server': {'bind': '0.0.0.0:9984', 'loglevel': 'info', 'workers': None}} != {'server': {'bind': 'localhost:9984', 'loglevel': 'info', 'workers': None}}                                                 
[...]

PR #2114, opened 2018-03-05
Problem: docker-compose-dev broken Revert "Migrate cli for tendermint integration"

muawiakh:
Reverts bigchaindb/bigchaindb#2092


ttmc:
Did you use the "Revert a previous pull request" feature?

ttmc:
If so, what do you plan to do next?

kansi:
Since the context of this PR is to revert to an older state because Travis tests are run using different compose file hence the breakage wasn't caught by the CI. I have one question:

**How do we catch such break-age in future?**

sbellem:
> Since the context of this PR is to revert to an older state because Travis tests are run using different compose file hence the breakage wasn't caught by the CI. I have one question:
> 
> How do we catch such break-age in future?

See https://github.com/bigchaindb/bigchaindb/issues/2109

sbellem:
> @sbellem #2109 details the tests which failed because of #2092 . I am more keen on understanding how can future PRs avoid crashing the tests in docker-compose-dev environment? I mean if contributers don't use docker-compose-dev they should be made aware that their PR shouldn't have any tests crashing in dev compose environment?

First: we should not modify the travis scripts to accommodate failing tests.

That is the root of the problem in this case.

PRs which modify the CI scripts, etc should go through a more thorough review process.

There are no silver bullets here.

Same problem could happen if someone modifies a test just so their patch passes.

Implementation changes need to be thoroughly distinguished from changes to tests.

muawiakh:
@kansi one way could be that Travis and Dev both use the same docker-compose file to run tests i.e. the `travis_script.sh` should run a common `docker-compose -f docker-compose.common.yml run --rm --no-deps bdb pytest -v --cov=bigchaindb`, instead of `docker-compose -f docker-compose.travis.yml run --rm --no-deps bdb pytest -v --cov=bigchaindb`.

I remember, I talked to @sbellem and he mentioned that they cannot be the same for some reasons. @sbellem can you explain here, so we are all on the same page.

sbellem:
Again, no silver bullet.

Thorough review is necessary.

Raymond Hettinger [sums it up very well](https://twitter.com/raymondh/status/622488057254539264):

> Python tip: When refactoring your code, tests are your safety net. When refactoring your tests, there is no safety net.

ttmc:
From my point of view, Docker Compose was supposed to make life easier, and then _this_ happens. Problems so complex that only Sylvain can understand them, and now he's gone, doubly. I mean, it's all probably blindingly obvious to him. We as a team need to decide what we're actually going to use for dev going forward. Preferably some dev process we all _understand_.

ttmc:
Meanwhile, what's the fix here? I suspect it's just:

- Don't touch them Docker Compose files. Danger! Put them all back the way they were.
- Redo all the other changes to migrated the CLI subcommands away.

muawiakh:
We are digressing here. I think there are a couple of problems and I will try to list them and be as objective as possible:

- Making changes to a CI infrastructure/Travis, is not a good practice if tests are failing. That was my bad and I concur to that, hence the reversion PR. I will make relevant changes to the codebase and come up with a better solution if possible.
- Surprisingly, there are no tests on travis that actually test BigchainDB service running i.e. the env vars that I removed should not allow for BigchainDB to run/be active but Travis passed for that. This is an important item and we should introduce tests that actually test an active/running BigchainDB, so we don't run into a scenario like this again.
- Another problem as @kansi pointed out, how do we test this type of scenario.
  - On the top of my head, I think they should be the same i.e. Travis and Dev should have the same docker-compose file which should test a running BigchainDB but @sbellem explained to me how he has spent considerable amount of time and it is very tricky. Which I understand now, where he is coming from. In short, running the same docker-compose file will fail on Travis.
  - But I am sure we can come up with a way to do it.

In terms of practice, In my experience, Infra/CI have different repositories and they require extensive review if there are changes made to them impacted by some core changes, since we don't have that, I personally want to be more cautious before making CI changes so that tests can pass.

One more note is that, since this PR impacts different stakeholders, besides review process being extensive(which goes without saying), peers should actively review. This PR had been around for ~10 days, most of the reviews came quite quickly in ~7 days back and I waited ~2 days to merge it so everyone impacted can review. We should not pick and choose good practices from a process.

I hope this is a sound summary of the concerns pointed out above. Let me know, if not. We can take this offline.

PR #2115, opened 2018-03-05
Problem: K8s deployment with Access Token Authorization expects Access Token for GET calls

shahbazn:
## Description
K8s deployment with `secret-access-token` based authorization instead of three-scale should expect Secret-Access-Token header to be set only for POST calls. But currently it expects the same header for GET calls as well.

### DIFF
--- a/k8s/nginx-https/container/nginx.conf.template
+++ b/k8s/nginx-https/container/nginx.conf.template
@@ -120,7 +120,7 @@ http {
         set $auth_check "${auth_check}1";
       }
 
-      # Block POST requests with invalid Secrete Access Token
+      # Block POST requests with invalid Secret Access Token
       if ( $auth_check = "01" ) {
         return 403;
       }
--- a/k8s/nginx-https/container/nginx.conf.template
+++ b/k8s/nginx-https/container/nginx.conf.template
@@ -86,6 +86,7 @@ http {
     # Forward other URL paths as per business logic/use case to BDB or
     # OpenResty instance.
     location / {
+      set $auth_check 1; # Flag to authorize POST requests
       proxy_ignore_client_abort on;
       proxy_set_header X-Real-IP  $remote_addr;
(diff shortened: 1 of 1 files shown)


k8s/nginx-https/container/nginx.conf.template
```diff
@@ -111,11 +112,21 @@ http {
...
       if ( $http_x_secret_access_token != "SECRET_ACCESS_TOKEN" ) {
+        set $auth_check 0;
+      }
+
+      # CODES: POST=1 and GET=0
+      if ( $request_method = POST ) {
+        set $auth_check "${auth_check}1";
+      }
+
+      # Block POST requests with invalid Secrete Access Token
```
muawiakh: Secret*

shahbazn:
@muawiakh fixed

ttmc:
Sigh, I don't understand NGINX, I mean could they be any more arcane? But if this works then good.

ttmc:
Maybe you could add some comments for future readers to understand what is happening, maybe with links to relevant documentation online.

k8s/nginx-https/container/nginx.conf.template
```diff
@@ -111,11 +112,21 @@ http {
 
       # Check for security header to authorize POST requests
       if ( $http_x_secret_access_token != "SECRET_ACCESS_TOKEN" ) {
+        set $auth_check 0;
+      }
+
+      # CODES: POST=1 and GET=0
```
muawiakh: You are not getting for GET-> 0 anywhere, might as well remove it.

shahbazn:
@muawiakh for three scale there is a separate conf file `nginx.conf.threescale.template` this file will only we used if you have set `AUTHORIZATION_MODE ` to `secrete-token`

shahbazn:
closing this in favor of https://github.com/bigchaindb/bigchaindb/pull/2134#pullrequestreview-103934212

PR #2123, opened 2018-03-09
Problem: Docs for BEP#3 implementation not updated

kansi:
Solution: Update docs for #2070 and #2106

### DIFF
--- a/docs/server/source/server-reference/bigchaindb-cli.md
+++ b/docs/server/source/server-reference/bigchaindb-cli.md
@@ -73,3 +73,24 @@ section of Python's documentation.
 For a more fine-grained control over the logging configuration you can use the
 configuration file as documented under
 [Configuration Settings](configuration.html).
+
+
+## bigchaindb upsert-validator
+
+Add/Update/Remove a validator from the validaors set of the local node. Below is the command line syntax,
+
+```bash
+$ bigchaindb upsert-validator PUBLIC_KEY_OF_VALIDATOR POWER
+```
+
+Example usage,
+
+```bash
+$ bigchaindb upsert-validator B0E42D2589A455EAD339A035D6CE1C8C3E25863F268120AA0162AD7D003A4014 10
+```
+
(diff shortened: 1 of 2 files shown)


docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -73,3 +73,24 @@ section of Python's documentation.
...
+$ bigchaindb upsert-validator PUBLIC_KEY_OF_VALIDATOR POWER
+```
+
+Example usage,
+
+```bash
+$ bigchaindb upsert-validator B0E42D2589A455EAD339A035D6CE1C8C3E25863F268120AA0162AD7D003A4014 10
+```
+
+If the command is returns without any error then a request to update the validator set has been successfully submitted. So, even if the command has been successfully executed it doesn't imply that the validator set has been updated. In order to check whether the change has been applied, the node operator can execute `curl http://node_ip:9985/api/v1/validators` which will list the current validators set. Refer [validators](/http-client-server-api.html#validators) section for more detail.
```
ttmc: Minor grammar and clarity fix:

Refer **to the** validators section **of the HTTP API docs** for more detail.

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -73,3 +73,24 @@ section of Python's documentation.
...
+
+Note:
+- When `POWER`is set to `0` then the validator will be removed from the validator set.
+- The node operator is allowed to submit only one validator upsert request at a time. Only after a pending request is processed can the node operator submit another request.
```
ttmc: Maybe you could say that upsert requests are handled once per block, to give them a sense of how long they have to wait. If Tendermint is in "no empty blocks" mode, and there are no new transactions coming in, it could be a while until the next block.

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -73,3 +73,25 @@ section of Python's documentation.
 For a more fine-grained control over the logging configuration you can use the
 configuration file as documented under
 [Configuration Settings](configuration.html).
+
+
+## bigchaindb upsert-validator
```
vrde: I would add a reference to the BEP you wrote.

PR #2126, opened 2018-03-12
Update setup.py for BigchainDB 2.0

ttmc:
I took the license name from this list of valid PyPI licenses:

https://jlesquembre.github.io/autopilot/license_list.html

About line 112: The docs will say it's production-ready depending on your use case so I figured we could update that. It's not like this is a contract or anything.

I'm not sure if we should keep line 120:

```python
'Programming Language :: Python :: 3.5',
```
or line 122:
```python
 'Operating System :: MacOS :: MacOS X',
```
It didn't actually work on mac os, not sure why we had that. Does it work on Macs now?


setup.py
```diff
@@ -105,17 +105,17 @@ def check_setuptools_features():
     url='https://github.com/BigchainDB/bigchaindb/',
     author='BigchainDB Contributors',
     author_email='dev@bigchaindb.com',
-    license='AGPLv3',
+    license='Apache Software License 2.0',
     zip_safe=False,
 
     classifiers=[
-        'Development Status :: 4 - Beta',
+        'Development Status :: 5 - Production/Stable',
```
vrde: How would this work when we release an `rc`? To me, we should keep `4 - Beta` for now.

setup.py
```diff
@@ -105,17 +105,17 @@ def check_setuptools_features():
...
         'Intended Audience :: Developers',
         'Topic :: Database',
         'Topic :: Database :: Database Engines/Servers',
         'Topic :: Software Development',
         'Natural Language :: English',
-        'License :: OSI Approved :: GNU Affero General Public License v3',
+        'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python :: 3 :: Only',
         'Programming Language :: Python :: 3.5',
```
vrde: I think it's fine to remove `3.5`. We are actually using features from `3.6`.

setup.py
ttmc: Let's discuss Thursday.

setup.py
ttmc: Okay I removed that line.

ttmc:
I left the Mac OS X line for now.

ttmc:
We will start the BigchainDB 2.0 releases at "BigchainDB 2.0 alpha" so I changed the production status line back to:

```text
        'Development Status :: 3 - Alpha',
```

setup.py
ttmc: See below: I changed the dev status back to "3 - Alpha" for now.

ttmc:
@vrde I think I addressed both your comments on this pull request.

PR #2132, opened 2018-03-14
Problem: UTXO implementation unfinished for release.

kansi:
The current UTXO implementation requires some more work in order for it to be production ready. So, it is better to rollback the feature for time being for the alpha release.

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -67,7 +67,7 @@ class BigchainDB(Bigchain):
     def store_transaction(self, transaction):
         """Store a valid transaction to the transactions collection."""
 
-        self.update_utxoset(transaction)
+        # self.update_utxoset(transaction)
         transaction = deepcopy(transaction.to_dict())
         if transaction['operation'] == 'CREATE':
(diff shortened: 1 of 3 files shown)


ttmc:
I wondered about this.

Let's get it right, e.g. write a BEP, understand how we store the UTXO data, and the app hash, make sure Sylvain is on board, maybe use some of the tools from the Tendermint team for storing/manipulating IAVL+ trees, etc.

Can you write a bit about what exists now? Maybe in a short "Informational" BEP? Don't spend too long on that, but it would be useful for context for everyone. We can then reference it when we write a long "Standard Track" BEP to describe it's replacement.

:+1: from me.

ttmc:
I will bump the bullet point about UTXO and app hash to the next release in the ROADMAP.md document.

kansi:
@ttmc That sounds good. I will write an information BEP to the best of my understanding.

kansi:
@ttmc Please refer bigchaindb/BEPs#20

ttmc:
I made some comments on https://github.com/bigchaindb/BEPs/pull/20

PR #2135, opened 2018-03-15
Problem: No queries to store pre-commit state

kansi:
Solution: Add backend logic to store and retireve pre-commit state

Refer [BEP#8](https://github.com/bigchaindb/BEPs/tree/master/8) for more details.


bigchaindb/backend/localmongodb/query.py
```diff
@@ -266,6 +266,22 @@ def get_unspent_outputs(conn, *, query=None):
                                                   projection={'_id': False}))
 
 
+@register_query(LocalMongoDBConnection)
+def put_pre_commit_state(conn, state):
```
vrde: All functions to save data have "store" as prefix, would it make sense to use it here as well so we are consistent?

PR #2130, opened 2018-03-13
Problem: standardize docker-compose workflows

muawiakh:
## Description
Clean up the deprecated workflow(s).
  - Remove `docker-compose.rdb.yml`.
  - Remove `docker-compose.benchmark.yml`.
  - Remove `docker-compose.docs.yml`, since building documentation should be part of a development environment.
  - Remove `docker-compose.ssl.yml`, because we are planning to remove SSL support for MongoDB.
  - Remove `docker-compose.network.yml` file, this workflow will be supported by another deployment method. Currently, we are not using this method for development. We will have Vagrant, Ansible and K8s network driver to offer this functionality.

- Update the workflows with the new deployment workflow(s).
  - Update `docker-compose.yml` to support the current deployment strategy i.e. BigchainDB + Tendermint + Stand alone MongoDB.
- Integrate the development and CI compose file(s) in `docker-compose.yml`
- Following tools `bigchaindb-driver`, `curl-client` can be brought up using:
    - `docker-compose up curl-client`
    - `docker-compose up bdb-driver`
- You can build docs using:
  - `docker-compose up bdocs`

- Update the MongoDB and Tendermint version(s) to the latest supported version:

  - mongo:3.4.13
  - tendermint/tendermint:0.13

- Deprecate the redundant Dockerfiles and only keep two:
  - Dockerfile
  - Dockerfile-dev

- Remove `compose/bigchaindb-server`, `compose-bigchaindb-driver`, and `compose/travis`.

- Introduce `bigchaindb/tools` directory to handle `Dockerfile`(s) for tools required for development i.e.
  - `bigchaindb-driver`
    - Python driver to interact and transact with the BigchainDB server.

- Change docker-compose version to 2.1
  - why, one might ask? because compose version `3.0` does not support `depends_on` and `inherits`
    like we want to and is more aimed towards migration to using `docker stack`, for our current strategy `2.1` is a better choice.
- change `bdb` service `bigchaindb` service
  - why? Introduced a new proxy service `bdb` which is just a dummy `busybox` image.
    - why? because this ensure via healthcheck of bigchaindb that BigchainDB
      has started properly and makes a `curl` to ensure HTTP API server is up
      and running.
      - why? Because we have had scenarios where BigchainDB is not started
        via docker compose and user has to check out the logs to find out what
        the problem might be. This ensures that bigchaindb is up and running, atleast the HTTP API endpoint.
- Does this change deployment workflow? No.
  - The only thing change is that if you want to run commands inside a bigchaindb
[...]

PR #2138, opened 2018-03-17
Problem: ABCI server not run for tests

kansi:
Solution: Create seperate test mark for which the ABCI server will be running.

### Description
Under the current testing environment the ABCI server is not purposely run for testing parts of BigchainDB which depend of Tendermint eg. Event stream, transaction modes etc. This PR addresses those issues i.e. a new test matrix has been introduced for which ABCI server is run for the whole session and tests can be written accodingly. Some of the circumstances which can be tested are as follows,

- Test event stream [DONE]
- Test transaction modes
   - Test the three modes (`async`, `sync`, `commit`)
   - Test posting invalid transaction using different modes 
   - ... etc
- Test adding a new validator.
- Anything that depends on Tendermint

### Why are these tests run in a different Travis martix?
The tests which depend on Tendermint cannot flush the MongoDB collections as that would cause Tendermint to panic. Hence they should be executed in a different matrix.

### Things to Ignore
The `docker-compose.travis.yml` should be ignored for this PR because @muawiakh is addressing that in #2130 . Once that is merged the changes will be merged back here.


docker-compose.travis.yml
```diff
@@ -24,11 +26,12 @@ services:
       BIGCHAINDB_START_TENDERMINT: 0
       BIGCHAINDB_TENDERMINT_HOST: tendermint
       BIGCHAINDB_TENDERMINT_PORT: 46657
-    command: bigchaindb start
+    command: bash -c "sleep 3600"
   tendermint:
-    image: tendermint/tendermint:0.13
+    image: tendermint/tendermint:0.12
```
kansi: NOTICE: This is important, `0.13` doesn't actually work.

.ci/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e -x
+
+if [[ ${BIGCHAINDB_CI_ABCI} == 'enable' ]]; then
```
muawiakh: shouldn't you get an `unbound` variable error when this script is executed for other matrices e.g. `3.5`? Same for other instances? Why not make it global?

.ci/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
...
+    sleep 3600
+else
+    bigchaindb start
```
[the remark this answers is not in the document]
kansi: This script for conditionally executing `bigchindb start` depending on which Travis test matrix is being executed.

.ci/travis-install.sh
```diff
@@ -6,6 +6,8 @@ pip install --upgrade pip
 
 if [[ -n ${TOXENV} ]]; then
     pip install --upgrade tox
+elif [[ ${BIGCHAINDB_CI_ABCI} == 'enable' ]]; then
+    docker-compose build --no-cache --build-arg abci_status=enable bigchaindb
 elif [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then
     docker-compose build --build-arg python_version=3.5 --no-cache bigchaindb
```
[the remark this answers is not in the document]
kansi: `bigchaindb`, please refer the `Dockerfile-dev` file.

.ci/entrypoint.sh
muawiakh: ignore my comment, if it is not exposed it will be treated with `''`

.ci/entrypoint.sh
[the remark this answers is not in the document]
muawiakh: makes sense. I thought this was for travis because it was under `.ci`

.ci/entrypoint.sh
kansi: Undeclared vairables seem to be replaced with empty string, [here](https://travis-ci.org/bigchaindb/bigchaindb/jobs/364526325#L1045)

.ci/entrypoint.sh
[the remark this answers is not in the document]
kansi: I didn't know the right place to put it. If you think it could be moved somewhere better let me know.

tests/conftest.py
```diff
@@ -555,6 +555,46 @@ def mocked_setup_sub_logger(mocker):
...
+
+        except Exception:
+            pass
+        time.sleep(1)
+
+    return False
+
+
+@pytest.yield_fixture(scope='session')
+def event_loop(request):
```
vrde: `pytest-asyncio` already includes an [`event_loop` fixture](https://github.com/pytest-dev/pytest-asyncio#fixtures).

tests/conftest.py
```diff
@@ -555,6 +555,46 @@ def mocked_setup_sub_logger(mocker):
...
+    import requests
+    import time
+
+    for i in range(5):
+        try:
+            uri = 'http://{}:{}/abci_info'.format(tendermint_host, tendermint_port)
+            requests.get(uri)
+            return True
+
+        except Exception:
```
vrde: Please catch only what you expect.

tests/conftest.py
```diff
@@ -555,6 +555,46 @@ def mocked_setup_sub_logger(mocker):
         'bigchaindb.log.setup.setup_sub_logger', autospec=True, spec_set=True)
 
 
+@pytest.fixture
+def abci_http(tendermint_host, tendermint_port):
```
vrde: I see this fixture is used in combination with `abci_server`, the question is: does it make sense to use `abci_http` alone? If not, I'd unify it with `abci_server`.

tests/conftest.py
kansi: Yes, but to set the scope of `event_loop` to `'session'` it has to be re-defined.

kansi:
@vrde I converted the fixutres `abci_server` and `abci_http` to marker `@pytest.mark.abci`. Let me know see any issues.

PR #2147, opened 2018-03-22
Update Makefile

vrde:
Solution: we already have a nice Makefile, but it's outdated. The idea is to revamp it and make it easy to use.

The description can be found in the updated `README.md` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ To know the IP of the instance running, execute in a new terminal `make ip`.
 There are also other commands you can execute:
 - `make start`: Run BigchainDB from source and daemonize it (stop it with `make stop`).
 - `make stop`: Stop BigchainDB.
-- `make ip`: Get the IP address of BigchainDB.
 - `make logs`: Attach to the logs.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -10,6 +10,30 @@
...
+```
+git clone https://github.com/bigchaindb/bigchaindb.git
+cd bigchaindb
+make run
+```
+
+To know the IP of the instance running, execute in a new terminal `make ip`.
+
+There are also other commands you can execute:
+- `make start`: Run BigchainDB from source and damonize it (stop it with `make stop`).
```
codegeschrei: daemonize ?

vrde: Weirdly enough, [yes](https://en.wikipedia.org/wiki/Daemon_(computing)).

codegeschrei: Ah, can you add an 'e' then? :D

vrde: damonizee?

codegeschrei: we can replace it with an evil emoji!

PR #2148, opened 2018-03-22
Problem: vim swap files are not ignored by git

vrde:
Solution: add patterns to ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,21 @@ __pycache__/
 # C extensions
 *.so
 
+# Swap -- copypasta from https://github.com/github/gitignore/blob/master/Global/Vim.gitignore
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,6 +6,21 @@ __pycache__/
 # C extensions
 *.so
 
+# Swap
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-v][a-z]
```
muawiakh: Shouldn't we just limit to `swp`, `swo` and `*~`? With these changes some other formats might be ignored too even though we don't use them! e.g. [some of these](https://en.wikipedia.org/wiki/List_of_filename_extensions_(S%E2%80%93Z)), I am approving it but just food for thought.  :angel:

vrde: I took it from here: https://github.com/github/gitignore/blob/master/Global/Vim.gitignore

PR #2152, opened 2018-03-23
Problem: Event stream test incomplete

kansi:
Solution: Complete the test by checking connection status returned in the `result` field


tests/tendermint/test_event_stream.py
```diff
@@ -78,4 +78,5 @@ def test_process_unknown_event():
     msg_data_dict = json.loads(msg.data)
     assert msg_data_dict['id'] == stream_id
     assert msg_data_dict['jsonrpc'] == '2.0'
+    assert msg_data_dict['result'] == {}
```
kansi: For a successfull connection the `result` field will be an empty dict.

PR #2159, opened 2018-03-28
Problem: CHANGELOG.md isn't updated for the 2.0 Alpha release

### LINKED ISSUE #2128: Problem: CHANGELOG.md doesn't cover changes in BigchainDB 2.0
Solution: List all changes that went into BigchainDB 2.0 in `CHANGELOG.md`. Or reference lists elsewhere (like in blog posts).

This update to `CHANGELOG.md` will be different from usual, because too much has happened to just put all the changes in there like usual. Here's what I plan to do:

- Link to all the blog posts about the various changes to the HTTP API, the Transaction Spec, BFT, the CLI, configuration settings, maybe more. Those will all be published on Medium by the time the release happens.

- Say that the _biggest_ change is the integration of Tendermint for doing consensus, replication and sync, as described in the linked blog post.

- Mention any external contributors, of course!

- Remove the whole front matter defining the **Public API** and what we do about Semantic Versioning and link to BEP-7 instead, which is where the Public API will now be defined, officially, i.e. in an actual spec! That's also where we specifically say how _we_ do semantic versioning, i.e. we do the Python form of Semantic Versioning.

https://github.com/bigchaindb/BEPs/pull/15#issue-174402354

@gautamdhameja You mentioned this yesterday so I thought I should write up my intent. I will do this before the release, of course!

ttmc:
Solution: Update CHANGELOG.md for 2.0

Resolves #2128


ttmc:
Note: The git tag for BigchainDB 2.0 Alpha will be 2.0.0a1 because we use the Python version of semantic versioning. See: https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme

2.0.0a1 is also what we should use use in `version.py`,

PR #2161, opened 2018-03-28
Problem: RELEASE_PROCESS.md doesn't follow C4

### LINKED ISSUE #2136: Problem: RELEASE_PROCESS.md not following C4 practices
Solution: Update `RELASE_PROCESS.md` to follow C4 practices, namely:

- Assume there is only one branch that matters: `master`
- Assume that each release (with a version number, whether major or minor) is just a specific commit in the `master` branch, identified by an _annotated_ Git Tag (which can be generated using GitHub's user interface for generating releases)
- Stop making a new branch for some releases!

ttmc:
Solution: Update RELEASE_PROCESS.md to use only the `master` branch to use Git tags (never branches) to identify new versions.

Also made several other changes:
- Added a note about how we follow C4 and linked to it in bigchaindb/BEPs.
- Added an example of how to label an Alpha release (because we are about to do one).
- Added the possibility of changing the "Development Status" item in `setup.py`.
- The release process is now the same for major and minor version releases!
- Several other minor edits.

Resolves #2136

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -1,73 +1,61 @@
 # Our Release Process
 
-The release process for BigchainDB server differs slightly depending on whether it's a minor or a patch release.
+## Notes
 
-BigchainDB follows 
+BigchainDB follows
 [the Python form of Semantic Versioning](https://packaging.python.org/tutorials/distributing-packages/#choosing-a-versioning-scheme)
 (i.e. MAJOR.MINOR.PATCH),
 which is almost identical
-to [regular semantic versioning](http://semver.org/)
-except release candidates are labelled like
-`3.4.5rc2` not `3.4.5-rc2` (with no hyphen).
+to [regular semantic versioning](http://semver.org/), but there's no hyphen, e.g.
 
+- `0.9.0` for a typical final release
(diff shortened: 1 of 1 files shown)


RELEASE_PROCESS.md
```diff
@@ -1,73 +1,61 @@
...
 
+- `0.9.0` for a typical final release
+- `4.5.2a1` not `4.5.2-a1` for the first Alpha release
+- `3.4.5rc2` not `3.4.5-rc2` for Release Candidate 2
 
-## Minor release
+We use `0.9` and `0.9.0` as example version and short-version values below. You should replace those with the correct values for your new version.
 
-A minor release is preceeded by a feature freeze and created from the 'master' branch. This is a summary of the steps we go through to release a new minor version of BigchainDB Server.
+We follow [C4, the Collective Code Construction Contract](https://github.com/bigchaindb/BEPs/tree/master/1), so a release is just a [tagged commit](https://git-scm.com/book/en/v2/Git-Basics-Tagging) on the `master` branch, i.e. a label for a particular Git commit.
```
vrde: Now that I just noticed, should we say *We follow BEP-1..." because it's our own version of C4?

ttmc: ok

PR #2164, opened 2018-03-28
Problem: BigchainDB server configuration causing failure

muawiakh:
## Solution
Update BigchainDB server configuration i.e. threads/workers to `1` and update log level of `tendermint` for better debugging.
- Update bigchaindb server threads/workers to `1`
- Configure `debug` log-level for tendermint

### DIFF
--- a/k8s/bigchaindb/bigchaindb-dep.yaml
+++ b/k8s/bigchaindb/bigchaindb-dep.yaml
@@ -84,6 +84,10 @@ spec:
             configMapKeyRef:
               name: bdb-config
               key: bigchaindb-log-level
+        - name: BIGCHAINDB_SERVER_WORKERS
+          value: "1"
+        - name: BIGCHAINDB_SERVER_THREADS
+          value: "1"
         - name: BIGCHAINDB_DATABASE_SSL
           value: "true"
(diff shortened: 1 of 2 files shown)


k8s/bigchaindb/bigchaindb-dep.yaml
```diff
@@ -84,6 +84,10 @@ spec:
             configMapKeyRef:
               name: bdb-config
               key: bigchaindb-log-level
+        - name: BIGCHAINDB_SERVER_WORKERS
+          value: "1"
+        - name: BIGCHAINDB_SERVER_THREADS
```
ttmc: `BIGCHAINDB_SERVER_WORKERS` is [documented in the BigchainDB Server docs](https://docs.bigchaindb.com/projects/server/en/master/server-reference/configuration.html#server-bind-server-loglevel-server-workers), but `BIGCHAINDB_SERVER_THREADS` isn't. Should we should add docs for the latter?

Maybe not. I get the feeling that this business of setting the number of threads to 1 is a temporary workaround and one wouldn't normally ever do it.

muawiakh: Yes, I agree. We won't need these parameters in the future as well, hopefully.

PR #2165, opened 2018-03-29
Problem: Multi-threading not configured

kansi:
Solution: Auto-generate multi-threading config with default threads and workers set to 1


ttmc:
Don't worry about documenting `server.threads`.

PR #2167, opened 2018-03-30
Problem: Block height in event stream is string.

### LINKED ISSUE #2166: Problem: block_id stored in a string but should be stored in an int
See https://github.com/bigchaindb/bigchaindb/pull/2163#discussion_r178039081

### LINKED ISSUE #2168: Problem: Inconsistent name for one of the top-level keys inside a block
The code and docs are inconsistent about whether a block (a BigchainDB block, not a Tendermint block) has the top-level key "height" or "block_id". For example, the [docs about **The Block Model**](https://docs.bigchaindb.com/projects/server/en/master/data-models/block-model.html) say that it's "height" but the code (or at least some of it) builds a block with top-level key "block_id" instead of "height", e.g. see line 53 in `bigchaindb/tendermint/event_stream.py`.

Proposal: Use "height" or "block height" everywhere.

kansi:
Solution: Change block height to integer.

Fixes #2166 
Fixes #2168

### DIFF
--- a/bigchaindb/tendermint/event_stream.py
+++ b/bigchaindb/tendermint/event_stream.py
@@ -50,7 +50,7 @@ def process_event(event_queue, event, stream_id):
         # Only push non empty blocks
         if block_txs:
             block_txs = [decode_transaction_base64(txn) for txn in block_txs]
-            new_block = {'id': str(block_id), 'transactions': block_txs}
+            new_block = {'id': block_id, 'transactions': block_txs}
             event = Event(EventTypes.BLOCK_VALID, new_block)
             event_queue.put(event)
 
--- a/bigchaindb/tendermint/event_stream.py
+++ b/bigchaindb/tendermint/event_stream.py
@@ -50,7 +50,7 @@ def process_event(event_queue, event, stream_id):
         # Only push non empty blocks
         if block_txs:
             block_txs = [decode_transaction_base64(txn) for txn in block_txs]
-            new_block = {'id': block_id, 'transactions': block_txs}
+            new_block = {'height': block_id, 'transactions': block_txs}
             event = Event(EventTypes.BLOCK_VALID, new_block)
             event_queue.put(event)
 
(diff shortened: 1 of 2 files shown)


tests/tendermint/test_event_stream.py
```diff
@@ -35,6 +35,8 @@ def test_process_event_new_block():
     event_queue = Queue()
     process_event(event_queue, event, 'test_stream_id')
     assert not event_queue.empty()
+    block = event_queue.get()
+    assert type(block.data['id']) == int
```
ttmc: From [the Python3 documentation about `type()`](https://docs.python.org/3/library/functions.html#type):

> The `isinstance()` built-in function is recommended for testing the type of an object, because it takes subclasses into account.

I'm not sure if that matters here.  I looked it up and, `int` _does_ have subclasses: apparently "[Booleans are a subtype of integers.](https://docs.python.org/3.6/library/stdtypes.html#numeric-types-int-float-complex)"

I'll let you decide what to use.

bigchaindb/tendermint/event_stream.py
```diff
@@ -50,7 +50,7 @@ def process_event(event_queue, event, stream_id):
         # Only push non empty blocks
         if block_txs:
             block_txs = [decode_transaction_base64(txn) for txn in block_txs]
-            new_block = {'id': str(block_id), 'transactions': block_txs}
+            new_block = {'id': block_id, 'transactions': block_txs}
```
ttmc: I just noticed an inconsistency between this and what the docs say is inside a block. The [docs about **The Block Model**](https://docs.bigchaindb.com/projects/server/en/master/data-models/block-model.html) say that the two top-level keys in a block are `"height"` and `"transactions"` but here it's building a block with `"id"` and `"transactions"` as the top-level keys.

This inconsistency might be elsewhere as well.

ttmc: Maybe we could standardize on using `block_height` and "block height" everywhere.

ttmc: I created issue #2168 about the inconsistent key name.

kansi:
Ok, I forgot to refresh the page so it only displayed to me the comments uptil 'request changes' so I made a those tiny fixes. Nevertheless the changes are very small.

ttmc:
Okay. I'll change the description of this PR to say that it also fixes that other issue.

PR #2173, opened 2018-04-03
Problem: Event stream api crashing

kansi:
Solution: Fix 'Exchange' initialization

### DIFF
--- a/bigchaindb/tendermint/commands.py
+++ b/bigchaindb/tendermint/commands.py
@@ -35,8 +35,6 @@ BANNER = """
 def start():
     # Exchange object for event stream api
     exchange = Exchange()
-    p_exchange = Process(name='exchange', target=exchange.run)
-    p_exchange.start()
 
     # start the web api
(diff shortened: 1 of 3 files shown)


bigchaindb/tendermint/commands.py
```diff
@@ -61,6 +59,9 @@ def start():
                                  args=(exchange.get_publisher_queue(),))
     p_websocket_client.start()
 
+    p_exchange = Process(name='exchange', target=exchange.run)
```
vrde: Why moving it after the websocket initialization and start?

kansi: The `exachge` object is somehow not updated after `exchange.get_subscriber_queue(EventTypes.BLOCK_VALID),)` is executed.

vrde: OK, I got what the problem is, and thanks a lot for debugging :clap:. I'll create a new PR after merging this one.

PR #2177, opened 2018-04-03
Problem: Exchange can fail silently

vrde:
Solution: Raise a RuntimeError if Exchange is running and a new subscriber queue is requested.

Related: #2173.

### DIFF
--- a/bigchaindb/events.py
+++ b/bigchaindb/events.py
@@ -1,3 +1,4 @@
+from queue import Empty
 from collections import defaultdict
 from multiprocessing import Queue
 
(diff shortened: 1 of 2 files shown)


vrde:
> Seem good. It will fail once when called and not on subsequent calls but that would be sufficent I guess.

Yes it will fail only once, and it should happen in the main thread, where an `exchange` is instantiated. Do you think it's better to fail every time?

kansi:
Ideally yes, but since we are not creating new queues on the fly this should suffice.

PR #2179, opened 2018-04-05
Problem: Python version not check before running

### LINKED ISSUE #2178: Problem: People Confused When BDB Won't Run with Python 3.5-
Here's one example: https://gitter.im/bigchaindb/bigchaindb?at=5ac3a0737c3a01610da33aa8

There have been other questions like that one, and we're about to get many more because the latest release of BigchainDB Server _requires_ Python 3.6+. I consider this a "UX bug."

Proposed Solution: If someone runs BigchainDB Server with Python 3.5 or less, then tell them they need to run it with Python 3.6 and then exit.

kansi:
Solution: Check python version

Fixes  #2178

### DIFF
--- a/bigchaindb/commands/bigchaindb.py
+++ b/bigchaindb/commands/bigchaindb.py
@@ -242,7 +242,4 @@ def create_parser():
 
 
 def main():
-    if sys.version_info[0] == 3 and sys.version_info[1] <= 5:
-        logger.info('Please use python version 3.6 or up')
-    else:
-        utils.start(create_parser(), sys.argv[1:], globals())
+    utils.start(create_parser(), sys.argv[1:], globals())
--- a/bigchaindb/commands/bigchaindb.py
+++ b/bigchaindb/commands/bigchaindb.py
@@ -242,4 +242,7 @@ def create_parser():
 
 
 def main():
-    utils.start(create_parser(), sys.argv[1:], globals())
+    if sys.version_info[0] == 3 and sys.version_info[1] <= 5:
+        logger.info('Please use python version 3.6 or up')
+    else:
+        utils.start(create_parser(), sys.argv[1:], globals())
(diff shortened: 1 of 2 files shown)


bigchaindb/commands/bigchaindb.py
```diff
@@ -242,4 +242,7 @@ def create_parser():
 
 
 def main():
-    utils.start(create_parser(), sys.argv[1:], globals())
+    if sys.version_info[0] == 3 and sys.version_info[1] <= 5:
+        logger.info('Please use python version 3.6 or up')
```
ttmc: 1. I'd write '...version 3.6 or higher'
1. What's the default log level? (I wonder if a default user will see this message.)
1. Could we just exit (shut down BigchainDB Server) if they're trying to use Python 3.5 or lower? If they are, it won't work, and they'll get a big long confusing message (in addition to the new and actually-helpful message).

kansi:
@vrde Will it also work in case users have an older pip version? (< `9.0.0`) ?

kansi:
I won't help will older pip versions. So I guess we have to keep the `if .. else`

setup.py
```diff
@@ -5,8 +5,12 @@
 
 """
 from setuptools import setup, find_packages
+import sys
 
 
+if sys.version_info[0] == 3 and sys.version_info[1] <= 5:
```
vrde: If `sys.version_info[0]` is `2`, then it won't generate the error. You can also use this simple construct: `sys.version_info < (3, 6)`.

Other nice examples:
```
 ~ ➜ python2 -c "import sys; print(sys.version_info<=(3, 6), sys.version_info)"
(True, sys.version_info(major=2, minor=7, micro=14, releaselevel='final', serial=0))

 ~ ➜ python3 -c "import sys; print(sys.version_info<=(3, 6), sys.version_info)"
False sys.version_info(major=3, minor=6, micro=3, releaselevel='final', serial=0)
```

PR #2188, opened 2018-04-09
Problem: Websocket connection fails and retries are exhausted

muawiakh:
In a Kubernetes deployment, websocket connection to tendermint fails and all the retries are burnt out, the reason is that BigchainDB spins up all the servers/modules asynchronously but the Tendermint connection with BigchainDB/abci is synchronous, hence before the abci connections are established the connections initiated from BigchainDB for websocket are exhausted.

Also adding `tm-instance-name` as `moniker` for Tendermint instances.
## Solution

- Make the minimum timeout to be 30 seconds(10(retries)*3(seconds)). 
- `moniker` should represent the `tm-instance-name` i.e. name of the Kubernetes service instead of `hostname`

### DIFF
--- a/k8s/tendermint/tendermint_container/tendermint_entrypoint.bash
+++ b/k8s/tendermint/tendermint_container/tendermint_entrypoint.bash
@@ -32,7 +32,7 @@ if [[ -z "${tm_seeds:?TM_SEEDS not specified. Exiting!}" || \
       -z "${tmhome:?TMHOME not specified. Exiting!}" || \
       -z "${tm_p2p_port:?TM_P2P_PORT not specified. Exiting!}" || \
       -z "${tm_abci_port:?TM_ABCI_PORT not specified. Exiting! }" || \
-      -z "${tm_instance_name:?TM_INSTANCE_NAME not specifiec. Exiting! }" ]]; then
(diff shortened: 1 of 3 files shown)


k8s/tendermint/tendermint_container/tendermint_entrypoint.bash
```diff
@@ -30,7 +31,8 @@ if [[ -z "${tm_seeds:?TM_SEEDS not specified. Exiting!}" || \
       -z "${tm_chain_id:?TM_CHAIN_ID not specified. Exiting!}" || \
       -z "${tmhome:?TMHOME not specified. Exiting!}" || \
       -z "${tm_p2p_port:?TM_P2P_PORT not specified. Exiting!}" || \
-      -z "${tm_abci_port:?TM_ABCI_PORT not specified. Exiting! }" ]]; then
+      -z "${tm_abci_port:?TM_ABCI_PORT not specified. Exiting! }" || \
+      -z "${tm_instance_name:?TM_INSTANCE_NAME not specifiec. Exiting! }" ]]; then
```
ttmc: There's a small typo here, actually: `specifiec.` should be `specified.`

PR #2198, opened 2018-04-12
Problem: Imprecise POST transaction responses

### LINKED ISSUE #2184: Problem: we don't know what happens with a transaction after it is posted
Solution: provide a feedback to see the status of a transaction

When sending a transaction it is not clear what happens to it. We can check the logs but that's a lot of effort for many transactions. Thus we should return some information if the tx is valid, in a block or if something went wrong (regarding the modes).

The feedback will be provided for the modes `commit` and `sync`.

kansi:
Solution: Improve responses

Fixes #2184


kansi:
@vrde I couldn't find any documentation on the same. The current implementation is based on trail and error.

PR #2207, opened 2018-04-13
Problem: No crash recovery mechanism

kansi:
Solution: Write crash recovery logic following [BEP-8](https://github.com/bigchaindb/BEPs/tree/master/8)

## BEPs Implemented

Refer to [BEP-8](https://github.com/bigchaindb/BEPs/tree/master/8)

### DIFF
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -5,6 +5,7 @@ from functools import singledispatch
 from bigchaindb.backend.exceptions import OperationError
 
 VALIDATOR_UPDATE_ID = 'a_unique_id_string'
+PRE_COMMIT_ID = 'a_unique_id_string'
 
 
 @singledispatch
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -553,22 +553,8 @@ def store_block(conn, block):
 
 
 @singledispatch
-def delete_zombie_transactions(conn):
-    """Delete transactions not included in any block"""
-
-    raise NotImplementedError
-
-
-@singledispatch
 def store_unspent_outputs(connection, unspent_outputs):
     """Store unspent outputs in ``utxo_set`` table."""
-
-    raise NotImplementedError
-
-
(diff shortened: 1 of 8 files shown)


bigchaindb/commands/bigchaindb.py
```diff
@@ -162,13 +150,26 @@ def run_drop(args):
         print("Cannot drop '{name}'. The database does not exist.".format(name=dbname), file=sys.stderr)
 
 
+def run_recover(b):
+    pre_commit = query.get_pre_commit_state(b.connection, PRE_COMMIT_ID)
+
+    # Initially the pre-commit collection would be empty
+    if pre_commit:
+        latest_block = query.get_latest_block(b.connection)
```
vrde: We store the `pre_commit` **before** storing the `block`.

In the unfortunate event of BigchainDB crashing before writing the very first block, `latest_block` would be `null`, and accessing `latest_block['height']` would generate a `KeyError`. Is this a possibility?

bigchaindb/backend/query.py
```diff
@@ -5,6 +5,7 @@
 from bigchaindb.backend.exceptions import OperationError
 
 VALIDATOR_UPDATE_ID = 'a_unique_id_string'
+PRE_COMMIT_ID = 'a_unique_id_string'
```
vrde: *No need to address this comment in this PR.*

I have the feeling those unique ids are necessary because we have two different collections for two specific use cases, one is to keep track of the last validator to update, the other one is for storing the pre commit id. I know it's outside the scope of this PR, but an idea is to have (in the future) a collection to keep the current state of the application, something like `state`. In `state` we have two documents, one for the validator to update, the other one for the pre commit id. The first one will have key `validator_update`, the second one will have key `pre_commit`.

bigchaindb/commands/bigchaindb.py
vrde: That was fast, thanks!

PR #2213, opened 2018-04-16
Problem: No debug statement for post transaciton

kansi:
Solution: write debug statement for response received from POST transaction

#### Description
The reponses received when POSTing a transaction to Tendermint are not fully documented in Tendermint's docs. So a debug statement is required for handling listing and handling unkown responses.


bigchaindb/tendermint/lib.py
```diff
@@ -61,6 +61,7 @@ def write_transaction(self, transaction, mode):
         return self._process_post_response(response.json(), mode)
 
     def _process_post_response(self, response, mode):
+        logger.info(response)
```
vrde: This can get quite chatty, did you consider using `debug`?

PR #2216, opened 2018-04-17
Problem: we don't have acceptance tests

vrde:
Solution: have a simple way to start a node and run scripts against it.

For major fun I've also added some documentation, try to run `make integration-test-doc` :gift:


docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
       interval: 3s
       timeout: 5s
       retries: 3
-    entrypoint: '.ci/entrypoint.sh'
+    entrypoint: bigchaindb start
```
kansi: `.ci/entrypoint.sh` is need for other travis matrix tests which use ABCI proxy app as fixture. Keep `entrypoint: '.ci/entrypoint.sh` should work too.

integration/python/src/test_basic.py
```diff
@@ -0,0 +1,123 @@
...
+            prepared_creation_tx,
+            private_keys=alice.private_key)
+
+    # We will use the `id` of this transaction several time, so we store it in
+    # a variable with a short and easy name
+    bike_id = fulfilled_creation_tx['id']
+
+
+    # Now she is ready to send it to the BigchainDB Network.
+    sent_transfer_tx = bdb.transactions.send(fulfilled_creation_tx, mode='commit')
```
codegeschrei: you're not using `sent_transfer_tx` again so we don't really need it.

codegeschrei: I just realised it is also called transfer, but we have a create

integration/python/src/test_double_spend.py
```diff
@@ -0,0 +1,41 @@
+# Double Spend testing
+# This test challenge the system with double spends.
```
codegeschrei: I don't have a line break in my docs, so it looks off since the first line has no punctuation at the end.
It's just a small detail, but if you touch the code you can check it again :)

integration/python/src/test_stream.py
```diff
@@ -0,0 +1,127 @@
...
+                    operation='CREATE',
+                    signers=alice.public_key,
+                    asset={'data': {'uuid': str(uuid4())}}),
+                private_keys=alice.private_key)
+        # We don't want to wait for each transaction to be in a block. By using
+        # `async` mode, we make sure that the driver returns as soon as the
+        # transaction is pushed to the BigchainDB API. Remember: we expect all
+        # transactions to be in the shared queue: this is a two phase test,
+        # first we send a bunch of transactions, then we check if they are
+        # valid (and, in this case, they should.
```
codegeschrei: missing closing parenthesis

vrde:
I noticed sometimes `make integration-test-doc` fails. It should be fixed once https://github.com/pycco-docs/pycco/pull/101 is merged.

tests/tendermint/test_event_stream.py
```diff
@@ -69,6 +69,7 @@ def test_process_unknown_event():
     assert event_queue.empty()
 
 
+@pytest.mark.skip
```
kansi: Reminder : re-enable this.

tests/tendermint/test_lib.py
```diff
@@ -352,6 +352,7 @@ def test_get_utxoset_merkle_root(b, utxoset):
     assert merkle_root == expected_merkle_root
 
 
+@pytest.mark.execute
```
kansi: Reminder: re-enable this.

Makefile
```diff
@@ -72,7 +72,14 @@ test: check-deps ## Run all tests once
 	@$(DC) exec bigchaindb pytest
 
 test-watch: check-deps ## Run all tests and wait. Every time you change code, tests will be run again
-	@$(DC) run --rm bigchaindb pytest -f -v
+	@$(DC) run --rm --no-deps bigchaindb pytest -f
+
+integration-test: check-deps ## Run all integration tests
+	@./run-integration-tests.sh
```
shahbazn: `run-integration-tests.sh` has been renamed below

PR #2233, opened 2018-04-24
Problem: unclear why vote.yaml is still around

ttmc:
Solution: explain why the vote.yaml file should not be deleted.

This resolves one of the checkboxed tasks in issue #1895


vrde:
What about deleting the file and point people to the tag `1.3` on GitHub?

ttmc:
If we delete `vote.yaml`, we'll be slightly happier because we cleaned old cruft from the `master` branch. But it would make it harder for people using older versions of BigchainDB to find `vote.yaml`. Yes, it's in the Git history, but not everyone is good with Git.

My inclination is to just leave `vote.yaml` there for now, and maybe in the future, once the old versions of BigchainDB aren't in use by _anyone_, we can quietly delete that file and nobody will care.

ttmc:
> What about deleting the file and point people to the tag 1.3 on GitHub?

Okay, I did that. If this PR is okay with you now, then please approve and merge it.

ttmc:
I deleted `VOTE_SCHEMA_PATH`, `VOTE_SCHEMA` and `validate_vote_schema()` from `__init__.py` and that broke some other things which relied on `validate_vote_schema()`, e.g. `tests/common/test_schema.py` and `bigchaindb/voting.py`

In other words, removing `vote.yaml` should probably be done as part of removing _all_ the voting-related code: a bigger cleanup task. It can't just be removed by itself.

I think what I'll do is wait for some other PR to delete the old voting-related code (maybe even `vote.yaml`) and then I'll resolve the merge conflicts in this PR, so that it all it does is add some text to `bigchaindb/common/schema/README.md` to explain where to find the old `vote.yaml` file.

ttmc:
I created a new pull request, #2319, to replace this one, so I closed this one.

PR #2235, opened 2018-04-25
Problem: Tests failing with py-abci 0.4.1

kansi:
Solution: Fix tests and Migrate to Tendermint 0.19.0


kansi:
@vrde The test matrix for python 3.5 fails because py-abci `0.4.1` is only supports `3.6` or up

PR #2239, opened 2018-04-26
Problem: Detailed transaction responses disabled

kansi:
Solution: Migrate the response to Tendermint 0.19.0 and enable the same

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -65,23 +65,31 @@ class BigchainDB(Bigchain):
         if response.get('error') is not None:
             return (500, 'Internal error')
 
-        return (202, '')
-        # result = response['result']
-        # if mode == MODE_LIST[2]:
-        #     return self._process_commit_mode_response(result)
-        # else:
-        #     status_code = result['code']
-        #     return self._process_status_code(status_code,
-        #                                      'Error while processing transaction')
+        result = response['result']
+        if mode == MODE_LIST[2]:
+            return self._process_commit_mode_response(result)
+        else:
+            status_code = result['code']
+            return self._process_status_code(status_code,
+                                             'Error while processing transaction')
 
-    # def _process_commit_mode_response(self, result):
-    #     check_tx_status_code = result['check_tx']['code']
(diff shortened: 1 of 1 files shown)


bigchaindb/tendermint/lib.py
```diff
@@ -65,23 +65,31 @@ def _process_post_response(self, response, mode):
...
+                return (500, 'Error while validating the transaction')
+        except KeyError:
+            pass
+
+        try:
+            deliver_tx_status_code = result['deliver_tx']['code']
+            if deliver_tx_status_code > 0:
+                return (500, 'Error while commiting the transaction')
+        except KeyError:
+            pass
```
vrde: In case both `try ... except` capture the `KeyError` exception, we return a `202`, am I correct? If so, we should return an error.

bigchaindb/tendermint/lib.py
```diff
@@ -65,23 +65,31 @@ def _process_post_response(self, response, mode):
...
+        if mode == MODE_LIST[2]:
+            return self._process_commit_mode_response(result)
+        else:
+            status_code = result['code']
+            return self._process_status_code(status_code,
+                                             'Error while processing transaction')
+
+    def _process_commit_mode_response(self, result):
+        """The implementation below is dumb but clear"""
+        try:
```
ldmberman: A good wat to make it more concise and probably more clear is:

```
check_tx_status_code = result.get('check_tx', {}).get('code', 0)
```

ttmc:
I didn't close this issue directly --- and didn't intend to close it at all! It got closed automatically when I deleted the old branch named `alpha-3`, which is the branch that this PR was merging _into_.

@kansi - You should probably still have the only commit in this PR, i.e.  https://github.com/bigchaindb/bigchaindb/pull/2239/commits/908bbe101d4173e0ba279d85d3902e1c732d16dd --- along with the branch `kansi:iss/txn-responses`. Can you create a new PR based on that? 

GitHub also has that commit.

kansi:
@ttmc This feature needs to be redone and I don't the code in this branch would be of much help. So, I think it is okay to close this one.

PR #2241, opened 2018-04-26
Problem: Arbitrary max_tries in bigchaindb when connecting to tendermint ws

### LINKED ISSUE #2240: Problem: Arbitrary max_tries in bigchaindb when connecting to tendermint ws
# Description
When bigchaindb service start it starts a pyabci server app and tries to connect with tendermint over a websocket and fails after 10 [max_tries](https://github.com/bigchaindb/bigchaindb/blob/119420785d092bc69d21befa3022dac3646d0906/bigchaindb/tendermint/event_stream.py#L75). 

Tendermint on the other hand when it starts tries to connect with bigchaindb pyabci app and waits indefinitely until the app server becomes available. 

So there is a cyclic dependency between two services but the arbitraty limit of `max_tries` from bigchaindb service adds un-necessary constraint for tendermint service to be started first.

shahbazn:
**Solution:** When bigchaindb service start it starts a pyabci server app and tries to connect with tendermint over a websocket and fails after 10 max_tries.

Tendermint on the other hand when it starts tries to connect with bigchaindb pyabci app and waits indefinitely until the app server becomes available.

So there is a cyclic dependency between two services but the arbitraty limit of max_tries from bigchaindb service adds un-necessary constraint for tendermint service to be started first.

Issues Resolved: #2240

### DIFF
--- a/bigchaindb/tendermint/event_stream.py
+++ b/bigchaindb/tendermint/event_stream.py
@@ -67,22 +67,19 @@ def subscribe_events(ws, stream_id):
 
 
 @asyncio.coroutine
-def try_connect_and_recv(event_queue, max_tries):
+def try_connect_and_recv(event_queue):
     try:
         yield from connect_and_recv(event_queue)
 
     except Exception as e:
-        if max_tries:
-            logger.warning('WebSocket connection failed with exception %s', e)
-            time.sleep(3)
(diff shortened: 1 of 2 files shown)


vrde:
Before merging this, let's make sure that the message in the squash commit follows [BEP1 patch requirements](https://github.com/bigchaindb/BEPs/tree/master/1#23-patch-requirements).

PR #2248, opened 2018-04-30
Problem: Insufficent logs

kansi:
Solution: Add debug log statements


bigchaindb/tendermint/core.py
```diff
@@ -140,6 +141,10 @@ def commit(self):
             # this effects crash recovery. Refer BEP#8 for details
             self.bigchaindb.store_block(block._asdict())
 
+        logger.debug('Commit-ing new block with hash: ',
```
vrde: You can use "committing". Please use the `%s` operator for logging.

`logger.debug('one=%s, two=%s', a, b)`

PR #2251, opened 2018-05-02
Problem: BigchainDB liveness probe should be HTTP API

muawiakh:
## Solution

Liveness probe for BigchainDB container changed to HTTP API responses instead of `abci_query`

### DIFF
--- a/k8s/bigchaindb/bigchaindb-ss.yaml
+++ b/k8s/bigchaindb/bigchaindb-ss.yaml
@@ -118,7 +118,7 @@ spec:
               key: bdb-rpc-port       
         resources:
           limits:
-            cpu: 200m
+            cpu: 1
             memory: 5G
         volumeMounts:
         - name: bdb-data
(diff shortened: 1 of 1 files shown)


k8s/bigchaindb/bigchaindb-ss.yaml
```diff
@@ -118,7 +118,7 @@ spec:
               key: bdb-rpc-port       
         resources:
           limits:
-            cpu: 200m
+            cpu: 1
```
ttmc: What does this mean (`cpu: 1`)?

ttmc:
It seems that the changes were made for the container with `- name: tendermint` and there is no container with `- name: bigchaindb`

k8s/bigchaindb/bigchaindb-ss.yaml
ttmc: Also, why wasn't the corresponding `cpu:` line (line 285) changed for the bigchaindb container?

shahbazn: does one core means more compute resource than 200m?

ttmc: :+1:

PR #2252, opened 2018-05-02
Problem: Upsert validator not migrated to Tendermint 0.19.0

kansi:
Solution: Migrate the same to Tendermint 0.19.0

### Description
In Tendermint `0.19.0` the public keys are now `base64` encoded but the ABCI response should be go-amino encoded ed25519 public key. The necessary decoding have been implemented.


bigchaindb/tendermint/core.py
```diff
@@ -148,9 +150,8 @@ def commit(self):
 
 
 def encode_validator(v):
-    pub_key = v['pub_key']['data']
-    # NOTE: tendermint expects public to be encoded in go-wire format
-    # so `01` has to be appended
-    pub_key = bytes.fromhex('01{}'.format(pub_key))
+    ed25519_public_key = v['pub_key']['data']
```
muawiakh: Shouldn't we handle the key accesses in a better way. Maybe something like this?
```
try:
    x = y.get('key')
except KeyError, e:
    raise something
```

kansi: @muawiakh you are right, but the way we [store the key](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/commands/bigchaindb.py#L103) when `upsert-validator PUBKEY POWER` command is executed ensures that there wouldn't be any key errors

PR #2256, opened 2018-05-02
Problem: Acceptance tests on CI failing

muawiakh:
## Description
Acceptance tests on Travis are failing because the following command fails during `travis-install.sh`

```
/bin/sh -c pip install --no-cache-dir --process-dependency-links -e .[dev]
```

More details in the [travis build](https://travis-ci.org/bigchaindb/bigchaindb/jobs/373963887).

## Solution

Update the travis matrix to enable acceptance tests for python36


vrde:
We should remove Python 3.5 from our Travis CI, instead of forcing Python 3.6 for a specific row of `matrix`.

PR #2259, opened 2018-05-03
Problem: How do I store files in BigchainDB?

ttmc:
Solution: Write a new root docs page about how to store files in BigchainDB.

Note: This is a frequently-asked question (FAQ) on Gitter and elsewhere.


ttmc:
@vrde I assigned you as a reviewer because you're going to do some tests to see how big a transaction can be before things break. We might want to add that factoid to this page.

PR #2261, opened 2018-05-03
Problem: pypi bigchaindb-abci not used

kansi:
Solution: Migrate to pypi abci


kansi:
@ttmc I deliberately didn't remove it as it generalizes the installation.

ttmc:
Okay, that seems okay to me.

PR #2270, opened 2018-05-07
Problem: Invalid transaction posted to Tendermint

### LINKED ISSUE #2269: HTTP API responds with 202 for certain invalid transactions
Happens because `BigchainDB::validate_transaction` in `tendermint/lib` does not raise exceptions as the old BigchainDB used to do, but returns a boolean.

A patch offering a failed `web` test is in the attachment - just remove the `.txt` part.

[0001-Problem-Amount-error-is-not-tested-on-the-HTTP-level.patch.txt](https://github.com/bigchaindb/bigchaindb/files/1979715/0001-Problem-Amount-error-is-not-tested-on-the-HTTP-level.patch.txt)

kansi:
Solution: Pass the exception to HTTP POST handler function.

Fixes #2269

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -259,21 +259,9 @@ class BigchainDB(Bigchain):
         if len(transactions) > 1:
             raise DoubleSpend('tx "{}" spends inputs twice'.format(txid))
         elif transactions:
-            transaction = transactions[0]
+            transaction = Transaction.from_db(self, transactions[0])
 
-        if transaction and transaction['operation'] == 'CREATE':
-            asset = backend.query.get_asset(self.connection, transaction['id'])
-
-            if asset:
-                transaction['asset'] = asset
-            else:
-                transaction['asset'] = {'data': None}
-
-            return Transaction.from_dict(transaction)
-        elif transaction and transaction['operation'] == 'TRANSFER':
-            return Transaction.from_dict(transaction)
-        else:
-            return None
+        return transaction
 
     def store_block(self, block):
         """Create a new block."""
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -19,7 +19,9 @@ import requests
 from bigchaindb import backend
 from bigchaindb import Bigchain
 from bigchaindb.models import Transaction
-from bigchaindb.common.exceptions import SchemaValidationError, ValidationError
+from bigchaindb.common.exceptions import (SchemaValidationError,
+                                          ValidationError,
+                                          DoubleSpend)
 from bigchaindb.tendermint.utils import encode_transaction, merkleroot
 from bigchaindb.tendermint import fastquery
 from bigchaindb import exceptions as core_exceptions
@@ -242,6 +244,10 @@ class BigchainDB(Bigchain):
         transactions = backend.query.get_spent(self.connection, txid,
                                                output)
         transactions = list(transactions) if transactions else []
+        if len(transactions) > 1:
+            raise core_exceptions.CriticalDoubleSpend(
+                '`{}` was spent more than once. There is a problem'
+                ' with the chain'.format(txid))
 
         for ctxn in current_transactions:
             for ctxn_input in ctxn.inputs:
@@ -251,9 +257,7 @@ class BigchainDB(Bigchain):
 
         transaction = None
         if len(transactions) > 1:
-            raise core_exceptions.CriticalDoubleSpend(
-                '`{}` was spent more than once. There is a problem'
-                ' with the chain'.format(txid))
+            raise DoubleSpend('tx "{}" spends inputs twice'.format(txid))
         elif transactions:
             transaction = transactions[0]
 
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -338,6 +338,9 @@ class BigchainDB(Bigchain):
 
         transaction = tx
 
+        # CLEANUP: The conditional below checks for transaction in dict format.
+        # It would be better to only have a single format for the transaction
+        # throught the code base.
         if not isinstance(transaction, Transaction):
             try:
                 transaction = Transaction.from_dict(tx)
@@ -347,12 +350,14 @@ class BigchainDB(Bigchain):
             except ValidationError as e:
                 logger.warning('Invalid transaction (%s): %s', type(e).__name__, e)
                 return False
+        return transaction.validate(self, current_transactions)
+
+    def is_valid_transaction(self, tx, current_transactions=[]):
         try:
-            return transaction.validate(self, current_transactions)
+            return self.validate_transaction(tx, current_transactions)
         except ValidationError as e:
             logger.warning('Invalid transaction (%s): %s', type(e).__name__, e)
             return False
-        return transaction
 
     @property
     def fastquery(self):
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -345,6 +345,8 @@ class BigchainDB(Bigchain):
         return transaction.validate(self, current_transactions)
 
     def is_valid_transaction(self, tx, current_transactions=[]):
+        # NOTE: the function returns the Transaction object in case
+        # the transaction is valid
         try:
             return self.validate_transaction(tx, current_transactions)
         except ValidationError as e:
--- a/tests/web/test_transactions.py
+++ b/tests/web/test_transactions.py
@@ -345,25 +345,22 @@ def test_post_invalid_transfer_transaction_returns_400(b, client, user_pk):
 @pytest.mark.tendermint
 def test_post_wrong_asset_division_transfer_returns_400(b, client, user_pk):
     from bigchaindb.models import Transaction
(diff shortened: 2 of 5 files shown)


bigchaindb/tendermint/lib.py
```diff
@@ -347,12 +351,18 @@ def validate_transaction(self, tx, current_transactions=[]):
             except ValidationError as e:
                 logger.warning('Invalid transaction (%s): %s', type(e).__name__, e)
                 return False
+        return transaction.validate(self, current_transactions)
+
+    def is_valid_transaction(self, tx, current_transactions=[]):
+        # The below execption handling is quite broad because this method
+        # is called by ABCI proxy app. An exception not caught in this method
+        # will cause a crash in the ABCI server leaving the system in a
+        # dis-functional state
```
vrde: typos: execption → exception, dis-functional → dysfunctional.

bigchaindb/tendermint/lib.py
```diff
@@ -347,12 +351,18 @@ def validate_transaction(self, tx, current_transactions=[]):
...
         try:
-            return transaction.validate(self, current_transactions)
-        except ValidationError as e:
-            logger.warning('Invalid transaction (%s): %s', type(e).__name__, e)
+            return self.validate_transaction(tx, current_transactions)
+        except BigchainDBError as e:
+            logger.error('Invalid transaction (%s): %s', type(e).__name__, e)
```
vrde: Any specific reason to use `logger.error`? `info` should be enough.

tests/web/test_transactions.py
```diff
@@ -342,6 +342,29 @@ def test_post_invalid_transfer_transaction_returns_400(b, client, user_pk):
...
+    assert res.status_code == 202
+
+    transfer_tx = Transaction.transfer(
+        create_tx.to_inputs(),
+        [([pub_key], 20)],  # 20 > 10
+        asset_id=create_tx.id,
+    ).sign([priv_key])
+    res = client.post(TX_ENDPOINT + '?mode=commit', data=json.dumps(transfer_tx.to_dict()))
+    # TODO identify the particular error in the response
+    assert res.status_code == 400
```
codegeschrei: why is the check for the error just a todo. Can you add an assert for the error / error message

bigchaindb/tendermint/lib.py
```diff
@@ -347,12 +351,18 @@ def validate_transaction(self, tx, current_transactions=[]):
...
+        return transaction.validate(self, current_transactions)
+
+    def is_valid_transaction(self, tx, current_transactions=[]):
+        # The below exception handling is quite broad because this method
+        # is called by ABCI proxy app. An exception not caught in this method
+        # will cause a crash in the ABCI server leaving the system in a
+        # dysfunctional state
         try:
-            return transaction.validate(self, current_transactions)
-        except ValidationError as e:
```
kansi: `ValidationError` doesn't handle `CriticalDoubleSpend` and this causes a crash. Both these exception types have a common parent `BigchainBDError`.

vrde: I think we should investigate why there was a `CriticalDoubleSpend`.

The exception is defines as:
```python
class CriticalDoubleSpend(BigchainDBError):
    """Data integrity error that requires attention"""
```

It is raised in [`get_spent`](https://github.com/bigchaindb/bigchaindb/blob/9cdd46c0844b442af738e83c89dcc24fae8d21c7/bigchaindb/core.py#L389). Do you have an idea on why this is happening?

ldmberman: Perhaps, there is a test reproducing it? Should `validate_transaction` never raise `BigchainBDError`?

bigchaindb/tendermint/lib.py
```diff
@@ -353,10 +354,14 @@ def validate_transaction(self, tx, current_transactions=[]):
         return transaction.validate(self, current_transactions)
 
     def is_valid_transaction(self, tx, current_transactions=[]):
+        # The below exception handling is quite broad because this method
+        # is called by ABCI proxy app. An exception not caught in this method
+        # will cause a crash in the ABCI server leaving the system in a
```
ldmberman: Under the circumstances, might makes sense to additionally catch all the exceptions here to be on the safe side. If it's not `BigchainBDError`, log an error and return False, I suppose?

kansi: As a general practice suggested by @vrde only expected exceptions should be handeled. Since `BigchainDBError` is the base class for all BigchainDB specific exceptions, so that has been handled here.

@vrde Any thoughts?

bigchaindb/tendermint/lib.py
kansi: @vrde My understanding is that [`get_spent`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/tendermint/lib.py#L241) is called with `current_transactions` from the un-commited block which probably contains duplicate transactions which raises an exception.

codegeschrei: How can we have an un-commited block with duplicate transactions?

kansi: @ldmberman this [test](https://github.com/bigchaindb/bigchaindb/blob/master/tests/tendermint/test_core.py#L105) check that.

@codegeschrei During the Tendermint [state machine](http://tendermint.readthedocs.io/projects/tools/en/master/introduction.html#intro-to-abci) until `commit` is executed the block is considered un-commited. If a transaction T2 spends T1 and there are two instances of T2 transaction included in the same block then we have a duplicate transactions in the same block.

codegeschrei: But we should reject a double spend right in the beginning, or am I wrong?

kansi: @codegeschrei You are right. The initial check for double spend compares the incoming transactions to the transactions which have been commited to the chain. It doesn't compare to the transaction that haven't yet been commited and are still in the mempool (this check is perfomed during the `deliver_tx` ABCI call)

bigchaindb/tendermint/lib.py
```diff
@@ -347,12 +342,14 @@ def validate_transaction(self, tx, current_transactions=[]):
             except ValidationError as e:
                 logger.warning('Invalid transaction (%s): %s', type(e).__name__, e)
                 return False
+        return transaction.validate(self, current_transactions)
+
+    def is_valid_transaction(self, tx, current_transactions=[]):
```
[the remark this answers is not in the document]
ldmberman: Why can'y we always return True/False? Then also docstring is not needed since the function is self-explanatory.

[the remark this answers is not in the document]
vrde: It is used in the code in [an assignment](https://github.com/bigchaindb/bigchaindb/pull/2270/files#diff-c45065100e5e308772764cdf833def41R83).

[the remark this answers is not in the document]
ldmberman: But the transaction hopefully does not change during validation. So that code can just use the object it passes to the function.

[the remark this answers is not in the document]
kansi: @vrde @ldmberman I think the concern raised is valid but I feel the context of this PR is getting too broad. Fixing the return value for `is_valid_transaction` would also require chainging the response values in the ABCI proxy app. I think this refactoring should be done in a seperate PR?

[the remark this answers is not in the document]
ldmberman: If it's only about the place @vrde pointed to, I would fix it now cause it's very simple.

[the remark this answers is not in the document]
kansi: @ldmberman Could you please explain the fix you have in mind?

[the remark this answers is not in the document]
ldmberman: Instead of:
```
transaction = self.bigchaindb.is_valid_transaction(
             decode_transaction(raw_transaction), self.block_transactions)
if not transaction:
    ...
```
do:
```
transaction = decode_transaction(raw_transaction)
if not self.bigchaindb.is_valid_transaction(transaction, self.block_transactions):
    ...
```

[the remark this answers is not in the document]
kansi: Hmm, the return value in the `transaction` variable is an object and not a dict. The function call `store_bulk_transactions` in the `commit` API expects an object not a dict. Also the corresponding tests for `store_bulk_transactions` will have to be fixed.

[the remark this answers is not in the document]
ldmberman: Ah, I see! Can we then do `Transaction.from_dict` there?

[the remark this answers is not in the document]
kansi: @ldmberman Yup. But as soon as we make that change we will need to modify another bunch of code which is out of the context of this PR (refer 2.3.2 in BEP#1).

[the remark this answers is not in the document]
ldmberman: Ideally:

```
if is_valid_transaction(tx_dict):
    transaction = Transaction.from_dict(tx)

if transaction:
   ...
```

But I see the mess around it now so I am fine with keeping it as is for now, sorry for holding that off for a little while.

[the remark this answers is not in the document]
kansi: @ldmberman No problem, I am happy to answer the queries.

PR #2275, opened 2018-05-09
Problem: abci_info liveness probe check is not enough

muawiakh:
## Description

Only using `abci_info` endpoint to check the status of the system is not enough, sometimes the connection is not broken between BigchainDB and Tendermint but Tendermint still behaves incorrectly and is not restarted appropriately.

## Solution

Add a `/status` liveness probe to the mix

### DIFF
--- a/k8s/bigchaindb/bigchaindb-ss.yaml
+++ b/k8s/bigchaindb/bigchaindb-ss.yaml
@@ -136,7 +136,8 @@ spec:
             - /bin/bash
             - "-c"
             - |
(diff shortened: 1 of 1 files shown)


ttmc:
I looked at the list of endpoints at http://tendermint.readthedocs.io/projects/tools/en/master/specification/rpc.html#endpoints

There are quite a few! I wonder about using the `/health` endpoint instead of the `/status` endpoint.

muawiakh:
@ttmc I checked `/health` as well, health endpoint is not useful for us. For some reason, it does not return the replies required for the operations we need to perform.

ttmc:
Okay, I will approve.

PR #2278, opened 2018-05-09
Problem: web transaction tests were broken or disabled.

ldmberman:
Solution: Fix or reenable the web transaction tests after the migration to BigchainDB 2.*.

### DIFF
--- a/bigchaindb/backend/schema.py
+++ b/bigchaindb/backend/schema.py
@@ -122,7 +122,7 @@ def validate_language_key(obj, key):
     """
     backend = bigchaindb.config['database']['backend']
 
-    if backend == 'mongodb':
+    if backend == 'localmongodb':
         data = obj.get(key, {})
         if isinstance(data, dict):
             validate_all_values_for_key(data, 'language', validate_language)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -350,6 +350,14 @@ def signed_create_tx(b, create_tx):
     return create_tx.sign([b.me_private])
 
 
+@pytest.mark.abci
+@pytest.fixture
+def posted_create_tx(b, signed_create_tx):
+    res = b.post_transaction(signed_create_tx, 'broadcast_tx_commit')
+    assert res.status_code == 200
+    return signed_create_tx
+
+
 @pytest.fixture
 def signed_transfer_tx(signed_create_tx, user_pk, user_sk):
     from bigchaindb.models import Transaction
--- a/tests/web/test_transactions.py
+++ b/tests/web/test_transactions.py
@@ -1,5 +1,5 @@
 import json
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import base58
 import pytest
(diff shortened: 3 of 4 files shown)


tests/web/test_transactions.py
```diff
@@ -12,18 +12,15 @@
...
 
 
-@pytest.mark.bdb
-@pytest.mark.usefixtures('inputs')
-def test_get_transaction_endpoint(b, client, user_pk):
-    input_tx = b.get_owned_ids(user_pk).pop()
-    tx = b.get_transaction(input_tx.txid)
-    res = client.get(TX_ENDPOINT + tx.id)
-    assert tx.to_dict() == res.json
+pytestmark = pytest.mark.abci
```
vrde: Why do those tests need this marker? /cc @kansi

bigchaindb/backend/schema.py
```diff
@@ -122,7 +122,7 @@ def validate_language_key(obj, key):
     """
     backend = bigchaindb.config['database']['backend']
 
-    if backend == 'mongodb':
+    if backend in ('mongodb', 'localmongodb'):
```
kansi: I think we can just use `'localmongodb'`, @vrde ?

tests/web/test_transactions.py
kansi: I would also like to understand why every test in this module needs this marker?

bigchaindb/backend/schema.py
ldmberman: Why would you only want to validate the language in tests?

tests/web/test_transactions.py
ldmberman: All of the tests in this module are API-level. Only a few of them do not require the running Tendermint - I was thinking it would not hurt to mark the whole module as such.

bigchaindb/backend/schema.py
kansi: @ldmberman The 'mongodb' backend will be gone after refactoring. So, I think we can just use `'localmongdb'` there.

ldmberman: I understand now, thanks! Yes, makes sense to only keep `localmongodb` then.

tests/web/test_transactions.py
```diff
@@ -270,10 +265,12 @@ def test_post_invalid_transaction(mock_logger, client, exc, msg, monkeypatch,):
     def mock_validation(self_, tx):
         raise exc_cls(msg)
 
+    TransactionMock = Mock(validate=mock_validation)
+
     monkeypatch.setattr(
         'bigchaindb.Bigchain.validate_transaction', mock_validation)
     monkeypatch.setattr(
-        'bigchaindb.models.Transaction.from_dict', lambda tx: None)
+        'bigchaindb.models.Transaction.from_dict', lambda tx: TransactionMock)
```
kansi: Could you please elaborate this?

tests/conftest.py
```diff
@@ -350,6 +350,13 @@ def signed_create_tx(b, create_tx):
     return create_tx.sign([b.me_private])
 
 
+@pytest.fixture
+def posted_create_tx(b, signed_create_tx):
```
kansi: Is this a static fixture?

tests/web/test_transactions.py
ldmberman: The test mocks the validation that happens inside `Transaction.validate` now. I can actually also remove the previous patch.

tests/conftest.py
ldmberman: Nope, good catch - added an `abci` mark.

PR #2282, opened 2018-05-11
Problem: acceptance tests are using old pydriver

vrde:
Solution: update to the new driver version, and simplify the code removing the `mode` parameter.

### DIFF
--- a/acceptance/python/Dockerfile
+++ b/acceptance/python/Dockerfile
@@ -5,4 +5,4 @@ RUN pip install --upgrade \
 	pycco \
 	websocket-client~=0.47.0 \
 	pytest~=3.0 \
-	bigchaindb-driver==0.5.0a2
+	bigchaindb-driver==0.5.0a4
--- a/acceptance/python/src/test_basic.py
+++ b/acceptance/python/src/test_basic.py
@@ -59,7 +59,7 @@ def test_basic():
 
 
     # Now she is ready to send it to the BigchainDB Network.
-    sent_transfer_tx = bdb.transactions.send(fulfilled_creation_tx, mode='commit')
+    sent_transfer_tx = bdb.transactions.send(fulfilled_creation_tx)
 
     # And just to be 100% sure, she also checks if she can retrieve
     # it from the BigchainDB node.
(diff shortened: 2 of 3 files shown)


acceptance/python/src/test_basic.py
```diff
@@ -59,7 +59,7 @@ def test_basic():
 
 
     # Now she is ready to send it to the BigchainDB Network.
-    sent_transfer_tx = bdb.transactions.send(fulfilled_creation_tx, mode='commit')
+    sent_transfer_tx = bdb.transactions.send(fulfilled_creation_tx)
```
ttmc: The default mode is `async` so this will change the mode. Is that your intent?

acceptance/python/Dockerfile
```diff
@@ -5,4 +5,4 @@ RUN pip install --upgrade \
 	pycco \
 	websocket-client~=0.47.0 \
 	pytest~=3.0 \
-	bigchaindb-driver==0.5.0a2
+	bigchaindb-driver==0.5.0a4
```
ttmc: I wonder if we should _always_ update this line as part of our [release process](https://github.com/bigchaindb/bigchaindb/blob/master/RELEASE_PROCESS.md)? We could add a sub-step to the release process (in step 1).

ttmc: Actually, maybe it's part of the Python driver release process.

acceptance/python/src/test_basic.py
vrde: The default is actually [`commit`](https://github.com/bigchaindb/bigchaindb-driver/blob/fd088013e0159ff2baf89fabbb78f1c6df3a95ba/CHANGELOG.rst#050a2-2018-04-18).

acceptance/python/Dockerfile
vrde: To me, this depends on the driver, should we put something there?

ttmc: If the driver had its own release process document, we could put a task in that, but it doesn't have one. However, there _is_ an open issue to create one. I'll add a comment to that issue, noting that the process should include updating this line in the bigchaindb/bigchaindb repo.

acceptance/python/src/test_basic.py
ttmc: Really? That's weird, because if you don't include the `?mode=async` string at the end of the `POST /api/v1/transactions?mode=async` endpoint, then the default is `async`, not `commit`. In other words, the endpoint treats `async` as the default, so the Python driver should follow its lead.

vrde: Oh, I see what you mean now. To me, there are many reasons why we should have `commit` as the default, also on the `POST /api/v1/transactions` endpoint. I'll open an issue for that.

ttmc: Okay, maybe post a link to that issue here in this discussion thread, for quick reference.

ttmc: I created an issue to discuss the default mode: issue #2307

PR #2284, opened 2018-05-11
Problem: Liveness probe too aggressive

muawiakh:

### DIFF
--- a/k8s/bigchaindb/bigchaindb-ss.yaml
+++ b/k8s/bigchaindb/bigchaindb-ss.yaml
@@ -149,7 +149,7 @@ spec:
                 exit 0
               fi
           initialDelaySeconds: 60
-          periodSeconds: 15
+          periodSeconds: 60
           failureThreshold: 3
           timeoutSeconds: 15
       # BigchainDB container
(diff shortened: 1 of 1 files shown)


shahbazn:
Whats the issue here caused by liveness probe being aggressive ?

muawiakh:
Sometimes, failures are caused by network delays or tendermint recovers itself. So keeping it a bit laid back.

ttmc:
I wanted to better-understand what `periodSeconds` means, so I looked it up. Apparently it is _how often_ the kubelet performs a liveness probe.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes

`timeoutSeconds` is "Number of seconds after which the probe times out." Maybe we should be increasing that?

muawiakh:
We already have a max timeout of 10 seconds here(https://github.com/bigchaindb/bigchaindb/blob/master/k8s/bigchaindb/bigchaindb-ss.yaml#L139), which should be sufficient because it is being resolved over `localhost`.  I am just changing the frequency of the the probe.

PR #2286, opened 2018-05-11
Problem: The pipelines library is not used anymore.

ldmberman:
Solution: Remove pipelines. This is part of the BigchainDB 2.* migration cleanup.


ttmc:
broom

ldmberman:
I removed `test_processes.py` but forgot about `processes.py` itself - I will remove it too.

ttmc:
Is the `multipipes` package used anywhere else (besides the pipelines)? If not, we can remove it as a dependency, e.g. in `setup.py`.

PR #2294, opened 2018-05-15
Problem: we don't have enough acceptance tests

codegeschrei:
Solution: test different scenarios for divisible assets


acceptance/python/src/test_divisible_asset.py
```diff
@@ -0,0 +1,176 @@
+# # Divisible assets integration testing
```
ldmberman: If we ever generate documentation out of these tests, it would come handy to have the descriptions enclosed in 
 the triple quotes `"""` so that the rendering tool parses it properly.

codegeschrei: we do. You can test it by running `make doc-acceptance`

acceptance/python/src/test_divisible_asset.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Oh look, it is Alice again and she brought her friend Bob along.
+    alice, bob = generate_keypair(), generate_keypair()
+
+
+    # ## Alice creates a time sharing token
+    # Alice wants to go on vacation, while Bobs bike just broke down.
+    # Alice decides to rent her bike to Bob while she is gone.
+    # So she creates a time sharing token for her bike.
+    # As you can see in the description one token can be used to ride
```
ldmberman: I was confused by this sentence at first because tokens are introduced later by `bdb.transactions.prepare`. What do you think about explaining the correspondence there?

Perhaps, like this:
```
# She prepares a `CREATE` transaction and issues 10 tokens. Bob and Alice agree that each token buys one hour of riding the bike.
```

acceptance/python/src/test_divisible_asset.py
```diff
@@ -0,0 +1,176 @@
...
+# of a given transaction.
+#
+# This integration test is a rip-off of our
+# [tutorial](https://docs.bigchaindb.com/projects/py-driver/en/latest/usage.html).
+
+# ## Imports
+# We need some utils from the `os` package, we will interact with
+# env variables.
+# We need the `pytest` package to catch the `AmountError` exception properly.
+# And of course, we also need the `AmountError`.
```
ldmberman: It's actually `BadRequest`, right?

codegeschrei: yes it is, that's also what I'm importing. I forgot to change this part.

acceptance/python/src/test_divisible_asset.py
```diff
@@ -0,0 +1,176 @@
...
+    # Oh Bob, what have you done?! You tried to spend more tokens than you had.
+    # Remember Bob, last time you spent 3 tokens already,
+    # so you only have 7 left.
+    with pytest.raises(BadRequest) as error:
+        bdb.transactions.send(fulfilled_transfer_tx, mode='commit')
+
+    # Now Bob gets an error saying that the amount he wanted to spent is
+    # higher than the amount of tokens he has left.
+    assert error.value.args[0] == 400
+    message = 'Invalid transaction (AmountError): The amount used in the ' \
```
ldmberman: In the future we need to introduce clean error codes to parse, like `amount_error`. It's inconvenient to rely on long strings aimed to explain the error to humans in the API clients.

codegeschrei:
good idea @ldmberman we can make another PR and add it.

ldmberman:
#2295

PR #2296, opened 2018-05-16
Problem: no instructions on simple network setup

vrde:
Solution: write a step-by-step tutorial on who should do what.

Note: this is a work in progress. It covers the first steps to set up a static network. The second part is on how to add validators dynamically.

### DIFF
--- a/docs/server/source/simple-network-setup.md
+++ b/docs/server/source/simple-network-setup.md
@@ -146,30 +146,30 @@ At this point the Coordinator should have received the data from all the Members
       "name": "<Member 1 name>"
     },
     {
-     "pub_key": {
-       "type": "AC26791624DE60",
-       "value": "<Member 2 public key>"
-     },
-     "power": 10,
+      "pub_key": {
+        "type": "AC26791624DE60",
+        "value": "<Member 2 public key>"
+      },
+      "power": 10,
       "name": "<Member 2 name>"
-    },
-    {
-     "...": { },
-    },
-    {
-     "pub_key": {
-       "type": "AC26791624DE60",
-       "value": "<Member N public key>"
      },
-     "power": 10,
-      "name": "<Member N name>"
-    }
+     {
+       "...": { },
+     },
+     {
+       "pub_key": {
+         "type": "AC26791624DE60",
+         "value": "<Member N public key>"
+       },
+       "power": 10,
+       "name": "<Member N name>"
+     }
   ],
   "app_hash": ""
 }
 ```
 
-The new `genesis.json` file contains the data that describes the Network.
+The new `genesis.json` file contains the data that describes the Network. The key `name` is the Member's moniker, it can be any valid string, put something human readable like: "Alice's Node Shop".
 
 At this point, the Coordinator must share the new `genesis.json` file with all Members.
 
--- a/docs/server/source/simple-network-setup.md
+++ b/docs/server/source/simple-network-setup.md
@@ -127,7 +127,8 @@ tendermint show_node_id
 
 Share with all other Members your `node_id`, `pub_key.value`, and the host name or IP of your Node.
 
-**Important note on security: each Member should verify the data they receive from the other Members.**
+**Important note on security: each Member should take extra steps to verify the public keys they receive from the other Members have not been tampered, e.g. a key signing party would help address that.**
+
 
 ## Coordinator: Initialize the Network
 At this point the Coordinator should have received the data from all the Members, and should combine them in the `.tendermint/config/genesis.json` file:
--- a/docs/server/source/simple-network-setup.md
+++ b/docs/server/source/simple-network-setup.md
@@ -39,10 +39,10 @@ BigchainDB Server requires **Python 3.6+**, make sure your system has it. Instal
 sudo apt install -y python3-pip libssl-dev
 ```
 
-Now install the latest version of BigchainDB. Check the [project page on PyPI][bdb:pypi] for the last version (`2.0.0a5` at the time of writing) and install it:
+Now install the latest version of BigchainDB. Check the [project page on PyPI][bdb:pypi] for the last version (`2.0.0a6` at the time of writing) and install it:
 
 ```
-sudo pip3 install bigchaindb==2.0.0a5
+sudo pip3 install bigchaindb==2.0.0a6
 ```
 
 Check you installed the correct version of BigchainDB Server with `bigchaindb --version`.
@@ -60,7 +60,7 @@ The package manager should take care of installing the startup script for MongoD
 Install a [recent version of Tendermint][tendermint:releases]. BigchainDB Server requires `0.19` or newer.
 
 ```
-sudo apt install unzip
+sudo apt install -y unzip
 wget https://github.com/tendermint/tendermint/releases/download/v0.19.3/tendermint_0.19.3_linux_amd64.zip
 unzip tendermint_0.19.3_linux_amd64.zip
 rm tendermint_0.19.3_linux_amd64.zip
--- a/docs/server/source/simple-network-setup.md
+++ b/docs/server/source/simple-network-setup.md
@@ -25,17 +25,23 @@ Every Member in the Network **must** set up its own Node. The process consists o
 **Important note on security: it's up to the Member to harden their system.**
 
 ### Install the required software
+Make sure your system is up to date.
+
+```
+sudo apt update
+sudo apt full-upgrade
+```
 
 #### Install BigchainDB Server
 BigchainDB Server requires **Python 3.6+**, make sure your system has it. Install the required packages:
 
-```bash
+```
 sudo apt install -y python3-pip libssl-dev
 ```
 
 Now install the latest version of BigchainDB. Check the [project page on PyPI][bdb:pypi] for the last version (`2.0.0a5` at the time of writing) and install it:
 
-```bash
+```
 sudo pip3 install bigchaindb==2.0.0a5
 ```
 
@@ -44,7 +50,7 @@ Check you installed the correct version of BigchainDB Server with `bigchaindb --
 #### Install MongoDB
 Install a recent version of MongoDB. BigchainDB Server requires `3.4` or newer.
 
-```bash
+```
 sudo apt install mongodb
 ```
 
@@ -53,7 +59,7 @@ The package manager should take care of installing the startup script for MongoD
 #### Install Tendermint
 Install a [recent version of Tendermint][tendermint:releases]. BigchainDB Server requires `0.19` or newer.
 
-```bash
+```
 sudo apt install unzip
 wget https://github.com/tendermint/tendermint/releases/download/v0.19.3/tendermint_0.19.3_linux_amd64.zip
 unzip tendermint_0.19.3_linux_amd64.zip
@@ -64,7 +70,7 @@ sudo mv tendermint /usr/local/bin
 ### Firewall settings
 Make sure to accept inbound connections to (`22`, you don't want to lock you out), `9984`, `9985`, and `46656`.
 
-```bash
+```
 sudo ufw allow 22/tcp
 sudo ufw allow 9984/tcp
 sudo ufw allow 9985/tcp
@@ -77,24 +83,24 @@ Some cloud providers, like Microsoft Azure, require you to go through their port
 ## Member: Configure BigchainDB Server
 To configure BigchainDB Server, run:
 
-```bash
+```
 bigchaindb configure
 ```
 
 The first question is ``API Server bind? (default `localhost:9984`)``, to expose the API to the public, bind the API Server to `0.0.0.0:9984`. Unless you have specific needs, you can keep the default value for all other questions.
 
-## Member: Generate the private key and address
-A Node is identified by the triplet `<hostname, address, public_key>`.
+## Member: Generate the private key and node id
+A Node is identified by the triplet `<hostname, node_id, public_key>`.
 
-As a Member, it's your duty to create and store securely your private key, and share your `hostname`, `address`, and `public_key` with the other members of the network.
+As a Member, it's your duty to create and store securely your private key, and share your `hostname`, `node_id`, and `public_key` with the other members of the network.
 
 To generate all of that, run:
 
-```bash
+```
 tendermint init
 ```
 
-The `public_key` and `address` are stored in the file `.tendermint/config/priv_validator.json`, and it should look like:
+The `public_key` is stored in the file `.tendermint/config/priv_validator.json`, and it should look like:
 
 ```json
 {
@@ -113,7 +119,13 @@ The `public_key` and `address` are stored in the file `.tendermint/config/priv_v
 }
 ```
 
-Share with all other Members your `address`, `pub_key.value`, and the host name or IP of your Node.
+To extract your `node_id`, run the command:
+
+```
+tendermint show_node_id
+```
+
+Share with all other Members your `node_id`, `pub_key.value`, and the host name or IP of your Node.
 
 **Important note on security: each Member should verify the data they receive from the other Members.**
 
@@ -166,7 +178,7 @@ At this point the Member should have received the `genesis.json` file.
 
 **Important note on security: each Member should verify that the `genesis.json` file contains the correct public keys.**
 
-The Member must copy the `genesis.json` file in the local `.tendermint/config` directory.
+The Member must copy the `genesis.json` file in the local `.tendermint/config` directory. Every Member now shares the same `chain_id`, `genesis_time`, used to identify the Network, and the same list of `validators`.
 
 The Member must edit the `.tendermint/config/config.toml` file and make the following changes:
 
@@ -176,14 +188,14 @@ The Member must edit the `.tendermint/config/config.toml` file and make the foll
 create_empty_blocks = false
 ...
 
-persistent_peers = "<Member 1 address>@<Member 1 hostname>:46656,\
-<Member 2 address>@<Member 2 hostname>:46656,\
-<Member N address>@<Member N hostname>:46656,"
+persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
+<Member 2 node id>@<Member 2 hostname>:46656,\
+<Member N node id>@<Member N hostname>:46656,"
 ```
 
 Now the Member must run two commands:
 - `bigchaindb start`
-- `tendermint`
+- `tendermint node`
 
 The commands will run on the foreground, you might want to use `nohup`, `tmux`, or `screen` to keep them running after you logout. The recommended approach is to create startup scripts for both BigchainDB and Tendermint, so they will start right after the boot of the operating system.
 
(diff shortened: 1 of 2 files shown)


docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+## Before we start
+This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine. The commands are tailored for an up-to-date *Debian-like* distribution (we use an **Ubuntu 17.10** Virtual Machine on Microsoft Azure): if you are on a different distribution you might need to adapt the name of the packages installed.
+
+We don't make any assumptions on **where** you run the Node.
+You can run the BigchainDB Server on a Virtual Machine on the cloud, on a machine in your data center, on a Raspberry Pi. Just make sure that your Node is reachable by the other Nodes, a **non-exhaustive list of examples**:
+- **good**: all Nodes running in the cloud using public IPs.
+- **bad**: some Nodes running in the cloud using public IPs, some Nodes in a private network.
+- **good**: all Nodes running in a private network.
+
+The rule of thumb is: if Nodes can ping each other, then you are good to go.
```
ldmberman: Makes sense to clarify that to understand if nodes see each other you may search for the logs like `Successful handshake with peer`.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+A Network will stop working if more than one third of the Nodes are down. The bigger a Network, the more failures it can handle. A Network of size 4 can tolerate only 1 failure, so if 3 out of 4 Nodes are online, everything will work as expected. Eventually, the Node that was offline will automatically sync with the others.
+
+
+## Before we start
+This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine. The commands are tailored for an up-to-date *Debian-like* distribution (we use an **Ubuntu 17.10** Virtual Machine on Microsoft Azure): if you are on a different distribution you might need to adapt the name of the packages installed.
+
+We don't make any assumptions on **where** you run the Node.
+You can run the BigchainDB Server on a Virtual Machine on the cloud, on a machine in your data center, on a Raspberry Pi. Just make sure that your Node is reachable by the other Nodes, a **non-exhaustive list of examples**:
+- **good**: all Nodes running in the cloud using public IPs.
+- **bad**: some Nodes running in the cloud using public IPs, some Nodes in a private network.
```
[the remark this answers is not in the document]
vrde: The nodes in the private network are behind a NAT, so they are not reachable by the ones in the cloud.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+  "pub_key": {
+    "type": "AC26791624DE60",
+    "value": "W3tqeMCU3d4SHDKqrwQWTahTW/wpieIAKZQxUxLv6rI="
+  },
+  "last_height": 0,
+  "last_round": 0,
+  "last_step": 0,
+  "priv_key": {
+    "type": "954568A3288910",
+    "value": "3sv9aExgME6MMjx0JoKVy7KtED8PBiPcyAgsYmVizslbe2p4wJTd3hIcMqqvBBZNqFNb/CmJ4gAplDFTEu/qsg=="
```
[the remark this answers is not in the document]
vrde: Of course!

docs/server/source/simple-network-setup.md
vrde: It's too early in the document to talk about that.

docs/server/source/simple-network-setup.md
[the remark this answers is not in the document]
ldmberman: True :)

docs/server/source/simple-network-setup.md
ldmberman: I see, thanks - at first I misunderstood which type of ping you meant.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+
+**Important note on security: each Member should verify the data they receive from the other Members.**
+
+## Coordinator: Initialize the Network
+At this point the Coordinator should has received the data from all the Members, and should combine them in the `.tendermint/config/genesis.json` file:
+
+```json
+{
+  "genesis_time": "0001-01-01T00:00:00Z",
+  "chain_id": "test-chain-la6HSr",
```
ldmberman: It's important that everyone shares the same `chain_id`.

vrde: True, I'll make it more explicit thanks!

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+
+The next sections are labelled with **Member** or **Coordinator**, depending on who should follow the instructions. Remember, a Coordinator is also a Member.
+
+## Member: Set up a Node
+Every Member in the Network **must** set up its own Node. The process consists of installing three components, BigchainDB Server, Tendermint Core, and MongoDB, and configure the firewall.
+
+**Important note on security: it's up to the Member to harden their system.**
+
+### Install the required software
+
```
ttmc: I wonder if we should include some basic commands here to get their VM up-to-date before installing anything else, so that even if a Linux newbie followed these instructions, they'd be in fairly good shape. I'm thinking of:
```text
sudo apt update
sudo apt full-upgrade
```

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+#### Install BigchainDB Server
+BigchainDB Server requires **Python 3.6+**, make sure your system has it. Install the required packages:
+
+```bash
```
ttmc: I find that using "backtick backtick backtick bash" causes the result to be rendered with a distracting rainbow mix of colors, so I prefer using "backtick backtick backtick text" instead.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+  "priv_key": {
+    "type": "954568A3288910",
+    "value": "3sv9aExgME6MMjx0JoKVy7KtED8PBiPcyAgsYmVizslbe2p4wJTd3hIcMqqvBBZNqFNb/CmJ4gAplDFTEu/qsg=="
+  }
+}
+```
+
+Share with all other Members your `address`, `pub_key.value`, and the host name or IP of your Node.
+
+**Important note on security: each Member should verify the data they receive from the other Members.**
```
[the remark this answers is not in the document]
vrde: The question here is: how do I make sure that the public key that a member shared with me is the real one? How do I make sure there hasn't been a MITM attack?

There are many ways to do that, one idea is to have something like a "key review party" (something I just made up, similar to a PGP key signing party). I can add some hints.

docs/server/source/simple-network-setup.md
vrde: Yes, I noticed the same, will remove the language hint.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+
+A **BigchainDB Network** (*Network* henceforth) is a set of **4 or more BigchainDB Nodes** (*Node* henceforth). Every Node is independently managed by a Member, and runs an instance of the [BigchainDB Server software][bdb:software]. At the **Genesis** of a Network, there **MUST** be at least **4** Nodes ready to connect. After the Genesis, a Network can dynamically add new Nodes or remove old Nodes.
+
+A Network will stop working if more than one third of the Nodes are down. The bigger a Network, the more failures it can handle. A Network of size 4 can tolerate only 1 failure, so if 3 out of 4 Nodes are online, everything will work as expected. Eventually, the Node that was offline will automatically sync with the others.
+
+
+## Before we start
+This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine. The commands are tailored for an up-to-date *Debian-like* distribution (we use an **Ubuntu 17.10** Virtual Machine on Microsoft Azure): if you are on a different distribution you might need to adapt the name of the packages installed.
```
ttmc: I would feel better about suggesting a long-term-support (LTS) version of Ubuntu. Ubuntu 18.04 LTS is out now.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+...
+
+persistent_peers = "<Member 1 address>@<Member 1 hostname>:46656,\
+<Member 2 address>@<Member 2 hostname>:46656,\
+<Member N address>@<Member N hostname>:46656,"
+```
+
+Now the Member must run two commands:
+- `bigchaindb start`
+- `tendermint`
```
kansi: `tendermint node`

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+
+The Member must edit the `.tendermint/config/config.toml` file and make the following changes:
+
+```
+...
+
+create_empty_blocks = false
+...
+
+persistent_peers = "<Member 1 address>@<Member 1 hostname>:46656,\
```
kansi: It would be good to mention that the member address or member node id can be generated using `tendermint show_node_id`

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,197 @@
...
+sudo pip3 install bigchaindb==2.0.0a5
+```
+
+Check you installed the correct version of BigchainDB Server with `bigchaindb --version`.
+
+#### Install MongoDB
+Install a recent version of MongoDB. BigchainDB Server requires `3.4` or newer.
+
+```bash
+sudo apt install mongodb
```
ttmc: After doing this, I checked the MongoDB version:
```text
$ mongod --version
db version v3.6.3
git version: 9586e557d54ef70f9ca4b43c26892cd55257e1a5
OpenSSL version: OpenSSL 1.1.0g  2 Nov 2017
allocator: tcmalloc
modules: none
build environment:
    distarch: x86_64
    target_arch: x86_64
```
Uh oh, it's MongoDB 3.6.3, but we only test BigchainDB on MongoDB 3.4.x. I'm not sure what to do here: We could start testing on MongoDB 3.6.x, or we could force the install of MongoDB 3.4.x.

ttmc: I started looking up how to install MongoDB 3.4.x on Ubuntu and found the page https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/ 

It says:
screenshot from 2018-05-17 11-41-06

In other words, we shouldn't be using `sudo apt install mongodb`

docs/server/source/simple-network-setup.md
vrde: I'll add this at the beginning of the tutorial because it can help avoiding basic mistakes.

One key thing of this tutorial is:
> This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine.

I'm leaving other things to the user, like hardening the system and keeping things up to date.

docs/server/source/simple-network-setup.md
vrde: I can update to 18.04 LTS

docs/server/source/simple-network-setup.md
ttmc: It seems MongoDB (the company) doesn't have an official `mongodb-org` 3.4.x package for Ubuntu 18.04 LTS (bionic) yet, as far as I can tell. I tried it but ran into this error message:

"The repository 'http://repo.mongodb.org/apt/ubuntu bionic/mongodb-org/3.4 Release' does not have a Release file."

Maybe the solution, for now, is to use Ubuntu 16.04 LTS.

docs/server/source/simple-network-setup.md
vrde: Thanks, I should change `address` to `node id`.

docs/server/source/simple-network-setup.md
vrde: Good point, I see in our `setup.py` that we are using an old version of `pymongo`. I'll make a PR to upgrade to the last version of the driver, that is compatible with [all versions of MongoDB](https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#reference-compatibility-mongodb-python).

docs/server/source/simple-network-setup.md
muawiakh: Sometimes/Mostly ICMP/ping is disabled on start  of VM(s) spawned by cloud providers.

vrde: @muawiakh true that, I can add a note that some cloud providers disable ICMP traffic, any other suggestion?

What I usually do in this case is `nmap -P0 -sL <hostname>`, it works like a charm.

docs/server/source/simple-network-setup.md
ttmc: Okay, but what version of MongoDB should we be installing here: 3.4 or 3.6?

I suppose we _could_ use Ubuntu 18.04 and `sudo apt install mongodb` (i.e. the MongoDB package from Canonical, not from the MongoDB company) because that installs MongoDB 3.6.3 (currently).

I'd _prefer_ to use the official `mongodb-org` package from MongoDB, but they don't have a version for Ubuntu 18.04 yet.

I think the best option, for now, is to use Ubuntu 16.04 LTS and to use the official `mongodb-org` package to install MongoDB 3.6 on that.

ttmc:
We just released BigchainDB 2.0.0 Alpha 6, which uses PyMongo 3.6, and PyMongo 3.6 works with MongoDB 3.6 and lower. Therefore we can use Ubuntu 18.04 and install MongoDB using `sudo apt install mongodb` for now: that's Canonical's package to install MongoDB. It currently installs MongoDB 3.6.3.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+BigchainDB Server requires **Python 3.6+**, make sure your system has it. Install the required packages:
+
+```
+sudo apt install -y python3-pip libssl-dev
+```
+
+Now install the latest version of BigchainDB. Check the [project page on PyPI][bdb:pypi] for the last version (`2.0.0a5` at the time of writing) and install it:
+
+```
+sudo pip3 install bigchaindb==2.0.0a5
```
ttmc: Let's change this to 2.0.0a6 and maybe not mention any version number in the line above

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+
+create_empty_blocks = false
+...
+
+persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
+<Member 2 node id>@<Member 2 hostname>:46656,\
+<Member N node id>@<Member N hostname>:46656,"
+```
+
+Now the Member must run two commands:
```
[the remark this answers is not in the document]
ttmc: Something like this?
```
sudo mkdir -p /data/db
sudo chmod -R 700 /data/db
sudo mongod
```

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+- `bigchaindb start`
+- `tendermint node`
+
+The commands will run on the foreground, you might want to use `nohup`, `tmux`, or `screen` to keep them running after you logout. The recommended approach is to create startup scripts for both BigchainDB and Tendermint, so they will start right after the boot of the operating system.
```
ttmc: Maybe we could offer a little more help here, to be in line with the level of help so far. Maybe give some _example_ commands that will run MongoDB, BigchainDB and Tendermint in the background, e.g.
```
sudo mongod --fork --logpath mongodb.log
sudo nohup tendermint node 2>&1 > tendermint.log &
sudo nohup bigchaindb start 2>&1 > bigchaindb.log &
```
and then mention "The recommended approach..." afterwards.

docs/server/source/simple-network-setup.md
vrde: Everything works well with Ubuntu 18.04 :1st_place_medal:

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+```
+tendermint show_node_id
+```
+
+Share with all other Members your `node_id`, `pub_key.value`, and the host name or IP of your Node.
+
+**Important note on security: each Member should verify the data they receive from the other Members.**
+
+## Coordinator: Initialize the Network
+At this point the Coordinator should has received the data from all the Members, and should combine them in the `.tendermint/config/genesis.json` file:
```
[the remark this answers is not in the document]
vrde: I can has grammar.

docs/server/source/simple-network-setup.md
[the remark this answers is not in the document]
vrde: This is handled by the operating system when installing MongoDB, in the section *Install MongoDB* I wrote:

> The package manager should take care of installing the startup script for MongoDB as well.

docs/server/source/simple-network-setup.md
vrde: I'd like to add the two `systemd` unit files here, if you don't mind I'll address this in another PR :)

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+     "power": 10,
+      "name": "<Member N name>"
+    }
+  ],
+  "app_hash": ""
+}
+```
+
+The new `genesis.json` file contains the data that describes the Network.
+
```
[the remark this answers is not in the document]
vrde: I think you can put there whatever you want, @kansi can you confirm?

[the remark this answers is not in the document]
kansi: Yup, I also think the same. We are using node IPs as names for the internal network.

docs/server/source/simple-network-setup.md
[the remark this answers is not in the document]
ttmc: Installing MongoDB doesn't also run it, does it?

docs/server/source/simple-network-setup.md
ttmc: So the idea would be that they change the startup behavior of the VM, so that on reboot, MongoDB, BigchainDB and Tendermint will all start? That should be explained.

docs/server/source/simple-network-setup.md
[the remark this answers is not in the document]
ttmc: Just to follow-up on and close this comment thread, it seems that our strategy will be to tell people to create startup scripts for MongoDB, BigchainDB and Tendermint, and then to reboot their VM, so that everything will start on reboot. Apparently installing MongoDB already creates a startup script.

docs/server/source/simple-network-setup.md
```diff
@@ -0,0 +1,209 @@
...
+To extract your `node_id`, run the command:
+
+```
+tendermint show_node_id
+```
+
+Share with all other Members your `node_id`, `pub_key.value`, and the host name or IP of your Node.
+
+**Important note on security: each Member should verify the data they receive from the other Members.**
```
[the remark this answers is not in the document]
vrde: Yup, a signing party is an option. I can put something like:
> **Important note on security**: each Member should take extra steps to verify the public keys they receive from the other Members have not been tampered, e.g. a _key signing party_ would help address that.

docs/server/source/simple-network-setup.md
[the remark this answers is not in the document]
vrde: Yes exactly. I'd like to give hints on how to do things on an Ubuntu machine, and also give some startup scripts. Then it's up to the end user to customize it.

docs/server/source/simple-network-setup.md
ttmc: It still says **Ubuntu 17.10**.

PR #2298, opened 2018-05-17
Problem: BigchainDB does not support newer MongoDB

vrde:
Solution: Update driver to the last version, that is compatible with all versions of MongoDB.

Related to @ttmc's [comment](https://github.com/bigchaindb/bigchaindb/pull/2296#pullrequestreview-120964718) on #2296.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ benchmarks_require = [
 install_requires = [
     # TODO Consider not installing the db drivers, or putting them in extras.
     'rethinkdb~=2.3',  # i.e. a version between 2.3 and 3.0
-    'pymongo~=3.4',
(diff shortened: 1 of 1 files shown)


vrde:
I changed the image in our `docker-compose.yml` to use Mongo 3.6. There is also the one in `k8s/mongodb/container/Dockerfile` to change, I can do that but I don't know much about k8s.

muawiakh:
It is not `k8s` related, just a normal Docker image file. You can change the version.

ttmc:
Should we release a new version of BigchainDB after this PR gets merged, so that `pip install bigchaindb...` will use the proper version of PyMongo?

PR #2299, opened 2018-05-17
Problem: dev env doesn't allow P2P connections

vrde:
Solution: expose Tendermint P2P port 46656

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,8 +51,8 @@ services:
     #   - ./tmdata:/tendermint
     entrypoint: ''
     ports:
-      - "46656"
-      - "46657"
+      - "46656:46656"
+      - "46657:46657"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -51,8 +51,8 @@ services:
     #   - ./tmdata:/tendermint
     entrypoint: ''
     ports:
-      - "46656"
-      - "46657"
+      - "46656:46656"
+      - "46657:46657"
```
ttmc: I though port 46657 was only used internally for RPC traffic. I don't understand why it needs to be exposed externally.

vrde: It might be useful for debugging the system without logging in the running container.

ttmc: Maybe we should add a big disclaimer somewhere to tell people not to use Docker Compose to deploy a production node. I'll make a note to do that. Meanwhile I guess this PR can be merged.

PR #2300, opened 2018-05-17
Problem: No automated way to set up a dev/test network

muawiakh:
The existing automated deployment methods to deploy BigchainDB dev/test instance or networks were broken i.e. Vagrant, Ansible, Stack script. All of these scripts were not updated for BigchainDB `2.0.0a5`.

## Solution
- Update all the scripts to handle `BigchainDB 2.0.0a5`.
- Enhance `stack.sh` to support
  - `local` -> Virtual Machine based deployment
  - `docker` -> docker based network deployment
  - `azure` -> docker based network deployment on a Cloud
- Cleanup BigchainDB driver and docker-compose roles from Ansible(not needed).


ttmc:
Wow, a big PR.

Remember to update the relevant docs. Currently, they are in the Appendices. We should probably reorganize all the docs about ways-to-deploy (e.g. to make these ones easier to find), but that can be done in a future PR.

muawiakh:
@ttmc yes, docs will be part of another PR.

Dockerfile-alpine
```diff
@@ -0,0 +1,30 @@
...
+RUN mkdir -p /usr/src/app
+COPY . /usr/src/app/
+WORKDIR /usr/src/app
+RUN apk --update add sudo \
+    && apk --update add python3 openssl ca-certificates git\
+    && apk --update add --virtual build-dependencies python3-dev \
+        libffi-dev openssl-dev build-base \
+    && apk add --no-cache libstdc++ \
+    && pip3 install --upgrade pip cffi \
+    && pip install --no-cache-dir --process-dependency-links -e .[dev] \
```
kansi: :+1:

docs/contributing/source/dev-setup-coding-and-contribution-process/run-node-with-docker.md
```diff
@@ -32,10 +32,10 @@ docker run \
   --env BIGCHAINDB_DATABASE_HOST=172.17.0.1 \
   bigchaindb/bigchaindb \
   -y configure \
-  [mongodb|rethinkdb]
+  localmongodb
```
kansi: Correct me if I am wrong but I don't think `localmongodb` is needed when `configure`ing with `-y`, as the default backend is set to `localmongodb`

pkg/configuration/roles/bigchaindb/defaults/main.yml
```diff
@@ -1,31 +1,22 @@
 ---
 dependencies_deb:
   - g++
-  - python3-dev
   - libffi-dev
   - build-essential
   - libssl-dev
-  - python3-pip
```
kansi: The instructions https://docs.bigchaindb.com/projects/server/en/master/simple-network-setup.html#install-bigchaindb-server are different.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-node-with-docker.md
muawiakh: yes not necessary. Just kept it for clarity.

pkg/configuration/roles/tendermint/files/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG tm_version=0.19.2
```
kansi: Shouldn't the version be updated to `0.19.3` ?

muawiakh: yes

pkg/configuration/roles/bigchaindb/defaults/main.yml
muawiakh: Removed this because, I install py36 and pip separately, in the `py36` role, under `pkg/configuration/roles/py36`

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
```diff
@@ -0,0 +1,296 @@
...
+$ export STACK_SIZE=4
+
+#Optional, since docker is the default type.
+$ export STACK_TYPE=docker
+
+#Optional, codebase to use for the network deployment
+$ export STACK_BRANCH=master
+
+#Optional, since 0.19.2 is the default tendermint version.
+$ export TM_VERSION=0.19.2
```
kansi: `0.19.3`, Might need to grep and replace it

ttmc:
@muawiakh Sphinx was throwing some warnings so the Contributing docs wouldn't build for me (locally). I made two small commits to make those warnings go away, and pushed them here. The git commit message for each one says the problem it resolved.

ttmc:
@muawiakh Maybe it would be okay if we just deleted the entire page about how to run a local node with Docker (i.e. `run-node-with-docker.md`). Why?

- The Docker Compose way is much more user-friendly, especially using `make start`, `make stop` and `make reset`. It's still using Docker, but the complexity is hidden.
- It currently uses the latest bigchaindb/bigchaindb image from Docker Hub, i.e. _not_ the latest local code, so as those instructions stand, they won't help someone who is doing local dev/test: any changes they make won't be reflected in changed behavior.
- It would save us from having to keep that page up-to-date!
- It would reduce the number of options for new folks, so they don't have to spend time figuring out which way to run a local node.

pkg/scripts/stack.sh
```diff
@@ -1,124 +1,234 @@
...
-TOP_DIR=$(cd $(dirname "$0") && pwd)
-BASE_DIR=${TOP_DIR}/../..
-
-# Check for uninitialized variables, a big cause of bugs
+# defaults
+stack_branch=${STACK_BRANCH:="master"}
+stack_size=${STACK_SIZE:=4}
+stack_type=${STACK_TYPE:="docker"}
+stack_type_provider=${STACK_TYPE_PROVIDER:=""}
+tm_version=${TM_VERSION:="0.19.2"}
```
ttmc: Maybe just do a project-wide search for "0.19.2" and change it to 0.19.3.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
```diff
@@ -0,0 +1,296 @@
...
+- Fedora >= 24
+- MacOSX
+
+## Minimum Requirements
+Minimum resource requirements for a single node BigchainDB dev setup. **The more the better**:
+- Memory >= 512MB
+- VCPUs >= 1
+
+## Download the scripts
+```text
```
ttmc: These are instructions for developers, some of whom may be working on BigchainDB Server code, so they may already have local recent versions of `stack.sh` and `unstack.sh`. Maybe add a sentence before these instructions saying:

> If you're working on BigchainDB Server code, on a branch based on recent code, then you already have local recent versions of `stack.sh` and `unstack.sh` in your `bigchaindb/pkg/scripts/` directory. Otherwise you can get them using:

Also, you might help them by saying they can use `master` in place of `${GIT_BRANCH}` to get the scripts from the current master branch.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
```diff
@@ -0,0 +1,296 @@
...
+$ curl -fOL https://raw.githubusercontent.com/bigchaindb/bigchaindb/${GIT_BRANCH}/pkg/scripts/stack.sh
+
+# Optional
+$ curl -fOL https://raw.githubusercontent.com/bigchaindb/bigchaindb/${GIT_BRANCH}/pkg/scripts/unstack.sh
+```
+
+## Quick Start
+If you run `stack.sh` out of the box i.e. without any configuration changes, you will be able to deploy a 4 node
+BigchainDB network with Docker containers.
```
ttmc: Maybe you should add that this is going to deploy a network where all four nodes are `Using bigchaindb branch 'master'` because if they're devs, they probably want to run _their_ code, not code from the `master` branch.

ttmc:
Here's what happened when I ran `stack.sh` on my Ubuntu 18.04 machine (with the latest Docker and Docker Compose):
```text
$ bash stack.sh
Capturing output to logs/install-20180528-192554.log
Installation started at 2018-05-28 19:25:54
Using bigchaindb branch 'master'
Cloning into 'bigchaindb'...
######...#### 100.0%
Configuring Dockers locally!
Unknown option: --operation
Installation finished at 2018-05-28 19:27:27
```
That didn't seem like success. I did a `docker ps` and no containers were listed.

The log line `Unknown option: --operation` hints at what might be wrong. `--operation` is supposed to be an option for `bootstrap.sh` and the code for `bootstrap.sh` seems to define what it means, but I'm not a Bash wizard, and I don't see what's wrong.

pkg/scripts/stack.sh
muawiakh: I changed it to `0.19.2` but Tendermint has not tagged a `0.19.3` image on Docker Hub, so I reverted back to `0.19.2`.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
muawiakh: I will add a note along these lines. Default is master.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+## Quick Start
+If you run `stack.sh` out of the box i.e. without any configuration changes, you will be able to deploy a 4 node
+BigchainDB network with Docker containers. 
+
+**Note**: Run `stack.sh` with either root or non-root user with sudo enabled.
+
+```text
+$ bash stack.sh
```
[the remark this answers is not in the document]
muawiakh: Will update the permissions for all the scripts to be executable. Good catch.

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
muawiakh: I will, but if they are devs, `stack.sh` will clone a master, they can make changes in the cloned project and they can be reflected since we are using a `-e` flag while installing BigchainDB or they can simply `unstack.sh` and `stack.sh`

muawiakh:
@ttmc on your comment regarding removing the `run-node-with-docker`, I understand that `docker-compose` hides the complexity of running an environment with the `make` wrapper, but I kept it so people can know what we are doing in the back end. I can remove it, if this seems like an overkill. 

About your latest comment about trying out the change, it will not work out of the box, why? Because these changes are not merged yet, hence when you clone `stack.sh`, it pulls the latest master which does not have the updated `bootstrap.sh` script which parses the `--operation` flag. 

One way to test it could be that:
```
$ git clone git@github.com:muawiakh/bigchaindb.git
$ cd bigchaindb/
$ git checkout update-dev-test-deploy
$ cd pkg/scripts/
$ # In that directory
$ git clone git@github.com:muawiakh/bigchaindb.git
$ cd bigchaindb/
$ git checkout update-dev-test-deploy
$ cd ..
$ bash stack.sh
$ # Verify
$ docker ps
$ bash unstack.sh
OR 

$ git clone git@github.com:bigchaindb/bigchaindb.git
$ cd bigchaindb/
$ git checkout -b muawiakh-update-dev-test-deploy master
$ git pull https://github.com/muawiakh/bigchaindb.git update-dev-test-deploy
$ # MERGE BRANCH
$ cd pkg/scripts/
$ git clone git@github.com:bigchaindb/bigchaindb.git
$ cd bigchaindb/
$ git checkout -b muawiakh-update-dev-test-deploy master
$ git pull https://github.com/muawiakh/bigchaindb.git update-dev-test-deploy
$ #MERGE BRANCH
$ cd ..
$ bash stack.sh
$ # Verify
$ docker ps
$ bash unstack.sh
```

docs/contributing/source/dev-setup-coding-and-contribution-process/run-dev-network-stack.md
ttmc: I wouldn't have guessed that as the workflow for devs. Please explain that in the docs.

ttmc:
@muawiakh Yes, let's remove `run-node-with-docker`. If people are curious how we get Docker working, they can look inside `Makefile`, `docker-compose.yml` and the Docker-related commands used (ultimately) by `stack.sh` and Ansible.

ttmc:
@muawiakh I tried the two approaches you suggested in your comment https://github.com/bigchaindb/bigchaindb/pull/2300#issuecomment-392625934 above. Both failed starting with a log ling that said `fatal: destination path 'bigchaindb' already exists and is not an empty directory.`

I have to move on to doing other things now for a while.

muawiakh:
@ttmc you can pull the latest changes now, it should work. Tested with a fresh `Ubuntu 18.04`. Make sure you have Docker already installed, even though the playbook installs it but with `18.04` but `docker-ce stable` is not officially released yet only `test` is available which can be installed using something like this:
```
curl -fsSL test.docker.com | sh
```

manolodewiner:
I have tested using this commands: https://github.com/bigchaindb/bigchaindb/pull/2300#issuecomment-392625934 and it works properly.

PR #2304, opened 2018-05-17
Problem: Current logging server keeps running in some tear down scenarios

shahbazn:
Solution: Remove the current pub/sub based logging infra as this model is no longer necessary. This can now be replaced by a simple system level logging config.


shahbazn:
updated the PR to remove relevant docs files as well.

kansi:
There are some logger config parameters in [`tests/log/test_setup.py`](https://github.com/bigchaindb/bigchaindb/blob/master/tests/log/test_setup.py)

shahbazn:
@kansi thanks for the review. removed irrelevant and inactive tests

kansi:
@shahbazn I would like to understand how to enable console debug logs like [these](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/tendermint/core.py#L58) i.e. `bigchaindb -l DEBUG start` doesn't seem to print the logs on the console anymore.

shahbazn:
@kansi good catch fixed the bug.

PR #2309, opened 2018-05-24
Problem: Docs page on using Docker out-of-date

ttmc:
Solution: Point readers to the page titled, “Notes on Running a Local Dev Node with Docker Compose” in the BigchainDB docs about Contributing to BigchainDB

I realize that will have them running the latest BigchainDB code, but I guess that should be okay.


ttmc:
@muawiakh Your pull request #2300 updates the existing page about running BigchainDB with Docker, so I'll close this pull request.

PR #2310, opened 2018-05-25
Problem: It is cumbersome to build Tendermint cluster.

ldmberman:
Solution: Add helpful scripts for running a cluster.

This is a very minimal update. I am happy to further adjust it as you consider helpful.


muawiakh:
Is this strictly to run Tendermint cluster? with any app or BigchainDB?

ldmberman:
@muawiakh yep, just Tendermint, with [the key-value store app from the examples](https://github.com/tendermint/abci/blob/master/example/kvstore/persistent_kvstore.go). With it you can upsert validators by posting transactions.

ldmberman:
Makes sense. Eventually, I'll find a better place for them.

PR #2312, opened 2018-05-26
Copyedit simple-network-setup completeness & grammar

ttmc:
The simple-network-setup page is being used a lot, so I did some copyediting:

- Added a few notes at the top to answer some common questions and to steer people away from setting up their own network if they don't need to.
- Make the instructions _complete_, e.g. so that if someone goes through them, they can end up with a running node. The main addition was some instructions about how to start MongoDB. (The recommended way is to start everything with startup scripts, but we don't provide any for BigchainDB or Tendermint yet, so we should throw our users a life-raft in the meantime, i.e. give them _some_ way to get the node running.)
- Added examples of a `node_id` and a hostname.
- Grammar edits
- Stylistic edits, e.g. all titles in Title Case, blank lines after titles

### DIFF
--- a/docs/server/source/simple-network-setup.md
+++ b/docs/server/source/simple-network-setup.md
@@ -1,16 +1,26 @@
-# How to setup a BigchainDB Network
+# How to Set Up a BigchainDB Network
+
+Note 1: These instructions will also work for a "network" with only one node.
+
+Note 2: You might not need to set up your own network yet. You should start by creating a proof-of-concept app that writes to [the BigchainDB Testnet](https://testnet.bigchaindb.com/), and if that goes well, then you can look into setting up your own network. 
+
+Note 3: If you want to set up a node or network so that you can contribute to developing and testing the BigchainDB code, then see [the docs about contributing to BigchainDB](https://docs.bigchaindb.com/projects/contributing/en/latest/index.html).
+
+<hr>
+
 The process to create a network is both *social* and *technical*: *social* because someone (that we will call **Coordinator**) needs to find at least three other **Members** willing to join the network, and coordinate the effort; *technical* because each member of the network needs to set up a machine running BigchainDB. (Note: a Coordinator is a Member as well.)
 
-A **BigchainDB Network** (*Network* henceforth) is a set of **4 or more BigchainDB Nodes** (*Node* henceforth). Every Node is independently managed by a Member, and runs an instance of the [BigchainDB Server software][bdb:software]. At the **Genesis** of a Network, there **MUST** be at least **4** Nodes ready to connect. After the Genesis, a Network can dynamically add new Nodes or remove old Nodes.
+A **BigchainDB Network** (or just *Network*) is a set of **4 or more BigchainDB Nodes** (or *Nodes*). Every Node is independently managed by a Member, and runs an instance of the [BigchainDB Server software][bdb:software]. At the **Genesis** of a Network, there **MUST** be at least **4** Nodes ready to connect. After the Genesis, a Network can dynamically add new Nodes or remove old Nodes.
 
-A Network will stop working if more than one third of the Nodes are down. The bigger a Network, the more failures it can handle. A Network of size 4 can tolerate only 1 failure, so if 3 out of 4 Nodes are online, everything will work as expected. Eventually, the Node that was offline will automatically sync with the others.
+A Network will stop working if more than one third of the Nodes are down or faulty _in any way_. The bigger a Network, the more failures it can handle. A Network of size 4 can tolerate only 1 failure, so if 3 out of 4 Nodes are online, everything will work as expected. Eventually, the Node that was offline will automatically sync with the others.
 
+## Before We Start
 
-## Before we start
-This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine. The commands are tailored for an up-to-date *Debian-like* distribution (we use an **Ubuntu 18.04 LTS** Virtual Machine on Microsoft Azure): if you are on a different distribution you might need to adapt the names of the packages installed.
+This tutorial assumes you have basic knowledge on how to manage a GNU/Linux machine. The commands are tailored for an up-to-date *Debian-like* distribution. (We use an **Ubuntu 18.04 LTS** Virtual Machine on Microsoft Azure.) If you are on a different Linux distribution then you might need to adapt the names of the packages installed.
 
-We don't make any assumptions on **where** you run the Node.
-You can run the BigchainDB Server on a Virtual Machine on the cloud, on a machine in your data center, on a Raspberry Pi. Just make sure that your Node is reachable by the other Nodes, a **non-exhaustive list of examples**:
+We don't make any assumptions about **where** you run the Node.
+You can run BigchainDB Server on a Virtual Machine on the cloud, on a machine in your data center, or even on a Raspberry Pi. Just make sure that your Node is reachable by the other Nodes. Here's a **non-exhaustive list of examples**:
+
 - **good**: all Nodes running in the cloud using public IPs.
 - **bad**: some Nodes running in the cloud using public IPs, some Nodes in a private network.
 - **good**: all Nodes running in a private network.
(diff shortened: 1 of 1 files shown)


docs/server/source/simple-network-setup.md
```diff
@@ -194,14 +217,37 @@ persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
...
+
+The _recommended_ approach is to reboot the machine and let the MongoDB startup script start MongoDB. Don't do that yet though…
+
+To start BigchainDB, one uses the command `bigchaindb start` but that will run it in the foreground. If you want to run it in the background (so it will continue running after you logout), you can use `nohup`, `tmux`, or `screen`. For example, `nohup bigchaindb start 2>&1 > bigchaindb.log &`
+
+The _recommended_ approach is to create a startup script for BigchainDB, so it will start right after the boot of the operating system. (As mentioned earlier, MongoDB should already have a startup script.)
+
+To start Tendermint, one uses the command `tendermint node` but that will run it in the foreground. If you want to run it in the background (so it will continue running after you logout), you can use `nohup`, `tmux`, or `screen`. For example, `nohup tendermint node 2>&1 > tendermint.log &`
+
+The _recommended_ approach is to create a startup script for BigchainDB, so it will start right after the boot of the operating system. (As mentioned earlier, MongoDB should already have a startup script.)
```
kansi: This seems to repeated, any particular reason its this way?

ttmc: Oh, good catch. It should say "...for Tendermint, so it will start..." I will fix that.

docs/server/source/simple-network-setup.md
```diff
@@ -194,14 +217,37 @@ persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
...
+
+If you followed the recommended approach and created startup scripts for BigchainDB and Tendermint, then you can reboot the machine now. MongoDB, BigchainDB and Tendermint should all start.
+
+## How Others Can Access Your Node
+
+If you followed the above instructions, then your node should be publicly-accessible with BigchainDB Root URL `http://hostname:9984` (where hostname is something like `bdb7.canada.vmsareus.net` or `17.122.200.76`). That is, anyone can interact with your node using the [BigchainDB HTTP API](http-client-server-api.html) exposed at that address. The most common way to do that is to use one of the [BigchainDB Drivers](./drivers-clients/index.html).
+
+## Refreshing Your Node
+
+If you want to refresh your node back to a fresh empty state, then your best bet is to terminate it and deploy a new virtual machine, but if that's not an option, then you can:
```
vrde: Good idea to give additional information on how to reset a node, but trashing a virtual machine to start with an empty state is too much IMO.

What about keeping just the paragraph below? You can mention that if you want to reset your node to join a new network you can also remove the `.tendermint` directory.

ttmc: For people who don't fully-understand what is going on, and where all the data is stored, trashing the virtual machine and starting a new one is a completely reasonable and believable way to get back to a fresh state, so I'm not going to delete that sentence.

docs/server/source/simple-network-setup.md
```diff
@@ -194,14 +217,37 @@ persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
...
 ```
 
-Now the Member must run two commands:
-- `bigchaindb start`
-- `tendermint node`
+## Member: Start MongoDB, BigchainDB and Tendermint
+
+To start MongoDB, one uses the command `mongod` but that will run it in the foreground. If you want to run it in the background (so it will continue running after you logout), you can use `mongod --fork --logpath /var/log/mongodb.log`. (You might have to create the `/var/log` directory if it doesn't already exist.)
+
+The _recommended_ approach is to reboot the machine and let the MongoDB startup script start MongoDB. Don't do that yet though…
```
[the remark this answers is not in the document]
ttmc: Since when does `apt install anything` _start_ the thing in question? Is that common? I had no idea that that sometimes happens. It seems like overloading the word "install" to mean "install and start too".

I know that installing MongoDB from the official MongoDB PPA does _not_ also start MongoDB. Moreover, we've always recommended installing MongoDB from the official PPA in the past, because the MongoDB you get by using `apt install mongodb` can get behind the latest official MongoDB. Now I'm feeling even more wary of the package that the Ubuntu people built. What _else_ does their so-called "installer" do?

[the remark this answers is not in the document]
ttmc: Anyway, I guess I'll add something about how the so-called installer also starts MongoDB, and I'll remove what I wrote about how to start it. In the future, once there's an official MongoDB package for Ubuntu 18.04, I think we should go back to using that. Maybe I'll also file a bug report with Ubuntu: "BUG: Installer doesn't just install, it also starts MongoDB. Do not want.".

docs/server/source/simple-network-setup.md
```diff
@@ -194,14 +219,40 @@ persistent_peers = "<Member 1 node id>@<Member 1 hostname>:46656,\
...
+
+If MongoDB isn't running, then you can start it using the command `mongod`, but that will run it in the foreground. If you want to run it in the background (so it will continue running after you logout), you can use `mongod --fork --logpath /var/log/mongodb.log`. (You might have to create the `/var/log` directory if it doesn't already exist.)
+
+If you installed MongoDB using `sudo apt install mongodb`, then a MongoDB startup script should already be installed (so MongoDB will start automatically when the machine is restarted). Otherwise, you should install a startup script for MongoDB.
+
+To start BigchainDB, one uses the command `bigchaindb start` but that will run it in the foreground. If you want to run it in the background (so it will continue running after you logout), you can use `nohup`, `tmux`, or `screen`. For example, `nohup bigchaindb start 2>&1 > bigchaindb.log &`
+
+The _recommended_ approach is to create a startup script for BigchainDB, so it will start right after the boot of the operating system. (As mentioned earlier, MongoDB should already have a startup script.)
```
kansi: MonogDB should be running before excuting BigchainDB's startup script else the command crashes. It would be helpful to explicitly state this?

docs/server/source/simple-network-setup.md
```diff
@@ -60,15 +60,17 @@ sudo pip3 install bigchaindb==2.0.0a6
...
 Install a recent version of MongoDB. BigchainDB Server requires version 3.4 or newer.
 
 ```
 sudo apt install mongodb
 ```
 
-The package manager should take care of installing the startup script for MongoDB as well. (That means MongoDB will automatically start if the machine is restarted.)
+If you install MongoDB using the above command (which installs the `mongodb` package), it also configures MongoDB, starts MongoDB (in the background), and installs a MongoDB startup script (so that MongoDB will be started automatically when the machine is restarted).
+
+Note: The `mongodb` package is _not_ the official MongoDB package from MongoDB the company. If you want to install the official MongoDB package, please see [the MongoDB documentation](https://docs.mongodb.com/manual/installation/). Note that installing the official package _doesn't_ also start MongoDB.
```
vrde: I guess you added this paragraph because of the notes on the [MongoDB installation guide](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/#overview). Every GNU/Linux distribution has its maintainers that are not necessarily connected to the original project. I'm wondering what MongoDB is trying to communicate there, because to me this sounds a bit like FUD. Ideas?

ttmc: I probably read that paragraph in the MongoDB installation guide in the past, but I was also writing from experience. In the past, e.g. in our Quickstart page, we always told people to install the official MongoDB package from MongoDB the company, and there were extra instructions about how to create the data directory and how to start MongoDB. That's also what this other page _still_ says to do:

https://docs.bigchaindb.com/projects/contributing/en/latest/dev-setup-coding-and-contribution-process/run-node-as-processes.html#install-and-run-mongodb

As for the motives of MongoDB, I suspect they don't want to have to support MongoDB packages made by outsiders. That seems reasonable.

docs/server/source/simple-network-setup.md
ttmc: If they have followed the instructions up to this point, in order, then MongoDB should be running by now.

PR #2314, opened 2018-05-28
Problem: upsert-validator feature is error prone

kansi:
Solution: Implement [BEP-19](https://github.com/bigchaindb/BEPs/pull/45) to fix the behavour.

TODO:
- [x] Implement `UpsertValidator` election class
    - [x] Implement and test election proposal.
    - [x] Implement and test election vote.
    - [x] Implement and test election conclusion.

NOTE: Even though BEP-19 (and [BEP-18](https://github.com/bigchaindb/BEPs/pull/44)) aren't stable yet, this PR implements the general ideas suggested in both the BEPs.

### DIFF
--- a/bigchaindb/tep/transactional_election.py
+++ b/bigchaindb/tep/transactional_election.py
@@ -1,0 +1,38 @@
+
+class TransactionalElection(object):
+
+    version = '0.0'
+    name = 'base-class'
+
+    def __init__(self):
+        pass
+
+    def propose(self, election_object):
+        """Initiate a new election"""
+
+        raise NotImplementedError
+
+    def list(self, election_id):
+        """List the election with id `election_id`"""
+
+        raise NotImplementedError
+
+    def is_valid_proposal(self, election_object):
+        """Check if the `election_object` is a valid election"""
+
+        raise NotImplementedError
+
+    def vote(self, election_id, node_key_path):
+        """Vote on the given `election_id`"""
+
+        raise NotImplementedError
+
+    def is_valid_vote(self, vote):
+        """Validate the casted vote"""
+
+        raise NotImplementedError
+
+    def is_concluded(self, election_id, current_votes=[]):
+        """Check if the Election with `election_id` has reached consensus"""
+
+        raise NotImplementedError
--- a/bigchaindb/tep/transactional_election.py
+++ b/bigchaindb/tep/transactional_election.py
@@ -7,13 +7,13 @@ class TransactionalElection(object):
     def __init__(self):
         pass
 
-    def propose(self, election_object):
+    def propose(self, election_object, node_key_path):
         """Initiate a new election"""
 
         raise NotImplementedError
 
-    def list(self, election_id):
-        """List the election with id `election_id`"""
+    def show(self, election_id):
+        """Show the election with id `election_id`"""
 
         raise NotImplementedError
 
@@ -32,7 +32,7 @@ class TransactionalElection(object):
 
         raise NotImplementedError
 
-    def is_concluded(self, election_id, current_votes=[]):
+    def get_election_status(self, election_id, current_votes=[]):
         """Check if the Election with `election_id` has reached consensus"""
 
         raise NotImplementedError
--- a/bigchaindb/tep/transactional_election.py
+++ b/bigchaindb/tep/transactional_election.py
@@ -1,4 +1,6 @@
 
+
+# Base class for creating Transaction Election Process
 class TransactionalElection(object):
 
     version = '0.0'
--- a/bigchaindb/tep/upsert_validator.py
+++ b/bigchaindb/tep/upsert_validator.py
@@ -82,7 +82,7 @@ class UpsertValidator(TransactionalElection):
         return (current_topology.items() == voters.items())
 
     @classmethod
-    def election_id(cls, tx_election_id):
+    def public_key(cls, tx_election_id):
         return base58.b58encode(bytes.fromhex(tx_election_id))
 
     def _execute_action(self, tx):
@@ -140,7 +140,7 @@ class UpsertValidator(TransactionalElection):
         (tx_vote_input,
          voting_power) = self._get_vote(tx_election, node_key.public_key)
 
-        election_id = self.election_id(tx_election_id)
+        election_id = self.public_key(tx_election_id)
 
         # NOTE: Node will spend all its allocated votes
         tx_vote = Transaction.transfer(tx_vote_input,
@@ -193,18 +193,32 @@ class UpsertValidator(TransactionalElection):
         if not self._is_same_topology(current_validators, tx_election.outputs):
             return (False, 'inconsistent_topology')
 
+        election_public_key = self.public_key(tx_election.id)
         total_votes = sum(current_validators.values())
-        votes = 0
-        #  Aggregate vote count for `election_id`
+        casted_votes = self.bigchain.get_asset_outputs_filtered(tx_election.id,
+                                                                election_public_key,
+                                                                spent=False)
+        #  Aggregate vote count for `election_public_key`
+        vote_count = 0
+        for vote in casted_votes:
+            if vote.public_keys == [election_public_key]:
+                vote_count = vote_count + vote.amount
 
-        for output_id, output in enumerate(tx_election.outputs):
-            vote = self.bigchain.get_spent(tx_election_id, output_id, current_votes)
+        #
+        # THE FOLLOWING IS A REALLY CRITICAL ASSUMPTION
+        # NOTE: It is assumed that the votes cannot be double spends because
+        # `is_valid_transaction` method of BigchainDB class ensures that
+        #
 
-            if vote:
-                votes = votes + vote.outputs[0].amount
+        # sum the vote count from the current votes
+        for vote in current_votes:
+            if vote.asset['id'] == tx_election.id:
+                for voted_for in vote.outputs:
+                    if voted_for.public_keys == [election_public_key]:
+                        vote_count = vote_count + voted_for.amount
 
         # NOTE: Check if >2/3 of total votes have been casted
-        if votes > (2/3)*total_votes:
+        if vote_count > (2/3)*total_votes:
             return (True, tx_election.asset['data'])
 
         return (False, 'insufficient_votes')
--- a/bigchaindb/tep/upsert_validator.py
+++ b/bigchaindb/tep/upsert_validator.py
@@ -87,7 +87,7 @@ class UpsertValidator(TransactionalElection):
         return (True if code == 202 else False)
 
     def propose(self, validator_data, node_key_path):
-        validator_public_key = validator_data['pub_key']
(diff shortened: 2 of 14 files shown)


bigchaindb/tep/transactional_election.py
```diff
@@ -0,0 +1,38 @@
+
+class TransactionalElection(object):
+
+    version = '0.0'
```
ldmberman: How do we use version and name?

kansi: As of now it might not be used. But in "future" a class and extend the `UpsertValidator` class and upgrade the version and re-implement necessary methods to achieve desired behavour

ldmberman: A new class inherited from `UpsertValidator` will have a different name so they will be distinguishable.

Honestly, I do not see how these fields are useful.

bigchaindb/tep/transactional_election.py
```diff
@@ -0,0 +1,38 @@
...
+        """Initiate a new election"""
+
+        raise NotImplementedError
+
+    def list(self, election_id):
+        """List the election with id `election_id`"""
+
+        raise NotImplementedError
+
+    def is_valid_proposal(self, election_object):
```
ldmberman: If I understood it correctly, `is_valid_proposal` checks if the node wants to vote for the given proposal. Even if `is_valid_proposal` returns `False`, the proposal can be still voted for by the majority and the validator will be added. 

What do you think about renaming it to better reflect this? I can think of e. g. `agree_on_proposal`. The current name makes me think that the function checks for the structural validity of the proposal, like general transaction validity.

ldmberman: True, I misunderstood it at first, sorry.

bigchaindb/tep/transactional_election.py
kansi: Consider a situation wherein a version `2.0` for `UpsertValidator` has to be released because of major changes. In such a case to maintain backward compaitability and validate/replay elections which were proposed using version `1.0` format we would need to have different classes addressing the logic. The current approach tries to store this version information in `version` variable of the class. This hasn't been used anywhere but I hope it would be usefull when need be. Needless to say, if you think it can be addressed in a better way then please let me know.

bigchaindb/tep/upsert_validator.py
```diff
@@ -0,0 +1,181 @@
...
+               tx_election.output[i].public_keys == [public_key]:
+                return ([tx_input], tx_election.output[i].amount)
+
+        return ([], 0)
+
+    def _is_same_topology(current_topology, election_topology):
+
+        voters = {}
+        for voter in election_topology:
+            [public_key] = voter.public_keys
```
ldmberman: There is a special case of `len(voter.public_keys) > 1`. But then if we handle it here, a better name for the function might be `_is_valid_election_topology`.

bigchaindb/tep/transactional_election.py
ldmberman: Could you, please, provide more details about how do you think the version can be used? An artificial code sample can help.

kansi: I imagine it to be used something like this,

```python
class NewUpsertValidator(UpsertValidator):
    version = '2.0'
    def __init__(self):
...    
```

Then at other places in the code,

```python
[for uv in upsert_validator_versions if uv.version == election_txn.asset['data']['version']][0].vote(...) 
```

bigchaindb/tep/upsert_validator.py
kansi: Didn't understanding the comment, could you please re-phrase?

ldmberman: I mean, in case `len(voter.public_keys) > 1`, it makes sense return `False` here instead of raising an unpacking exception.

bigchaindb/tep/transactional_election.py
ttmc: @kansi Maybe you could add more information/explanation about "version" in BEP-20, i.e. https://github.com/bigchaindb/BEPs/pull/45

I noticed that the example there sets "version" to "1.0".

kansi: Added some details here https://github.com/bigchaindb/BEPs/pull/45/files#diff-8e56cd5a0d784d78d68efd53fa940687R52

ldmberman: Regarding versioning and transaction types, I expressed my concerns in the [election BEP](https://github.com/bigchaindb/BEPs/pull/44#issuecomment-393113609).

bigchaindb/tep/transactional_election.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __init__(self):
+        pass
+
+    def propose(self, election_object):
+        """Initiate a new election"""
+
+        raise NotImplementedError
+
+    def list(self, election_id):
```
ldmberman: You did not implement it.. Do we really need this method? What do you think about calling it "get"?

bigchaindb/tep/upsert_validator.py
```diff
@@ -0,0 +1,216 @@
...
+           value as the `voting_power`
+        """
+
+        validators = {}
+        for validator in self.bigchain.get_validators():
+            if validator['pub_key']['type'] == GO_AMINO_ED25519:
+                public_key = public_key_from_ed25519_key(key_from_base64(validator['pub_key']['value']))
+                validators[public_key] = validator['voting_power']
+            else:
+                # TODO: use appropriate exception
```
ldmberman: `UnsupportedPublicKeyType` may be?

bigchaindb/tep/upsert_validator.py
```diff
@@ -0,0 +1,216 @@
...
+        * If an election cannot be concluded because of topology changes then a new
+          election can be initiated.
+        """
+
+        tx_election = self.bigchain.get_transaction(tx_election_id)
+        current_validators = self._validators()
+
+        # NOTE: Check if network topology is exactly same to when the election was initiated
+        if not self._is_same_topology(current_validators, tx_election.outputs):
+            return (False, 'inconsistant_topology')
```
ldmberman: `is_concluded` suggests the function returns a boolean. I would either keep it simple and not return any statuses or rename it into something like `get_election_status`, does it make sense?

There is a typo - inconsist**e**nt_topology.

tests/conftest.py
```diff
@@ -626,3 +631,67 @@ def utxoset(dummy_unspent_outputs, utxo_collection):
...
+        'last_height': 0,
+        'last_round': 0,
+        'last_step': 0,
+        'priv_key': {
+            'type': '954568A3288910',
+            'value': private_key
+        }
+    }
+
+    with open('/tmp/priv_validator.json', 'w') as outfile:
```
ldmberman: Makes sense to use `tempfile.mkstemp` to better isolate test cases.

tests/conftest.py
```diff
@@ -626,3 +631,67 @@ def utxoset(dummy_unspent_outputs, utxo_collection):
...
+        json.dump(priv_validator, outfile)
+
+    return '/tmp/priv_validator.json'
+
+
+@pytest.fixture
+def network_validators(node_keys):
+    validator_pub_power = {}
+    i = 0
```
ldmberman: `i` is not used, right?

bigchaindb/tep/transactional_election.py
kansi: This method call will allow implementation for `bigchaindb upsert-validator show-request` which will be implemented in a seperate PR. @ttmc suggested that the sub-command should be called `show-request` so I think `show(self, election_id)` would be more appropriate?

ldmberman: Yes, I like `show` better than `list`.

bigchaindb/tep/upsert_validator.py
```diff
@@ -0,0 +1,216 @@
...
+            'public_key': validator_public_key,
+            'power': validator_power,
+            'node_id': validator_node_id
+        }
+
+        tx = Transaction.create([node_key.public_key],
+                                self._recipients(),
+                                asset=self._new_election_object(args))\
+                        .sign([node_key.private_key])
+        return self._execute_action(tx)
```
ldmberman: Can we make propose return actual error details? It should be quite helpful.

tests/tep/test_upsert_validator.py
```diff
@@ -0,0 +1,139 @@
...
+    voters = {}
+    for output in tx_election.outputs:
+        [output_public_key] = output.public_keys
+        voters[output_public_key] = output.amount
+
+    valid_asset = {
+        'type': 'election',
+        'name': 'upsert-validator',
+        'version': '1.0',
+        'args': new_validator
```
ldmberman: Makes sense to describe the validator anew or make a copy of it here, otherwise the test won't notice any undesired changes made to the `new_validator` object inside `propose`.

tests/conftest.py
ldmberman: Oops, sorry :)

ldmberman: You can alternatively do `voting_power.pop()`.

tests/tep/test_upsert_validator.py
kansi: I thought you were suggesting to create a `deepcopy` of `new_validator`? Its a little defensive but I made the change.

ldmberman: Sorry, that's not what I meant - copy it in the test, like:

```
'version': '1.0',
'args': {
    'public_key': ..
..
```

In the code I would not be defensive without a good reason - it can significantly affect the performance.

bigchaindb/tep/upsert_validator.py
ldmberman: In the PR you mentioned you are returning different error messages although not very specific yet. The HTTP API reports the error message it gets from `write_transaction` as well - why not returning the message here?

kansi: Pleae have a look [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/tendermint/lib.py#L67). The error response details contain information regarding the internal exceptions that Tendermint faced when executing the transaction i.e. whenever the json rpc api returns `500` the response is usually some internal exception within Tendermint with a sepecific error code in the json response. IMHO returning these internal errors to the admin might not be of much help but if you think it might be of some use then I can return the response.

ldmberman: I would return the error message from the start so that if we expose this functionality in the HTTP API or in the command line we do not have to do backwards-incompatible changes in the future when the error handling is improved.

kansi:
This PR is being closed in favor of an alternative approach discussed in [BEP-21](https://github.com/bigchaindb/BEPs/tree/master/21). The logic implemented in this PR has been re-implemented in the following PRs
- #2392 
- #2428
- #2445

Refer issue #2372 for more details.

PR #2318, opened 2018-05-29
Problem: Current spending transaction cause an exception

kansi:
Solution: Differentiate between stored spent txns and current spending txns


kansi:
@ttmc The transaction-in-waiting were being checked but the when returned the spending transaction the code didn't consider the current transactions differently and tried to build then using `Transactions.from_db` which would cause an exception.

PR #2321, opened 2018-05-30
Problem: API returns 404 for existing empty blocks

### LINKED ISSUE #2190: Report empty block instead of 404
Currently BighcainDB doesn't store empty blocks so whenever the blocks API is called a `404` is retuned. This can be confusing as it is not clear wheather the block was empty or doesn't exist. 

#### Soltuion
Return `200` with an empty block in response.

### LINKED ISSUE #2204: Problem: Empty blocks not being retrieved from BigchainDB
## Description

With the current deployment of the BigchainDB test network, if we query the blocks API e.g. 
```
$ curl -X GET https://test.bigchaindb.com/api/v1/blocks/2
{
  "message": "Not found",
  "status": 404
}
```

And if we query a valid transaction and its block height e.g.
```
$ curl -X GET https://test.bigchaindb.com/api/v1/blocks/?transaction_id=b3de778a417cc099ce6e36a3bfcd611701a6a0392626aced8857962ee697c620

[93850]

$ curl -X GET https://test.bigchaindb.com/api/v1/blocks/93850 | jq "."
{
  "height": 93850,
  "transactions": [
    {
      "inputs": [
        {
          "owners_before": [
            "4pzVMeZEox32agiQLH6cqWy34X1g9FuyA3joNH39keNm"
          ],
          "fulfills": null,
          "fulfillment": "pGSAIDjeXJjk_AmCH5PSe-J_jLlgR2_EZC-4fmkzMhEtvUYagUB7sr3-QuUcL914RMt0gphuFO1kZ1S9eyq4z64SJkQW9rWs4wztGywJjBowRgAyl1H-eGYxYlCZ8_9dCDwO1EkO"
        }
      ],
      "outputs": [
        {
          "public_keys": [
            "4pzVMeZEox32agiQLH6cqWy34X1g9FuyA3joNH39keNm"
          ],
          "condition": {
            "details": {
              "type": "ed25519-sha-256",
              "public_key": "4pzVMeZEox32agiQLH6cqWy34X1g9FuyA3joNH39keNm"
            },
            "uri": "ni:///sha-256;kovGWunC6vrpyRaLQkw3KQ7oV8oy1nWivyXoOhaql04?fpt=ed25519-sha-256&cost=131072"
          },
          "amount": "1"
        }
      ],
      "operation": "CREATE",
      "metadata": {
        "planet": "earth"
      },
      "asset": {
        "data": {
          "bicycle": {
            "serial_number": "abcd1234",
            "manufacturer": "bkfab"
          }
        }
      },
      "version": "2.0",
      "id": "b3de778a417cc099ce6e36a3bfcd611701a6a0392626aced8857962ee697c620"
    }
  ]
}
```

The reason we get a 404 in the first case is because block `2` does not have any transactions(empty) and BigchainDB does not return/returns 404 if a block is empty(without transactions), but tendermint creates those blocks i.e. if we query tendermint:
```
$ curl http://<tendermint-instance-name>:46657/block?height=2 | jq "."
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "block_meta": {
      "block_id": {
        "hash": "263365E6F66F7260B5A0288FDC2A56AE1F12530D",
        "parts": {
          "total": 1,
          "hash": "4609AE738BD88871353A1AC928F7F53800A61D89"
        }
      },
      "header": {
        "chain_id": "test-chain-l1h3rY",
        "height": 2,
        "time": "2018-04-09T12:06:45.526Z",
        "num_txs": 0,
        "last_block_id": {
          "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
          "parts": {
            "total": 1,
            "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
          }
        },
        "last_commit_hash": "0ED72DBC89EB375FE16777B0C7EAE8B55474B3F0",
        "data_hash": "",
        "validators_hash": "DED29B546877BF75F7C3243A5EE6AAD8502D6264",
        "app_hash": ""
      }
    },
    "block": {
      "header": {
        "chain_id": "test-chain-l1h3rY",
        "height": 2,
        "time": "2018-04-09T12:06:45.526Z",
        "num_txs": 0,
        "last_block_id": {
          "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
          "parts": {
            "total": 1,
            "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
          }
        },
        "last_commit_hash": "0ED72DBC89EB375FE16777B0C7EAE8B55474B3F0",
        "data_hash": "",
        "validators_hash": "DED29B546877BF75F7C3243A5EE6AAD8502D6264",
        "app_hash": ""
      },
      "data": {
        "txs": []
      },
      "last_commit": {
        "blockID": {
          "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
          "parts": {
            "total": 1,
            "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
          }
        },
        "precommits": [
          {
            "validator_address": "2AA866E6CFF15EFEFC6F0D09354B239A937B5BB1",
            "validator_index": 0,
            "height": 1,
            "round": 0,
            "type": 2,
            "block_id": {
              "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
              "parts": {
                "total": 1,
                "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
              }
            },
            "signature": {
              "type": "ed25519",
              "data": "9AC44DAE518919D80CB486873294227998F0DEB2EEC5316AADB60C95517CF4236AA68BAD39D9E0767D73836FF2CF83BD00A9FD3E2D5B99FF43D2FDEA8B60EF0A"
            }
          },
          {
            "validator_address": "3C5042EA7A63117A42B21C0D3F56A89C0D241296",
            "validator_index": 1,
            "height": 1,
            "round": 0,
            "type": 2,
            "block_id": {
              "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
              "parts": {
                "total": 1,
                "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
              }
            },
            "signature": {
              "type": "ed25519",
              "data": "E793A0BD427E5A0ABDDADBDEEF666CC50A0E00577260AF7B69E2F83D8D3A86F09C17BAAFBD3349F28EE1BE9F08784180867E3D4C85F9A25D8DBD60294D2BE50D"
            }
          },
          null,
          {
            "validator_address": "EFCF5346EB05899572BBC656E6C4F7F647CFD071",
            "validator_index": 3,
            "height": 1,
            "round": 0,
            "type": 2,
            "block_id": {
              "hash": "D234CF59D76588A6E313FB43B00340D58B715023",
              "parts": {
                "total": 1,
                "hash": "6384AD0DA94C8867EBD4AE31E02978E80A0B66A7"
              }
            },
            "signature": {
              "type": "ed25519",
              "data": "3FE0F4DFF6C5944C7FC19925A91CA913A944D72FA9C300D8B69FE0EDDBA162EC876E956E34F78E7A245570B740F33592E90716B003F44BB6CE9BA090C2449005"
            }
          }
        ]
      }
    }
  }
}
```

To keep a consistent state between what BigchainDB and Tendermint, run Tendermint with a configuration such that it does not create empty blocks and that can be done by setting `--consensus.create_empty_blocks` to `false`.

## Solution
Start tendermint node with the above mentioned flag in our `bigchaindb/tendermint` containers.
```
tendermint node --p2p.seeds="$seeds" --moniker="`hostname`" -proxy_app="tcp://$tm_proxy_app:$tm_abci_port" --consensus.create_empty_blocks=false --log_level debug
```

vrde:
Solution: Empty blocks are not store in MongoDB because Tendermint does not notify BigchainDB about them. In case a user requests a block that is not in our MongoDB, we check if the requested height is less than or equal to `latest_block_height`. If that's the case, we return an empty block. If the requested height is greater than `latest_block_height`, then we return 404.

Close #2190, Close #2204.

### DIFF
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -88,10 +88,6 @@ class BigchainDB(Bigchain):
     def _process_status_code(self, status_code, failure_msg):
         return (202, '') if status_code == 0 else (500, failure_msg)
 
-    def get_latest_block_height_from_tendermint(self):
-        r = requests.get(ENDPOINT + 'status')
-        return r.json()['result']['sync_info']['latest_block_height']
-
     def store_transaction(self, transaction):
         """Store a valid transaction to the transactions collection."""
 
@@ -284,8 +280,10 @@ class BigchainDB(Bigchain):
         """
 
         block = backend.query.get_block(self.connection, block_id)
+        latest_block = self.get_latest_block()
+        latest_block_height = latest_block['height'] if latest_block else 0
 
-        if not block and block_id > self.get_latest_block_height_from_tendermint():
+        if not block and block_id > latest_block_height:
             return
 
         result = {'height': block_id,
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -280,7 +280,7 @@ class BigchainDB(Bigchain):
         found.
 
         Args:
-            block_id (str): block id of the block to get
+            block_id (int): block id of the block to get.
         """
 
         block = backend.query.get_block(self.connection, block_id)
--- a/bigchaindb/tendermint/lib.py
+++ b/bigchaindb/tendermint/lib.py
@@ -90,7 +90,7 @@ class BigchainDB(Bigchain):
 
     def get_latest_block_height_from_tendermint(self):
         r = requests.get(ENDPOINT + 'status')
-        return r.json()['result']['latest_block_height']
+        return r.json()['result']['sync_info']['latest_block_height']
 
     def store_transaction(self, transaction):
         """Store a valid transaction to the transactions collection."""
@@ -273,41 +273,29 @@ class BigchainDB(Bigchain):
 
         return backend.query.get_latest_block(self.connection)
 
-    def get_block(self, block_id, include_status=False):
-        """Get the block with the specified `block_id` (and optionally its status)
+    def get_block(self, block_id):
+        """Get the block with the specified `block_id`.
 
         Returns the block corresponding to `block_id` or None if no match is
         found.
 
         Args:
             block_id (str): block id of the block to get
-            include_status (bool): also return the status of the block
-                       the return value is then a tuple: (block, status)
         """
-        # get block from database
-        if isinstance(block_id, str):
-            block_id = int(block_id)
 
         block = backend.query.get_block(self.connection, block_id)
+
+        if not block and block_id > self.get_latest_block_height_from_tendermint():
+            return
+
+        result = {'height': block_id,
+                  'transactions': []}
+
         if block:
             transactions = backend.query.get_transactions(self.connection, block['transactions'])
-            transactions = Transaction.from_db(self, transactions)
+            result['transactions'] = [t.to_dict() for t in Transaction.from_db(self, transactions)]
 
-            block = {'height': block['height'],
-                     'transactions': []}
-            block_txns = block['transactions']
-            for txn in transactions:
-                block_txns.append(txn.to_dict())
-
-        status = None
-        if include_status:
-            # NOTE: (In Tendermint) a block is an abstract entity which
-            # exists only after it has been validated
-            if block:
-                status = self.BLOCK_VALID
-            return block, status
-        else:
-            return block
+        return result
 
     def get_block_containing_tx(self, txid):
         """Retrieve the list of blocks (block ids) containing a
--- a/bigchaindb/web/routes.py
+++ b/bigchaindb/web/routes.py
@@ -29,7 +29,7 @@ ROUTES_API_V1 = [
     r('/', info.ApiV1Index),
     r('assets/', assets.AssetListApi),
     r('metadata/', metadata.MetadataApi),
-    r('blocks/<string:block_id>', blocks.BlockApi),
+    r('blocks/<int:block_id>', blocks.BlockApi),
     r('blocks/', blocks.BlockListApi),
     r('transactions/<string:tx_id>', tx.TransactionApi),
     r('transactions', tx.TransactionListApi),
--- a/tests/tendermint/test_lib.py
+++ b/tests/tendermint/test_lib.py
@@ -61,6 +61,13 @@ def test_get_latest_block(tb):
     assert block['height'] == 9
 
 
+@pytest.mark.bdb
+@patch('bigchaindb.backend.query.get_block', return_value=None)
+@patch('bigchaindb.tendermint.lib.BigchainDB.get_latest_block_height_from_tendermint', return_value=10)
+def test_get_empty_block(_0, _1, tb):
+    assert tb.get_block(5) == {'height': 5, 'transactions': []}
+
+
 def test_validation_error(b):
     from bigchaindb.models import Transaction
     from bigchaindb.common.crypto import generate_key_pair
--- a/tests/tendermint/test_lib.py
+++ b/tests/tendermint/test_lib.py
@@ -63,7 +63,7 @@ def test_get_latest_block(tb):
 
 @pytest.mark.bdb
 @patch('bigchaindb.backend.query.get_block', return_value=None)
-@patch('bigchaindb.tendermint.lib.BigchainDB.get_latest_block_height_from_tendermint', return_value=10)
+@patch('bigchaindb.tendermint.lib.BigchainDB.get_latest_block', return_value={'height': 10})
 def test_get_empty_block(_0, _1, tb):
     assert tb.get_block(5) == {'height': 5, 'transactions': []}
 


bigchaindb/tendermint/lib.py
```diff
@@ -90,7 +90,7 @@ def _process_status_code(self, status_code, failure_msg):
 
     def get_latest_block_height_from_tendermint(self):
         r = requests.get(ENDPOINT + 'status')
-        return r.json()['result']['latest_block_height']
+        return r.json()['result']['sync_info']['latest_block_height']
```
kansi: Does `'sync_info'` paramter hold the syncing state with network or the abci proxy app?
On my FDN node I see the following information,

`/abci_info`
```json
{
  "jsonrpc": "2.0",
  "id": "",
  "result": {
    "response": {
      "last_block_height": 135,
      "last_block_app_hash": "NmJjM2FiZjY0M2JhMGQwYjY0ODIwODk5N2UwZGVkYTY4MzYwMjFmODI3YTRlNmQyNTlmMThmNTVkMTM3MzgyOQ=="
    }
  }
}
```

`/status`

```json
    "sync_info": {
      "latest_block_hash": "0C37B9B3CCD6B243096B971F329342993DD4C29E",
      "latest_app_hash": "36626333616266363433626130643062363438323038393937653064656461363833363032316638323761346536643235396631386635356431333733383239",
      "latest_block_height": 136,
      "latest_block_time": "2018-05-30T14:13:49.76567914Z",
      "syncing": false
    }
```

bigchaindb/tendermint/lib.py
```diff
@@ -273,41 +273,29 @@ def get_latest_block(self):
...
-        # get block from database
-        if isinstance(block_id, str):
-            block_id = int(block_id)
 
         block = backend.query.get_block(self.connection, block_id)
+
+        if not block and block_id > self.get_latest_block_height_from_tendermint():
+            return
+
+        result = {'height': block_id,
```
ldmberman: I think, if the replay is still in progress, we can hit the following:

```
last_stored_block < block < last_tendermint_block
```

That's why we should check for the last block stored in MongoDB.

bigchaindb/tendermint/lib.py
```diff
@@ -273,41 +273,29 @@ def get_latest_block(self):
...
-    def get_block(self, block_id, include_status=False):
-        """Get the block with the specified `block_id` (and optionally its status)
+    def get_block(self, block_id):
+        """Get the block with the specified `block_id`.
 
         Returns the block corresponding to `block_id` or None if no match is
         found.
 
         Args:
             block_id (str): block id of the block to get
```
ldmberman: It's an integer.

bigchaindb/tendermint/lib.py
vrde: Good point, I'm not sure if we should use `/abci_info` or `status`, will check now. But we cannot use `bigchain.get_latest_block()` because it checks our local MongoDB, that does not contain empty blocks.

bigchaindb/tendermint/lib.py
vrde: I think we should disable the HTTP API while BigchainDB is syncing. Tendermint is not available while syncing, if you try to push a transaction the endpoint is not reachable.

Would this solve the issue?

ldmberman: I think, it would not. What if a new block arrives to Tendermint just in time so that the syncing has not yet begun while the RPC endpoint returns the new height?

Also, not accepting transactions "at times" is bad user experience apparently. If Tendermint is really not available while syncing, we can accept and store transactions ourselves.

bigchaindb/tendermint/lib.py
vrde: OK I'll go for `/status` :100: 

image

bigchaindb/tendermint/lib.py
kansi: > I think, it would not. What if a new block arrives to Tendermint just in time so that the syncing has not yet begun while the RPC endpoint returns the new height?

This seems like a valid concern, in a situation wherein 2/3 of the network has voted in new blocks, the height retrieved from Tendermint might be way into the future and fetching the block from MongoDB may return `None`.

> Also, not accepting transactions "at times" is bad user experience apparently. If Tendermint is really not available while syncing, we can accept and store transactions ourselves.
 
I don't necessarily agree with this i.e. the system shouldn't give the illusion of being in a working state when its actually not. (Tendermint follows this approach)

vrde: > I think, it would not. What if a new block arrives to Tendermint just in time so that the syncing has not yet begun while the RPC endpoint returns the new height?

OK now I understand, good point. I guess this is what @kansi highlighted as well. The only inconsistency that can happen is that BigchainDB latest height is one block behind Tendermint latest height. This happens if and only if the last block created by Tendermint is an empty one. I think it's not an issue.

> Also, not accepting transactions "at times" is bad user experience apparently. If Tendermint is really not available while syncing, we can accept and store transactions ourselves.

_It might be that we are using sync in different ways, for me sync happens when a node was offline for a while and needs to download the missing blocks from its peers, or when a new node is added to the network._

If a node is syncing then it does not have enough information to answer correctly to client's requests, so it should not be reachable. We can move the conversation to a specific issue.

ldmberman: It's not an illusion - it's an eventually consistent system. Unless you use the synchronous mode of posting transactions, you do not immediately see your transactions in the blockchain.

The fact Tendermint follows this approach does not imply the app has to do the same. On the contrary - they could have decided to do that to let the apps handle these problems.

ldmberman: > This happens if and only if the last block created by Tendermint is an empty one. I think it's not an issue.

And what if the last block was created by other nodes? It's still fine to return 404 as long as this node does not have the information about it.

> If a node is syncing then it does not have enough information to answer correctly to client's requests, so it should not be reachable

At `DeliverTx` the situation may change so we never have enough information until the transaction is committed. We may offer an interface saying smth like "ok, we can accept your transactions but we __do not__ promise anything about them apart from the minimal validity; later your can check their status" or we can say "come back later" what can be convenient for us but not necessarily convenient for the clients. But I agree - it's out of scope of this PR.

bigchaindb/tendermint/lib.py
```diff
@@ -273,41 +273,29 @@ def get_latest_block(self):
...
-                       the return value is then a tuple: (block, status)
+            block_id (int): block id of the block to get.
         """
-        # get block from database
-        if isinstance(block_id, str):
-            block_id = int(block_id)
 
         block = backend.query.get_block(self.connection, block_id)
+
+        if not block and block_id > self.get_latest_block():
```
kansi: `self.get_latest_block()['height']`?

bigchaindb/tendermint/lib.py
kansi: > It's not an illusion - it's an eventually consistent system

I think the discussion is going off topic but just to clarify Tendermint is a strongly consistent system with consensus guranteed at each block unlike bitcoin or ethereum. So this has nothing to do with consistency but with user experience as you pointed out earlier.

ldmberman: We kept different types of consistency in mind. Tendermint blocks, unlike bitcoin blocks, can not change when committed - in this sense Tendermint is strongly consistent. But it's eventually consistent in the conventional sense of a distributed system consistency - the nodes may disagree on what is the last block of the blockchain.

PR #2326, opened 2018-06-01
Problem: we don't have any tests for malicious strings

z-bowen:
## Solution

Use a parameterized test that fuzzes over a library of potentially hazardous strings. I'm using the 'big_list_of_naughty_strings', which includes some nice sql-injections, some crazy unicode sequences, and some invocations to elder gods... This up shot is, the work of maintaining the 'naughty strings' is off-loaded, and the test file generates individual tests for each string, so we should stay compliant if any new attack vectors open up.

## Issues Resolved

Contributing as part of the stability push for release of beta

## BEPs Implemented

NA

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -33,6 +33,5 @@ RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN pip install --no-cache-dir --process-dependency-links -e .[dev]
-RUN pip install blns
 RUN bigchaindb -y configure
 
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -33,5 +33,6 @@ RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN pip install --no-cache-dir --process-dependency-links -e .[dev]
-RUN bigchaindb -y configure "$backend"
+RUN pip install blns
+RUN bigchaindb -y configure
 
--- a/acceptance/python/src/test_naughty_strings.py
+++ b/acceptance/python/src/test_naughty_strings.py
@@ -1,0 +1,87 @@
+# ## Testing potentially hazardous strings
+# This test uses a library of `naughty` strings (code injections, weird unicode chars., etc.) as both keys and values.
+# We look for either a successful tx, or in the case that we use a naughty string as a key, and it violates some key
+# constraints, we expect to receive a well formatted error message.
+
+# ## Imports
+# We need some utils from the `os` package, we will interact with
+# env variables.
+import os
+
+# Since the naughty strings get encoded and decoded in odd ways, we'll use a regex to sweep those details under the rug.
+import re
+
+# We'll use a nice library of naughty strings...
+from blns import blns
+
+# And parameterize our test so each one is treated as a separate test case
+import pytest
+
+# For this test case we import and use the Python Driver.
+from bigchaindb_driver import BigchainDB
+from bigchaindb_driver.crypto import generate_keypair
+from bigchaindb_driver.exceptions import BadRequest
+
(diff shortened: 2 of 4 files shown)


Dockerfile-dev
```diff
@@ -33,5 +33,6 @@ RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN pip install --no-cache-dir --process-dependency-links -e .[dev]
-RUN bigchaindb -y configure "$backend"
+RUN pip install blns
```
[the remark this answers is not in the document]
z-bowen: I tested that, adding `blns` to both `dev_require` and `tests_require`, but in both cases, without the changes to `acceptance/python/Dockerfile`, I can't get the import to work in a new Docker image.

Dockerfile-dev
```diff
@@ -33,5 +33,5 @@ RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN pip install --no-cache-dir --process-dependency-links -e .[dev]
-RUN bigchaindb -y configure "$backend"
+RUN bigchaindb -y configure
```
ttmc: I generally follow the advice, "If it ain't broke, don't fix it." so I wouldn't have made this change or the corresponding one in `docker-compose.yml`. We might add support for other backends in the future.

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,87 @@
...
+        assert status_code == 400, status_code
+        assert re.fullmatch(regex, error), error
+    else:
+        raise TypeError(sent_transaction)
+
+
+@pytest.mark.parametrize("naughty_string", naughty_strings, ids=naughty_strings)
+def test_naughty_keys(naughty_string):
+
+    asset = {'data': {naughty_string: 'naughty_value'}}
```
ttmc: This is a minor quibble about semantics...

Here, the value is `'naughty_value'`, which is actually just a normal, non-strange string, a non-naughty value. In other words, `'naughty_value'` is not a naughty value, and this melts my brain a little bit.

treachery

Maybe replace `'naughty_value'` with `'normal_value'`

Similar comments can be made for the following line, and for the `'naughty_key'` in the `test_naughty_values()` function.

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,87 @@
...
+    # Then she checks that either her transaction was written to the block chain,
+    # or she got back a nice error message
+    if type(sent_transaction) == dict:
+        if 'id' in sent_transaction.keys():
+            tx_id = sent_transaction['id']
+            assert bdb.transactions.retrieve(tx_id), sent_transaction
+    elif type(sent_transaction) == BadRequest:
+        status_code = sent_transaction.status_code
+        error = sent_transaction.error
+        regex = '\{"message":"Invalid transaction \\(ValidationError\\): Invalid key name .* in asset object. The key name cannot contain characters .* or null characters","status":400\}\n'
```
ttmc: Also, a key can't start with `$`. I wonder why that case is not mentioned.

Dockerfile-dev
z-bowen: This does, indeed, break the test. Removing this one change causes the execution to fail while building the Docker images. I'm not totally clear on why this happens, but unless removing the `"$backend"` breaks something else, I think we should pull it.

acceptance/python/src/test_naughty_strings.py
z-bowen: Fixed

Dockerfile-dev
ttmc: Oh, interesting! I guess you should keep this change then.

kansi:
@z-bowen Why was this PR closed?

Issue #2254, opened 2018-05-02
Problem: `backlog_reassign_delay` config setting is defunct


ttmc:
Before BigchainDB 2.0, there was a "backlog" of transactions waiting to be put into blocks.

As of BigchainDB 2.0, the backlog is gone. Tendermint has something similar called the mempool, but configuring that is not a concern of BigchainDB Server.

Therefore we should remove the `backlog_reassign_delay` configuration setting, and update the associated docs, i.e. remove the section https://docs.bigchaindb.com/projects/server/en/master/server-reference/configuration.html#backlog-reassign-delay

ttmc:
@vrde I think maybe this configuration setting (`backlog_reassign_delay`) could be removed in the pull request that removes all the old pipelines code, since the backlog is tied up with that. (That might take more than one pull request!)

ttmc:
This issue was resolved by pull request #2332

PR #2332, opened 2018-06-03
Problem: backlog_reassign_delay is defunct but still listed

ttmc:
`backlog_reassign_delay` is a defunct BigchainDB configuration parameter, but it was being treated as if it was still in use.

I removed all code involving `backlog_reassign_delay` (or at least I think I did).

This PR resolves issue #2254


ttmc:
Note: I only did one commit (7c481e4) to create this branch. I don't know why GitHub thinks it should list those other commits. The changes in those other commits don't show up in the "Files changed."

PR #2334, opened 2018-06-04
Problem: we don't have any tests to cover malicious strings

z-bowen:
## Solution

Implemented a test that fuzzes over a library of malicious strings. The strings cover code injection, weird unicode characters like NUL, and also 'nasty' ascii characters like '.', '$', and '\'. Since we're using an external resource, keeping the list of naughty strings up to date should be handled automatically.

## Issues Resolved

Keeps us covered against code injections


Dockerfile-dev
```diff
@@ -33,5 +33,5 @@ RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/
 WORKDIR /usr/src/app
 RUN pip install --no-cache-dir --process-dependency-links -e .[dev]
-RUN bigchaindb -y configure "$backend"
+RUN bigchaindb -y configure
```
z-bowen: Rebuilding Docker images breaks if I leave in the `"$backend"`

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,93 @@
...
+        prepared_transaction,
+        private_keys=alice.private_key)
+
+    # The fulfilled tx gets sent to the BDB network
+    try:
+        sent_transaction = bdb.transactions.send(fulfilled_transaction)
+    except BadRequest as e:
+        sent_transaction = e
+
+    # If her key contained a '.', began with a '$', or contained a NUL character
```
z-bowen: @ttmc Made the conditions under which we look for an error more explicit.

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,93 @@
...
+        assert bdb.transactions.retrieve(tx_id), sent_transaction
+    # If neither condition was true, then something weird happened...
+    else:
+        raise TypeError(sent_transaction)
+
+
+@pytest.mark.parametrize("naughty_string", naughty_strings, ids=naughty_strings)
+def test_naughty_keys(naughty_string):
+
+    asset = {'data': {naughty_string: 'nice_value'}}
```
z-bowen: @ttmc Fixed the string names for the static keys/values

ttmc: Nice.

acceptance/python/src/test_naughty_strings.py
ttmc: :+1:

ttmc:
From my [earlier review](https://github.com/bigchaindb/bigchaindb/pull/2326#pullrequestreview-125519671):

> How does one run this test? Is that documented somewhere?

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,93 @@
...
+        sent_transaction = bdb.transactions.send(fulfilled_transaction)
+    except BadRequest as e:
+        sent_transaction = e
+
+    # If her key contained a '.', began with a '$', or contained a NUL character
+    regex = '.*\..*|\$.*|.*\x00.*'
+    key = next(iter(metadata))
+    if re.match(regex, key):
+        # Then she expects a nicely formatted error code
+        assert sent_transaction.status_code, sent_transaction
```
kansi: What is the purpose of this `assert`? There is another one below with `assert`s the `status_code`

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,93 @@
...
+        status_code = sent_transaction.status_code
+        assert sent_transaction.error, sent_transaction
+        error = sent_transaction.error
+        regex = '\{"message":"Invalid transaction \\(ValidationError\\): Invalid key name .* in asset object. ' \
+                'The key name cannot contain characters .* or null characters","status":400\}\n'
+        assert status_code == 400, status_code
```
kansi: `assert status_code == 400` should suffice? or am I missing something?

acceptance/python/src/test_naughty_strings.py
```diff
@@ -0,0 +1,93 @@
...
+        sent_transaction = e
+
+    # If her key contained a '.', began with a '$', or contained a NUL character
+    regex = '.*\..*|\$.*|.*\x00.*'
+    key = next(iter(metadata))
+    if re.match(regex, key):
+        # Then she expects a nicely formatted error code
+        assert sent_transaction.status_code, sent_transaction
+        status_code = sent_transaction.status_code
+        assert sent_transaction.error, sent_transaction
```
kansi: IMHO the regex `assert` below on the error message would be sufficient.

z-bowen: See my previous comment. Since I'm explicitly trying to return this error, I want as much information as possible for diagnostics if something is wrong with the error message I get back.

acceptance/python/src/test_naughty_strings.py
kansi: > I don't get a status code for some reason 

There are two situations here:
1. The HTTP request doesn't return, which in itself is an issue and cannot be caught by this assert
2. The HTTP api returns with a response that doesn't have an error! This would imply that the HTTP server is not written properly i.e. it doesn't follow the HTTP properly and I don't think we are trying to test these senarios here.

Also I don't understand why there is a `, sent_transaction` towards the end of each assert. What is that trying to achieve?

acceptance/python/src/test_naughty_strings.py
kansi: > I want as much information as possible for diagnostics

I don't think one can return more error information than what has been received in the HTTP response itself i.e. the regex would capture all there is to the error message.

Also I would like to understand the purpose of having `, sent_transaction` in the assertion when `sent_transaction.error` is being asserted?

acceptance/python/src/test_naughty_strings.py
z-bowen: Changed this to report the full `sent_transaction` in case of an error. I'm being a bit paranoid with this test, since it's meant to update dynamically to catch all bad strings, now and future. So I want a ton of verbosity in case of any unexpected behavior.

acceptance/python/src/test_naughty_strings.py
z-bowen: The `, sent_transaction` is there to explicitly include the tx object in the error message, so that I don't get an error message that looks like `AssertaionError: 400 != 200` and need to dig through a stack trace to see what failed. For diagnostics, I want to return the full object, so that whoever sees the test fail can understand what went wrong with the least amount of digging.

ttmc:
(I just fixed a minor merge conflict and that's where the above merge came from.)

ttmc:
Oops, sorry about removing `blns`! I see you fixed it, so now I will now squash and merge.

PR #2339, opened 2018-06-06
Problem: Stack installation on Centos 7 caused issues with some packages

muawiakh:
## Solution
- Add `EPEL` repo on CentOS
- Fix `home_path` for cloud based stack.


muawiakh:
Yes, thats how I found out it was an issue in the first place. Doesn't happen with the image that I used earlier for testing.

PR #2342, opened 2018-06-08
Problem: Tendermint configuration not present in BigchainDB config

### LINKED ISSUE #2192: Problem: The BigchainDB Root URL is returning some wrong values
Today, with BigchainDB 2.0 Alpha, when one sends an HTTP request to the BigchainDB Root URL, such as https://test.bigchaindb.com/ they get back a response like:
```json
{
   "api": {
      "v1": {
         "assets": "/api/v1/assets/",
         "docs": "https://docs.bigchaindb.com/projects/server/en/v2.0.0a1/http-client-server-api.html",
         "metadata": "/api/v1/metadata/",
         "outputs": "/api/v1/outputs/",
         "streams": "wss://test.bigchaindb.com:443/api/v1/streams/valid_transactions",
         "transactions": "/api/v1/transactions/",
         "validators": "/api/v1/validators"
      }
   },
   "docs": "https://docs.bigchaindb.com/projects/server/en/v2.0.0a1/",
   "keyring": [],
   "public_key": null,
   "software": "BigchainDB",
   "version": "2.0.0a1"
}
```

The problem is that "keyring" and "public_key" used to be the BigchainDB node's keyring and public key, but BigchainDB doesn't have an associated keyring or public key any more --- the equivalents are all maintained by Tendermint now.

**Solution Idea 1** - Remove the "keyring" and "public_key" from the HTTP response.

**Solution Idea 2** - Change "keyring" to "tendermint_validators" and return the same list as the `GET /api/v1/validators` endpoint. Change the "public_key" to return the "tendermint_public_key".

### LINKED ISSUE #2255: Problem: Unsure if we need new config settings for Tendermint
Do we need new BigchainDB configuration settings in BigchainDB 2.0? For example, do we need to add:

- a setting for the FQDN of Tendermint or
- a setting for the port of the Tendermint?

If so, we also need to update the docs about the BigchainDB Server configuration settings at https://docs.bigchaindb.com/projects/server/en/master/server-reference/configuration.html

muawiakh:
## Solution

Add Tendermint host, port to BigchainDB configuration

## Issues Resolved

Resolves #2255


ldmberman:
As far as I can see, the config is not taken into account when the Tendermint endpoint is buillt in `tendermint/lib.py`.

bigchaindb/__init__.py
```diff
@@ -99,6 +99,10 @@
         'advertised_host': 'localhost',
         'advertised_port': 9985,
     },
+    'tendermint': {
+        'host': 'localhost',
+        'port': 46657,
+    },
```
kansi: Where is this config being used?

bigchaindb/core.py
```diff
@@ -25,7 +25,7 @@ class Bigchain(object):
     TX_IN_BACKLOG = 'backlog'
     """return if transaction is in backlog"""
 
-    def __init__(self, connection=None):
+    def __init__(self, tendermint_host=None, tendermint_port=None, connection=None):
```
kansi: I don't think we need this, the user can use the config file `.bigchaindb` or the environment variables.

bigchaindb/core.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, connection=None):
         """
         config_utils.autoconfigure()
 
+        self.tendermint_host = tendermint_host or bigchaindb.config['tendermint']['host']
+        self.tendermint_port = tendermint_port or bigchaindb.config['tendermint']['port']
+        self.endpoint = 'http://{}:{}/'.format(self.tendermint_host, self.tendermint_port)
+        self.mode_list = ('broadcast_tx_async',
+                          'broadcast_tx_sync',
+                          'broadcast_tx_commit')
```
kansi: This code has been added to the old `core.py`. As a part of the cleanup, we decided not to add any code to the older files. So, it would be better to move this to `tendermint/lib.py`

ttmc:
I approved because GitHub was still showing my review status with the red X even though it was supposed to be showing the speech balloon to indicate that I'd made comments. It should still be the speech balloon, but I don't know how to make that happen without making a new comment. GitHub!

docs/server/source/server-reference/configuration.md
```diff
@@ -426,3 +428,27 @@ logging of the `core.py` module to be more verbose, you would set the
...
+
+**Example using environment variables**
+```text
+export BIGCHAINDB_TENDERMINT_HOST=tendermint
+export BIGCHAINDB_TENDERMINT_PORT=46657
+```
+
+**Default values**
+
+If (no environment variables were set and there's no local config file), or you used `bigchaindb -y configure localmongodb` to create a default local config file for a `localmongodb` backend, then the defaults will be:
```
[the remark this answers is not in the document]
ttmc: If you like, you could add a short line to every subsection about default values, saying something like, "To understand how the default value of a setting affects the run-time value of a setting, see the top of this page."

muawiakh:
Docs server are failing, will be fixed by #2360

PR #2346, opened 2018-06-11
Problem: No diagrams of transactions in docs

ttmc:
Solution: Put some simple diagrams of BigchainDB transactions in the docs

We used to have some diagrams showing BigchainDB transactions, inputs, outputs, etc. in the docs, but somehow they all went away without us noticing. Last week, someone asked me on Gitter if there is any standard pictoral way to illustrate transactions. There is, but there were no examples in the docs!

I put two simple diagrams in the **Transaction Concepts** section of the root docs. They're very simple diagrams, but I think that they're enough for someone to figure out how to diagram even the most complicated set of transactions.


amed83:
It looks fine to me.

PR #2347, opened 2018-06-12
Problem: The write-a-bep page/process is daunting

ttmc:
Solution: Make the BEP writing process easier and more approachable

- Give them options to float/propose an idea for a BEP, so they can get feedback on it before writing a full BEP.
- Note that there's now an "official" way to discuss existing BEPs.
- When it comes to writing the BEP, don't make them read BEP-1 and BEP-2 before doing anything. Suggest that they look at existing BEPs to get an idea of the expected structure.
- They must write the BEP before making a pull request, so they don't have to read BEP-1 (our process for handling pull requests) until after they write their BEP.

@vrde - I know you're on vacation but I put you as a reviewer since I know you care about this. You can review it when you get back.

### DIFF
--- a/docs/contributing/source/ways-to-contribute/write-a-bep.md
+++ b/docs/contributing/source/ways-to-contribute/write-a-bep.md
@@ -1,7 +1,14 @@
-# Write a BigchaindB Enhancement Proposal (BEP)
+# Write a BigchainDB Enhancement Proposal (BEP)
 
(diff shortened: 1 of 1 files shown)


ttmc:
I also switched to using terminology like "BEP-2" rather than "2/COSS," as suggested by issue bigchaindb/BEPs#32.

codegeschrei:
I like it! 
Just a tiny style change. Is it maybe better to have 'BEP' always written all caps? Labels are case sensitive so it's possible.

ttmc:
@codegeschrei approved this PR so I will merge it now.

PR #2349, opened 2018-06-13
Problem: There is no logging for benchmark stats

shahbazn:
Solution: Add a new log handler to capture benchmark stats to a separate file.

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -72,8 +72,7 @@ config = {
         'datefmt_logfile': log_config['formatters']['file']['datefmt'],
         'fmt_console': log_config['formatters']['console']['format'],
         'fmt_logfile': log_config['formatters']['file']['format'],
-        'granular_levels': {},
-        'benchmark': False
+        'granular_levels': {}
     },
 }
 
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -73,6 +73,7 @@ config = {
         'fmt_console': log_config['formatters']['console']['format'],
         'fmt_logfile': log_config['formatters']['file']['format'],
         'granular_levels': {},
+        'benchmark': False
     },
 }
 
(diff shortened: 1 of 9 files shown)


bigchaindb/__init__.py
```diff
@@ -73,6 +73,7 @@
         'fmt_console': log_config['formatters']['console']['format'],
         'fmt_logfile': log_config['formatters']['file']['format'],
         'granular_levels': {},
+        'benchmark': False
```
kansi: Instead of having a flag to enable benchmark it would be better to enable it via log level instead i.e. something as following

```
    "level_logfile": "benchmark",
```

bigchaindb/log.py
```diff
@@ -1,8 +1,14 @@
...
+import functools
 import logging
+
+from bigchaindb.common.exceptions import ConfigurationError
+from logging.config import dictConfig as set_logging_config
 from os.path import expanduser, join
 
 
 DEFAULT_LOG_DIR = expanduser('~')
+BENCHMARK_LOG_LEVEL = 60
```
kansi: After discussion with @shahbazn it seems that having `BENCHMARK_LOG_LEVEL=15` would be more suitable as the benchmark logs would not show up when a production system is started with levels, `INFO`, `WARNING`, `ERROR` or `CRITICAL`.

docs/server/source/server-reference/configuration.md
```diff
@@ -289,7 +289,7 @@ defined by [Python](https://docs.python.org/3.6/library/logging.html#levels),
 but case insensitive for convenience's sake:
 
 ```
-"critical", "error", "warning", "info", "debug", "notset"
+"critical", "error", "warning", "info", "debug", "notset", "benchmark"
```
ttmc: Shouldn't "benchmark" (logging level 15) be between "info" (logging level 20) and logging level "debug" (logging level 10)?

docs/server/source/server-reference/configuration.md
```diff
@@ -311,7 +311,7 @@ defined by [Python](https://docs.python.org/3.6/library/logging.html#levels),
...
```
ttmc: Same comment --- Shouldn't "benchmark" (logging level 15) be between "info" (logging level 20) and logging level "debug" (logging level 10)?

PR #2355, opened 2018-06-18
Problem: no codecoverage

codegeschrei:
Solution: enable codecov again

### DIFF
--- a/.ci/travis-after-success.sh
+++ b/.ci/travis-after-success.sh
@@ -3,5 +3,5 @@
 set -e -x
 
 if [[ -z ${TOXENV} ]]; then
-    codecov -v
+    codecov -v -f htmlcov/coverage.xml
 fi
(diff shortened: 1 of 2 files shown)


.ci/travis-after-success.sh
```diff
@@ -2,6 +2,6 @@
 
 set -e -x
 
-if [[ -z ${TOXENV} ]]; then
-    codecov -v
+if ! [[ -n ${TOXENV} ]] && ! [[ ${BIGCHAINDB_CI_ABCI} == 'enable' ]] && ! [[ ${BIGCHAINDB_ACCEPTANCE_TEST} == 'enable' ]]; then
```
ldmberman: The expression would be a little more readable if you avoid double negation by using `-z` instead of `-n` and `!=` instead of `==`.

codegeschrei: ah, I see! Thanks for the input

PR #2356, opened 2018-06-19
Problem: Upsert validator not marked experimental

kansi:
Solution: Update documentation to mark it experimental

### DIFF
--- a/docs/server/source/server-reference/bigchaindb-cli.md
+++ b/docs/server/source/server-reference/bigchaindb-cli.md
@@ -76,6 +76,12 @@ configuration file as documented under
 
 
 ## bigchaindb upsert-validator
+
+<div class="admonition warning">
+<p class="first admonition-title">Warning</p>
+<p class="last">This is an <strong>experimental</strong> feature. Users are advised <strong> not </strong> to use it in production.</p>
+</div>
+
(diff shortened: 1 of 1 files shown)


kansi:
I couldn't find a way to create the warning block in markdown file like the ones in rst files.

kansi:
We write the it in the bold but the rest of documentation has the `warning` blocks so I just ported that. Should I change it to bold?

ldmberman:
The "warning" class apparently does not do anything. +1 making it bold. I would also write it in uppercase.

kansi:
@ldmberman  The warning makes it look like follows:
screenshot-2018-6-19 command line interface cli bigchaindb server 2 0 0b1 documentation

ldmberman:
Ah, indeed! It just does not work on Github, but I guess it's fine for this doc.

vrde:
> We write the it in the bold but the rest of documentation has the warning blocks so I just ported that. Should I change it to bold?

My point is that the *warning* directive is a reStructuredText feature, it looks like this:
```
.. warning:: My warning text
```

If one day Read the Docs decides to change the style of the warnings, or changes the rendering engine for `rst`, the hardcoded HTML and the CSS classes we have in this `md` file might become obsolete or wrong.

I don't want to be picky on this simple PR, but I worry about consistency. If we have a Markdown file then I'd go for bold, otherwise we can move (for this file) from Markdown to reStructuredText and use the `.. warning::` directive.

kansi:
@vrde as per your recommendation I changed it to bold.

PR #2357, opened 2018-06-19
Problem: voting code is not used anymore

codegeschrei:
Solution: remove all voting related code


kansi:
There is vote spec in `common/schema/vote.yml` and `web/view/votes.py`

codegeschrei:
Thanks @kansi ! If this PR is important then maybe someone else want's to remove it. I'm off until next Wednesday. If not I'll remove it next week :)

bigchaindb/models.py
```diff
@@ -158,7 +158,7 @@ def from_db(cls, bigchain, tx_dict_list):
 
 # TODO: Remove node_pubkey as part of cleanup II
 class Block(object):
-    """Bundle a list of Transactions in a Block. Nodes vote on its validity.
```
kansi: This block was is deprecated in favour of [this](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/tendermint/lib.py#L374)

tests/test_block_model.py
```diff
@@ -2,6 +2,9 @@
 from pytest import raises
 
 
+pytestmark = pytest.mark.tendermint
```
kansi: I don't think we need to fix these tests for `Block` class it's going to be removed soon.

codegeschrei:
@z-bowen the comment is Tendermint related and has to stay :)

bigchaindb/fastquery.py
```diff
@@ -1,60 +0,0 @@
-from bigchaindb.utils import condition_details_has_owner
-from bigchaindb.backend import query
-from bigchaindb.common.transaction import TransactionLink
-
-
-class FastQuery:
```
kansi: This class doesn't have much to do with votes but If it is being removed in this PR it would be better to remove it from the tests too i.e. `tests/test_fastquery.py` and remove its mocking from `tests/db/test_bigchain_api.py`

codegeschrei: I think it's better to do it in a separate PR.
I'll remove it and also take care of the new PR

PR #2360, opened 2018-06-28
Problem: docs are not building anymore

codegeschrei:
Solution: add wget to the requirements for docs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ docs_require = [
     'sphinxcontrib-httpdomain>=1.5.0',
     'sphinxcontrib-napoleon>=0.4.4',
     'aafigure>=0.6',
+    'wget'
 ]
 
 tests_require = [


ldmberman:
@muawiakh @shahbazn looks like documentation builds started to fail after the base Docker image was updated (wget was excluded). Should we fix the image versions that we are using to avoid such incidents?

muawiakh:
I am not entirely sure when it was removed, also I don't entirely understand by what you mean by fix image versions that we are using? Installing as part of the docker image would be the ideal scenario IMO, not in the `setup.py`

codegeschrei:
oh, if that's the case I can change it

ldmberman:
> what you mean by fix image versions that we are using

I mean, in the `Dockerfile` put `FROM <os>:<paricular-version>`.

shahbazn:
>Installing wget inside the docker image would be the ideal scenario IMO, not in the setup.py
Agreed

codegeschrei:
We have the `wget` requirement in [here](https://github.com/bigchaindb/bigchaindb/blob/master/docs/contributing/requirements.txt) 
In all `docs` folders we have different requirements, which are also in `setup.py` except for the wget. So to me it makes sense to just add it to the `setup.py` instead of `Dockerfile`

shahbazn:
@ldmberman possible that `wget` installed as part of the image by default before i am not sure how it got broken. will merge this now

PR #2365, opened 2018-06-28
Problem: 2 implementations of fastquery exist.

ldmberman:
Solution: Remove the old deprecated implementation. Update the tests.


tests/db/test_bigchain_api.py
```diff
@@ -989,41 +989,51 @@ def test_get_owned_ids_calls_get_outputs_filtered():
     assert res == gof()
 
 
+@pytest.mark.tendermint
```
codegeschrei: I don't know if we need to mark them for tendermint yet. Maybe fixing them but still skipping them is better, @kansi ?

codegeschrei: And some parts of this file are also changed by @z-bowen s PR: https://github.com/bigchaindb/bigchaindb/pull/2366/files#diff-5dbcdadd05006f510ee75ac7dae5db47

ldmberman: It does not make sense to change tests that are not executed. Also, if the tests have value, why disable them?

codegeschrei: As far as I know, making tests 'work again' was supposed to be an extra step. I don't mind if we change them in here and make them work.

tests/db/test_bigchain_api.py
```diff
@@ -978,41 +978,51 @@ def test_get_owned_ids_calls_get_outputs_filtered():
...
 def test_get_outputs_filtered_only_unspent():
     from bigchaindb.common.transaction import TransactionLink
-    from bigchaindb.core import Bigchain
-    with patch('bigchaindb.fastquery.FastQuery.get_outputs_by_public_key') as get_outputs:
+    from bigchaindb.tendermint.lib import BigchainDB
+
+    go = 'bigchaindb.tendermint.fastquery.FastQuery.get_outputs_by_public_key'
```
muawiakh: :+1:, but not entirely sure why its named `go`.

ldmberman: A short of **g**et_**o**utputs :)

tests/db/test_bigchain_api.py
kansi: I concur with @ldmberman , we have been modifying tests in other PRs which aren't enabled. That doesn't give any insight into whether the test actually passes or not. Also, I really don't mind having more tests :smile:

tests/db/test_bigchain_api.py
muawiakh: Yes , figured it out, thats why updated. 👍

PR #2366, opened 2018-06-28
Problem: Bigchain class is unused and redundant

z-bowen:
## Solution

Remove Bigchain class and BigchainDB's dependency on it.


bigchaindb/models.py
```diff
@@ -208,8 +208,8 @@ def validate(self, bigchain):
         """Validate the Block.
 
         Args:
-            bigchain (:class:`~bigchaindb.Bigchain`): An instantiated Bigchain
-                object.
+            bigchain (:class:`~bigchaindb.tendermint.BigchainDB`): An
+                instantiated BigchainDB object.
```
codegeschrei: This whole class is going to be removed in https://github.com/bigchaindb/bigchaindb/pull/2368

bigchaindb/tendermint/lib.py
```diff
@@ -16,15 +16,16 @@
 
 import requests
 
-from bigchaindb import backend
-from bigchaindb import Bigchain
+import bigchaindb
```
codegeschrei: Do we need this import?

z-bowen: It was like that in the old `Bigchain` class, and it causes my local unit tests to fail if I change it. I agree that it's ugly, and we should fix it, but I think it's outside the scope of this PR.

tests/db/test_bigchain_api.py
```diff
@@ -981,36 +981,36 @@ def test_get_spent_multiple_owners(self, b, user_sk, user_pk, alice):
 
 
 def test_get_owned_ids_calls_get_outputs_filtered():
-    from bigchaindb.core import Bigchain
+    from bigchaindb.tendermint import BigchainDB
     with patch('bigchaindb.core.Bigchain.get_outputs_filtered') as gof:
```
codegeschrei: I think this needs to be changed too

codegeschrei:
For codecov I wouldn't mind merging it with a decrease of .3
We need to check the coverage later anyways

tests/db/test_bigchain_api.py
z-bowen: Yup. Good thing this test is deactivated! lol

bigchaindb/tendermint/lib.py
```diff
@@ -39,12 +40,55 @@
...
+    BLOCK_INVALID = 'invalid'
+    """return if a block has been voted invalid"""
+
+    BLOCK_VALID = TX_VALID = 'valid'
+    """return if a block is valid, or tx is in valid block"""
+
+    BLOCK_UNDECIDED = TX_UNDECIDED = 'undecided'
+    """return if block is undecided, or tx is in undecided block"""
+
+    TX_IN_BACKLOG = 'backlog'
```
kansi: The parameters `BLOCK_INVALID`, `BLOCK_VALID`, `BLOCK_UNDECIDED` , `TX_IN_BACKLOG` are not required anymore and should be removed.

muawiakh:
There are some conflicts in this change now, probably because of #2342

bigchaindb/tendermint/lib.py
z-bowen: It breaks a bunch of tests when I remove those parameters. I suggest we excavate them in a separate PR.

codegeschrei: I think this is a good idea. This PR is already big enough.

kansi:
I think the changes haven't been merged from the master carefully, i.e. please have a look at [this](https://github.com/bigchaindb/bigchaindb/pull/2366/files#diff-c6511560546a7dc577e7e647b5bfdace) vs [this](https://github.com/bigchaindb/bigchaindb/pull/2342/files#diff-c6511560546a7dc577e7e647b5bfdace)

bigchaindb/tendermint/lib.py
codegeschrei: I'm taking care of removing the parameters in a different PR

bigchaindb/tendermint/lib.py
```diff
@@ -14,38 +14,76 @@
...
 from bigchaindb.tendermint import fastquery
 from bigchaindb import exceptions as core_exceptions
+from bigchaindb.consensus import BaseConsensusRules
 
 logger = logging.getLogger(__name__)
 
 
-class BigchainDB(Bigchain):
+class BigchainDB(object):
+    """Bigchain API
```
ttmc: It's called Bigchain (instead of BigchainDB) in some places because the project began its life as "Bigchain". The "DB" was added a couple of months after the start of the project (i.e. over two years ago). It would be okay to change "Bigchain" to "BigchainDB".

PR #2368, opened 2018-06-29
 Problem: The Block class is deprecated

codegeschrei:
Solution: remove the block class and all occurrences


tests/test_fastquery.py
```diff
@@ -1,26 +1,10 @@
 import pytest
 
 from bigchaindb.common.transaction import TransactionLink
-from bigchaindb.models import Block, Transaction
```
ldmberman: I deleted the file in the [fastquery cleanup](https://github.com/bigchaindb/bigchaindb/pull/2365). Can you exclude it from here so that we avoid some git conflicts?

codegeschrei: I can merge your PR. That's the only solution. I needed to remove it because otherwise my tests and linter were failing.

ldmberman: Ah, did not get that, sorry. We can resolve the conflict then.

codegeschrei: I think it depends when we merge the PRs
If your PR is merged before mine, I can just merge master into mine. If mine is merged before it should actually not be a conflict (I think)

PR #2374, opened 2018-07-03
 Problem: Block parameters are not required anymore

codegeschrei:
Solution: remove Block parameters

remove parameters mentioned [here](https://github.com/bigchaindb/bigchaindb/pull/2366#discussion_r199154433)
depending on PR https://github.com/bigchaindb/bigchaindb/pull/2366
(when https://github.com/bigchaindb/bigchaindb/pull/2366 is merged I'll update this PR so that it can be merged)


bigchaindb/tendermint/lib.py
```diff
@@ -35,17 +35,11 @@ class BigchainDB(object):
...
-    BLOCK_VALID = TX_VALID = 'valid'
-    """return if a block is valid, or tx is in valid block"""
-
-    BLOCK_UNDECIDED = TX_UNDECIDED = 'undecided'
-    """return if block is undecided, or tx is in undecided block"""
-
-    TX_IN_BACKLOG = 'backlog'
-    """return if transaction is in backlog"""
+    TX_UNDECIDED = 'undecided'
+    """return if tx is in undecided block"""
```
kansi: These are not needed anymore.

codegeschrei: ah, so the only one that stays is the `TX_VALID`?

kansi: both `TX_IN_BACKLOG` and `TX_VALID` are needed anymore.

codegeschrei: ok, will remove it.

codegeschrei:
@ttmc yes, for the ones that are currently removed. If you find occurrences of `BLOCK_INVALID` or `BLOCK_VALID` they should be event types.

vrde:
> I think we can get rid of status? Since transactions can only be 'valid' there seems to be no reason to explicitly return 'valid'. /cc @vrde

:+1:

bigchaindb/tendermint/lib.py
```diff
@@ -271,19 +259,15 @@ def get_transaction(self, transaction_id, include_status=False):
             transaction = Transaction.from_dict(transaction)
 
         if include_status:
-            return transaction, self.TX_VALID if transaction else None
+            return transaction, 'valid' if transaction else None
```
kansi: I think the `'valid'` should go away here too.

codegeschrei: Then I need to change the whole function and remove `include_status` too. I can do it if we want that.

kansi: I think `include_status` is only used in the above `if` condition?

codegeschrei: yes, it's only in the if.

bigchaindb/lib.py
```diff
@@ -269,20 +257,13 @@ def get_transaction(self, transaction_id, include_status=False):
...
 
     def get_transactions_filtered(self, asset_id, operation=None):
         """Get a list of transactions filtered on some criteria
         """
-        txids = backend.query.get_txids_filtered(self.connection, asset_id,
-                                                 operation)
-        for txid in txids:
-            tx, status = self.get_transaction(txid, True)
-            if status == self.TX_VALID:
-                yield tx
```
kansi: The original code seems to return tx class but the new code returns dict ?

bigchaindb/lib.py
```diff
@@ -421,16 +402,8 @@ def text_search(self, search, *, limit=0, table='assets'):
...
-        objects = backend.query.text_search(self.connection, search, limit=limit,
-                                            table=table)
-
-        # TODO: This is not efficient. There may be a more efficient way to
-        #       query by storing block ids with the assets and using fastquery.
-        #       See https://github.com/bigchaindb/bigchaindb/issues/1496
-        for obj in objects:
-            tx, status = self.get_transaction(obj['id'], True)
-            if status == self.TX_VALID:
-                yield obj
```
kansi: This code also originally returned tx class but now it returns tx dict?

PR #2382, opened 2018-07-16
Last PR before the BigchainDB 2.0 beta 2 release

ttmc:
Following [the BigchainDB release process](https://github.com/bigchaindb/bigchaindb/blob/master/RELEASE_PROCESS.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,51 @@ For reference, the possible headings are:
 * **Known Issues**
 * **Notes**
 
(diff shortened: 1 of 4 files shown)


vrde:
Thanks @ttmc! Would it make sense to highlight in the `Changed` section that the BigchainDB ABCI server now listens to port `26657` instead of `46657`?

ttmc:
@vrde Yes that's a good idea. I will add a note about that.

Issue #2381, opened 2018-07-16
Problem: Tests cleanup incomplete


kansi:
Following is a comprehensive list of test modules which need to be cleaned/removed/activated (each module has been annotated with the possible action item):

- [ ] test_utils.py (activate/remove tests)
- [ ] test_txlist.py (activate/remove tests)
- [ ] test_core.py (activate/remove tests)
- [ ] test_block_model.py (Depricate/Remove)
- [ ] test_config_utils.py (activate/remove tests)
- [ ] remove test_consensus.py
- [ ] test_bigchain_api.py (activate/remove tests)
- [ ] common/test_schema.py (activate/remove tests)
- [ ] commands/test_utils.py (activate/remove tests)
- [ ] common/conftest.py (remove unused fixtures)
- [ ] clean up fixtures from conftest.py (remove unused fixtures)
- [ ] remove `pytest.mark.genesis` marker
- [ ] remove `pytest.mark.tendermint` marker

ttmc:
@kansi As I write this, on August 12, there is a task that reads:

> rename `localmongodb` to `mongodb`

While I can understand why that might seem like a nice idea, I'm not sure it's necessary. It will break all existing deployment tooling, including external tooling created by our users. (For example, the command `bigchaindb configure localmongodb` would stop working.)

In short, it's not broken, so let's not fix it.

Issue #1779, opened 2017-10-09
Use a phrase other than "pluggable consensus" for that feature


ttmc:
## Background

BigchainDB Server has a nice, albeit somewhat hidden, feature that allows one to change the rules for validating transactions and blocks via a plugin. It used to be documented in the docs, e.g. [in the docs for v0.6.0](https://docs.bigchaindb.com/projects/server/en/v0.6.0/appendices/consensus.html).

## The Problem

The above-described feature was added by an external developer, so we didn't want to be too critical. Unfortunately, they called it "pluggable consensus" and that is a badly misleading name. One can _not_ use it to change the consensus protocol used by BigchainDB Server (e.g. to Raft, Paxos, PBFT, or Uncle Joes' Awesome BFT Protocol). As explained above, one can change the rules to check if a transaction or block is valid.

Other blockchain solutions out there (such as Hyperledger Fabric) _do_ have pluggable consensus. We don't want to confuse people into thinking that BigchainDB Server has the same feature. Moreover, we might want to offer _actual_ pluggable consensus in the future. What would we call _that_?

## Proposal

Use a phrase other than "consensus" to describe this functionality, in the code, the docstrings, and the docs (if any).

How about "pluggable validation"?

The relevant code is the code mentioning the `BaseConsensusRules` class, e.g.

https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/consensus.py

ttmc:
I mentioned this issue in today's sprint planning meeting. Some notes:

* Some users are using the "pluggable consensus" feature but a name change would be fairly trivial for them to handle.
* Maybe we should wait a couple weeks, until we have a better picture of roadmap details, before we do this.
* Can we use Python's [deprecation library](https://pypi.python.org/pypi/deprecation) to help us manage the change?

sbellem:
I think we can address this is issue now using a deprecation warning.

That means we support both names (old and new) for the time being until the version by the deprecation warning.

Of relevant interest, in [Blockchain Consensus Protocols in the Wild](https://arxiv.org/abs/1707.01873) by [Christian Cachin](https://arxiv.org/find/cs/1/au:+Cachin_C/0/1/0/all/0/1), [Marko Vukolić](https://arxiv.org/find/cs/1/au:+Vukolic_M/0/1/0/all/0/1):

> ### 3.4  Validation
[...]

PR #2384, opened 2018-07-16
Problem: consensus and blocks are deprecated

codegeschrei:
Solution: remove test files for consensus and blocks

removes two items from https://github.com/bigchaindb/bigchaindb/issues/2381


ttmc:
I was curious what used to be in `tests/test_consensus.py` (which is currently empty) so I looked at its Git history. Apparently Scott removed all its remaining contents in https://github.com/bigchaindb/bigchaindb/commit/89e76ffec206b670aef249af8784f6ab1b943711#diff-08da6ad7500d470a74a8613a1a03a22b

If you look at that link, you'll see that all the removed stuff was related to voting, and of course BigchainDB doesn't do _any_ voting now, so I guess we really don't need that file.

(I also looked because the transaction validation plugin system is currently, and incorrectly, called "pluggable consensus" and we just want to _rename_ that, not delete it. See issue #1779.)

Issue #2222, opened 2018-04-18
Cannot post a simple transaction of more than 9300 bytes. No warning not error in the response


manolodewiner:
Running latest master branch of BigchainDB I cannot post a transaction of ~200 KB

Steps to reproduce:
From https://github.com/bigchaindb/project-jannowitz/tree/code-examples/js-examples
Run `node differentPayloadSize.js`

The logs of the server when inserting a transaction of this size is:

```
bigchaindb_1   | [2018-04-18 11:46:59] [DEBUG] (gunicorn.error) GET / (webapi     - pid: 23)
bigchaindb_1   | [2018-04-18 11:47:00] [DEBUG] (gunicorn.error) POST /api/v1/transactions (webapi     - pid: 23)
bigchaindb_1   | [2018-04-18 11:47:00] [DEBUG] (urllib3.connectionpool) Starting new HTTP connection (1): tendermint (webapi     - pid: 23)
bigchaindb_1   | [2018-04-18 11:47:00] [DEBUG] (urllib3.connectionpool) http://tendermint:46657 "POST / HTTP/1.1" 200 190 (webapi     - pid: 23)
bigchaindb_1   | [2018-04-18 11:47:00] [DEBUG] (bigchaindb.tendermint.lib) {'jsonrpc': '2.0', 'id': '35d5eb20-6ad3-431b-8c65-8943d8987439', 'result': {'code': 0, 'data': '', 'log': '', 'hash': '0FADA44F5302012E41D83B81AD1CD93841004196'}} (webapi     - pid: 23)
```

I do not see any log for creating a block, not any other warning.

Also cannot find the transaction searching for it with the id and the `transactions` endpoint

vrde:
I found some interesting parameters in the Tendermint source code, specifically for:
- [0.12.1](https://github.com/tendermint/tendermint/blob/v0.12.1/types/params.go)
- [0.19](https://github.com/tendermint/tendermint/blob/v0.19.0/types/params.go)

The limits are 21MB for a block, and 10kB for a single transaction.
```go
const (
	MaxBlockSizeBytes = 104857600 // 100MB
)

// DefaultBlockSize returns a default BlockSize.
func DefaultBlockSize() BlockSize {
	return BlockSize{
		MaxBytes: 22020096, // 21MB
		MaxTxs:   100000,
		MaxGas:   -1,
	}
}

// DefaultTxSize returns a default TxSize.
func DefaultTxSize() TxSize {
[...]

PR #2387, opened 2018-07-17
Problem: arbitrary big transactions might fail

vrde:
Solution: add a new configuration parameter `max_transaction_size` to
limit the size of transactions accepted by the web API. The parameter is
also exposed by the root API endpoint using the same key
`max_transaction_size`, so drivers can act accordingly and eventually
fail early if the transaction to be sent is bigger than the upper limit.
This issue was first reported in #2222.

By default Flask will accept file uploads to an unlimited amount of
memory, but you can limit that by setting the [`MAX_CONTENT_LENGTH`
config key][1]. Spoofing the `Content-Length` is also considered: see the
integration test under
`acceptance/python/src/test_max_transaction_size.py`.

I had troubles creating a unit test to check this feature. It seems that
the pytest fixture for testing Flask ignores the `MAX_CONTENT_LENGTH`
setting. If you have better ideas on how to test that, please make a
patch.

To be honest, I'm not sure that having the parameter
`max_transaction_size` in the BigchainDB configuration file is a good
idea. It should be a Network wide configuration parameter, shared and
accepted by every non-byzantine node. We should have something like the
`genesis.json` file in Tendermint, and it should be updated using the
Election Process specified by [BEP-18][2]. If we do that, then
`max_transaction_size` should be checked also in the validation rules.

[1]: http://flask.pocoo.org/docs/1.0/patterns/fileuploads/#improving-uploads
[2]: https://github.com/bigchaindb/BEPs/tree/master/18

### DIFF
--- a/bigchaindb/__init__.py
+++ b/bigchaindb/__init__.py
@@ -63,6 +63,7 @@ config = {
         'host': 'localhost',
         'port': 26657,
     },
+    'max_transaction_size': 1048576,
     # FIXME: hardcoding to localmongodb for now
     'database': _database_map['localmongodb'],
     'log': {
--- a/acceptance/python/src/test_max_transaction_size.py
+++ b/acceptance/python/src/test_max_transaction_size.py
@@ -1,0 +1,93 @@
+# # Transaction Size Tes
+# BigchainDB allows to limit the maximum size of a transaction. This acceptance
+# test make sure that this is happening.
+
+import os
+
+import pytest
+import requests
+
+from bigchaindb_driver import exceptions, BigchainDB
+from bigchaindb_driver.crypto import generate_keypair
+
+## Helper functions
+
+# Helper function to normalize the URL to connect to a BigchainDB node.
+def normalize_url(url):
+    from urllib.parse import urlparse, urlunparse
+    if '://' not in url:
+        url = 'http://' + url
+    p = urlparse(url)
+    if not p.port:
+        p = p._replace(netloc=p.hostname + ':9984')
+    return urlunparse(p)
+
+
+BIGCHAINDB_ENDPOINT = normalize_url(os.environ.get('BIGCHAINDB_ENDPOINT', 'http://localhost:9984'))
+bdb = BigchainDB(BIGCHAINDB_ENDPOINT)
+
+
+# Helper function to generate a transaction with an approximated size. Note
+# that the final size of the transaction will be bigger, since this function
+# does not consider the size of all other fields of the transaction model such
+# as `id`, `inputs`, `outputs`, and more.
+def generate_transaction_by_size(size):
+    keypair = generate_keypair()
+
+    asset = {'data': {'_': 'x' * size}}
+
+    prepared_creation_tx = bdb.transactions.prepare(
+            operation='CREATE',
+            signers=keypair.public_key,
+            asset=asset)
+
+    fulfilled_creation_tx = bdb.transactions.fulfill(
+            prepared_creation_tx,
+            private_keys=keypair.private_key)
+
+    return fulfilled_creation_tx
+
+## Testing time!
+def test_max_transaction_size():
+    # First, we retrieve the maximum size of a transaction from the API.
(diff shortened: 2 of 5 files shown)


ttmc:
Please add some documentation about this new parameter in the page `docs/server/source/server-reference/configuration.md`:

- Note that the number is in bytes
- Link to the docs for the associated Flask configuration setting (MAX_CONTENT_LENGTH): http://flask.pocoo.org/docs/1.0/config/#MAX_CONTENT_LENGTH

I don't know how we could _enforce_ all nodes having the same setting value. A Byzantine node could just lie about its value of `max_transaction_size` when asked. Each node could test another node, but then we have to ask which one, and how often it does the test (i.e. a real rabbit hole of complexity).

ttmc:
On a different note, it seems that this new setting is under the same umbrella as `server.bind`, `server.loglevel` & `server.workers`, so maybe it should be `server.max_transaction_size`.

bigchaindb/__init__.py
```diff
@@ -63,6 +63,7 @@
         'host': 'localhost',
         'port': 26657,
     },
+    'max_transaction_size': 1048576,
```
ttmc: Where did this number come from?

ttmc:
I'm okay with having the `max_transaction_size` as a top-level setting (for each node). I just wondered your reasoning for it.

> I'm arguing that this setting should be something implemented on a Network level, be written somewhere on the blockchain, and integrated at a validation logic.

Oh, I understand now: a transaction validation rule of the form 

`size(transaction) <= network_max_transaction_size`

That's a nice idea. We could consider it for the v3 transaction model (or later). To ensure we don't forget, we could create an issue with the `tx_model` label so we don't forget about it: https://github.com/bigchaindb/bigchaindb/labels/tx%20model

For now, letting each node implement its own size check (using the web server) is probably okay, so long as there's a good default (less than Infinity).

ttmc:
In issue #2393, @CosmicCoder96 posted an example `genesis.json` file, for the latest Tendermint: see https://github.com/bigchaindb/bigchaindb/issues/2393#issuecomment-406074148 

I noticed some new things in there:

```javascript
{
  "genesis_time": "2018-07-18T15:48:46.24269529Z",
  "chain_id": "test-chain-cl61op",
  "consensus_params": {
    "block_size_params": {
      "max_bytes": "22020096",
      "max_txs": "10000",
      "max_gas": "-1"
    },
    "tx_size_params": {
      "max_bytes": "10240",
      "max_gas": "-1"
    },
    "block_gossip_params": {
      "block_part_size_bytes": "65536"
    },
    "evidence_params": {
      "max_age": "100000"
    }
  },
  "validators": [...],
  "app_hash": ""
}
```

In particular, there's a `tx_size_params.max_bytes` parameter. Since it's in the `genesis.json` file, it's the same for the whole (Tendermint) network.

Does this mean we can just rely on Tendermint to limit the size of a transaction (and also the size of a block)? Does Tendermint actually _use_ these parameters?

If so, we should still point our users to the Tendermint documentation about all these parameters, i.e. from the BigchainDB documentation.

ttmc:
On August 12, I asked a question in the dev-tendermint channel on Slack:

> In newer `genesis.json` files, there is a `consensus_params.tx_size_params.max_bytes` parameter. Is that maximum enforced by Tendermint?

and one of the core devs answered:

> not currently
> will be in the next release

(The latest release, at that point in time, was Tendermint v0.23.0.)

Therefore I guess we don't need to have a similar parameter in the BigchainDB configuration settings (especially since it would be per node, whereas `genesis.json` is for the whole network).

ttmc:
## Update

Now it seems that Tendermint might not add the new max transaction size parameter after all! See https://github.com/tendermint/tendermint/issues/2347

ttmc:
This is interesting: there's another limit on the size of a transaction, currently hard-coded into Tendermint...

@ebuchman [wrote](https://github.com/tendermint/tendermint/issues/2347#issuecomment-419095990):
> Also note that the mempool must have an upper bound on message size - it's currently hardcoded to 1MB but perhaps we can make it dependent on the block size (ie. it should be the max size of data for the block)

Issue #1271, opened 2017-03-10
double spend incorrectly reported


sbellem:
From [jmduque on gitter](https://gitter.im/bigchaindb/bigchaindb?at=58c20249de504908226c4835) (with small edits to help clarifying):

> This is an issue with the implementation of the [`get_spent()`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/backend/mongodb/query.py#L151-L167) query for MongoDB.
> For the RethinkDB implementation the issue does NOT arise.
>
> ### How to duplicate the bug?
> 1. Create a divisible asset (outputs: tx1-0)
> 2. Transfer 2 amount to user A and 2 amount to user B (input: tx1-0, outpus: tx2-0/tx2-1)
> 3. Transfer another 1 Amount to A (inputs: tx1-0, ouputs: tx3-0/tx3-1)
> 4. Transfer 3 Amount from user A to user B (inputs: tx2-1/tx3-0, outputs: tx4-0/tx4-1)
> 5. Transfer 2 Amount from user B to user A (inputs: tx2-0, outputs: tx5-0)

See report by jmduque on gitter: https://gitter.im/bigchaindb/bigchaindb?at=58c12bf200c00c3d4fe02c88 for all details.

/cc @jmduque

ttmc:
The bug arises when you ask for all of A's unspent outputs. The answer should be tx5-0 (i.e. only one unspent output), but the response is an error about invalid blockchain because of double-spend of output tx2-0.

ttmc:
I drew a little diagram to understand how all the transactions are related:

![office lens 20170310-142642](https://cloud.githubusercontent.com/assets/1210951/23797022/23ce4988-059e-11e7-9239-73a41b6b36df.jpg)

sbellem:
fyi: a test that attempts to reproduce the issue is at https://github.com/sbellem/bigchaindb/blob/bug-1271-false-double-spend/tests/test_core.py#L331

It must be noted that the test should fail but does not right now. There are two possibilities:

1. The test is incorrect.
2. There's no bug.

PR #2389, opened 2018-07-17
Problem: a valid transaction is detected as double spend

codegeschrei:
Solution: query the wanted data per input

see #1271

To add more details:
In the test we have 5 transactions, one create and 4 transfer. In the first 4 transaction Bob only receives tokens and in the fifth he wants to spend the first two tokens he got.

When validating the fifth transaction in the query we check if the `transaction_id` and the `output` are in `inputs`. The `inputs` contains two `input`s, one that matches the `transaction_id` and the other one that matches the `output`, thus we found a match and the fifth transaction is a double spend. 
Which it is obviously not. 
So @kansi and me changed the query accordingly and we now match the `transaction_id` and the `output` for every `input` in `inputs`.

I needed to skip the other tests in this file, they're not related to this issue and should be activated separately (related to https://github.com/bigchaindb/bigchaindb/issues/2381)

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -99,11 +99,13 @@ def get_assets(conn, asset_ids):
 
 @register_query(LocalMongoDBConnection)
 def get_spent(conn, transaction_id, output):
+    query = {'inputs.fulfills': {
+        'transaction_id': transaction_id,
+        'output_index': output}}
+
     return conn.run(
         conn.collection('transactions')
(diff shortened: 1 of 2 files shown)


ttmc:
Some notes: A test got skipped while we were doing Tendermint integration, then we changed how a certain query worked, and the new code had a bug, but it didn't get detected because the test was skipped. When we turned on the test again, the bug was caught. This PR fixes the query (and the associated test).

The bug will say (incorrectly) that certain valid transactions are invalid (double spends), in complex situations. That is annoying, but not as bad as invalid transactions being considered valid.

PR #2390, opened 2018-07-17
Problem: some tests are not activated

codegeschrei:
Solution: activate and remove or fix tests

see #2381

WARNING: I skipped the `test_get_spent_issue_1271` test because it is handled in #2389
if #2389 is merged before this PR then the master needs to be merged in here and the skip mark needs to be removed


tests/test_txlist.py
```diff
@@ -54,8 +35,8 @@ def txlist(b, user_pk, user2_pk, user_sk, user2_sk, genesis_block):
 @pytest.mark.bdb
 def test_get_txlist_by_asset(b, txlist):
     res = b.get_transactions_filtered(txlist.create1.id)
```
kansi: [`get_transaction_filtered`](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/lib.py#L284) uses `status`, I think we can skip that and directly yield i.e.

z-bowen:
There are still tests in `tests/test_core.py` that are not marked for execution. Are these going to be fixed in a future PR, or should they be removed?

codegeschrei:
I think it should not be skipped but I have to check again. Will remove it another PR anyways.

Issue #2372, opened 2018-07-03
Upsert validator feature incomplete


kansi:
This issue tracks the various tasks (in order) which need to be completed in order to implement `upsert-validator` i.e. [BEP-21](https://github.com/bigchaindb/BEPs/pull/45/files#diff-b4ca8692e6608b9b622c49bfaba4a825).

- [x] Implement [`VALIDATOR_ELECTION` spec](https://github.com/bigchaindb/BEPs/blob/master/21/transaction_validator_election_v2.0.yaml). See PR #2392
- [x] Implement [`VALIDATOR_VOTE` spec](https://github.com/bigchaindb/BEPs/blob/master/21/transaction_validator_election_vote_v2.0.yaml) See PR #2428
- [x] Track validator set in BigchainDB #2436 
- [x] Implement `upsert-validator` transaction election process.  See PR ~~#2314~~ #2445 
- [x] Implement `upsert-validator new E_PUBKEY E_POWER E_NODE_ID` See PR #2446 
- [x] Implement `upsert-validator show <election_id>` #2457
- [x] Implement `upsert-validator approve <election_id>` #2458
- ~~Implement `upsert-validator status <election_id>`~~
- [x] Add required tests for `upsert-validator new` #2460
- [x] Update corresponding documentation, including the section titled [Member: Dynamically Add a New Member to the Network](http://docs.bigchaindb.com/projects/server/en/master/simple-deployment-template/network-setup.html#member-dynamically-add-a-new-member-to-the-network) --- See pull request https://github.com/bigchaindb/bigchaindb/pull/2570 
- [x] Update BEP-18 to reflect what was _actually_ implemented
- [x] Update BEP-21 to reflect what was _actually_ implemented https://github.com/bigchaindb/BEPs/pull/80
- [x] Update BEP-13 to reflect what was _actually_ implemented https://github.com/bigchaindb/BEPs/pull/90
- [x] standardize public key format i.e. Tendermint encodes keys in base64 format when generating configs but during `end_block` it expects the public key to be base58 encoded. #2480 #2508
- [x] remove experimental tag from documentation of upsert-validator

ttmc:
@kansi Should the code to implement the [GET /api/v1/validators](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#get--api-v1-validators) endpoint also be updated?

PR #2392, opened 2018-07-18
Problem: (BEP-21) no way to propose validator election

kansi:
Solution: Integrate the new election spec specified in [BEP-21](https://github.com/bigchaindb/BEPs/tree/master/21)

Implements one of the tasks in issue #2372


bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+                                      TX_SCHEMA_VERSION,
+                                      TX_SCHEMA_COMMON,
+                                      TX_SCHEMA_CREATE)
+
+
+_, TX_SCHEMA_VALIDATOR_ELECTION = _load_schema('transaction_validator_election_' +
+                                               TX_SCHEMA_VERSION, __file__)
+
+
+class ValidatorElection(Transaction):
```
ldmberman: Looks like the `unspent_outputs`, `inputs_valid`, and `get_asset_id` methods are broken now.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+
+    VALIDATOR_ELECTION = 'VALIDATOR_ELECTION'
+    # NOTE: this transaction class extends create so the operation inheritence is achieved
```
ldmberman: May be it is cleaner to override `cls.create` to raise an exception, what do you think? Or do you want to use `ValidatorElection` for creating election votes later?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @classmethod
+    def generate(cls, initiator, voters, election_data, metadata=None):
+        election = cls.create(initiator, voters, metadata, asset=election_data)
+        cls.validate_schema(election.to_dict())
+        return election
+
+    @classmethod
+    def from_dict(cls, tx):
+        cls.validate_id(tx)
```
ldmberman: Does it make sense to update the base `from_dict` instead?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+            if len(voter.public_keys) > 1:
+                return False
+
+            [public_key] = voter.public_keys
+            voting_power = voter.amount
+            voters[public_key] = voting_power
+
+        # Check whether the voters and their votes is same to that of the
+        # validators and their voting power in the network
+        return (current_topology.items() == voters.items())
```
ldmberman: Although the new Python guarantees the key order, I would compare dicts directly `current_topology == voters`, for simplicity.

bigchaindb/tendermint/lib.py
```diff
@@ -326,7 +327,7 @@ def get_spent(self, txid, output, current_transactions=[]):
         if len(transactions) + len(current_spent_transactions) > 1:
             raise DoubleSpend('tx "{}" spends inputs twice'.format(txid))
         elif transactions:
-            transaction = Transaction.from_db(self, transactions[0])
+            transaction = operation_class(transactions[0]).from_db(self, transactions[0])
```
ldmberman: To make the core transaction module independent from the plugin modules, we can do DI the following way:

```
class Transaction:
    type_registry = {}

    @classmethod
    def register_type(typ, clas):
        type_registry[typ] = clas

    @classmethod
    def from_whatever(tx):
        return cls(type_registry[tx['operation']])
```

Then, inside `a_plugin.py` one can do smth like:

```
from core import Transaction

Transaction.register_type('election', ValidatorElection)
```

What do you think?

tests/upsert_validator/test_validator_election.py
```diff
@@ -0,0 +1,48 @@
...
+    valid_election = ValidatorElection.generate([node_key.public_key],
+                                                voters,
+                                                new_validator, None).sign([node_key.private_key])
+
+    with pytest.raises(DuplicateTransaction):
+        valid_election.validate(b_mock, [valid_election])
+
+    b_mock.store_bulk_transactions([valid_election])
+
+    with pytest.raises(DuplicateTransaction):
```
ldmberman: Looks like these two `with pytest.raises(DuplicateTransaction):` are following the same code path.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+        # NOTE: Check if the proposer is a validator.
+        [election_initiator_node_pub_key] = self.inputs[0].owners_before
+        if election_initiator_node_pub_key not in current_validators.keys():
+            return False
+
+        # NOTE: Check if all validators have been assigned votes equal to their voting power
+        return self.is_same_topology(current_validators, self.outputs)
+
+    @classmethod
+    def generate(cls, initiator, voters, election_data, metadata=None):
```
ldmberman: Is there a particular reason for having both the constructor and `generate`?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,140 @@
...
+        if len(self.inputs) != 1:
+            return False
+
+        # NOTE: Check if the proposer is a validator.
+        [election_initiator_node_pub_key] = self.inputs[0].owners_before
+        if election_initiator_node_pub_key not in current_validators.keys():
+            return False
+
+        # NOTE: Check if all validators have been assigned votes equal to their voting power
+        return self.is_same_topology(current_validators, self.outputs)
```
ldmberman: It would be helpful to also consider elections which add up more than 1/3 of the current total validator power to be invalid.

tests/upsert_validator/test_validator_election.py
kansi: Not quite, the previous exception will be generated because `valid_election` is a part of `current_transactions` list and the second exception would be generated because `valid_election` has been stored in MongoDB and should be read from there to raise the exception.

ldmberman: I think, you just need to call `valid_election.validate(b_mock)` in the second case. Otherwise you depend on the order of the checks.

bigchaindb/upsert_validator/validator_election.py
kansi: Why?

bigchaindb/upsert_validator/validator_election.py
kansi: > May be it is cleaner to override cls.create to raise an exception, what do you think? 
Cloud you please elaborate? What would be achieved if we override `cls.create` and raise an exception here?

bigchaindb/upsert_validator/validator_election.py
ldmberman: These functions expect `self.operation` to be one of `Transaction.CREATE, Transaction.TRANSFER`, what is not the case for `ValidatorElection`.

bigchaindb/upsert_validator/validator_election.py
kansi: I think we can improve this by calling `super().from_dict()`?

bigchaindb/upsert_validator/validator_election.py
ldmberman: Currently, there are three methods for instantiating an election transaction - `ValidatorElection.create`, `ValidatorElection.generate`, and the constructor. It is easier to read and use the code if we only keep one. The problem with `create` though is that it comes from the parent so raising an exception is a way to explain that this method is not for use. Ideally, we should have a base transaction class without `create` and `transfer`.

tests/upsert_validator/test_validator_election.py
kansi: oh yeah, my bad.

bigchaindb/upsert_validator/validator_election.py
kansi: > Ideally, we should have a base transaction class without `create` and `transfer`.

That is the target. In this PR I am trying to consolidate the methods which are required in the base `Transaction` class so what other new transaction classes can be written by extending it.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,139 @@
...
+                                       .format(self.id))
+
+        if not self.inputs_valid(input_conditions):
+            raise InvalidSignature('Transaction signature is invalid.')
+
+        current_validators = self.current_validators(bigchain)
+
+        # NOTE: Proposer should be a single node
+        if len(self.inputs) != 1:
+            raise TypeError('`tx_signers` must be a list instance or length one')
```
ldmberman: It makes sense to have a specifc exception in the case like this so that a caller can easily distinguish exceptions.

kansi: You mean a custom exception?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # NOTE: change more than 1/3 of the current power is not allowed
+        if self.asset['data']['power'] > (1/3)*sum(current_validators.values()):
```
ldmberman: Isn't it `>= 1/3`?

bigchaindb/upsert_validator/validator_election.py
ldmberman: Yes.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,148 @@
...
+        NOTE:
+        * A valid election is initiated by an existing validator.
+
+        * A valid election is one where voters are validators and votes are
+          alloacted according to the voting power of each validator node.
+
+        Args:
+            bigchain (BigchainDB): an instantiated bigchaindb.tendermint.BigchainDB object.
+
+        Returns:
```
ldmberman: The "Returns" section is outdated.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,148 @@
...
+    def validate(self, bigchain, current_transactions=[]):
+        """Validate election transaction
+        NOTE:
+        * A valid election is initiated by an existing validator.
+
+        * A valid election is one where voters are validators and votes are
+          alloacted according to the voting power of each validator node.
+
+        Args:
+            bigchain (BigchainDB): an instantiated bigchaindb.tendermint.BigchainDB object.
```
ldmberman: The location of `BigchainDB` has been [changed](https://github.com/bigchaindb/bigchaindb/pull/2401).

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -0,0 +1,148 @@
...
+            voters[public_key] = voting_power
+
+        # Check whether the voters and their votes is same to that of the
+        # validators and their voting power in the network
+        return (current_topology == voters)
+
+    def validate(self, bigchain, current_transactions=[]):
+        """Validate election transaction
+        NOTE:
+        * A valid election is initiated by an existing validator.
```
ldmberman: What do you think about putting a reference to the BEP here? Then, a reader will get the complete context and we won't have to maintain the validation documentation in two places.

kansi:
Removing @vrde from review as he is on holiday.

bigchaindb/common/exceptions.py
```diff
@@ -102,3 +102,19 @@ class GenesisBlockAlreadyExistsError(ValidationError):
...
+
+class MultipleInputsError(ValidationError):
+    """Raised if there were multiple inputs when only one was expected"""
+
+
+class InvalidProposer(ValidationError):
+    """Raised if the public key is not a part of the validator set"""
+
+
+class UnequalValidatorSet(ValidationError):
```
muawiakh: nitpick: `InvalidValidatorSet` makes more sense for this exception?

kansi: I thought about that name `InvalidValidatorSet`, but the code [here](https://github.com/bigchaindb/bigchaindb/pull/2392/files#diff-dff01128fef040477df965185b717a15R119) tries to detect if the validator sets are the same or different. The only thing we wish to imply is their inequality and not validity.

Issue #2393, opened 2018-07-18
Bigchain DB nodes in a 4 node cluster unable to replicate the data


CosmicCoder96:
I have set up a cluster of 4 nodes.
The operating system on each of those is **Ubuntu 16.04** 
The bigchaindb version is : **2.0.0b3**
Tendermint version: **0.22.3**

The nodes were set up as per instructions in 
https://docs.bigchaindb.com/projects/server/en/latest/simple-network-setup.html

The ports mentioned in the above link are open on each of those machines.

### Current Behavior
- I make a simple hello world create transaction on a particular node (say node 1).
- Transaction is completed and a transaction ID is returned.
- I query node 1 (using http api or directly accessing database)
- I am able to find the transaction
- I query the other nodes' (node2, node3, node4) HTTP API or their database, I am unable to find the transaction.

### Expected Behavior
- The transaction data should be replicated among the nodes of the cluster.

### What I want
Some help in debugging. I expect bigchaindb should raise a warning or an error, when replication fails in a cluster. It might even be raising it, but I am not sure what is the right place to look for. (There were no errors on output of the `bigchaindb start` output.)

### What I have tried so far
- Tested ping for nodes - they were able to ping each other
- Unblocked the required ports and double checked them

output of ` sudo lsof -i -P -n `
![image](https://user-images.githubusercontent.com/17252805/42906321-22325952-8adb-11e8-818d-98f9002feef9.png)

kansi:
@CosmicCoder96 It seems to me that something from the above description is missing i.e. what you are trying to imply here is that only 1 node commits the transaction in a 4 node network. This is not possbile as a transaction is considered commited if at least 3 nodes commit the transaction. I would suggest making sure that all the nodes are infact connected and are part of the same blockchain.

CosmicCoder96:
@kansi how can I check that - presently when I say it's a 4 node network, it's based on the following:
- They share the same **genesis.JSON**
- In **config.toml** - all 4 nodes are listed as `persistant peers`

ttmc:
@vrde wrote some tips to help you get diagnostics on your Tendermint network, at https://github.com/vrde/notes/tree/master/tendermint 

Note that port 46657 has changed to port 26657 in the latest release. (This change was made by Tendermint.)

CosmicCoder96:
@ttmc Thanks for sharing this, this has certainly given me the information I was looking for.
This is the output
```json
{
  "jsonrpc": "2.0",
[...]

PR #2394, opened 2018-07-19
 Problem: Docs re/ resetting Tendermint are wrong

ttmc:
Solution: Update them and also link to the relevant Tendermint docs

Note 0: I don't know why there are so many commits in this PR. The last one should be the only one.

Note 1: I noticed this when helping @CosmicCoder96 resolve issue #2393


kansi:
I am confused here, following is the output for both commands,

```bash
➜  ~ tendermint-0.22.0 unsafe_reset_all 
I[07-19|19:11:27.546] Reset private validator file to genesis state module=main file=/home/kansi/.tendermint/config/priv_validator.json
I[07-19|19:11:27.549] Removed all blockchain history               module=main dir=/home/kansi/.tendermint/data

➜  ~ tendermint-0.22.0 unsafe_reset_priv_validator
I[07-19|19:11:37.302] Reset private validator file to genesis state module=main file=/home/kansi/.tendermint/config/priv_validator.json
```

It seems to me that the `unsafe_reset_priv_validator` is implicitly executed when `unsafe_reset_all ` is called.

ttmc:
@kansi Maybe it would be enough to just do `tendermint unsafe_reset_all` but the current Tendermint docs at https://tendermint.readthedocs.io/projects/tools/en/master/using-tendermint.html#reset
say to do both:

> To reset a blockchain, stop the node, remove the ~/.tendermint/data directory and run
>
> `tendermint unsafe_reset_priv_validator`

ttmc:
The instructions, as they are, should be enough, so I'm closing this PR (not merging it).

PR #2395, opened 2018-07-20
Problem: BigchainDB main/parent process does not handle signals

### LINKED ISSUE #1279: [shutdown] Error in atexit._run_exitfuncs
When running bigchaindb with the command `bigchaindb start` and interrupting the process with `CTRL-C`, the following error shows up:

```bash
$ bigchaindb start
# ...
# ...
# ...
# CTRL-C
^CError in atexit._run_exitfuncs:
  File "/usr/local/lib/python3.5/multiprocessing/popen_fork.py", line 29, in poll
    pid, sts = os.waitpid(self.pid, flag)
KeyboardInterrupt
[2017-03-12 23:29:25 +0000] [34] [INFO] Handling signal: int
```
#### Related links
Listing a few links just in case it may be useful:

* Python docs: [signal](https://docs.python.org/3/library/signal.html#module-signal), [atexit](https://docs.python.org/3/library/atexit.html#module-atexit), [multiprocessing](https://docs.python.org/3/library/multiprocessing.html)
* http://stackoverflow.com/questions/9139127/why-is-my-threading-multiprocessing-python-script-not-exiting-properly#9143486
* http://stackoverflow.com/questions/2546276/python-process-wont-call-atexit#2546397
* http://stackoverflow.com/questions/39884898/large-amount-of-multiprocessing-process-causing-deadlock
* http://stackoverflow.com/questions/19652446/python-program-with-thread-cant-catch-ctrlc
* http://stackoverflow.com/questions/37692262/python-multiprocessing-assertionerror-can-only-join-a-child-process

#### Noteworthy
Quoting Martin Sústrik (original architect and lead developer of ØMQ), in http://aosabook.org/en/zeromq.html (thanks to @krish7919 for the link):

> It turns out that shutting down a fully asynchronous system in a clean way is a dauntingly complex task. Trying to shut down a thousand moving parts, some of them working, some idle, some in the process of being initiated, some of them already shutting down by themselves, is prone to all kinds of race conditions, resource leaks and similar. The shutdown subsystem is definitely the most complex part of ØMQ. A quick check of the bug tracker indicates that some 30--50% of reported bugs are related to shutdown in one way or another.
>
> Lesson learned: When striving for extreme performance and scalability, consider the actor model; it's almost the only game in town in such cases. However, if you are not using a specialised system like Erlang or ØMQ itself, you'll have to write and debug a lot of infrastructure by hand. **Additionally, think, from the very beginning, about the procedure to shut down the system.** It's going to be the most complex part of the codebase and if you have no clear idea how to implement it, you should probably reconsider using the actor model in the first place.

### LINKED ISSUE #1662: bigchaindb's main process doesn't use signal handlers
Starting bigchaindb in the terminal and pressing Crtl+C works just fine. All child processes are terminated.

However, starting bigchaindb (e.g. with nohup) and then killing it (either with SIGTERM or SIGINT) will not result in bigchaindb shutting down properly. Only the main process exits. The child processes stay hanging around.

In python, catching KeyboardInterrupt doesn't suffice to respond to signals like SIGINT. Signal handlers must be set to ensure a proper shutdown of bigchaindb.

muawiakh:
## Problem
Starting bigchaindb in the terminal and pressing Crtl+C works just fine. All child processes are terminated.

However, starting bigchaindb (e.g. with nohup) and then killing it (either with SIGTERM or SIGINT) will not result in bigchaindb shutting down properly. Only the main process exits. The child processes stay hanging around.

## Solution

Handling the following signals SIGTERM, SIGINT, SIGQUIT in the BigchainDB parent process. The parent process is named `bigchaindb`.

```
$ pstree -p 
...
           ├─bigchaindb(19610)─┬─bigchaindb_exch(19631)───{bigchaindb_exch}(19633)
           │                   ├─bigchaindb_ws(19629)───{bigchaindb_ws}(19634)
           │                   ├─bigchaindb_ws_t(19630)
           │                   ├─gunicorn: maste(19628)─┬─gunicorn: worke(19632)
           │                   │                        ├─gunicorn: worke(19635)
           │                   │                        ├─gunicorn: worke(19636)
           │                   │                        ├─gunicorn: worke(19637)
           │                   │                        └─gunicorn: worke(19638)
           │                   ├─{bigchaindb}(19639)
           │                   └─{bigchaindb}(19640)
...
```

To kill the parent and all the linked child processes now the USER can send any of the above mentioned signals.

```
# Sending SIGINT
$ kill -2 <bigchaindb_parent_pid>

OR
# Sending SIGTERM
$ kill -15 <bigchaindb_parent_pid>

OR
# Sending SIGQUIT
$ kill -3 <bigchaindb_parent_pid>

OR
# If you want to kill all the processes by name yourself
$ pgrep bigchaindb | xargs kill -2 bigchaindb 
```

The purpose of handling SIGTERM, SIGQUIT and SIGINT is to terminate a (parent)process and give it an opportunity to handle graceful shut down, which is the use case, we are trying to address here. So IMO, might as well handle all the signals which allow for graceful shutdown.

e.g. We cannot handle `SIGKILL` i.e. `kill -9 <pid>` because it terminates a process immediately and is sort of a last resort. Similarly, `SIGSTOP`. For `SIGKILL` and `SIGSTOP` the user can use the following i.e.

```
$ pgrep bigchaindb | xargs kill -9

OR
$ pkill bigchaindb
```

## Issues Resolved

Resolves #1662
Resolves #1279 (This issue is not relevant anymore)


bigchaindb/tendermint/commands.py
```diff
@@ -31,17 +34,27 @@
...
     exchange = Exchange()
-
     # start the web api
     app_server = server.create_server(
         settings=bigchaindb.config['server'],
         log_config=bigchaindb.config['log'],
         bigchaindb_factory=BigchainDB)
-    p_webapi = Process(name='bigchaindb_webapi', target=app_server.run)
+    p_webapi = Process(name='bigchaindb_webapi', target=app_server.run, daemon=True)
+    PROCESSES.append(p_webapi)
```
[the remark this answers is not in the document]
muawiakh: My understanding is that we are starting all processes from this module. What do you mean by *as part of BigchainDB* , do you mean the BigchainDB class? Where do you suggest we define it?

bigchaindb/tendermint/commands.py
```diff
@@ -69,8 +85,17 @@ def start():
 
     setproctitle.setproctitle('bigchaindb')
 
+    # Start the ABCIServer and watch for signals to stop
+    # the server and the relevant child bigchaindb.BIGCHAINDB_PROCESSES.
     app = ABCIServer(app=App())
-    app.run()
+    gevent.signal(signal.SIGTERM, sig_handler, app.server)
```
shahbazn: isn't just SIGINT sufficient? why do we need to handle SIGTERM and SIGQUIT?

muawiakh: The purpose of handling `SIGTERM`, `SIGQUIT` and `SIGINT`  is to terminate a (parent)process and give it an opportunity to handle graceful shut down, which is the use case, we are trying to address here. So IMO, might as well handle all the signals which allow for `graceful` shutdown.

e.g. We cannot handle `SIGKILL` i.e. `kill -9 <pid>` because it terminates a process immediately and is sort of a last resort. Similarly, `SIGSTOP`. For `SIGKILL` and `SIGSTOP` the user can use the  following i.e.

```
$ pgrep bigchaindb | xargs kill -9 bigchaindb 
```

ldmberman:
I tried a simpler patch locally - it successfully handles startup and shuts down gracefully when the root process receives SIGTERM, SIGQUIT or SIGINT.

Those signals are propagated to the child processes without an explicit signal handler (`daemon=True` does the trick); `p.join` avoids zombies in the situation when `gevent` is too quick to shutdown; the ABCI server has some signal handling inside `app.run` that works.

Could you check if it works for you too?

screen shot 2018-07-23 at 11 57 09

muawiakh:
@ldmberman yes, it works for me. Seems like I don't need to do:

```
for p in bigchaindb.BIGCHAINDB_PROCESSES:
    p.join()
```

As well as:

```
for p in bigchaindb.BIGCHAINDB_PROCESSES:
    p.terminate()
```

ldmberman:
@muawiakh cool! Also, custom signal handling and custom ABCI server startup does not seem to be required.

kansi:
@muawiakh Could you please update the docs (like simple_network_setup) and add a little description about how to stop BigchainDB?

muawiakh:
@kansi done, please review the latest changes.

ttmc:
Let me try rendering the docs locally before you merge this PR.

docs/server/source/simple-network-setup.md
```diff
@@ -252,6 +252,66 @@ If you want to refresh your node back to a fresh empty state, then your best bet
 - reset Tendermint using `tendermint unsafe_reset_all`
 - delete the directory `$HOME/.tendermint`
 
+## Shutting down BigchainDB
```
vrde: This section is way too detailed for a "simple network setup". You can move some of the details in other parts of the documentation.

ttmc: Maybe just create a new stand-alone page about how to shut down BigchainDB, and then link to that page from here?

muawiakh: @ttmc @vrde should I stick to creating a new page and referring it here?

PR #2401, opened 2018-07-24
Refactor tendermint directory to project root

z-bowen:
**Problem: The BigchainDB class should be defined in project root, not in the /tendermint directory.**

## Solution

Moved the BigchainDB class to lib.py in project root, as well as the rest of the contents from the /tendermint directory, and removed /tendermint.


docs/server/source/appendices/tendermint-integration.rst
```diff
@@ -1,26 +0,0 @@
-######################
```
ttmc: You will also have to remove the reference to this page from the file that generates the table of contents for the Appendices, i.e. `docs/server/source/appendices/index.rst`

Currently, when I go to the `/docs/server/` directory and do `make html`, I get the error message:
```text
Warning, treated as error:
/home/troy/repos/bigchaindb/docs/server/source/appendices/index.rst:4:toctree contains reference to nonexisting document 'appendices/tendermint-integration'
Makefile:56: recipe for target 'html' failed
make: *** [html] Error 2
```

z-bowen: Fixed

ttmc:
You fixed the problem with index.rst, but now there is some other problem with building the docs:
```text
Warning, treated as error:
autodoc: failed to import class 'BigchainDB' from module 'bigchaindb.tendermint'; the following exception was raised:
No module named 'bigchaindb.tendermint'
```

The bigchaindb.tendermint module is referenced in the file `docs/server/source/appendices/the-bigchaindb-class.rst`.

Also there are some docstrings which still refer to the bigchaindb.tendermint module, and that might cause problems with documentation generation based on those docstrings.

PR #2402, opened 2018-07-24
Problem: config utils tests are disabled.

ldmberman:
Solution: enable them back, cleanup no longer relevant parts.


tests/test_config_utils.py
```diff
@@ -185,12 +169,7 @@ def test_autoconfigure_read_both_from_file_and_env(monkeypatch, request, ssl_con
                                            'BIGCHAINDB_WSSERVER_ADVERTISED_HOST': WSSERVER_ADVERTISED_HOST,
                                            'BIGCHAINDB_WSSERVER_ADVERTISED_PORT': WSSERVER_ADVERTISED_PORT,
                                            'BIGCHAINDB_KEYRING': KEYRING,
```
kansi: Is `KEYRING` still being used?

tests/test_config_utils.py
```diff
@@ -185,12 +169,7 @@ def test_autoconfigure_read_both_from_file_and_env(monkeypatch, request, ssl_con
...
-                                           'BIGCHAINDB_LOG_FILE': LOG_FILE,
-                                           'BIGCHAINDB_DATABASE_CA_CERT': ssl_context.ca,
-                                           'BIGCHAINDB_DATABASE_CRLFILE': ssl_context.crl,
-                                           'BIGCHAINDB_DATABASE_CERTFILE': ssl_context.cert,
-                                           'BIGCHAINDB_DATABASE_KEYFILE': ssl_context.key,
-                                           'BIGCHAINDB_DATABASE_KEYFILE_PASSPHRASE': None})
```
kansi: Are we no longer supporting ssl for MongoDB?

tests/test_config_utils.py
```diff
@@ -31,21 +30,6 @@ def test_bigchain_instance_is_initialized_when_conf_provided(request):
     assert bigchaindb.config['CONFIGURED'] is True
 
 
-def test_bigchain_instance_raises_when_not_configured(request, monkeypatch):
```
kansi: I couldn't understand why don't we need this test anymore?

ldmberman: Where do you expect `ConfigurationError` to come from? From the first glance, without a config assigned, it will fail with a different exception. Also, the test setup assigns a config so it does not work either way.

kansi: It seems that this test wasn't of much help before. I guess  we can do without it then

z-bowen: It seems like the intent of the test is to make sure we fail in a controlled fashion if the config gets dropped in some way. Couldn't we modify it to use a patched `autoconfigure` with `force=True` to set a bad config and produce the error that way?

ldmberman: Currently, the function does not raise `ConfigurationError` in case the config is not set or empty. The contract that was tested here is lost.

PR #2412, opened 2018-07-26
Problem: JSON schema allows output amount "0"

ttmc:
Solution: Change the regex to allow "1" to "9"x20 but not "0"

TODO:

- [ ] Write a unit test for the "0" case, i.e. make sure a transaction containing an output amount "0" is not considered valid. @z-bowen could you do that? It could be added as another commit to this pull request.

### DIFF
--- a/bigchaindb/common/schema/transaction_v2.0.yaml
+++ b/bigchaindb/common/schema/transaction_v2.0.yaml
@@ -79,7 +79,7 @@ definitions:
     properties:
       amount:
         type: string
-        pattern: "^[0-9]{1,20}$"
+        pattern: "^[1-9][0-9]{0,19}$"
       condition:
         type: object
         additionalProperties: false


ttmc:
As it happens, there is already a unit test which checks a transaction with an output amount of 0, except it expects that to be okay, and so the test is failing in Travis CI. For example, see https://travis-ci.org/bigchaindb/bigchaindb/jobs/408487035

ttmc:
I'm closing this pull request and will create another issue because there are already some checks on the size of amount elsewhere in the code. For example, as of today, there's code in `bigchaindb/bigchaindb/common/transaction.py` which says:

```python
        if amount < 1:
            raise AmountError('`amount` must be greater than 0')
        if amount > self.MAX_AMOUNT:
            raise AmountError('`amount` must be <= %s' % self.MAX_AMOUNT)
```

PR #1823, opened 2017-11-09
 Add Web Server bindings to post a transaction to Tendermint

vrde:
Free goodies: `GET` transactions work as well!!1!

Merge #1822 first.

PR #2413, opened 2018-07-26
Problem: websocket might be unresponsive

vrde:
Solution: It might be related on how aiohttp 2.x handles `send_str`.
AFAIK the call `send_str` was not "fully async". In aiohttp 3.x they
[changed it][1]:

> send_str(), send_bytes(), send_json(), ping() and pong() are genuine
async functions now. (#2475)

So this patch adds support for aiohttp 3.x, and uses the `send_str`
coroutine.

[1]: https://github.com/aio-libs/aiohttp/blob/master/HISTORY.rst#deprecations-and-removals-1

### DIFF
--- a/bigchaindb/web/websocket_server.py
+++ b/bigchaindb/web/websocket_server.py
@@ -105,7 +105,7 @@ class Dispatcher:
 
             for _, websocket in self.subscribers.items():
                 for str_item in str_buffer:
-                    websocket.send_str(str_item)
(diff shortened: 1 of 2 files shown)


vrde:
Related issue: [ What happens when a WebSocket Server is publishing faster than a Client? #1823](https://github.com/aio-libs/aiohttp/issues/1823).

PR #2415, opened 2018-07-27
Reenable test bigchain api - part 1

z-bowen:
**Problem: Tests in class `TestBigChainAPI` from `tests/db/test_bigchain_api.py` were disabled**

## Solution

Fixed the tests that still made sense, and deleted the rest.


tests/conftest.py
```diff
@@ -325,17 +325,20 @@ def _get_height(b):
 def inputs(user_pk, b, alice):
     from bigchaindb.models import Transaction
     # create blocks with transactions for `USER` to spend
-    for block in range(4):
+    print('Fuck you')
```
kansi: I think we should remove this.

tests/db/test_bigchain_api.py
```diff
@@ -7,196 +7,73 @@
...
-@pytest.mark.skipif(reason='Some tests throw a ResourceWarning that might result in some weird '
-                           'exceptions while running the tests. The problem seems to *not* '
-                           'interfere with the correctness of the tests. ')
-def test_remove_unclosed_sockets():
-    pass
-
-
 class TestBigchainApi(object):
 
     @pytest.mark.genesis
```
kansi: The `genesis` marker should not be needed anymore

tests/conftest.py
z-bowen: XD

tests/conftest.py
```diff
@@ -325,17 +325,19 @@ def _get_height(b):
 def inputs(user_pk, b, alice):
     from bigchaindb.models import Transaction
     # create blocks with transactions for `USER` to spend
-    for block in range(4):
+    for height in range(1, 4):
```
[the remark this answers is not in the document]
kansi: height `0` is assumed for genesis block although change to `range(0,3)` shouldn't matter either

tests/db/test_bigchain_api.py
```diff
@@ -7,196 +7,68 @@
...
+        with pytest.raises(OperationError):
+            b.store_bulk_transactions([transfer_tx2])
 
-        # Vote both block2 and block3 valid
-        vote = b.vote(block2.id, b.get_last_voted_block().id, True)
-        b.write_vote(vote)
-        vote = b.vote(block3.id, b.get_last_voted_block().id, True)
-        b.write_vote(vote)
-
-        with pytest.raises(CriticalDoubleInclusion):
```
ldmberman: Now when you updated these tests you can remove `CriticalDoubleInclusion` from `bigchaindb.exceptions`.

tests/db/test_bigchain_api.py
```diff
@@ -7,196 +7,68 @@
...
-        b.write_block(block2)
+        b.store_bulk_transactions([transfer_tx])
 
-        monkeypatch.setattr('time.time', lambda: 1000000030)
         transfer_tx2 = Transaction.transfer(tx.to_inputs(), [([alice.public_key], 1)],
                                             asset_id=tx.id)
         transfer_tx2 = transfer_tx2.sign([alice.private_key])
-        block3 = b.create_block([transfer_tx2])
-        b.write_block(block3)
+        with pytest.raises(OperationError):
```
ldmberman: Judging by the name and the previous code of the test, we need to test that `BigchainDB.get_spent` raises`CriticalDoubleSpend` here.

tests/db/test_bigchain_api.py
```diff
@@ -7,196 +7,68 @@
...
-        block3 = b.create_block([transfer_tx2])
-        b.write_block(block3)
 
-        # Vote both block2 and block3 valid
-        vote = b.vote(block2.id, b.get_last_voted_block().id, True)
-        b.write_vote(vote)
-        vote = b.vote(block3.id, b.get_last_voted_block().id, True)
-        b.write_vote(vote)
+        with pytest.raises(DoubleSpend):
+            b.validate_transaction(transfer_tx2, [transfer_tx])
```
ldmberman: Also judging by the previous state, we should test how `BigchainDB.get_spent` raises `DoubleSpend` when called with `current_transactions=[transfer_tx]`.

tests/db/test_bigchain_api.py
z-bowen: No, this isn't a test for a double spend. It's testing that you can't have two txs with the same tx_id

tests/db/test_bigchain_api.py
```diff
@@ -7,196 +7,68 @@
...
-                                 metadata={'msg': 2})
-        tx2 = tx2.sign([alice.private_key])
-        block2 = b.create_block([tx2])
-        b.write_block(block2)
+        b.store_bulk_transactions([tx])
 
-        # vote the first block invalid
-        vote = b.vote(block1.id, b.get_last_voted_block().id, False)
-        b.write_vote(vote)
+        with pytest.raises(OperationError):
```
ldmberman: Does this test differ somehow from the one before the one above?

tests/db/test_bigchain_api.py
```diff
@@ -232,236 +99,22 @@ def test_text_search(self, b, alice):
...
-
-        assert len(block['block']['transactions']) == 1
-        assert block['block']['transactions'][0]['operation'] == 'GENESIS'
-        assert block['block']['transactions'][0]['inputs'][0]['fulfills'] is None
-
-    @pytest.mark.genesis
-    def test_create_genesis_block_fails_if_table_not_empty(self, b):
-        from bigchaindb.common.exceptions import GenesisBlockAlreadyExistsError
-
-        with pytest.raises(GenesisBlockAlreadyExistsError):
```
ldmberman: This exception can be removed from `bigchaindb/common/exceptions.py` now.

tests/db/test_bigchain_api.py
```diff
@@ -232,236 +99,22 @@ def test_text_search(self, b, alice):
...
-            tx = Transaction.create([alice.public_key], [([user_pk], 1)],
-                                    metadata={'msg': i}) \
-                            .sign([alice.private_key])
-            b.write_transaction(tx)
-
-        assert query.count_backlog(b.connection) == 4
+            tx.validate(b)
 
 
 class TestTransactionValidation(object):
```
ldmberman: There are tests still disabled starting from this one. Do you plan to enable them in a separate PR?

tests/db/test_bigchain_api.py
z-bowen: Cool! Done.

tests/db/test_bigchain_api.py
ldmberman: It's called `test_get_spent_with_double_inclusion_detected` and it was calling `b.get_spent(tx.id, 0)` expecting it to raise `CriticalDoubleInclusion`, which is like double spend that somehow sneaked into the database. Currently, `BigchainDB.get_spent` raises `core_exceptions.CriticalDoubleSpend` and `DoubleSpend`, which is what we by analogy need to test now.

tests/db/test_bigchain_api.py
z-bowen: Other than that the first test is for a transfer and this is for a new asset, no. I don't know if that's a significant enough difference to justify a second test, but it's already written, so...

tests/db/test_bigchain_api.py
z-bowen: Yes. I'll keep working on this until I've done all the tests in this file, but this seemed like a natural break point where I could commit some functionality.

tests/db/test_bigchain_api.py
z-bowen: Done

tests/db/test_bigchain_api.py
kansi: It seems that this test tries to brute force storing a duplicate transaction and the above one tries to check if the double spend transaction in fact raises an exception when `validate_transaction` is called.

tests/db/test_bigchain_api.py
codegeschrei: What works for me is keeping everything as it is. I change all the block creation to a `store_bulk_transaction` and in the end when I call 
```
with pytest.raises(CriticalDoubleSpend):
     b.get_spent(tx.id, 0)
```
the `CriticalDoubleSpend` is raised and the test passes.

tests/db/test_bigchain_api.py
```diff
@@ -216,12 +88,7 @@ def test_text_search(self, b, alice):
                                  asset=asset3).sign([alice.private_key])
 
         # create the block
```
codegeschrei: this needs to be removed or changed

z-bowen: Are you referring to the comment about creating the block, or something in the code?

codegeschrei: the comment, we don't create a block here anymore

tests/db/test_bigchain_api.py
```diff
@@ -232,204 +99,22 @@ def test_text_search(self, b, alice):
...
-        # should only return one asset
-        assert len(assets) == 1
-        # should return the asset created by tx1
-        assert assets[0] == {
-            'data': {'msg': 'Hello BigchainDB!'},
-            'id': tx1.id
-        }
-
-    @pytest.mark.usefixtures('inputs')
-    def test_write_transaction(self, b, user_pk, user_sk):
```
codegeschrei: why did you remove this one?

tests/db/test_bigchain_api.py
```diff
@@ -232,204 +99,22 @@ def test_text_search(self, b, alice):
...
-                                  asset_id=input_tx.id)
-        tx = tx.sign([user_sk])
-        b.write_transaction(tx)
-
-        tx_from_db, status = b.get_transaction(tx.id, include_status=True)
-
-        assert tx_from_db.to_dict() == tx.to_dict()
-
-    @pytest.mark.usefixtures('inputs')
-    def test_read_transaction(self, b, user_pk, user_sk):
```
codegeschrei: same here, why is it removed?

tests/db/test_bigchain_api.py
codegeschrei: I think we should merge this one and make a part two PR instead

tests/db/test_bigchain_api.py
codegeschrei: I think later on in the file we test an actual DoubleSpend. With this it's doubled.

codegeschrei: But then this one will be doubled?

tests/db/test_bigchain_api.py
z-bowen: I replaced the first test. This one seems very similar to the first... Basically, we write something to the DB, get it back, and check that nothing has changed. I don't see a way to keep these tests distinct with BDB2.0

codegeschrei: I see. So where is the first test? You said you replaced it?

PR #2419, opened 2018-07-30
Problem: Docs fail to build

kansi:
Solution: ReadTheDocs cannot find the ValidatorElection yaml spec. The same
should be included in the setup.py so that file is available when BigchainDB is installed


ttmc:
Why can't all the JSON Schema files live in `bigchaindb/common/schema/`? That's where the current BigchainDB Transactions Spec v2 says they live:

https://github.com/bigchaindb/BEPs/blob/master/13/README.md#json-schema-validation

kansi:
The reason I put the spec inside `bigchaindb/upsert_validator` is so that feature is self-contained. Having said that I have no particular inclination for either of those paths. So, if you wish I can move the spec to `bigchaindb/common/schema`?

ttmc:
> So, if you wish I can move the spec to `bigchaindb/common/schema`?

Yes please.

PR #2422, opened 2018-07-31
Problem: Stateful validation doesn't raise double spend exception

kansi:
Solution: Transaction.validate should raise exception DoubleSpend if the given that the transaction is already a part of the database.

### DIFF
--- a/bigchaindb/models.py
+++ b/bigchaindb/models.py
@@ -59,7 +59,7 @@ class Transaction(Transaction):
 
                 spent = bigchain.get_spent(input_txid, input_.fulfills.output,
                                            current_transactions)
-                if spent and spent.id != self.id:
+                if spent:
                     raise DoubleSpend('input `{}` was already spent'
                                       .format(input_txid))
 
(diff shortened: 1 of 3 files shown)


bigchaindb/models.py
```diff
@@ -59,7 +59,7 @@ def validate(self, bigchain, current_transactions=[]):
 
                 spent = bigchain.get_spent(input_txid, input_.fulfills.output,
                                            current_transactions)
-                if spent and spent.id != self.id:
+                if spent:
```
kansi: This is a more strict condition. IMHO there is no need to match ids i.e. if there a txn which spends the current input then regardless of the transaction id it should be considered a double spend.

ttmc: Yes, because we also check to see if the current transaction (self) has more than one input trying to spend the same output (below). Actually, the line:

`raise DoubleSpend('tx "{}" spends inputs twice'.format(self.id))`

should say

`raise DoubleSpend('tx "{}" spends the same output more than once'.format(self.id))`

Maybe make that change in this PR as well, because it helps to make the logic more clear.

ttmc:
I don't know what Codecov is upset about. I'm going to ignore it.

PR #2423, opened 2018-07-31
Problem: No wait when getting value from queue

kansi:
Solution: Random failure of test `tests/test_events.py::test_event_handler_raises_when_called_after_start` seem to be because of `get_nowait` which returns way to quickly even when the queue
has data.

### DIFF
--- a/bigchaindb/events.py
+++ b/bigchaindb/events.py
@@ -67,7 +67,7 @@ class Exchange:
         """
 
         try:
-            self.started_queue.get_nowait()
+            self.started_queue.get(timeout=0.01)
             raise RuntimeError('Cannot create a new subscriber queue while Exchange is running.')
         except Empty:
             pass


ttmc:
Could you say a bit more? When you described this to me in person, you said that `get_nowait()` was returning the wrong exception?

bigchaindb/events.py
```diff
@@ -67,7 +67,7 @@ def get_subscriber_queue(self, event_types=None):
         """
 
         try:
-            self.started_queue.get_nowait()
+            self.started_queue.get(timeout=0.01)
```
ldmberman: A bigger timeout (say, a second) won't hurt - a successful build will take the same amount of time anyway while  if there is an issue, waiting a second is not a big deal.

kansi: I really don't know what would be an appropriate value of this timeout i.e. 10ms, 100ms, 1sec? But I do understand the having no timeout at all would lead us back to the situation wherein the test failed randomly. Also one must keep in mind that while the timeout will help the test case it will also impact the code performance.

ldmberman: Since CI often fails for us here it makes sense to stay on the safe side and set a second. It only impacts the performance of the unsuccessful case. Do you expect failed retrievals to happen a lot? I suppose, we should not expect them to happen often so it's ok to wait for a second if it happens.

kansi: > Do you expect failed retrievals to happen a lot?

The only place where `get_subscriber_queue` is called in the code base is [here](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/start.py#L53). So it seems that it might not effect much.

PR #2424, opened 2018-07-31
Stand-alone container for a BigchainDB node

### LINKED ISSUE #2238: Problem: bigchaindb and tendermint are not packaged as a single service unit
bigchaindb version: `v2.0.0a2`
tendermint version: `0.12`
setup: `4 node network with docker-compose`

# Description
Tendermint polls to bigchaindb pyabci interface only on startup. So if at any point bigchaindb service crashes and restarts tendermint never tries to re-establish connection with bigchaindb, and bigchaindb in absence of a connection request form tendermint crashes again.

This tight coupling between tendermint and bigchaindb services requires them to be packaged together as a single unit to improve the crash recovery mechanism. 

# Solution
 - package tendermint binary as part of bigchaindb python package. This binary will be installed as part of bigchaindb package. (This needs to be scoped yet)
 - Add supervisord configs for spawning bigchaindb and tendermint services together and restart both in the expected order if one crashes. Supervisord will allow similar behavior across platforms and can be packaged with un-privileged docker containers as compared to other process managers like systemd which are distribution specific and can only work with privileged containers.

muawiakh:
## Background

- Some customers have asked for a single stand-alone BigchainDB image.
- Easier for early users of BigchainDB

## Solution

- Make BigchainDB, MongoDB and Tendermint be part of the same container, even though it is an anti-pattern but this has been requested at several instances by customers/partners.
- Easier for early adopters of BigchainDB i.e. 2 step approach:
  - docker pull
  - docker run

## TODOs

- [x] Add documentation of this workflow.
- [ ] Tag image once this PR is merged.

## Issues this Resolves

Resolves #2238 (or at least I think it does. -Troy)

### DIFF
--- a/Dockerfile-all-in-one
+++ b/Dockerfile-all-in-one
@@ -1,4 +1,4 @@
-FROM alpine:latest
+ FROM alpine:latest
 LABEL maintainer "dev@bigchaindb.com"
 
 ARG TM_VERSION=0.22.8
@@ -54,6 +54,5 @@ VOLUME /data/db /data/configdb /tendermint
 
 EXPOSE 27017 28017 9984 9985 26656 26657 26658
 
-COPY pkg/scripts/all-in-one.bash /usr/src/app/
 WORKDIR $HOME
-ENTRYPOINT ["/usr/src/app/all-in-one.bash"]
+ENTRYPOINT ["/usr/src/app/pkg/scripts/all-in-one.bash"]
--- a/Dockerfile-all-in-one
+++ b/Dockerfile-all-in-one
@@ -1,4 +1,4 @@
- FROM alpine:latest
+FROM alpine:latest
 LABEL maintainer "dev@bigchaindb.com"
 
 ARG TM_VERSION=0.22.8
@@ -13,7 +13,7 @@ RUN apk --update add sudo bash \
         libffi-dev openssl-dev build-base jq \
     && apk add --no-cache libstdc++ dpkg gnupg \
     && pip3 install --upgrade pip cffi \
-    && pip install --no-cache-dir --process-dependency-links -e .[dev] \
+    && pip install --no-cache-dir --process-dependency-links . \
     && apk del build-dependencies \
     && rm -f /var/cache/apk/*
 
@@ -34,10 +34,6 @@ RUN addgroup tmuser \
 # Set permissions required for mongodb
 RUN mkdir -p /data/db /data/configdb \
         && chown -R mongodb:mongodb /data/db /data/configdb
-
-# When developing with Python in a docker container, we are using PYTHONBUFFERED
-# to force stdin, stdout and stderr to be totally unbuffered and to capture logs/outputs
-ENV PYTHONUNBUFFERED 0
 
 # BigchainDB enviroment variables
 ENV BIGCHAINDB_DATABASE_PORT 27017
--- a/Dockerfile-all-in-one
+++ b/Dockerfile-all-in-one
@@ -13,7 +13,7 @@ RUN apk --update add sudo bash \
         libffi-dev openssl-dev build-base jq \
     && apk add --no-cache libstdc++ dpkg gnupg \
     && pip3 install --upgrade pip cffi \
-    && pip install --no-cache-dir --process-dependency-links . \
+    && pip install --no-cache-dir --process-dependency-links -e . \
     && apk del build-dependencies \
     && rm -f /var/cache/apk/*
 
@@ -27,9 +27,6 @@ RUN wget https://github.com/tendermint/tendermint/releases/download/v${TM_VERSIO
     && rm tendermint_${TM_VERSION}_linux_amd64.zip
 
 ENV TMHOME=/tendermint
-
-RUN addgroup tmuser \
-    && adduser -S -G tmuser tmuser -h "$TMHOME"
 
 # Set permissions required for mongodb
 RUN mkdir -p /data/db /data/configdb \
--- a/Dockerfile-all-in-one
+++ b/Dockerfile-all-in-one
@@ -1,0 +1,59 @@
+FROM alpine:latest
+LABEL maintainer "dev@bigchaindb.com"
+
+ARG TM_VERSION=0.22.8
+RUN mkdir -p /usr/src/app
+ENV HOME /root
+COPY . /usr/src/app/
+WORKDIR /usr/src/app
+
+RUN apk --update add sudo bash \
+    && apk --update add python3 openssl ca-certificates git \
+    && apk --update add --virtual build-dependencies python3-dev \
+        libffi-dev openssl-dev build-base jq \
+    && apk add --no-cache libstdc++ dpkg gnupg \
+    && pip3 install --upgrade pip cffi \
+    && pip install --no-cache-dir --process-dependency-links -e .[dev] \
+    && apk del build-dependencies \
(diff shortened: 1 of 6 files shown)


docs/server/source/appendices/all-in-one-bigchaindb.md
```diff
@@ -0,0 +1,84 @@
+# Run BigchainDB with all-in-one Docker
+
+**NOT for Production Use**
```
ldmberman: What are the production use criteria? We do not have such warnings in the docs of other setups.

docs/server/source/appendices/all-in-one-bigchaindb.md
```diff
@@ -0,0 +1,84 @@
...
+Docker image and a
+`Dockerfile-all-in-one` that can be used to build an image for `bigchaindb`.
+
+This image contains all the services required for a BigchainDB node i.e.
+
+- BigchainDB Server
+- MongoDB
+- Tendermint
+
+*We understand that it is an anti-pattern to run multiple services inside a docker.*
```
ldmberman: It makes sense to recommend a multi-image alternative, add a link to its description, and briefly mention its benefits. `it is an anti-pattern` is not really an explanation so I would omit it.

Dockerfile-all-in-one
```diff
@@ -0,0 +1,58 @@
...
+RUN addgroup tmuser \
+    && adduser -S -G tmuser tmuser -h "$TMHOME"
+
+# Set permissions required for mongodb
+RUN mkdir -p /data/db /data/configdb \
+        && chown -R mongodb:mongodb /data/db /data/configdb
+
+# When developing with Python in a docker container, we are using PYTHONBUFFERED
+# to force stdin, stdout and stderr to be totally unbuffered and to capture logs/outputs
+ENV PYTHONUNBUFFERED 0
```
ldmberman: Where is this env variable used?

kansi:
executed `docker build - < Dockerfile-all-in-on`, got the following error

```
  Running setup.py install for pycparser: started
    Running setup.py install for pycparser: finished with status 'done'
  Running setup.py install for cffi: started
    Running setup.py install for cffi: finished with status 'done'
Successfully installed cffi-1.11.5 pip-18.0 pycparser-2.18
Directory '.' is not installable. File 'setup.py' not found.
The command '/bin/sh -c apk --update add sudo bash     && apk --update add python3 openssl ca-certificates git     && apk --update add --virtual build-dependencies python3-dev         libffi-dev openssl-dev build-base jq     && apk add --no-cache libstdc++ dpkg gnupg     && pip3 install --upgrade pip cffi     && pip install --no-cache-dir --process-dependency-links -e .[dev]     && apk del build-dependencies     && rm -f /var/cache/apk/*' returned a non-zero code: 1
```

Dockerfile-all-in-one
```diff
@@ -0,0 +1,58 @@
+ FROM alpine:latest
```
kansi: Nitpick: Indentation is not right here.

Dockerfile-all-in-one
muawiakh: Actually this should be removed, since I will be removing:

`pip install --no-cache-dir --process-dependency-links -e .[dev]`

But this parameter is mostly required for dev use cases and since monit is maintaining the services, it does not make sense.

docs/server/source/appendices/all-in-one-bigchaindb.md
kansi: I think we all acknowledge that having all services running in a single container is an anti-pattern. This `all-in-one` container is just a convenience and we should have that warning here if not anywhere else.

ldmberman: `anti-pattern` is not really an explanation.

Moreover, any setup we have might be labeled "not for production" use at least for a reason that we do not have a stable release yet.

Therefore, to avoid disorienting the users I would refrain from such labels. Instead, we can add a recommendation to use a multi-image setup with a brief explanation on how it is better.

pkg/scripts/bigchaindb-monit-config
```diff
@@ -93,7 +93,7 @@ case \$1 in
   start_bigchaindb)
 
     pushd \$4
-      nohup bigchaindb start >> \$3/bigchaindb.out.log 2>> \$3/bigchaindb.err.log &
+      nohup bigchaindb -l DEBUG start >> \$3/bigchaindb.out.log 2>> \$3/bigchaindb.err.log &
```
kansi: Should the log level be a variable? if `bigchaindb-monit-config` is intended to run and monitor production system then user should be able to select the necessary/required log level.

docs/server/source/appendices/all-in-one-bigchaindb.md
kansi: > `anti-pattern` is not really an explanation.

Why?

muawiakh: `deployment anti-pattern`? I have updated the docs, maybe that is more generic?

ldmberman: > > anti-pattern is not really an explanation.

> Why?

`anti-pattern` is just a label.. Does such a setup perform worse? If yes, why? Is it more difficult to maintain such a setup? My point is, we need to describe why we recommend the multi-image setup.

On the other hand, I simply have doubts as to whether this approach has significant drawbacks. So I am asking to prove me wrong. We aim for our components to have a look and feel of a single program and might actually package them into a single binary in the near future - when we do it, there would be simply no choice but to have a single image.

muawiakh: Can you comment on the updated docs changes? I kept it more generic now. Also, anti-pattern implies, the suggested micro-service architecture that Docker introduces.

ldmberman: The doc changes are fine, thank you! Now the discussion is about removing the "production use" label. Docker is just a tool, the architecture should be determined by the use-cases.

kansi: > `anti-pattern` is just a label.. Does such a setup perform worse? If yes, why? Is it more difficult to maintain such a setup?

I have no experience regarding this. There are some [sources which suggest](https://blog.couchbase.com/docker-container-anti-patterns/) this is an anti-pattern but the docker documentation has [no mention](https://docs.docker.com/config/containers/multi-service_container/) of it.

I am not really sure if we can say anything about the production readiness of either of these approaches until we try them ourselves?

muawiakh: If someone wants to use just a single node in production then maybe yes, because in terms of performance or maintenance it is not a big overhead. If you want to have a network it will be more complex.

Maybe if we have dynamically adding validators implemented, i.e. you deploy one node with this and dynamically add using `upsert`, there are some questions yet to be answered. That is why I suggested not to be used in production.

ldmberman: Ah, makes sense!

In other words, a single-node cluster configuration is hardcoded in the image. It is not currently well suited for deploying a network. Could you, perhaps, add such a clarification?

ldmberman:
Created https://github.com/bigchaindb/bigchaindb/issues/2434, restarted the build.

PR #2430, opened 2018-08-01
Last pr before 2.0 beta 5

muawiakh:


setup.py
```diff
@@ -112,7 +112,7 @@ def check_setuptools_features():
     zip_safe=False,
     python_requires='>=3.6',
     classifiers=[
-        'Development Status :: 4 - Beta',
+        'Development Status :: 5 - Beta',
```
ttmc: This line should stay 'Development Status :: 4 - Beta'. You can find a list of all the allowed classifier values at:

https://pypi.org/pypi?%3Aaction=list_classifiers

CHANGELOG.md
```diff
@@ -18,6 +18,18 @@ For reference, the possible headings are:
...
+
+Tag name: v2.0.0b5
+
+### Changed
+
+* Supported version of Tendermint `0.22.3` -> `0.22.8`. [Pull request #2429](https://github.com/bigchaindb/bigchaindb/pull/2429).
+
+### Fixed
+
+* Raise DoubleSpend exception if transaction is already part of the database. [Pull request #2422](https://github.com/bigchaindb/bigchaindb/pull/2422).
```
ttmc: This isn't quite right. The DoubleSpend is now raised if there's any other transaction that spends the same output, _even if it has the same transaction ID_ as that "other" transaction.

ttmc: @kansi Is this right? Is this how you would describe the change?

PR #2433, opened 2018-08-02
Problem: get_validator doesn't sepcify height which leads to ambiguity

kansi:
Solution: `get_validator` should specify the `height` of the latest block witnessed by BigchainDB so that in case tendermint is ahead of BigchainDB correct validator set can be fetched.

### DIFF
--- a/bigchaindb/lib.py
+++ b/bigchaindb/lib.py
@@ -462,7 +462,10 @@ class BigchainDB(object):
 
     def get_validators(self):
         try:
-            resp = requests.get('{}validators'.format(self.endpoint))
+            block = self.get_latest_block()
+            query_args = ('?height=' + str(block['height'])) if block else ''
+
+            resp = requests.get('{}validators{}'.format(self.endpoint, query_args))
             validators = resp.json()['result']['validators']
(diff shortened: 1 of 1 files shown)


bigchaindb/lib.py
```diff
@@ -462,7 +462,10 @@ def fastquery(self):
 
     def get_validators(self):
         try:
-            resp = requests.get('{}validators'.format(self.endpoint))
+            block = self.get_latest_block()
```
ldmberman: Worth adding a test case in `tests/tendermint/test_integration.py`. Something like:

```
query.store_validator_update -> query.store_validator_update -> assert -> query.delete_validator_update  -> assert
```

What do you think?

kansi:
After some discussion with @ttmc and @ldmberman, I have come to the conclusion that we need to track the validator set inside BigchainDB i.e. a separate collection which tracks validator set at given heights is required. This is primarily because Tendermint during replay doesn't enable the RPC endpoints, hence re-validating an election transaction (which depends on knowing the validator set at a given height) would not be possible via Tendermint PRC endpoint `/validators?height=`

PR #2436, opened 2018-08-03
Problem: Validator set not tracked by BigchainDB

kansi:
Solution: BigchainDB depends on Tendermint's RPC API to get the validator set which is not available during replay so the validators set should be tracked inside BigchainDB.

NOTE: This PR will reduce coverage because some tests have been disabled as the new approach requires those tests to re-written. The same will be addressed in a future PR.

### DIFF
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -341,13 +341,6 @@ def store_pre_commit_state(connection, commit_id, state):
 
 
 @singledispatch
-def store_validator_update(conn, validator_update):
-    """Store a update for the validator set"""
-
-    raise NotImplementedError
-
-
-@singledispatch
 def get_pre_commit_state(connection, commit_id):
     """Get pre-commit state where `id` is `commit_id`.
 
@@ -362,14 +355,14 @@ def get_pre_commit_state(connection, commit_id):
 
 
 @singledispatch
-def get_validator_update(conn):
-    """Get validator updates which are not synced"""
+def store_validator_set(conn, validator_update):
+    """Store updated validator set"""
 
     raise NotImplementedError
 
 
 @singledispatch
-def delete_validator_update(conn, id):
-    """Set the sync status for validator update documents"""
+def get_validator_set(conn, height):
+    """Get validator set at which are not synced"""
 
     raise NotImplementedError
--- a/bigchaindb/backend/query.py
+++ b/bigchaindb/backend/query.py
@@ -363,6 +363,7 @@ def store_validator_set(conn, validator_update):
 
 @singledispatch
 def get_validator_set(conn, height):
-    """Get validator set at which are not synced"""
+    """Get validator set for a given `height`, if `height` is not specified
+    then return the latest validator set"""
 
     raise NotImplementedError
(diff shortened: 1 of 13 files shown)


bigchaindb/backend/query.py
```diff
@@ -362,14 +355,14 @@ def get_pre_commit_state(connection, commit_id):
...
+    """Store updated validator set"""
 
     raise NotImplementedError
 
 
 @singledispatch
-def delete_validator_update(conn, id):
-    """Set the sync status for validator update documents"""
+def get_validator_set(conn, height):
+    """Get validator set at which are not synced"""
```
ldmberman: The docstring is not very clear.

bigchaindb/core.py
```diff
@@ -176,3 +179,12 @@ def encode_validator(v):
     return Validator(pub_key=pub_key,
                      address=b'',
                      power=v['power'])
+
+
+def decode_validator(v):
+    validator = {'address': '', 'pub_key': {'type': '', 'data': ''}, 'voting_power': 0}
```
ldmberman: A neatpick:

```
return {
    'address': codecs.encode(v.address, 'hex').decode().upper().rstrip('\n'),
    'pub_key': {
         'type': v.pub_key.type,
         'data': codecs.encode(v.pub_key.data, 'base64').decode().rstrip('\n'),
    },
    'voting_power': v.power,
}
```

kansi: this is much cleaner, thanks.

bigchaindb/backend/query.py
```diff
@@ -362,14 +355,15 @@ def get_pre_commit_state(connection, commit_id):
 
 
 @singledispatch
-def get_validator_update(conn):
-    """Get validator updates which are not synced"""
+def store_validator_set(conn, validator_update):
```
muawiakh: This will be implemented once we have `upsert-validator`?

bigchaindb/core.py
```diff
@@ -42,11 +43,13 @@ def __init__(self, bigchaindb=None):
...
         self.new_height = None
 
-    def init_chain(self, validators):
+    def init_chain(self, genesis):
         """Initialize chain with block of height 0"""
 
+        validator_set = [decode_validator(v) for v in genesis.validators]
         block = Block(app_hash='', height=0, transactions=[])
         self.bigchaindb.store_block(block._asdict())
+        self.bigchaindb.store_validator_set(1, validator_set)
```
muawiakh: we just store the validator in init? What about later?

bigchaindb/lib.py
```diff
@@ -484,6 +478,10 @@ def delete_validator_update(self):
     def store_pre_commit_state(self, state):
         return backend.query.store_pre_commit_state(self.connection, state)
 
+    def store_validator_set(self, height, validators):
```
muawiakh: docstring for this function?

bigchaindb/core.py
kansi: We store the validator set at a given height. In the above case since the chain is being initialized so we store the `validator_set` at height `1`.

bigchaindb/backend/query.py
kansi: This is already implemented [here](https://github.com/bigchaindb/bigchaindb/pull/2436/files#diff-069119e14639b0dfce674ffef65f812dR281)

PR #2437, opened 2018-08-03
Problem: Docs say little re/ privacy & handling private data

ttmc:
Solution:

- In the root docs, add a new page titled **BigchainDB, Privacy and Private Data**
- In the server docs, add a new page titled **Production Node Security & Privacy**
- Other minor copy-editing

### DIFF
--- a/docs/root/source/private-data.rst
+++ b/docs/root/source/private-data.rst
@@ -1,0 +1,100 @@
+BigchainDB, Privacy and Private Data
+------------------------------------
+
+Basic Facts
+===========
+
+#. One can store arbitrary data (including encrypted data) in a BigchainDB network, within limits: there’s a maximum transaction size. Every transaction has a ``metadata`` section which can store almost any Unicode string (up to some maximum length). Similarly, every CREATE transaction has an ``asset.data`` section which can store almost any Unicode string.
+#. The data stored in certain BigchainDB transaction fields must not be encrypted, e.g. public keys and amounts. BigchainDB doesn’t offer private transactions akin to Zcoin.
+#. Once data has been stored in a BigchainDB network, it’s best to assume it can’t be change or deleted.
+#. Every node in a BigchainDB network has a full copy of all the stored data.
+#. Every node in a BigchainDB network can read all the stored data.
+#. Everyone with full access to a BigchainDB node (e.g. the sysadmin of a node) can read all the data stored on that node.
+#. Everyone given access to a node via the BigchainDB HTTP API can find and read all the data stored by BigchainDB. The list of people with access might be quite short.
+#. If the connection between an external user and a BigchainDB node isn’t encrypted (using HTTPS, for example), then a wiretapper can read all HTTP requests and responses in transit.
+#. If someone gets access to plaintext (regardless of where they got it), then they can (in principle) share it with the whole world. One can make it difficult for them to do that, e.g. if it is a lot of data and they only get access inside a secure room where they are searched as they leave the room.
+
+Storing Private Data Off-Chain
+==============================
+
+A system could store data off-chain, e.g. in a third-party database, document store, or content management system (CMS) and it could use BigchainDB to:
+
+- Keep track of who has read permissions (or other permissions) in a third-party system. An example of how this could be done is described below.
+- Keep a permanent record of all requests made to the third-party system.
+- Store hashes of documents-stored-elsewhere, so that a change in any document can be detected.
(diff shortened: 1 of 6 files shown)


docs/root/source/private-data.rst
```diff
@@ -0,0 +1,100 @@
...
+#. Once data has been stored in a BigchainDB network, it’s best to assume it can’t be change or deleted.
+#. Every node in a BigchainDB network has a full copy of all the stored data.
+#. Every node in a BigchainDB network can read all the stored data.
+#. Everyone with full access to a BigchainDB node (e.g. the sysadmin of a node) can read all the data stored on that node.
+#. Everyone given access to a node via the BigchainDB HTTP API can find and read all the data stored by BigchainDB. The list of people with access might be quite short.
+#. If the connection between an external user and a BigchainDB node isn’t encrypted (using HTTPS, for example), then a wiretapper can read all HTTP requests and responses in transit.
+#. If someone gets access to plaintext (regardless of where they got it), then they can (in principle) share it with the whole world. One can make it difficult for them to do that, e.g. if it is a lot of data and they only get access inside a secure room where they are searched as they leave the room.
```
gautamdhameja: is this needed? I guess it's implicit.

docs/root/source/private-data.rst
```diff
@@ -0,0 +1,100 @@
...
+
+To revoke the read permission, DocPile could create a TRANSFER transaction, to spend the one output on the original CREATE transaction, with a metadata field to say that the user in question no longer has read permission on that document.
+
+This can be carried on indefinitely, i.e. another TRANSFER transaction could be created by DocPile to indicate that the user now has read permissions again.
+
+DocPile can figure out if a given user has read permissions on a given document by reading the last transaction in the CREATE → TRANSFER → TRANSFER → etc. chain for that user+document pair.
+
+There are other ways to accomplish the same thing. The above is just one example.
+
+You might have noticed that the above example didn’t treat the “read permission” as an asset owned (controlled) by a user. The reason is that read permissions must be controlled by DocPile or a document owner, not by the user.
```
gautamdhameja: we can add,

```
... because if the permission asset is given to (transferred to or created by) the user then it cannot be controlled any further until the user transfers it back to DocPile.
```

docs/root/source/private-data.rst
```diff
@@ -0,0 +1,100 @@
...
+
+If k < N of the key-holders gets and decrypts k of the pieces, they can reconstruct the original plaintext. Less than k would not be enough.
+
+System Example 4
+~~~~~~~~~~~~~~~~
+
+This setup could be used in an enterprise blockchain scenario where a special node should be able to see parts of the data, but the others should not.
+
+- The special node generates an X25519 keypair (or similar asymmetric *encryption* keypair).
+- A BigchainDB end user finds out the X25519 public key (encryption key) of the special node.
```
gautamdhameja: `finds out` can be `queries`. just to make it clear that the encryption public key of the node will be published by the node.

docs/root/source/private-data.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Below we describe some example system setups, using various crypto primitives, to give a sense of what’s possible.
+
+Please note:
+
+- Ed25519 keypairs are designed for signing and verifying cryptographic signatures, `not for encrypting and decrypting messages <https://crypto.stackexchange.com/questions/27866/why-curve25519-for-encryption-but-ed25519-for-signatures>`_. For encryption, you should use keypairs designed for encryption, such as X25519.
+- If someone (or some group) publishes how to decrypt some encrypted data on-chain, then anyone with access to that encrypted data will be able to get the plaintext. The data can’t be deleted.
+- Encrypted data can’t be indexed or searched by MongoDB. (It can index and search the ciphertext, but that’s not very useful.) One might use homomorphic encryption to index and search encrypted data, but MongoDB doesn’t have any plans to support that any time soon.
```
gautamdhameja: We can add,

```
if there is indexing or keyword search needed, then some fields of the asset.data or metadata objects can be left as plain text and the sensitive information can be stored in an encrypted child-object
```

docs/root/source/private-data.rst
ttmc: It's something that I've noticed people often raise as an issue, so I figured it would be worthwhile to mention it explicitly, lest they think we never thought about it.

docs/root/source/private-data.rst
ttmc: :+1:

docs/root/source/private-data.rst
ttmc: I wanted to leave it vague (how the end user finds out the key). Maybe they get it from the node via SMS, Slack or fax.

docs/root/source/private-data.rst
ttmc: Good point!

PR #2439, opened 2018-08-06
Problem: k8s docs not clear about who they're for

ttmc:
Solution: Add a note atop k8s docs pages to clarify who they're for

Notes:

- This PR mostly adds a "Note" box at the top of several k8s docs pages, to clarify who those docs are for (and not for).
- It also does some minor edits of other text, so the reader doesn't get the impression that this is how _we_ do deployment (in the future).
- I wasn't able to _link_ to the page titled **How to Set Up a BigchainDB Network** because that page is written in Markdown. I didn't want to link to a specific URL, because it should link to the **How to Set Up a BigchainDB Network** page in whatever version of the docs that the reader is currently reading.
- I intend to rename the **Production Deployment Template** page to **Kubernetes Deployment Template** and do move it later in the contents, but I'll do that in a follow-up PR. I didn't want to do that in this PR because it wouldn't have been as clear what I changed in the text of the k8s docs.

### DIFF
--- a/docs/server/source/production-deployment-template/architecture.rst
+++ b/docs/server/source/production-deployment-template/architecture.rst
@@ -1,13 +1,25 @@
-Architecture of a BigchainDB Node
-==================================
+Architecture of a BigchainDB Node Running in a Kubernetes Cluster
+=================================================================
 
-A BigchainDB Production deployment is hosted on a Kubernetes cluster and includes:
+.. note::
+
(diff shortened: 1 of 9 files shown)


muawiakh:
Will merge once travis is 🥗

ttmc:
@muawiakh Travis is now :green_salad:

PR #2445, opened 2018-08-08
Problem: Cannot conclude validator election

kansi:
Solution: Count validator election votes and conclude election when a supermajority is achieved for the first time.


bigchaindb/core.py
```diff
@@ -134,9 +135,13 @@ def end_block(self, request_end_block):
 
         # TODO: calculate if an election has concluded
         # NOTE: ensure the local validator set is updated
-        # validator_updates = self.bigchaindb.get_validator_update()
-        # validator_updates = [encode_validator(v) for v in validator_updates]
-        validator_updates = []
+        validator_updates = self.bigchaindb.get_validator_update(self.block_transactions)
```
ldmberman: The logic of figuring out the new validator set is currently spread across `BigchainDB.get_validator_update`, `new_validator_set`, and `ValidatorElection.conclude`, what makes it hard to reason about how to implement new types of validator set updates in the future.

What do you think about the following alternative?

```
validator_set = get_validator_set()
new_validator_set = ValidatorElection.get_new_validator_set(txs, validator_set)
store_validator_set(new_validator_set)
```
This way, all the details of deciding how the new validator set should look like based on the block are hidden inside the method specific to this particular election process.

It is then extendable in a way that multiple processes can be stacked up:

```
yet_new_validator_set = RemoveValidatorsElection.get_new_validator_set(txs, new_validator_set)
store_validator_set(yet_new_validator_set)
```

bigchaindb/backend/localmongodb/query.py
```diff
@@ -301,3 +303,15 @@ def get_validator_set(conn, height=None):
     )
 
     return list(cursor)[0]
+
+
+@register_query(LocalMongoDBConnection)
+def get_received_votes_for_election(conn, election_id, election_public_key):
```
ldmberman: A more extensible approach might be to use `get_filtered_txids` and check for public keys outside. Otherwise, the database interface gets concerned with the election logic (in this case the logic of what is the election vote). What do you think?

bigchaindb/lib.py
```diff
@@ -438,15 +440,22 @@ def fastquery(self):
...
-    def get_validator_update(self):
-        update = backend.query.get_validator_update(self.connection)
-        return [update['validator']] if update else []
+    def get_validator_update(self, txns):
+        votes = {}
+        for txn in txns:
+            if isinstance(txn, ValidatorElectionVote):
+                election_id = txn.asset['id']
+                election_votes = votes.get(election_id, [])
+                votes[election_id] = election_votes.append(txn)
```
ldmberman: You are assigning `None` to `votes[election_id]` here. In case there are two votes in the block and the second one concludes the election, this code will raise an `AttributeError`. Makes sense to add the corresponding test case.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -114,7 +117,7 @@ def validate(self, bigchain, current_transactions=[]):
         if not self.is_same_topology(current_validators, self.outputs):
             raise UnequalValidatorSet('Validator set much be exactly same to the outputs of election')
 
-        return True
+        return self
```
ldmberman: Why did you change it?

Do not forget to update the docstring as well or perhaps remove that part of the docstring to avoid the outdated documentation.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -141,3 +144,59 @@ def create(cls, tx_signers, recipients, metadata=None, asset=None):
...
         raise NotImplementedError
+
+    @classmethod
+    def to_public_key(cls, election_id):
+        return base58.b58encode(bytes.fromhex(election_id))
+
+    @classmethod
+    def count_votes(cls, election_pk, txns):
+        votes = 0
+        for txn in txns:
```
ldmberman: A way to not duplicate the checks is to accept a getter.

```
def count_votes(cls, election_pk, txns, getter=getattr):
    ...
    if getter(txn, 'operation') == 'VALIDATOR_ELECTION_VOTE':
       ...
```
Then you just call `count_votes(election_pk, txns)` for objects and `count_votes(election_pk, txns, operator.itemgetter)` for dictionaries.

On a side note, it would be nice to deal with dictionaries everywhere in the future.

bigchaindb/backend/localmongodb/query.py
kansi: The only thing that makes this function a little specialized is including `operation: 'VALIDATIOR_ELECTION_VOTE'`. Do you want me to make that as a variable?

ldmberman: `operation` is already a variable of `get_filtered_txids` btw. `get_received_votes_for_election` is "special" in a sense that different types of elections can have different ways of identifying their votes.

bigchaindb/core.py
kansi: > The logic of figuring out the new validator set is currently spread across BigchainDB.get_validator_update, new_validator_set

I agree. But the approach you suggested has some issues i.e. `ValidatorElection.get_new_validator_set(txs, validator_set)` internally would look as follows (in `upsert_validator/validator_election.py`),

```python
        votes = {}
        for txn in txns:
            if isinstance(txn, ValidatorElectionVote):
                election_id = txn.asset['id']
                election_votes = votes.get(election_id, [])
                votes[election_id] = election_votes.append(txn)

                election = ValidatorElection.conclude(self, election_id, election_votes)
                # Once an election concludes any other conclusion for the same
                # or any other election is invalidated
                if election:
                    return [election.asset['data']]
        return []
```

The issue is that I cannot use `ValidatorElectionVote` because of import issues. The only way out I see is to create a base election class?

ldmberman: What if we make `get_new_validator_set` a method of `ValidatorElectionVote` instead? Then it should be fine, right?

kansi: > election = ValidatorElection.conclude(self, election_id, election_votes)

Tried it, could not include `ValidatorElection` class there because of import error.

bigchaindb/backend/query.py
```diff
@@ -370,11 +370,13 @@ def get_validator_set(conn, height):
 
 
 @singledispatch
-def get_received_votes_for_election(connection, election_id, election_public_key):
+def get_asset_tokens_for_public_keys(connection, asset_id,
+                                     public_keys, operation):
     """Retrieve a list of `VALIDATOR_ELECTION_VOTE`s `txids` that are owned by `owner`.
```
ldmberman: The docstring requires an update.

bigchaindb/core.py
```diff
@@ -135,13 +133,14 @@ def end_block(self, request_end_block):
 
         # TODO: calculate if an election has concluded
```
ldmberman: Don't forget to remove the TODO/comment leftovers.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -141,3 +146,86 @@ def create(cls, tx_signers, recipients, metadata=None, asset=None):
...
+                   (votes_commited + votes_current > (2/3)*total_votes):
+                    return election
+
+        return False
+
+    @classmethod
+    def get_validator_update(cls, bigchain, new_height, txns):
+        votes = {}
+        for txn in txns:
+            if isinstance(txn, ValidatorElectionVote):
```
ldmberman: A neatpick - it might be slightly easier to read the code that avoids nesting by reversing the condition checks.

```
for txn in txns:
    if not isintance .. :
        continue
    ....
    if not election:
        continue
    ....
```

bigchaindb/backend/query.py
```diff
@@ -372,10 +372,10 @@ def get_validator_set(conn, height):
...
-    """Retrieve a list of `VALIDATOR_ELECTION_VOTE`s `txids` that are owned by `owner`.
+    """Retrieve a list of tokens of type `asset_id` that are owned by `public_keys`.
     Args:
-        election_id (str): Id of the election.
-        election_public_key (str): base58 encoded public key of the election
+        asset_id (str): Id of the token.
+        public_keys (str): list of public keys
         operation: filter transaction based on `operation`
     Returns:
         Iterator of transaction that list given owner in conditions.
```
ldmberman: What does this line mean?

kansi: It means the those transactions are returned wherein the public keys are the owner of the tokens.

ldmberman: An alternative - "A list of transactions satisfying the specified criteria".

muawiakh:
👍  lgtm

bigchaindb/backend/query.py
```diff
@@ -372,3 +372,17 @@ def get_validator_set(conn, height):
     """
 
     raise NotImplementedError
+
+
+@singledispatch
+def get_asset_tokens_for_public_keys(connection, asset_id,
```
vrde: When do we need to query for multiple `public_keys`?

I see you use this function to extract all the votes, so the `operation` is in a way implicit with the `asset_id`.

I'm wondering if we can simplify it using a single `public_key` and removing `operation`.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -32,15 +37,15 @@ def __init__(self, operation, asset, inputs, outputs,
         super().__init__(operation, asset, inputs, outputs, metadata, version, hash_id)
 
     @classmethod
-    def current_validators(cls, bigchain):
+    def current_validators(cls, bigchain, height=None):
```
vrde: This method should belong to the `BigchainDB` class, what do you think?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -145,3 +150,79 @@ def create(cls, tx_signers, recipients, metadata=None, asset=None):
     @classmethod
     def transfer(cls, tx_signers, recipients, metadata=None, asset=None):
         raise NotImplementedError
+
+    @classmethod
+    def to_public_key(cls, election_id):
+        return base58.b58encode(bytes.fromhex(election_id))
+
+    @classmethod
+    def count_votes(cls, election_pk, txns, getter=getattr):
```
vrde: When do we need to change `getter`?

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -145,3 +150,79 @@ def create(cls, tx_signers, recipients, metadata=None, asset=None):
...
+        if election_pk is None:
+            election_pk = self.to_public_key(self.id)
+        txns = list(backend.query.get_asset_tokens_for_public_keys(bigchain.connection,
+                                                                   self.id,
+                                                                   [election_pk],
+                                                                   'VALIDATOR_ELECTION_VOTE'))
+        return self.count_votes(election_pk, txns, dict.get)
+
+    @classmethod
+    def conclude(cls, bigchain, txn_id, current_votes=[], height=None):
```
vrde: `conclude` sounds like an action, like you want to conclude the election right now, while this function is a check. I'd call it `is_concluded` or `has_concluded`.

I'm not sure about the semantic, the docstring says "Check if the given election", but there is no election explicitly passed to the function.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -282,13 +282,15 @@ def get_pre_commit_state(conn, commit_id):
...
-    try:
-        return conn.run(
-            conn.collection('validators')
-            .insert_one(validator_update))
-    except DuplicateKeyError:
-        raise MultipleValidatorOperationError('Validator update already exists')
+def store_validator_set(conn, validators_update):
+    height = validators_update['height']
+    return conn.run(
+        conn.collection('validators').replace_one(
```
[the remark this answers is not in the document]
kansi: Consider a situation wherein there is a crash after the validator set has been updated but before the response to `EndBlock` has been returned. In such a case Tendermint will try to replay `EndBlock` at which time `insert_one` will case an exception wherein `replace_one` re-write the same value.

bigchaindb/backend/query.py
kansi: Are you suggesting someting as follows?

```python
def get_asset_tokens_for_public_key(conn, asset_id, public_key):
    query = {'outputs.public_keys': [public_key],
             'asset.id': asset_id}

    cursor = conn.run(
        conn.collection('transactions').aggregate([
            {'$match': query},
            {'$project': {'_id': False}}
        ]))
    return cursor
```

bigchaindb/upsert_validator/validator_election.py
kansi: The `BigchainDB` already has a method `get_validators` this function is to just type cast the incomptabile formats of base58 and base64 public keys.

bigchaindb/upsert_validator/validator_election.py
kansi: The function can be called with either `dict` or `ValidatorElectionVote` class objects in `txns` list, so different `getattr`s are needed

bigchaindb/upsert_validator/validator_election.py
kansi: renamed to `has_concluded` and updated the doc string.

bigchaindb/backend/query.py
vrde: Yes, it seems simpler and easier to use.

bigchaindb/upsert_validator/validator_election.py
vrde: Can we make `BigchainDB::get_validators` return `base58` instead of `base64`? AFAIK `get_validators` is only used for changing the validator set.

I think we are still in time to change the interface. What do you think?

bigchaindb/upsert_validator/validator_election.py
vrde: I saw @ldmberman's comment after writing mine.

I don't like this solution much because we are delegating polymorphism to `getter`. I agree with @ldmberman that ideally we should deal only with dictionaries, and avoid custom objects.

Another idea is to have the class `Transaction` implement `__getitem__`.

Maybe we can keep it like this for now.

I'd still change `txns` to `transactions`.

bigchaindb/upsert_validator/validator_election.py
kansi: `get_validators` is used to list current validators in the `/api/v1/validators` API. We use base64 here so that users can actually cross verify the values in their Tendermint configs which are also base64 encoded.

vrde: Then why not adding an `encoding` parameter to `get_validators`? Also, `current_validators` should be called differently since it accepts a `height` parameter. `get_validators` would be a good name, that is the same we have in the `BigchainDB` class, that's why I'm pushing to use that methods. Moreover, this method is just taking the result of the method of another class, encoding the result in another way.

kansi: > Then why not adding an encoding parameter to `get_validators` ? 

I imagine that would be one of the tasks when standardizing the interface.

> Also, current_validators should be called differently since it accepts a height parameter. get_validators would be a good name, that is the same we have in the BigchainDB class, that's why I'm pushing to use that methods. Moreover, this method is just taking the result of the method of another class, encoding the result in another way.

:+1:

PR #2446, opened 2018-08-08
Create dynamic upsert validator commands

z-bowen:
**We need new methods to handle the dynamic upsert-validator elections**

## Solution

Created a new subcommand `new` for `upsert-validator` that initiates a new election

## BEPs Implemented

Related to BEP-21 and BEP-18


Dockerfile-dev
```diff
@@ -8,30 +7,3 @@ RUN apt-get update \
     && pip install pynacl \
     && apt-get autoremove \
     && apt-get clean
-
```
muawiakh: Why are we removing everything from here?

z-bowen: We shouldn't be. Vansh and I were screwing around trying to get my test env working, and I forgot to put these lines back in...

bigchaindb/utils.py
```diff
@@ -167,3 +167,15 @@ def run(self, instance):
 
         self.stack = []
         return last
+
+
+# Load Tendermint's public and private key from the file path
+def load_node_key(path):
+    import json
+    from bigchaindb.tendermint_utils import key_from_base64
+    from bigchaindb.common.crypto import key_pair_from_ed25519_key
```
kansi: Why are the imports done inside the function?

bigchaindb/commands/bigchaindb.py
```diff
@@ -95,21 +95,35 @@ def run_configure(args):
 
 @configure_bigchaindb
```
muawiakh: Do we need to set the `configure_bigchaindb` decorator here? Everything is already configured. Or am I missing something?

bigchaindb/commands/bigchaindb.py
```diff
@@ -95,21 +95,35 @@ def run_configure(args):
...
-    except MultipleValidatorOperationError:
-        logger.error('A validator update is pending to be applied. '
-                     'Please re-try after the current update has '
-                     'been processed.')
+    """Initiate and manage elections to change the validator set"""
+
+    b = BigchainDB()
+
+    # Call the function specified by args.action, as defined above
+    globals()[f'run_upsert_validator_{args.action}'](args, b)
```
muawiakh: Will there be any other action? Can you also add a docs string for what the args can be?

bigchaindb/commands/bigchaindb.py
```diff
@@ -208,16 +222,35 @@ def create_parser():
...
 
+    # parser for managing validator elections
     validator_parser = subparsers.add_parser('upsert-validator',
-                                             help='Add/update/delete a validator')
+                                             help='Add/update/delete a validator.')
+
+    validator_subparser = validator_parser.add_subparsers(title='Action',
+                                                          dest='action')
+
+    new_election_parser = validator_subparser.add_parser('new',
```
muawiakh: upsert already implies add/update/delete, should we mention `new`?

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
+    proposer_key = b.get_validators()[0]['pub_key']['value']
 
-    assert mock_store_validator_update.called
+    args = Namespace(action='new',
+                     public_key=proposer_key,
+                     power=1,
+                     node_id='12345',
+                     sk=priv_validator_path,
+                     config={})
+    print("*****ARGS*****")
```
muawiakh: left over?

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
+    print(args)
```
muawiakh: same

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
+    resp = run_upsert_validator_new(args, b)
+    print("*****TX ID*****")
+    print(resp)
```
muawiakh: same?

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
+    args = Namespace(action='new',
+                     public_key=proposer_key,
+                     power=1,
+                     node_id='12345',
+                     sk=priv_validator_path,
+                     config={})
+    print("*****ARGS*****")
+    print(args)
+    resp = run_upsert_validator_new(args, b)
+    print("*****TX ID*****")
```
muawiakh: applies to the remaining prints in the test and file.

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
+    print(resp)
+    time.sleep(3)
```
muawiakh: why is the sleep necessary? Is this a unit test or are we running it as an actual node?

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
...
-@pytest.mark.skip
-@patch('bigchaindb.config_utils.autoconfigure')
-@patch('bigchaindb.backend.query.store_validator_update')
-@pytest.mark.tendermint
-def test_upsert_validator(mock_autoconfigure, mock_store_validator_update):
-    from bigchaindb.commands.bigchaindb import run_upsert_validator
+# @pytest.mark.execute
+# @patch('bigchaindb.lib.BigchainDB.get_validators')
+# @pytest.mark.abci
+def test_upsert_validator_new_with_tendermint(b, priv_validator_path, user_sk, monkeypatch):
```
kansi: It would be good to explicitly skip this test with `@pytest.mark.skip`

tests/commands/test_commands.py
kansi: Please remove these print statements.

bigchaindb/utils.py
z-bowen: Moved to top

bigchaindb/commands/bigchaindb.py
```diff
@@ -208,16 +222,35 @@ def create_parser():
...
+    new_election_parser.add_argument('node_id',
+                                     help='The node_id of the validator.')
+
+    new_election_parser.add_argument('--private-key',
+                                     dest='sk',
+                                     help='Path to the private key of the election initiator.')
 
-    validator_parser.add_argument('public_key',
-                                  help='Public key of the validator.')
+    show_election_subparser = validator_subparser.add_parser('show',
```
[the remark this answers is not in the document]

bigchaindb/commands/bigchaindb.py
z-bowen: No idea. It's not part of my PR

bigchaindb/commands/bigchaindb.py
z-bowen: Yes. There is also going to be a `run_upsert_validator_show`, `_approve`, and `_status`, as detailed in BEP-21.

Added a docstring detailing the args

bigchaindb/commands/bigchaindb.py
z-bowen: I don't understand the question. `new` refers to a new election, not a new validator. Is this not clear from the help message/parser name?

tests/commands/test_commands.py

tests/commands/test_commands.py
z-bowen: Those are in a WIP test that is still in development, for diagnostic purposes. It's the next item on the agenda for this branch, so please just ignore them for now.

tests/commands/test_commands.py
z-bowen: This is a test intended to be run with an `abci` marker, so it will use a real Tendermint. The sleep is to give TM a chance to write the block

bigchaindb/commands/bigchaindb.py
```diff
@@ -95,21 +95,47 @@ def run_configure(args):
...
+
+    :param args: dict
+        args = {
+            'public_key': the public key of the proposed peer, (str)
+            'power': the proposed validator power for the new peer, (str)
+            'node_id': the node_id of the new peer (str)
+            'sk': the path to the private key of the node calling the election (str)
+        }
+    :param bigchain: an instance of BigchainDB
+    :return: election_id (tx_id)
```
codegeschrei: I don't know how detailed we want to be with doc strings. I like when there is a lot of info, so maybe you can add `:raises:` ?

z-bowen: Will do

bigchaindb/commands/bigchaindb.py
muawiakh: I see, didn't check the BEP before. looks good.

tests/commands/test_commands.py
```diff
@@ -341,15 +343,74 @@ def json(self):
         return {'result': {'latest_block_height': self.height}}
 
 
+# @pytest.mark.execute
+# @patch('bigchaindb.lib.BigchainDB.get_validators')
+# @pytest.mark.abci
```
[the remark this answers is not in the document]
z-bowen: *Need* is a strong word, but if I remove them, the first thing I'll do is put them back, so...

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -80,22 +80,27 @@ configuration file as documented under
 **This is an experimental feature. Users are advised not to use it in production.**
 
 
-Add, update, or remove a validator from the validators set of the local node. The command implements [3/UPSERT-VALIDATORS](https://github.com/bigchaindb/BEPs/tree/master/3), check it out if you need more details on how this is orchestrated.
+Manage elections to add, update, or remove a validator from the validators set of the local node. The commands implement [21/UPSERT-VALIDATORS](https://github.com/bigchaindb/BEPs/tree/master/21), check it out if you need more details on how this is orchestrated.
```
kansi: I think `Create elections to add ...` would be more appropriate here.

ttmc: I'd change "The commands implement [21/UPSERT-VALIDATORS]..." to "The upsert-validator subcommands implement [BEP-21]..."

(We now refer to BEPs with shortnames like BEP-21. See https://github.com/bigchaindb/BEPs/pull/51 )

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -80,22 +80,27 @@ configuration file as documented under
 **This is an experimental feature. Users are advised not to use it in production.**
```
ttmc: Can we remove this warning line now, or should we wait until later? If we should wait until later, maybe put it as another item near the end of the list in issue #2372

tests/commands/test_commands.py
```diff
@@ -341,15 +342,74 @@ def json(self):
...
 
-    args = Namespace(public_key='CJxdItf4lz2PwEf4SmYNAu/c/VpmX39JEgC5YpH7fxg=',
-                     power='10', config={})
-    run_upsert_validator(args)
 
-    assert mock_store_validator_update.called
+@pytest.mark.tendermint
+@pytest.mark.bdb
+def test_upsert_validator_new_without_tendermint(b, priv_validator_path, user_sk, monkeypatch):
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_new
```
ttmc: Why is there an import statement inside this function definition?

tests/commands/test_commands.py
```diff
@@ -341,15 +342,74 @@ def json(self):
...
-@patch('bigchaindb.config_utils.autoconfigure')
-@patch('bigchaindb.backend.query.store_validator_update')
-@pytest.mark.tendermint
-def test_upsert_validator(mock_autoconfigure, mock_store_validator_update):
-    from bigchaindb.commands.bigchaindb import run_upsert_validator
+def test_upsert_validator_new_with_tendermint(b, priv_validator_path, user_sk, monkeypatch):
+    """WIP: Will be fixed and activated in the next PR
+    """
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_new
+    import time
```
ttmc: Why is there an import statement inside this function definition?

docs/server/source/server-reference/bigchaindb-cli.md
kansi: I think we should wait, so I updated the issue description here #2372

tests/commands/test_commands.py
z-bowen: This is the code style used in all our other tests. I'm just trying to be consistent. I think it makes sense for tests, anyway. This way, if I run a single test, it depends on the minimal amount of imports. If I do all my imports at the top, then every test needs to import every module, whether it needs it or not.

tests/commands/test_commands.py
z-bowen: See my response above

docs/server/source/server-reference/bigchaindb-cli.md
z-bowen: We should certainly leave it until we have BEP-21 fully implemented. We can consider removing it when we have fully tested the new upsert-validator election process.

docs/server/source/server-reference/bigchaindb-cli.md

PR #2450, opened 2018-08-08
Problem: No SPDX license info in source files

ttmc:
Solution: Write a script to add such info to all source files

Notes:
- I ran the bash script (`add-spdx`) locally and it seems it worked fine. The unit tests all passed. The modified docs files all built fine and the rendered docs didn't show the added comments (as hoped).
- I didn't push the resulting changes (affecting over 250 files) in this pull request. This pull request just has the script, so it can be inspected for any problems. I'll push the modified files in another pull request.

### DIFF
--- a/add-spdx
+++ b/add-spdx
@@ -14,14 +14,36 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-# All files named *.py or *.sh or *.bash
-find . -type f \( -name "*.py" -o -name "*.sh" -o -name "*.bash" \) | while read fname; do
+# All files named *.py or *.sh or *.bash or *.yaml or *.yml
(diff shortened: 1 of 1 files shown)


ttmc:
I'm going to use the same script for other BigchainDB repositories, so:

- Please don't merge this pull request into the bigchaindb/bigchaindb repo; I'll put it somewhere else.
- I will add code for commenting JavaScript and Go code (`//`).
- I will add code for commenting Java code.

muawiakh:
We don't need this for files without extensions? Also the *.yaml files? I tried it and works for me.

ttmc:
I don't have time to go through the files-without-extensions one at a time (because the comment style depends on what it is). My thinking is that if we comment the files with known extensions, then that should be good enough. Technically, it's not required by copyright law; I'm doing it for another reason, as mentioned elsewhere.

I will also add comments to the YAML files.

PR #2452, opened 2018-08-09
Problem: several tests are skipped

codegeschrei:
Solution: activate or remove tests

#2381 

- [x] commands/test_utils.py (activate/remove tests)
- [x] store_transaction is deprecated, replace it in tests with store_bulk_transaction
- [x] unskipped one test in `test_core`


tests/tendermint/test_lib.py
```diff
@@ -39,10 +40,12 @@ def test_asset_is_separated_from_transaciton(b):
                             asset=asset)\
                     .sign([alice.private_key])
 
-    b.store_transaction(tx)
+    tx_dict = copy.deepcopy(tx.to_dict())
```
z-bowen: Is `deepcopy` really needed here? I assume `store_bulk_transactions` will just write the contents of `tx` to mongo. That should leave `tx` unchanged, and you can still perform the comparison in the last assert.

What breaks if you skip the `deepcopy`?

tests/web/test_block_tendermint.py
```diff
@@ -11,18 +11,20 @@
 @pytest.mark.bdb
 @pytest.mark.usefixtures('inputs')
 def test_get_block_endpoint(tb, client, alice):
+    import copy
     b = tb
     tx = Transaction.create([alice.public_key], [([alice.public_key], 1)], asset={'cycle': 'hero'})
     tx = tx.sign([alice.private_key])
-    b.store_transaction(tx)
+    tx_dict = copy.deepcopy(tx.to_dict())
```
z-bowen: Same question as above, regarding `deepcopy`

tests/tendermint/test_lib.py
```diff
@@ -176,54 +183,6 @@ def test_update_utxoset(tb, signed_create_tx, signed_transfer_tx, db_context):
     assert utxo['output_index'] == 0
 
 
-@pytest.mark.bdb
-def test_store_transaction(mocker, tb, signed_create_tx,
```
kansi: Why was this test removed?

codegeschrei: Because we use the `store_bulk_transactions` now. I can put it in again and switch the `store_transaction` to `store_transactions` if you want

kansi: @codegeschrei This test is also trying to assert if the transaction was appropriately split and the metadata and asset were stored in their respective collections. It would good to have this test back.

PR #2456, opened 2018-08-13
Problem: tests in test_bigchain_api are not activated

codegeschrei:
Solution: remove or active tests

#2381

### DIFF
--- a/tests/db/test_bigchain_api.py
+++ b/tests/db/test_bigchain_api.py
@@ -1,4 +1,3 @@
-from time import sleep
 from unittest.mock import patch
 
 import pytest
@@ -8,27 +7,6 @@ pytestmark = pytest.mark.bdb
 
 
 class TestBigchainApi(object):
-
-    @pytest.mark.tendermint
-    def test_get_spent_with_double_inclusion_detected(self, b, alice):
-        from bigchaindb.models import Transaction
-        from bigchaindb.backend.exceptions import OperationError
-
-        tx = Transaction.create([alice.public_key], [([alice.public_key], 1)])
-        tx = tx.sign([alice.private_key])
-
-        b.store_bulk_transactions([tx])
-
-        transfer_tx = Transaction.transfer(tx.to_inputs(), [([alice.public_key], 1)],
-                                           asset_id=tx.id)
-        transfer_tx = transfer_tx.sign([alice.private_key])
-        b.store_bulk_transactions([transfer_tx])
-
-        transfer_tx2 = Transaction.transfer(tx.to_inputs(), [([alice.public_key], 1)],
-                                            asset_id=tx.id)
-        transfer_tx2 = transfer_tx2.sign([alice.private_key])
-        with pytest.raises(OperationError):
-            b.store_bulk_transactions([transfer_tx2])
 
     @pytest.mark.tendermint
     def test_get_spent_with_double_spend_detected(self, b, alice):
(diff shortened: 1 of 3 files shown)


tests/db/test_bigchain_api.py
```diff
@@ -169,93 +151,31 @@ def test_non_create_valid_input_wrong_owner(self, b, user_pk):
...
-        b.write_transaction(signed_transfer_tx)
-        block = b.create_block([signed_transfer_tx])
-        b.write_block(block)
-
-        # vote block valid
-        vote = b.vote(block.id, b.get_last_voted_block().id, True)
-        b.write_vote(vote)
+        b.store_bulk_transactions([signed_create_tx, signed_transfer_tx])
 
         sleep(1)
```
ldmberman: Should be no need for this sleep, let's remove it.

tests/db/test_bigchain_api.py
```diff
@@ -9,27 +9,6 @@
 
 class TestBigchainApi(object):
 
-    @pytest.mark.tendermint
-    def test_get_spent_with_double_inclusion_detected(self, b, alice):
```
kansi: Could you please explain why this test was removed?

codegeschrei: because we don't have a double inclusion anymore. The test was supposed to test the exception `CriticalDoubleInclusion` that was removed in #2415. Now with the changes, everything that happens in this test, happens in some other tests too, so we don't need it anymore. If you have an idea how to test a double inclusion I'll gladly rewrite it :)

tests/db/test_bigchain_api.py
```diff
@@ -736,8 +542,9 @@ def test_cant_spend_same_input_twice_in_tx(b, genesis_block, alice):
         tx_transfer_signed.validate(b)
 
 
-@pytest.mark.bdb
+@pytest.mark.tendermint
 def test_transaction_unicode(b, alice):
```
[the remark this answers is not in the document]
codegeschrei: I don't know if it's valuable or not. But I do think it's good to check if unicode works.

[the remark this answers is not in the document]
ldmberman: The test does not even check it. The string we put into the database consists of the ASCII characters and the test does not assert they are interpreted as unicode.

tests/db/test_bigchain_api.py
```diff
@@ -62,7 +41,7 @@ def test_get_spent_with_double_spend_detected(self, b, alice):
             b.get_spent(tx.id, 0)
 
     @pytest.mark.tendermint
-    def test_get_block_status_for_tx_with_double_inclusion(self, b, alice):
+    def test_double_inclusion(self, b, alice):
```
ldmberman: Could you, please, also update the test below (`test_text_search`)?

We need to replace

```
# this query only works with MongoDB
try:
    assets = list(b.text_search('bigchaindb'))
except OperationError as exc:
    assert not isinstance(b.connection, LocalMongoDBConnection)
else:
    assert len(assets) == 3
```
with
```
assets = list(b.text_search('bigchaindb'))
assert len(assets) == 3
```
Using `LocalMongoDBConnection` is part of the test setup, we should not assert for that.

tests/db/test_bigchain_api.py
kansi: I think this test should just work as it is i.e. the test tries to insert two transactions with the same transaction ids so it would make sense to keep this test to ensure the transactions are uniquely indexed by their ids.

codegeschrei: We still tests the `OperationError` in another test: https://github.com/bigchaindb/bigchaindb/blob/f12734237e4fb32a899e3e0554f10e39f618e069/tests/db/test_bigchain_api.py#L47
and the existing double spend errors in the test below: https://github.com/bigchaindb/bigchaindb/blob/f12734237e4fb32a899e3e0554f10e39f618e069/tests/db/test_bigchain_api.py#L16

PR #2457, opened 2018-08-13
Create dynamic upsert validator commands

z-bowen:
**`upsert-validator` needs a command to show election data**

## Solution

Wrote `upsert-validator show` according to BEP-21

## BEPs Implemented

[BEP 21](https://github.com/bigchaindb/BEPs/tree/master/21)


bigchaindb/commands/bigchaindb.py
```diff
@@ -139,6 +139,19 @@ def run_upsert_validator_new(args, bigchain):
         raise OperationError('Failed to commit election')
 
 
+def run_upsert_validator_show(args, bigchain):
```
codegeschrei: Can you add a nice doc string again?

codegeschrei:
Can you add the command to the docs?

bigchaindb/commands/bigchaindb.py

ttmc:
It looks okay to me, but I'm not as steeped in BEP-21 as Vansh, so make sure you get his review before merging this PR.

tests/commands/test_commands.py
```diff
@@ -409,3 +409,44 @@ def mock_write(tx, mode):
...
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_show, run_upsert_validator_new
+
+    def mock_get():
+        return [
+            {'pub_key': {'value': 'zL/DasvKulXZzhSNFwx4cLRXKkSM9GPK7Y0nZ4FEylM=',
+                         'type': 'tendermint/PubKeyEd25519'},
+             'voting_power': 10}
+        ]
+
+    def mock_write(tx, mode):
```
kansi: What is the usage of this mock? I don't think there are any writes.

tests/commands/test_commands.py
```diff
@@ -409,3 +409,44 @@ def mock_write(tx, mode):
...
+    power = 1
+    node_id = '12345'
+
+    new_args = Namespace(action='new',
+                         public_key=public_key,
+                         power=1,
+                         node_id='12345',
+                         sk=priv_validator_path,
+                         config={})
+    election_id = run_upsert_validator_new(new_args, b)
```
kansi: the election should be directly written using `b.write_bulk_transactions`

bigchaindb/commands/bigchaindb.py
```diff
@@ -139,6 +139,28 @@ def run_upsert_validator_new(args, bigchain):
...
+
+    election = bigchain.get_transaction(args.election_id)
+
+    new_validator = election.asset['data']
+
+    public_key = new_validator['public_key']
+    power = new_validator['power']
+    node_id = new_validator['node_id']
+
+    return public_key, power, node_id
```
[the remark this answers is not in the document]
z-bowen: done

bigchaindb/commands/bigchaindb.py
```diff
@@ -181,6 +181,28 @@ def run_upsert_validator_approve(args, bigchain):
...
+
+    election = bigchain.get_transaction(args.election_id)
+
+    new_validator = election.asset['data']
+
+    public_key = new_validator['public_key']
+    power = new_validator['power']
+    node_id = new_validator['node_id']
+
+    return f'public_key={public_key}\npower={power}\nnode_id={node_id}'
```
kansi: I had a discussion with @ldmberman, in that we decided that we can definitely state whether the election has concluded or is ongoing or cannot be concluded i.e. the following statuses for a given election should be listed,
- `ongoing`: Implies that the election is in progress and that a supermajority vote has been received yet.
- `concluded`: Implies that the election was concluded and the validator change was applied.
- `inconclusive`: implies that the election received a supermajority vote but couldn't conclude because the validator set at the time of conclusion was different from that when the election was proposed.

So, in view of the above following would be the output of the command,
```bash
$ bigchaindb upsert-validator show ELECTION_ID
public_key=<e_pub_key>
power=<e_power>
node_id=<e_node_id>
status=ongoing
```

bigchaindb/commands/bigchaindb.py
```diff
@@ -191,6 +191,29 @@ def run_upsert_validator_approve(args, bigchain):
...
+    election = bigchain.get_transaction(args.election_id)
+
+    new_validator = election.asset['data']
+
+    public_key = new_validator['public_key']
+    power = new_validator['power']
+    node_id = new_validator['node_id']
+    status = election.get_status(bigchain)
+
+    return f'public_key={public_key}\npower={power}\nnode_id={node_id}\nstatus={status}'
```
kansi: I don't think returning the value will print it on the console?

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -139,3 +139,17 @@ $ bigchaindb upsert-validator approve 04a067582cf03eba2b53b82e4adb5ece424474cbd4
...
+
+Below is the command line syntax and the return value,
+
+```bash
+$ bigchaindb upsert-validator show ELECTION_ID
+public_key=<e_pub_key>
+power=<e_power>
+node_id=<e_node_id>
+status=<status>
+```
```
kansi: It would be good to explain the output of the command.

tests/commands/test_commands.py
```diff
@@ -347,34 +347,22 @@ def json(self):
...
 
 
-# @pytest.mark.execute
-# @patch('bigchaindb.lib.BigchainDB.get_validators')
-# @pytest.mark.abci
-@pytest.mark.skip
-def test_upsert_validator_new_with_tendermint(b, priv_validator_path, user_sk, monkeypatch):
+@pytest.mark.abci
+def test_upsert_validator_new_with_tendermint(b, priv_validator_path, user_sk, validators):
     """WIP: Will be fixed and activated in the next PR
```
kansi: remove WIP?

docs/server/source/server-reference/bigchaindb-cli.md

bigchaindb/commands/bigchaindb.py

bigchaindb/commands/bigchaindb.py
```diff
@@ -181,6 +181,28 @@ def run_upsert_validator_approve(args, bigchain):
...
+        args = {
+        'election_id': the transaction_id for an election (str)
+        }
+    :param bigchain: an instance of BigchainDB
+    :return: (public_key, power, node_id) for the proposed change to the validator set
+    """
+
+    election = bigchain.get_transaction(args.election_id)
+
+    new_validator = election.asset['data']
```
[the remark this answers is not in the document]
z-bowen: Fixed

tests/commands/test_commands.py
z-bowen: Honestly, it feels like the simplest way to get the data structures I need in place... What with creating the election, and the votes etc, it seems like a bunch of mocking to end up with a data structure that I can get by just using the real code.

bigchaindb/commands/bigchaindb.py
```diff
@@ -191,6 +191,32 @@ def run_upsert_validator_approve(args, bigchain):
...
+    """
+
+    election = bigchain.get_transaction(args.election_id)
+    if not election:
+        logger.error(f'No election found with election_id {args.election_id}')
+        return
+
+    new_validator = election.asset['data']
+
+    public_key = new_validator['public_key']
```
kansi: The public keys in election objects are stored in base58 format which differs from Tendermint's base64 format. I think we should display the public key in base64 format so the user can compare the key if wish to do so.

kansi:
@z-bowen It would better to pull the changes from master as there are some key changes to public key formating.

bigchaindb/commands/bigchaindb.py

bigchaindb/commands/bigchaindb.py
```diff
@@ -191,6 +191,32 @@ def run_upsert_validator_approve(args, bigchain):
...
 
+def run_upsert_validator_show(args, bigchain):
+    """Retrieves information about an upsert-validator election
+
+    :param args: dict
+        args = {
+        'election_id': the transaction_id for an election (str)
+        }
+    :param bigchain: an instance of BigchainDB
+    :return: (public_key, power, node_id) for the proposed change to the validator set
```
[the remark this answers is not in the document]
z-bowen: Removed

tests/commands/test_commands.py
z-bowen: Refactored the test to use the election fixtures. It no longer uses `run_upsert_validator_new`

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -218,9 +224,58 @@ def get_validator_update(cls, bigchain, new_height, txns):
...
+        return votes
+
+    def initial_height(self, bigchain):
+        heights = bigchain.get_block_containing_tx(self.id)
+        initial_height = 0
+        if len(heights) != 0:
+            initial_height = min(bigchain.get_block_containing_tx(self.id))
+        return initial_height
+
+    def get_status(self, bigchain, height=None):
```
ldmberman: An alternative of recording the election ID and election status with every validator update seems significantly simpler to me. We can extend the `validators` collection or introduce a new `validator_elections` collection to store election statuses.

Recomputing the status is quite complicated and partly repeats what's calculated upon election conclusion.

kansi: @ldmberman recall that this code is executed only via command line. I think the node operator won't notice a delay of few milliseconds.

ldmberman: I do not think performance is a concern either. It's about maintainability.

tests/upsert_validator/test_validator_election.py
```diff
@@ -95,3 +94,74 @@ def test_upsert_validator_invalid_election(b_mock, new_validator, node_key):
...
+    election_id = ongoing_election.id
+    public_key = public_key_to_base64(ongoing_election.asset['data']['public_key'])
+    power = ongoing_election.asset['data']['power']
+    node_id = ongoing_election.asset['data']['node_id']
+    status = ValidatorElection.ONGOING
+
+    show_args = Namespace(action='show',
+                          election_id=election_id)
+
+    # TODO: Turn this back on once tests/commands/test_utils.test_configure_bigchaindb_logging.py is fixed
```
ldmberman: I think, it's a good test already. We do not gain much by also testing that `msg` is also sent to the logger.

bigchaindb/commands/bigchaindb.py
```diff
@@ -191,6 +190,36 @@ def run_upsert_validator_approve(args, bigchain):
...
+    public_key = public_key_to_base64(new_validator['public_key'])
+    power = new_validator['power']
+    node_id = new_validator['node_id']
+    status = election.get_status(bigchain)
+
+    response = f'public_key={public_key}\npower={power}\nnode_id={node_id}\nstatus={status}'
+
+    logger.info(response)
+
+    # TODO: Remove this once tests/commands/test_utils.test_configure_bigchaindb_logging.py is fixed
```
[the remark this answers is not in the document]
z-bowen: We don't actually need to return anything since this is executed from the command line. We just log the info so they see it in `stdout`

The return value is just there to make this testable, since the damn `utils` test is breaking the `logger`

tests/commands/test_utils.py
```diff
@@ -72,6 +72,11 @@ def test_configure(args):
      logging.CRITICAL)
 )))
 def test_configure_bigchaindb_logging(log_level):
+    # TODO: See following comment:
```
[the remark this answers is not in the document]
z-bowen: I don't fully understand the importance of this test, and I'm hesitant to remove it until I know it doesn't expose a critical hole in our coverage.

As of now, I have a work around that lets me test the new code, and I've created an issue in GitHub to fix this test (or remove it.) In any case, whatever work needs to be done on this test should be tackled in another PR

tests/upsert_validator/test_validator_election.py
z-bowen: See my above response about the `return` value. I don't think we actually need it, so the commented out code is there so that we can easily align the test once that goes away.

bigchaindb/commands/bigchaindb.py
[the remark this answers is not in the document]
ldmberman: My point is, it's perfectly fine to have a return value and use it in tests. Asserting logs, on the other hand is very cumbersome. So, there is no need in these TODOs.

tests/commands/test_utils.py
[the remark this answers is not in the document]
ldmberman: I do not suggest to remove it, but to add a TODO about removing it. Currently, the TODO note suggests to continue fighting the beast :)

bigchaindb/commands/bigchaindb.py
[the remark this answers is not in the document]
z-bowen: Fair point... Gone.

tests/commands/test_utils.py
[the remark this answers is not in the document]
z-bowen: Fair enough. Like I said, I haven't looked into this test deeply enough to make a call as to what exactly it's covering and whether or not we can disable it, but I updated the TODO to suggest that removal is an alternative to fixing.

PR #2458, opened 2018-08-13
Problem: There is no way to vote for an election

codegeschrei:
Solution:  add `approve` command

#2372


tests/commands/test_commands.py
```diff
@@ -409,3 +409,39 @@ def mock_write(tx, mode):
...
+                         power=1,
+                         node_id='12345',
+                         sk=priv_validator_path,
+                         config={})
+    election_id = run_upsert_validator_new(new_args, b)
+    args = Namespace(action='approve',
+                     election_id=election_id,
+                     sk=priv_validator_path,
+                     config={})
+    approve = run_upsert_validator_approve(args, b)
```
kansi: The test should `assert` that the election is in fact stored in the blockchain by fetching i.e. `b.get_transaction`

tests/commands/test_commands.py
```diff
@@ -409,3 +409,39 @@ def mock_write(tx, mode):
     resp = run_upsert_validator_new(args, b)
 
     assert b.get_transaction(resp)
+
+
+@pytest.mark.tendermint
+@pytest.mark.abci
```
kansi: This is not the intended usage of the `abci` marker i.e. a test which uses the `abci` marker doesn't need to mock `write_transaction` api as Tendermint is running during tests and should be used to posting transactions.

tests/commands/test_commands.py
```diff
@@ -409,3 +409,28 @@ def mock_write(tx, mode):
     resp = run_upsert_validator_new(args, b)
 
     assert b.get_transaction(resp)
+
+
+@pytest.mark.abci
+@pytest.mark.bdb
+def test_upsert_validator_approve(b, priv_validator_path, user_sk, validators):
```
z-bowen: This test is leaving the test environment in a dirty state. Any test that happens after this one is executed get a wrong validator set, and hangs. We need a strategy for testing `upsert-validator` that can act on a real Tendermint instance, but restore it to it's original state when the test is done.

We should have a strategy session about this, since we're going to need this when we start testing Tendermint migrations, and so on.

z-bowen: Fixed. The problem was some weird interaction between the `abci` and `bdb` decorators. Still don't know the root cause, but we don't seem to need the `bdb` marker for this test.

tests/commands/test_commands.py
z-bowen: Resolved

tests/commands/test_commands.py
z-bowen: Resolved

kansi:
@z-bowen  PR #2445  has been merged now. You can pull the changes from the master.

tests/commands/test_commands.py
```diff
@@ -24,6 +24,8 @@ def test_make_sure_we_dont_remove_any_command():
     assert parser.parse_args(['start']).command
     assert parser.parse_args(['upsert-validator', 'new', 'TEMP_PUB_KEYPAIR', '10', 'TEMP_NODE_ID',
                               '--private-key', 'TEMP_PATH_TO_PRIVATE_KEY']).command
+    assert parser.parse_args(['upsert-validator', 'approve', 'ELECTION_ID', '--private-key',
+                              'TEMP_PATH_TO+PRIVATE_KEY']).command
```
kansi: `'TEMP_PATH_TO+PRIVATE_KEY'`, is the `+` intentional or did you mean to put `_`

z-bowen: That's a typo...

tests/commands/test_commands.py
```diff
@@ -411,3 +413,27 @@ def mock_write(tx, mode):
     resp = run_upsert_validator_new(args, b)
 
     assert b.get_transaction(resp)
+
+
+@pytest.mark.abci
```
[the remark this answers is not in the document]
z-bowen: I split this into two test cases, one with a real Tendermint and one with a mock.

tests/conftest.py
```diff
@@ -665,7 +665,7 @@ def ed25519_node_keys(node_keys):
     return node_keys_dict
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
```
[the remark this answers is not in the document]
z-bowen: Do we need it loaded for the whole test session? I think its better to keep a few fixtures in place as possible, to minimize the possibility for cross contamination between tests.

Consider the case where we write a test in future that loads the node keys, changes a private key in the file, and tests to make sure we detect the mismatched public/private pair. Scoped for the session, we'd then have that dirty pair loaded for all tests that run afterwards.

tests/conftest.py
```diff
@@ -677,7 +677,12 @@ def node_keys():
             'uz8bYgoL4rHErWT1gjjrnA+W7bgD/uDQWSRKDmC8otc95wnnxJo1GxYlmh0OaqOkJaobpu13BcUcvITjRFiVgw=='}
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
+def private_validator_keys(node_keys):
```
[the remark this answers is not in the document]
z-bowen: Nowhere anymore. Removed.

tests/conftest.py
```diff
@@ -677,7 +677,12 @@ def node_keys():
...
+    return list(node_keys.items())[0]
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
z-bowen: Are you asking why I moved some fixtures from `tests/upsert_validator/conftest.py` to `tests/conftest.py`? I needed them for some tests that were more appropriately stored in `tests/commands/` (ie testing the execution of `upsert-validator foo` on the command line), so I moved the fixtures so I'd have access to them across all the test directories.

bigchaindb/commands/bigchaindb.py
```diff
@@ -143,6 +144,36 @@ def run_upsert_validator_new(args, bigchain):
...
+        'sk': the path to the private key of the signer (str)
+        }
+    :param bigchain: an instance of BigchainDB
+    :return: a success message
+    :raises: OperationError if the write transaction fails for any reason
+    """
+
+    key = load_node_key(args.sk)
+    tx = bigchain.get_transaction(args.election_id)
+    voting_power = ValidatorElection.get_validators(bigchain)[key.public_key]
```
kansi: The voting power should be fetched from election transaction from the election `outputs`

bigchaindb/commands/bigchaindb.py
```diff
@@ -143,6 +144,36 @@ def run_upsert_validator_new(args, bigchain):
...
+
+    approval = ValidatorElectionVote.generate(tx.to_inputs(), [
```
kansi: The vote transaction cannot spend all the inputs, it can only spend the inputs which are allocated to the given user who is creating this vote.

tests/conftest.py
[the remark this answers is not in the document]
kansi: > Consider the case where we write a test in future that loads the node keys, changes a private key in the file, and tests to make sure we detect the mismatched public/private pair. Scoped for the session, we'd then have that dirty pair loaded for all tests that run afterwards.

This is a completely hypothetical case which I am not really sure when will it happen. Could you please elaborate when will such a situation occur with respect to our current codebase or near term feature set?

tests/upsert_validator/conftest.py
```diff
@@ -7,45 +7,6 @@
 from bigchaindb.upsert_validator import ValidatorElection
 
 
-@pytest.fixture
-def b_mock(b, network_validators):
```
kansi: why have these been moved?

bigchaindb/commands/bigchaindb.py
z-bowen: Done. But let me know if you can see a slicker way to do this.

bigchaindb/commands/bigchaindb.py
```diff
@@ -143,6 +144,36 @@ def run_upsert_validator_new(args, bigchain):
...
+        'sk': the path to the private key of the signer (str)
+        }
+    :param bigchain: an instance of BigchainDB
+    :return: a success message
+    :raises: OperationError if the write transaction fails for any reason
+    """
+
+    key = load_node_key(args.sk)
+    tx = bigchain.get_transaction(args.election_id)
+    voting_power = [v.amount for v in tx.outputs if key.public_key in v.public_keys][0]
```
kansi: In case there are no votes allocated to the public key, the above code raises an exception. It would be better to catch the exception and inform the user that they don't have the right to vote on this election.

tests/conftest.py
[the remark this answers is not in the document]
z-bowen: It isn't a problem in the current code base. But as a general practice, we shouldn't scope things to `session` unless we have an explicit use case. It's not good testing practice, which is why the `scope` defaults to single tests, and why you need to specify `session` as a value.

Since nothing breaks when we remove the `session` level scoping, we don't actually need it, and it makes the tests dirtier.

tests/upsert_validator/conftest.py
z-bowen: I needed them for some tests that were more appropriately stored in `tests/commands/` (ie testing the execution of `upsert-validator foo` on the command line), so I moved the fixtures so I'd have access to them across all the test directories.

kansi: Could you please point out where the following fixtures are being used in your tests i.e. `b_mock`, `valid_election`, `mock_get_validators`?

tests/commands/test_commands.py
```diff
@@ -9,6 +9,9 @@
 
 import pytest
 
+from bigchaindb import ValidatorElection
+from tests.conftest import node_keys
```
kansi: `node_keys` is a fixture. There is no need to import it

tests/conftest.py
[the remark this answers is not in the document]
kansi: Can you please give an example of how the `node_keys` fixture can become dirty? Its response is static so I want to understand how will it become dirty.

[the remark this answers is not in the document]
kansi: > It's not good testing practice, which is why the scope defaults to single tests, and why you need to specify session as a value.

Could please point me to where did you come across this explanation in the pytest docs?

docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -110,3 +110,17 @@ $ bigchaindb upsert-validator new B0E42D2589A455EAD339A035D6CE1C8C3E25863F268120
...
 If the command succeeds, it will create an election and return an `election_id`. Elections consist of one vote token per voting power, issued to the members of the validator set. Validators can cast their votes to approve the change to the validator set by spending their vote tokens. The status of the election can be monitored by providing the `election_id` to the `show` subcommand.
+
+#### upsert-validator approve
+ Approve an election by voting for it.
+ Below is the command line syntax and the return value,
+ ```bash
+$ bigchaindb upsert-validator approve <election_id> --private-key PATH_TO_YOUR_PRIVATE_KEY
+```
+ Here, `<election_id>` is the transaction id of the election the approval should be given for. `--private-key` should be
+the path to wherever the private key for your validator node is stored, (*not* the private key itself).
```
kansi: It would be good to explicitly state that the private key discussed/required above is the one generated by Tendermint's process (BigchainDB doesn't have its own private key).

tests/commands/test_commands.py
```diff
@@ -411,3 +403,111 @@ def mock_write(tx, mode):
...
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_approve
+    from argparse import Namespace
+
+    def mock_write(tx, mode):
+        b.store_bulk_transactions([tx])
+        return (202, '')
+
+    # patch the validator set. We now have one validator with power 10
+    b.get_validators = mock_get
+    b.write_transaction = mock_write
```
kansi: The code is being repeated in `test_upsert_validator_approve_without_tendermint` also. It would make sense to create a fixture and patch `b` and use it in both the functions.

tests/conftest.py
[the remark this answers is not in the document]
z-bowen: I've already come up with a hypothetical scenario where this could be a problem.

Let me turn the question around. What is the argument for keeping this scoped as `session`? I make the following argument:

1) Smaller code base is better than larger. Since the tests run fine without this parameter, we should remove it
2) We already have a problem with tests bleeding into one another. `session` level scoping only increases this risk. If we don't need data to be shared between tests, we shouldn't share data between tests.

Is there a good argument for scoping this as session that I'm missing?

tests/upsert_validator/conftest.py
z-bowen: I ended up building my own elections, so I guess I can move this back

docs/server/source/server-reference/bigchaindb-cli.md
z-bowen: Can you suggest a way to phrase this other than 'the private key for your validator node'?

tests/commands/test_commands.py

tests/conftest.py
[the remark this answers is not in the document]
kansi: > I've already come up with a hypothetical scenario where this could be a problem.

> Can you please give an example of how the node_keys fixture can become dirty? Its response is static so I want to understand how will it become dirty.

I was asking for the code of this hypothetical scenario. I would really help if you can find such an example.

> 1. Smaller code base is better than larger. Since the tests run fine without this parameter, we should remove it

The line is about 32 characters, I don't really see how is that any reduction or increase in the size of the code base. Our code is more than 10K+ Loc adding or subtracting 1 to that doesn't really have any impact there.

> 2. We already have a problem with tests bleeding into one another. session level scoping only increases this risk. If we don't need data to be shared between tests, we shouldn't share data between tests.

The tests you are talking about is `abci` tests and that is the point of these tests. They are supposed to bleed into one another. They have to share state.

> Is there a good argument for scoping this as session that I'm missing?

My argument is pretty simple i.e. the fixture returns a static value so there is no point re-evaluating the fixture every time its called, so we can scope it to the session. But since you [mentioned](https://github.com/bigchaindb/bigchaindb/pull/2458#discussion_r212280422) that this a bad practice and the reason why we don't have `session` as the default scope, I just wanted to know where the `pytest` docs state that and understand it better.

docs/server/source/server-reference/bigchaindb-cli.md
kansi: Maybe we can something as follows,

> Here, `<election_id>` is the transaction id of the election the approval should be given for and `--private-key` should 
> be the path to Tendermint's private key which can be generally found at `/home/user/.tendermint/config/priv_validator.json`

tests/commands/test_commands.py
```diff
@@ -9,6 +9,10 @@
 
 import pytest
 
+from bigchaindb import ValidatorElection
+from bigchaindb.common.exceptions import KeypairMismatchException
+from tests.conftest import node_keys
```
kansi: Reposting because the comment collapsed,

> `node_keys` is a fixture. There is no need to import it

tests/conftest.py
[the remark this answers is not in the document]
z-bowen: I'm not finding anything online to state one way or another how we should scope this.

However, speaking with a few years of experience as a test engineer, my experience tells me that cross contamination on the unit and integration test level is bad news.

We want tests that, in the event of failure, give us clear indicators of what is broken. This means tests should cover a specific interaction in the code, as narrowly as possible, with as few side effects as possible. This way, if something turns red, we know exactly where to look.

If we want to check for side effects that happen between code paths, those deserve their own tests. We shouldn't be checking for them as a 'by product' of normal test execution. The reason for this is two fold. First, with dedicated tests for cross contamination, we get better coverage of corner cases. As it stands, we have a relatively static order of execution on a deterministic system, so while it may seem that we're covering 'weird db states', we really only cover one random walk through our product, which we do over and over. Much better for exploratory testing to have something like ChaosMonkey, which does some truly exploratory testing.

Second, as it stands now, even if we do uncover a weird interaction due to broad scoping, our test reporting isn't designed in a way that would allow for a meaningful investigation. If I (or someone else) did run into a case where a session level fixture caused a test to break, the most likely result would just be to down-scope the fixture and move on. With overly broad, poorly defined tests, there is no goo way to tell when a failure is due to a broken test, or due to broken code. Again, if we want to test for weird cross interactions in our code (a generally good idea!) we should have some dedicated tests that do this.

Finally, as a further argument against test overlap, I've already run into the situation once this week where a problem in Test A caused a failure in Test Q (somewhere far away in the test execution order.) This was a debugging nightmare, and cost me about a day of time to dig out the problem. In the end, it was exactly the problem where fixtures didn't clean up after themselves that caused the issue (in this case, changing the validator set and not restoring it after the test ended.)

To reiterate, I agree that we should be testing for weird interactions, but I think this is a better candidate for scenario style acceptance tests. What we're writing here is somewhere between a unit and an integration test. In that context, my experience has been that over complicated tests lead to nightmare debugging when something breaks, and cost more time than the fraction of a second it saves to share fixtures between tests.

tests/commands/test_commands.py
z-bowen: I'm using it in the helper function `mock_get`. Since it's not a `test_foo`, it doesn't import fixtures from it's arguments.

tests/conftest.py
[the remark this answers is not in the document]
z-bowen: You're right, I am generalizing the scope of the discussion, because I'm trying to have a conversation about best practices in testing, and I'm making the argument that as a general rule, the scope of testing assets should be as small as possible while still allowing the test to cover it's intended target.

Given that I think that's a good rule of thumb, and given that no change in coverage occurs when I remove the `session` level scoping from this fixture, I consider this to be an improvement to our code, in the same way that removing unused imports is an improvement. Even if an unused import doesn't break anything, per se, it's still a good idea to clean them up when you see them.

If you have a good counter argument to my rule of thumb, I'm open to hearing it, but if not, I think it justifies this change to the code.

tests/commands/test_commands.py
kansi: But you can pass this function `node_keys` argument from the function where `mock_get` is called.

z-bowen: Yes, but `mock_get` overrides the `bigchaindb.get_validators` method, so I'd like to keep their input signatures the same, for the sake of easier patching. (Same reason I have `height` as a dummy parameter)

tests/conftest.py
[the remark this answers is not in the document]
kansi: > If you have a good counter argument to my rule of thumb, I'm open to hearing it, but if not, I think it justifies this change to the code.

Well, my argument is pretty straightforward, I don't see any way a static fixture can violate your rule of thumb. Your rule of thumb is derived from the experience that larger scopes can end up contaminating state, what I am saying is that `node_keys` is static and will never contaminate state so it satisfies the underlying reasoning for your rule of thumb.

PR #2461, opened 2018-08-14
Problem: Config settings docs missing some configs & inconsistent

ttmc:
Solution: Document some undocumented config settings and make the config settings docs page more consistent

- Removed the long list of environment variables at the beginning of the docs page about configuration settings. Explained the relationship between config-file names and environment variable names.
- Added documentation for some configuration settings which are still around (e.g. `database.replicaset`) but which weren't documented.
- Made the whole page more consistent. For example, every subsection has a name like **server.\*** and the **log.\*** subsection is now much more like the other subsections.
- All code blocks now specify the programming language.
- Lots of minor copy-editing.


muawiakh:
restarted the build, seems unrelated. Will merge when it is 💚

PR #2464, opened 2018-08-16
Problem: simple-network-setup page getting long

ttmc:
Solution: Start refactoring it into a new section

This first pull request in the refactoring of that page mainly just moves it into a page in new section. I wanted to retain as much Git history (about who contributed what) as possible, and I think this will work.

Note: The old URL will break, but if enough people complain, we can create a redirect (using ReadTheDocs).


ttmc:
This is strange... GitHub says that the Travis checks haven't completed yet, but when I go over to Travis, it says it's done.

PR #2471, opened 2018-08-17
Problem: Docs say "BigchainDB cluster" but IRL we say "BigchainDB network"

ttmc:
Solution: Edit the docs to say "BigchainDB network" instead

Notes:
- I couldn't just replace all instances of "cluster" with "network" because there are many cases where "cluster" is the right word, especially when referring to a Kubernetes cluster.
- I also wanted to add the word "BigchainDB" before "network" in some cases, for clarity.
- In the page about production node assumptions, I removed the line (assumption) saying, "1. Production nodes use MongoDB (not RethinkDB, PostgreSQL, Couchbase or whatever)." That line was a bit silly because MongoDB is currently the only real option.
- I also did some other minor copy-editing.

### DIFF
--- a/docs/root/source/assets.rst
+++ b/docs/root/source/assets.rst
@@ -8,7 +8,7 @@ How BigchainDB is Good for Asset Registrations & Transfers
 
 BigchainDB can store data of any kind (within reason), but it's designed to be particularly good for storing asset registrations and transfers:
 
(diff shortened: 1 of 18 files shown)


ttmc:
Travis builds the docs and all checks have passed, so I will merge this PR.

PR #2476, opened 2018-08-23
Problem: we have unused and outdated fixtures

codegeschrei:
Solution: clean up fixtures and tests accordingly

#2381

### DIFF
--- a/tests/commands/test_commands.py
+++ b/tests/commands/test_commands.py
@@ -179,7 +179,7 @@ def test_run_configure_when_config_does_exist(monkeypatch,
                                               mock_bigchaindb_backup_config):
     value = {}
 
-    def mock_write_config(newconfig, filename=None):
+    def mock_write_config(newconfig):
         value['return'] = newconfig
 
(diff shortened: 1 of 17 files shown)


tests/commands/test_commands.py
```diff
@@ -280,8 +277,7 @@ def test_calling_main(start_mock, base_parser_mock, parse_args_mock,
 @patch('bigchaindb.config_utils.autoconfigure')
 @patch('bigchaindb.commands.bigchaindb.run_recover')
 @patch('bigchaindb.start.start')
-def test_recover_db_on_start(mock_autoconfigure,
```
kansi: why was `mock_autoconfigure` removed?

codegeschrei: oh, that might have been a mistake. I see that the patch is still in. I need to check. If I removed it then because it was not needed, but I need to check again.

codegeschrei: ok, so I can confirm now, that it is not needed. I'll remove the patch in the next commit

PR #2480, opened 2018-08-24
Problem: Upsert validator 'approve' command not transfering vote to election public key

kansi:
Solution: Transfer vote to the election public key

### DIFF
--- a/bigchaindb/commands/bigchaindb.py
+++ b/bigchaindb/commands/bigchaindb.py
@@ -168,8 +168,10 @@ def run_upsert_validator_approve(args, bigchain):
         )
 
     inputs = [i for i in tx.to_inputs() if key.public_key in i.owners_before]
-    approval = ValidatorElectionVote.generate(inputs, [
-        ([key.public_key], voting_power)], tx.id).sign([key.private_key])
+    election_pub_key = ValidatorElection.to_public_key(tx.id)
+    approval = ValidatorElectionVote.generate(inputs,
+                                              [([election_pub_key], voting_power)],
+                                              tx.id).sign([key.private_key])
     approval.validate(bigchain)
 
     resp = bigchain.write_transaction(approval, 'broadcast_tx_commit')


bigchaindb/upsert_validator/validator_election.py
```diff
@@ -220,6 +220,7 @@ def get_validator_update(cls, bigchain, new_height, txns):
                 updated_validator_set = new_validator_set(curr_validator_set,
                                                           new_height, validator_updates)
 
+                updated_validator_set = [v for v in updated_validator_set if v['voting_power'] > 0]
```
ldmberman: When do you expect the voting power to be zero here?

kansi: When someone wishes to remove a validator they will set the validator power to `0`

ldmberman: Should we not allow them doing that?

kansi: Do you wish to imply that we should not allow users to remove the validators?

ldmberman: I believe, this restriction does not make much sense. Imagine a network where for some reason one thousand is a common voting power. Setting a voting power of 1 is then effectively the same as 0 so we as well just allow 0 to be voted for.

bigchaindb/commands/bigchaindb.py
```diff
@@ -124,21 +128,28 @@ def run_upsert_validator_new(args, bigchain):
...
-
-    voters = ValidatorElection.recipients(bigchain)
+    try:
+        key = load_node_key(args.sk)
+        voters = ValidatorElection.recipients(bigchain)
+        election = ValidatorElection.generate([key.public_key],
+                                              voters,
+                                              new_validator, None).sign([key.private_key])
+        election.validate(bigchain)
+    except ValidationError as e:
```
ldmberman: We return `False` when some errors occur and raise an exception in some other cases. Should we make it consistent?

bigchaindb/web/websocket_server.py
```diff
@@ -102,7 +102,8 @@ def publish(self):
                 block = event.data
 
                 for tx in block['transactions']:
-                    asset_id = tx['id'] if tx['operation'] == 'CREATE' else tx['asset']['id']
+                    asset_id = tx['asset'].get('id', None)
```
ldmberman: Might be just `tx['asset'].get('id', tx['id'])`.

bigchaindb/commands/bigchaindb.py
```diff
@@ -124,24 +126,32 @@ def run_upsert_validator_new(args, bigchain):
...
+        logger.error(e)
+        return False
```
ldmberman: Could you update the docstring as well (both of `new` and `approve`)?

bigchaindb/commands/bigchaindb.py
```diff
@@ -163,13 +173,17 @@ def run_upsert_validator_approve(args, bigchain):
     if len(voting_powers) > 0:
         voting_power = voting_powers[0]
     else:
-        raise KeypairMismatchException(
-            'The key you provided does not match any of the eligible voters in this election.'
-        )
+        logger.error('The key you provided does not match any of the eligible voters in this election.')
+        return False
 
+    # NOTE: The code below assumes no vote delegation has taken place. Vote delegation
```
ldmberman: BEP-21 still mentions delegation:

> A validator can choose to vote/delegate/burn their vote.

The mention is not a bold one but might be better to remove it.

tests/commands/test_commands.py
```diff
@@ -392,24 +391,67 @@ def mock_write(tx, mode):
...
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_new
+
+    b.get_validators = mock_get
+    args = Namespace(action='new',
+                     public_key='CJxdItf4lz2PwEf4SmYNAu/c/VpmX39JEgC5YpH7fxg=',
+                     power=10,
+                     node_id='fb7140f03a4ffad899fabbbf655b97e0321add66',
+                     sk=priv_validator_path,
+                     config={})
+    assert not run_upsert_validator_new(args, b)
```
ldmberman: If we return a particular response or raise a particular exception we can then test for particular outcomes. Currently, the tests do not guarantee the correct outcome (now expressed as logs) for these cases - they only guarantee a failure occurs.

bigchaindb/commands/bigchaindb.py
kansi: Is there a reason we should remove it? The `approve` command doesn't support delegation but that was never the scope for `approve` command.

kansi: > we should not mention this inside the approve code .
Any particular reason you recommend doing that?

ldmberman: If delegation is not supported by the current implementation, I suggest to remove the `A validator can choose to vote/delegate/burn their vote.` note from BEP-21 because it suggests the opposite.

If delegation is supported but the `approve` command is simply not concerned with it, I suggest to remove the `#NOTE` comment because it suggests that the `approve` function actually needs to be adopted to make delegation work.

tests/commands/test_commands.py
```diff
@@ -392,24 +392,67 @@ def mock_write(tx, mode):
...
+    from bigchaindb.commands.bigchaindb import run_upsert_validator_new
+
+    args = Namespace(action='new',
+                     public_key='CJxdItf4lz2PwEf4SmYNAu/c/VpmX39JEgC5YpH7fxg=',
+                     power=10,
+                     node_id='fb7140f03a4ffad899fabbbf655b97e0321add66',
+                     sk='/tmp/invalid/path/key.json',
+                     config={})
+
+    with caplog.at_level(logging.ERROR):
```
ldmberman: If `run_upsert_validator_new` returns an error code (can be stored inside the exception class), asserting is as simple as:

```
assert run_upsert_validator_new(args, b) == SomeException.error_code
```

It is then easier to read and comprehend.

Though technically, the current approach is good too.

ldmberman: Off course - the comprehensive output is important. It's just a little cumbersome to assert logs in tests. But I do not insist.

bigchaindb/commands/bigchaindb.py
kansi: In case of exceptions during execution of the command the return value is `False` and suitable log messages are generated for the users.

bigchaindb/commands/bigchaindb.py
kansi: Updated the docstring

PR #2481, opened 2018-08-27
Problem: HTTP API docs wrong regarding the mode parameter

ttmc:
Solution: Edit those docs to be accurate

In particular, note that the BigchainDB node always does a check of transaction validity before ever sending it to Tendermint and before doing any HTTP response.

Thanks to @ldmberman who noticed this error.

### DIFF
--- a/docs/server/source/http-client-server-api.rst
+++ b/docs/server/source/http-client-server-api.rst
@@ -142,7 +142,7 @@ Transactions
    `Tendermint broadcast API
    <https://tendermint.com/docs/tendermint-core/using-tendermint.html#broadcast-api>`_.
    
-   The meaning of the ``mode`` query parameter is inhereted from the mode parameter in
+   The meaning of the ``mode`` query parameter is inherited from the mode parameter in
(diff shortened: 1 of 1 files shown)


docs/server/source/http-client-server-api.rst
```diff
@@ -135,13 +135,21 @@ Transactions
...
-   ``mode=sync`` means the HTTP response will come back once the node has
-   checked the validity of the transaction.
+   Once the posted transaction arrives at a BigchainDB node,
+   that node will check to see if the transaction is valid.
+   If it's invalid, the node will return an HTTP 400 (error).
+   Otherwise, the node will send the transaction to Tendermint (in the same node) using the
+   `Tendermint broadcast API
+   <https://tendermint.com/docs/tendermint-core/using-tendermint.html#broadcast-api>`_.
+   
+   The meaning of the ``mode`` query parameter is inhereted from the mode parameter in
```
ldmberman: inher**i**ted

PR #2489, opened 2018-08-29
Problem: validation is slow

vrde:
Solution: parallelize the validation of transactions. This patch adds a
new flag for the `start` command, namely
`--experimental-parallel-validation`, that can be used to enable this
experimental feature.

Depends on #2529

### DIFF
--- a/tests/test_parallel_validation.py
+++ b/tests/test_parallel_validation.py
@@ -1,0 +1,133 @@
+# Copyright BigchainDB GmbH and BigchainDB contributors
+# SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+# Code is Apache-2.0 and docs are CC-BY-4.0
+
+import pytest
+
+from bigchaindb.common.crypto import generate_key_pair
+from bigchaindb.models import Transaction
+
+
+pytestmark = pytest.mark.tendermint
+
+
+def generate_create_and_transfer(keypair=None):
+    if not keypair:
+        keypair = generate_key_pair()
+    priv_key, pub_key = keypair
+    create_tx = Transaction.create([pub_key], [([pub_key], 10)]).sign([priv_key])
+    transfer_tx = Transaction.transfer(
+            create_tx.to_inputs(),
+            [([pub_key], 10)],
+            asset_id=create_tx.id).sign([priv_key])
+    return create_tx, transfer_tx
+
+
+def test_validation_worker_process_multiple_transactions(b):
+    import multiprocessing as mp
+    from bigchaindb.parallel_validation import ValidationWorker, RESET, EXIT
+
+    keypair = generate_key_pair()
+    create_tx, transfer_tx = generate_create_and_transfer(keypair)
+    double_spend = Transaction.transfer(
+            create_tx.to_inputs(),
+            [([keypair.public_key], 10)],
+            asset_id=create_tx.id).sign([keypair.private_key])
+
+    in_queue, results_queue = mp.Queue(), mp.Queue()
+    vw = ValidationWorker(in_queue, results_queue)
+
+    # Note: in the following instructions, the worker will encounter two
+    # `RESET` messages, and an `EXIT` message. When a worker processes a
+    # `RESET` message, it forgets all transactions it has validated. This allow
+    # us to re-validate the same transactions. This won't happen in real life,
+    # but it's quite handy to check if the worker actually forgot about the
+    # past transactions (if not, it will return `False` because the
+    # transactions look like a double spend).
+    # `EXIT` makes the worker to stop the infinite loop.
+    in_queue.put((0, create_tx.to_dict()))
+    in_queue.put((10, transfer_tx.to_dict()))
+    in_queue.put((20, double_spend.to_dict()))
+    in_queue.put(RESET)
+    in_queue.put((0, create_tx.to_dict()))
+    in_queue.put((5, transfer_tx.to_dict()))
+    in_queue.put(RESET)
+    in_queue.put((20, create_tx.to_dict()))
+    in_queue.put((25, double_spend.to_dict()))
+    in_queue.put((30, transfer_tx.to_dict()))
+    in_queue.put(EXIT)
+
+    vw.run()
+
+    assert results_queue.get() == (0, create_tx)
+    assert results_queue.get() == (10, transfer_tx)
+    assert results_queue.get() == (20, False)
+    assert results_queue.get() == (0, create_tx)
+    assert results_queue.get() == (5, transfer_tx)
+    assert results_queue.get() == (20, create_tx)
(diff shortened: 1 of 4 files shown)


vrde:
Can we ignore Codecov here? This is an experimental feature that is not *production ready*. We can iterate later on improving the coverage when we want to make it production ready.

bigchaindb/experimental/core.py
```diff
@@ -0,0 +1,26 @@
+from bigchaindb import App
```
ldmberman: Moving the files around produces huge diffs, I would just give a real name to this folder from the start.

bigchaindb/experimental/core.py
```diff
@@ -0,0 +1,26 @@
...
+from bigchaindb.experimental.utils import ParallelValidator
+from abci.types_pb2 import ResponseDeliverTx
+
+
+CodeTypeOk = 0
+
+
+class ParallelValidationApp(App):
```
ldmberman: It is difficult to introduce updates to the ABCI interface this way. For instance, take a look the [chain migration PR](https://github.com/bigchaindb/bigchaindb/pull/2488). Can we abstract away only the validation part somehow?

ldmberman: Actually, this may work too, but you need to call `super` in `deliver_tx` as well.

tests/experimental/test_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+    pv = ParallelValidator(number_of_workers=2)
+    pv.start()
+
+    # ParallelValidator is instantiated once, and then used several times.
+    # Here we simulate this scenario by running it an arbitrary number of
+    # times.
+    # Note that the `ParallelValidator.result` call resets the object, and
+    # makes it ready to validate a new set of transactions.
+    for _ in range(10):
```
ldmberman: Since we distribute transactions deterministically, there does not seem to be a need in repeating assertions. Repeating assertions helps to test random algorithms, by asserting results accumulated over numerous test runs.

bigchaindb/experimental/core.py
vrde: Calling `super().deliver_tx(...)` would trigger the old validation again. Regarding the comment on the chain migration PR, we can drop compatibility for now since parallel validation is just an experiment. If there are conflicts or if this code needs to be updated, I can maintain it.

tests/experimental/test_utils.py
vrde: Yes the distribution is deterministic. This test is to check that `ParallelValidator` behaves correctly in a loop, but I understand that the title of this test can be misleading. This test runs it an arbitrary number of times (10). It can also run it 2 times. I'll move it.

bigchaindb/experimental/utils.py
```diff
@@ -0,0 +1,95 @@
...
+            asset_id = dict_transaction['asset']['id']
+        except KeyError:
+            asset_id = dict_transaction['id']
+
+        transaction = self.bigchaindb.is_valid_transaction(
+                dict_transaction,
+                self.validated_transactions[asset_id])
+
+        if transaction:
+            self.validated_transactions[asset_id].append(transaction)
```
z-bowen: Why do we need the `defaultdict(list)` construction for `validated_transactions` if we end up storing by `asset_id`? Surely we can't validate more than one transaction with the same `id`, so we shouldn't ever have more than one tx stored for each key.

Can't we use an ordinary `dict` here?

tests/experimental/test_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    priv_key, pub_key = keypair
+    create_tx = Transaction.create([pub_key], [([pub_key], 10)]).sign([priv_key])
+    transfer_tx = Transaction.transfer(
+            create_tx.to_inputs(),
+            [([pub_key], 10)],
+            asset_id=create_tx.id).sign([priv_key])
+    return create_tx, transfer_tx
+
+
+def test_validation_worker_process_multiple_transactions(b):
```
z-bowen: If I understand this test correctly, you set up one worker and have it process all the txs. I'd also like to see a test with 2 workers, where `worker_1` processes `transfer_tx` and `worker_2` processes `double_spend`, and we verify that the parallel processes still catch the double spend.

tests/experimental/test_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    # makes it ready to validate a new set of transactions.
+    for _ in range(10):
+        for transaction in transactions:
+            pv.validate(dumps(transaction).encode('utf8'))
+        # We expect the result to have the same order as the input data.
+        assert pv.result(timeout=1) == transactions
+        pids = set()
+        for _ in transactions:
+            pids.add(validation_called_by.get())
+        assert len(pids) == 2
```
z-bowen: We have deterministic routing based on the `tx_id`, so you could further assert that `tx_0` and `tx_2` get routed to the same worker.

bigchaindb/experimental/utils.py
vrde: `validated_transaction` maps `asset_id` to one or more `transaction`s. As an example, within the same block we can have a `CREATE` and some subsequent `TRANSFER`s. In this case, we need to keep track of those dependencies.

tests/experimental/test_utils.py
vrde: I see your point, but given the Routing Algorithm™ (that is nothing more than the modulo operator) all transactions of a specific `asset_id` are routed to the same worker.

I'll try to write a test around it.

tests/experimental/test_utils.py
vrde: This might be something for the test I mentioned in my [previous comment](https://github.com/bigchaindb/bigchaindb/pull/2489#discussion_r216616139).

tests/test_parallel_validation.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Now we analize the transaction processed by the workers
+        worker_to_transactions = defaultdict(list)
+        for _ in transactions:
+            worker_pid, transaction_id = validation_called_by.get()
+            worker_to_transactions[worker_pid].append(transaction_id)
+
+        # The transactions are stored in two buckets.
+        for _, transaction_ids in worker_to_transactions.items():
+            assert len(transaction_ids) == 2
```
z-bowen: We have deterministic assignments between `tx`s and workers. Can we strengthen this test to assert that worker0 gets `tx_0, 2` and worker1 gets `tx_1, 3`?

vrde: https://github.com/bigchaindb/bigchaindb/pull/2489/commits/df824b07c1134891fb4802f84264f2aee02c04c2#diff-2772784c4ecd91053e0a2efcef70d208R125

PR #2490, opened 2018-08-29
Problem: Validation code not optimized

kansi:
Solution: Use memoization for functions with static validation


bigchaindb/common/memoize.py
```diff
@@ -0,0 +1,42 @@
+import functools
+import codecs
+from functools import lru_cache
+
+
+class HDict(dict):
+    def __hash__(self):
+        return  int.from_bytes(codecs.decode(self['id'], 'hex'), 'big')
```
vrde: I had a similar problem recently. While your code converts the hex string representing the `transaction.id` to a number, a simpler approach is to just use `int(self['id'], 16)` (I was actually quite surprised of its simplicity when I found it out).
```python
In [1]: int('437752a2c5c3cf2ab8ff6254ca8c0fb417a0951ab651c42481474dd9347971a7', 16) == int.from_bytes(codecs.decode('437752a2c5c3cf2ab8ff6254ca8c0fb417a0951ab651c42481474dd9347971a7', 'hex'), 'big')
Out[1]: True
```

I was curious about your approach so I compared performance of the two approaches:
```python
In [2]: %timeit int('437752a2c5c3cf2ab8ff6254ca8c0fb417a0951ab651c42481474dd9347971a7', 16)
382 ns ± 3.24 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit int.from_bytes(codecs.decode('437752a2c5c3cf2ab8ff6254ca8c0fb417a0951ab651c42481474dd9347971a7', 'hex'), 'big')
1.72 µs ± 8.16 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

bigchaindb/common/memoize.py
```diff
@@ -0,0 +1,42 @@
...
+@lru_cache(maxsize=16384)
+def from_dict(func, *args, **kwargs):
+    return func(*args, **kwargs)
+
+
+def memoize_from_dict(func):
+
+    @functools.wraps(func)
+    def memoized_func(*args, **kwargs):
+        print(args)
```
vrde: Leftover `print`? (same two lines below)

bigchaindb/common/transaction.py
```diff
@@ -1003,6 +1012,8 @@ def validate(i, output_condition_uri=None):
         return all(validate(i, cond)
                    for i, cond in enumerate(output_condition_uris))
 
+    # @memoize_input_valid
```
vrde: Do we need this comment?

bigchaindb/lib.py
```diff
@@ -224,6 +224,13 @@ def delete_unspent_outputs(self, *unspent_outputs):
             return backend.query.delete_unspent_outputs(
                                         self.connection, *unspent_outputs)
 
+    def is_commited(self, transaction_id):
```
vrde: typo: `is_committed`

bigchaindb/lib.py
```diff
@@ -224,6 +224,13 @@ def delete_unspent_outputs(self, *unspent_outputs):
...
+        transaction = backend.query.get_transaction(self.connection, transaction_id)
+        if transaction:
```
vrde: `return bool(transaction)`

bigchaindb/common/memoize.py
```diff
@@ -0,0 +1,42 @@
...
+        new_args = (args[0], HDict(args[1]), args[2])
```
vrde: That's difficult to understand, can you please add some comments around this code (and the equivalent for `memoize_to_dict`.

bigchaindb/common/memoize.py
kansi: Switched to `hash()`, following are the stats
```python
In [1]: %timeit hash('437752a2c5c3cf2ab8ff6254ca8c0fb417a0951ab651c42481474dd9347971a7')
70.4 ns ± 0.625 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

vrde:
I run `prof.py` to check the improvement, this transaction validation speed on `master`:
```
 /tmp ➜ python prof.py
Create 1000 transactions
Start parallel validation
  Total time: 1.809801
  Time per transaction: 0.001810
Start serial validation
  Total time: 4.285186
  Time per transaction: 0.004285
```

This is with this patch:
```
 /tmp ➜ python prof.py
Create 1000 transactions
Start parallel validation
  Total time: 1.407967
  Time per transaction: 0.001408
Start serial validation
  Total time: 3.363722
  Time per transaction: 0.003364
```

:horse_racing: :horse_racing: :horse_racing:

PR #2491, opened 2018-08-29
Problem: BigchainDB has un-necessary code to initialize a replica set and check if MongoDB was started with replicaSet

### LINKED ISSUE #2449: Problem: Backend for MongoDB does not work with hosted MongoDB deployed with replicaset
If you create user with the monog shell as follows
```
use bigchain;
db.createUser({user: "testUser", pwd: "superstrongpassword", roles: [{role: "clusterAdmin", db: "bigchain"}]});
```

And then set the following parameter in `.bigchaindb` config file

shahbazn:
Solution: Remove un-necessary code. Deployment of MongoDB with or without replicaSet should be the responsibility of MongoDB admin which can and cannot be a BigchainDB node operator. As far as BigchainDB is concerned replicaset, if provided in bigchaindb configs, should be used to establish connection with MongoDB.

Resolves #2449


shahbazn:
@vrde Those instances of `replicaset` do not need to be removed as `replicaset` config is still going to be used to login to a `MongoDB` which was deployed as part of a `replicaset`. This PR merely relieves BigchainDB of initializing a `replicaset` or checking if replica set was properly started as that should be the responsibility of DB admin.

PR #2495, opened 2018-08-30
Problem: No ducumentation for hosted MongoDB

### LINKED ISSUE #2449: Problem: Backend for MongoDB does not work with hosted MongoDB deployed with replicaset

shahbazn:
Solution: Add documentation for production nodes.

Resolves #2449

### DIFF
--- a/docs/server/source/production-nodes/node-requirements.md
+++ b/docs/server/source/production-nodes/node-requirements.md
@@ -14,6 +14,19 @@ BigchainDB Server requires Python 3.5+ and Python 3.5+ [will run on any modern O
 
 _Don't use macOS_ (formerly OS X, formerly Mac OS X), because it's not a server-grade operating system. Also, BigchaindB Server uses the Python multiprocessing package and [some functionality in the multiprocessing package doesn't work on Mac OS X](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.Queue.qsize).
 
+## Using Hosted MongoDB
+
+If you are using already existing MongoDB deployment or hosted MongoDB service (e.g. MongoDB Atlas) you should create a new database (e.g. `bigchain`) and create a user with `readWrite` access. You can perform above steps through MongoDB shell as follows:
+
+Login to your MongoDB deployment as Admin and run following commands
+
+```
+use <databse_name>
+db.createUser({user: "<username>", pwd: "<password>", roles: [{role: "readWrite", db: "<database_name>"}]})
+```
+
+Now configure the details of your database and newly created user in your BigchainDB configs before starting BigchainDB
+
 ## General Considerations
 
 BigchainDB Server runs many concurrent processes, so more RAM and more CPU cores is better.


docs/server/source/production-nodes/node-requirements.md
```diff
@@ -14,6 +14,19 @@ BigchainDB Server requires Python 3.5+ and Python 3.5+ [will run on any modern O
 
 _Don't use macOS_ (formerly OS X, formerly Mac OS X), because it's not a server-grade operating system. Also, BigchaindB Server uses the Python multiprocessing package and [some functionality in the multiprocessing package doesn't work on Mac OS X](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.Queue.qsize).
 
+## Using Hosted MongoDB
+
+If you are using already existing MongoDB deployment or hosted MongoDB service (e.g. MongoDB Atlas) you should create a new database (e.g. `bigchain`) and create a user with `readWrite` access. You can perform above steps through MongoDB shell as follows:
+
+Login to your MongoDB deployment as Admin and run following commands
+
+```
```
ttmc: \```text

docs/server/source/production-nodes/node-requirements.md
```diff
@@ -14,6 +14,19 @@ BigchainDB Server requires Python 3.5+ and Python 3.5+ [will run on any modern O
 
 _Don't use macOS_ (formerly OS X, formerly Mac OS X), because it's not a server-grade operating system. Also, BigchaindB Server uses the Python multiprocessing package and [some functionality in the multiprocessing package doesn't work on Mac OS X](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.Queue.qsize).
 
+## Using Hosted MongoDB
```
ttmc: I'm not sure this page is the best place to put this documentation. Maybe put it on the page about BigchainDB configuration settings (i.e. http://docs.bigchaindb.com/projects/server/en/master/server-reference/configuration.html#database ), since that's the only other place in the docs where the backend database username/login and password are mentioned.

shahbazn: The only usecase I see for this is a production one, which is the reason I did not add it to simple network guide either. So I thought it was appropriate to add it under production nodes notes. If you agree I think we should keep it here as well. I will add some detail in config page but keeping in mind `login` field is still used if we use SSL cert based `x.509` authentication method. `password` is not needed in `x.509` method.

ttmc: Using a hosted MongoDB isn't required, even in production nodes, and I don't want to give that impression by including it in the "Production Node Requirements" page. (We would also have to include instructions to set up MongoDB to use the X.509 auth system, and I really don't want to do that.)

It's good to know that `password` isn't needed when using the X.509 method. Maybe add a note about that to the documentation about `password`.

PR #2496, opened 2018-08-30
Rework `upsert-validator show status`

z-bowen:
**Problem: `upsert-validator show status` uses convoluted logic**

## Solution

Reworked the logic to be cleaner and more readable
-changes to the validator set now include `election_id`, so they are searchable by the election that approved them
-added a few helper methods to query validator updates in Mongo by `election_id`

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -308,12 +308,7 @@ def get_validator_set_by_election_id(conn, election_id):
         .find(query, projection={'_id': False})
     )
 
-    results = list(cursor)
-
-    if len(results) > 0:
-        return results[0]
-    else:
-        return None
+    return next(cursor, None)
 
 
 @register_query(LocalMongoDBConnection)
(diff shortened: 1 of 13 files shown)


bigchaindb/backend/localmongodb/query.py
```diff
@@ -299,6 +299,23 @@ def get_validator_set(conn, height=None):
...
+@register_query(LocalMongoDBConnection)
+def get_validator_set_by_election_id(conn, election_id):
+    query = {'election_id': election_id}
+
+    cursor = conn.run(
+        conn.collection('validators')
+        .find(query, projection={'_id': False})
+    )
+
+    results = list(cursor)
```
ldmberman: You can just:

```
return next(cursor, None)
```

bigchaindb/lib.py
```diff
@@ -421,24 +421,32 @@ def get_metadata(self, txn_ids):
...
-    def store_validator_set(self, height, validators):
+    def store_validator_set(self, height, validators, election_id):
         """Store validator set at a given `height`.
            NOTE: If the validator set already exists at that `height` then an
            exception will be raised.
         """
         return backend.query.store_validator_set(self.connection, {'height': height,
-                                                                   'validators': validators})
+                                                                   'validators': validators,
+                                                                   'election_id': election_id})
```
kansi: Please create a unique index on `election_id` field.

PR #2498, opened 2018-08-31
Create abstract election class

z-bowen:
**Problem: `ValidatorElection` and `MigrationElection` need to inherit from a generic `Election` class

## Solution

Factored the common logic out of `ValidatorElection` and moved it to a common parent class, `Election`


bigchaindb/backend/localmongodb/query.py
```diff
@@ -300,11 +300,11 @@ def get_validator_set(conn, height=None):
 
 
 @register_query(LocalMongoDBConnection)
-def get_validator_set_by_election_id(conn, election_id):
+def get_result_by_election_id(conn, election_id, table):
```
kansi: 1. A more appropriate function name could be `get_election_result`/`get_election_result_by_id` ?
2. What is the point of this generalization?

ldmberman: Creating a new collection for every election might be cumbersome. Instead, we can introduce an `elections` collection storing documents of the form:

```
{
    'election_id': '...',
    'type': '...',
    'height': X
}
```

bigchaindb/common/election.py
```diff
@@ -0,0 +1,248 @@
...
+
+        :return: {
+            'height': <block_height>,
+            'asset': {
+                'height': <block_height>,
+                'validators': <validator_set>,
+                'election_id': <election_id_that_approved_the_change>
+            }
+        }
+        """
```
kansi: I think this doc string is wrong? 

```

        :return: {
                'height': <block_height>,
                'validators': <validator_set>,
                'election_id': <election_id_that_approved_the_change>
        }
```

bigchaindb/backend/localmongodb/query.py
kansi: What would be the use of `'type'`?

bigchaindb/common/election.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @classmethod
+    def to_public_key(cls, election_id):
+        return base58.b58encode(bytes.fromhex(election_id))
+
+    @classmethod
+    def count_votes(cls, election_pk, transactions, getter=getattr):
+        votes = 0
+        for txn in transactions:
+            if getter(txn, 'operation') == 'VALIDATOR_ELECTION_VOTE':
```
kansi: The generalized value cannot contain hard-coded value.

bigchaindb/common/election.py
```diff
@@ -0,0 +1,248 @@
...
+            return self.INCONCLUSIVE
+        else:
+            return self.ONGOING
+
+    def get_election_result(self, election_id, bigchain):
+        result = bigchain.get_result_by_election_id(election_id, self.DB_TABLE)
+        return result
+
+    @classmethod
+    def is_approved(cls, bigchain, new_height, txns):
```
ldmberman: `is_approved` as a name suggests the function returns a boolean, while in fact it returns either `[]` or an arbitrary `on_approve` result.

I would add a function called `get_approved_election_or_none`, which returns what its name says. Additionally, in the `ValidatorElection` class I would keep `get_validators_update`, which in turn would use the generic `get_approved_election_or_none`.

What do you think?

bigchaindb/backend/localmongodb/query.py
ldmberman: `type` is what is `operation` in the transaction schema. For instance, if you want to find out if there was a validator update caused by the given election ID, you need to filter by type.

kansi: I don't think you need `type` once you have the election id, there shouldn't be two elections with same election id and yet have different `operations`

z-bowen: 1. Good suggestion. Changed the name.

2. When checking the election status, the first thing I want to do is look and see if anything with this `election_id` has been written to the relevant Mongo table, (ie has the vote been previoously concluded.)

ldmberman: Yes, but when you receive an election ID in the command line, you do not know its type for sure. A user may look for a validator election status providing a migration election ID by mistake.

ldmberman:
Regarding the failed tests, see https://github.com/bigchaindb/bigchaindb/pull/2507.

bigchaindb/common/election.py
```diff
@@ -0,0 +1,248 @@
...
+from bigchaindb.common.schema import (_validate_schema,
+                                      TX_SCHEMA_COMMON,
+                                      TX_SCHEMA_CREATE)
+
+
+class Election(Transaction):
+
+    # NOTE: this transaction class extends create so the operation inheritance is achieved
+    # by setting an ELECTION_TYPE and renaming CREATE = ELECTION_TYPE and ALLOWED_OPERATIONS = (ELECTION_TYPE,)
+    ELECTION_TYPE = None
```
kansi: In an attempt to make the code more uniform it would good to rename `ELECTION_TYPE` to `OPERATION` so that further in future when `CREATE` and `TRANSFER` both become an independent class they would also have `OPERATION` field and the registeration would look as follows,

```
Transaction.register_type(Create.OPERATION, models.Transaction)
Transaction.register_type(Transfer.OPEATION, models.Transaction)
Transaction.register_type(ValidatorElection.OPERATION, ValidatorElection)
Transaction.register_type(ValidatorElectionVote.OPERATION, ValidatorElectionVote)
```

bigchaindb/common/election.py
z-bowen: Nope. I checked it. It's correct.

kansi: Don't think so, `bigchain.get_validator_change(height)` calls [`backend.query.get_validator_set(self.connection, height)`](https://github.com/bigchaindb/bigchaindb/blob/cb22557771c2b75fa24607f58a1526e802fbe96f/bigchaindb/lib.py#L430) which returns the validator object of the following format,

```python
{'height': <block_height>,
 'validators': <validator_set>,
 'election_id': <election_id_that_approved_the_change>
}
```

bigchaindb/common/election.py
kansi: But `'VALIDATOR_ELECTION_VOTE'` still exists

bigchaindb/common/election.py
z-bowen: Done

bigchaindb/common/schema/transaction_vote_v2.0.yaml
```diff
@@ -5,14 +5,14 @@
...
-title: Validator Election Vote Schema - Vote on a validator set change
+title: Vote Schema - Vote on an election
 required:
 - operation
 - outputs
 properties:
   operation:
     type: string
-    value: "VALIDATOR_ELECTION_VOTE"
+    value: "OPERATION"
```
kansi: Value is supposed to be `VOTE`

bigchaindb/common/vote.py
```diff
@@ -6,16 +6,16 @@
...
+class Vote(Transaction):
 
-    VALIDATOR_ELECTION_VOTE = 'VALIDATOR_ELECTION_VOTE'
+    VOTE = 'VOTE'
     # NOTE: This class inherits TRANSFER txn type. The `TRANSFER` property is
     # overriden to re-use methods from parent class
-    TRANSFER = VALIDATOR_ELECTION_VOTE
-    ALLOWED_OPERATIONS = (VALIDATOR_ELECTION_VOTE,)
+    TRANSFER = VOTE
+    ALLOWED_OPERATIONS = (VOTE,)
```
kansi: `OPERATION = VOTE` should also be specified.

bigchaindb/common/election.py
```diff
@@ -0,0 +1,249 @@
...
+    def get_election(self, election_id, bigchain):
+        result = bigchain.get_election(election_id)
+        return result
+
+    @classmethod
+    def store_election(cls, bigchain, election, height):
+        bigchain.store_election(height, election)
+
+    @classmethod
+    def is_approved(cls, bigchain, new_height, txns):
```
kansi: `is_approved` as the name suggests should return `True` or `False`

bigchaindb/lib.py
```diff
@@ -483,6 +480,14 @@ def migrate_abci_chain(self):
 
         self.store_abci_chain(block['height'] + 1, new_chain_id, False)
 
+    def store_election(self, height, election):
```
kansi: The doc string says `Store election results` but the function name says `store_election`. It would be better to name the function too as `store_election_results`

bigchaindb/common/vote.py
```diff
@@ -39,20 +39,20 @@ def validate(self, bigchain, current_transactions=[]):
     @classmethod
     def generate(cls, inputs, recipients, election_id, metadata=None):
         (inputs, outputs) = cls.validate_transfer(inputs, recipients, election_id, metadata)
-        election_vote = cls(cls.VALIDATOR_ELECTION_VOTE, {'id': election_id}, inputs, outputs, metadata)
+        election_vote = cls(cls.VOTE, {'id': election_id}, inputs, outputs, metadata)
```
kansi: `cls.OPERATION`

bigchaindb/common/schema/transaction_vote_v2.0.yaml
kansi: It is weird that the CI tests are still passing even when the election created using `.generate` would not match the above schema!

bigchaindb/common/election.py
ldmberman: [A comment was folded](https://github.com/bigchaindb/bigchaindb/pull/2498#discussion_r214612707).

z-bowen: renamed

bigchaindb/lib.py
z-bowen: done

bigchaindb/common/vote.py
z-bowen: done

bigchaindb/common/vote.py
z-bowen: done

bigchaindb/common/schema/transaction_vote_v2.0.yaml
z-bowen: done

kansi: but the value is still `"OPERATION"`?

bigchaindb/common/election.py
```diff
@@ -0,0 +1,249 @@
...
+                                      TX_SCHEMA_CREATE)
+
+
+class Election(Transaction):
+
+    # NOTE: this transaction class extends create so the operation inheritance is achieved
+    # by setting an ELECTION_TYPE and renaming CREATE = ELECTION_TYPE and ALLOWED_OPERATIONS = (ELECTION_TYPE,)
+    OPERATION = None
+    # the model for votes issued by the election
+    VOTE_TYPE = Vote
```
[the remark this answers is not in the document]
z-bowen: I'd still rather leave `Vote` as a default parameter inherited from the base `Election` class. It costs nothing for the derived classes that choose to leave it alone, but also allows us the flexibility to have a specialized `VOTE_TYPE` in the future that builds on the base `Vote`.

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -2,257 +2,53 @@
...
 
-    @classmethod
-    def generate(cls, initiator, voters, election_data, metadata=None):
-        (inputs, outputs) = cls.validate_create(initiator, voters, election_data, metadata)
-        election = cls(cls.VALIDATOR_ELECTION, {'data': election_data}, inputs, outputs, metadata)
-        cls.validate_schema(election.to_dict(), skip_id=True)
-        return election
-
     @classmethod
     def validate_schema(cls, tx, skip_id=False):
```
ldmberman: @kansi do we still need `skip_id` here?

Can we make the schema another parameter and do

```
_validate_schema(TX_SCHEMA_COMMON, tx)
_validate_schema(TX_SCHEMA_CREATE, tx)
_validate_schema(cls.TX_SCHEMA_CUSTOM, tx)
```
in `Election` and
```
cls.validate_id(tx)
_validate_schema(TX_SCHEMA_COMMON, tx)
_validate_schema(TX_SCHEMA_CREATE, tx)
_validate_schema(TX_SCHEMA_CUSTOM, tx)
```
in `Vote`?

bigchaindb/backend/localmongodb/schema.py
```diff
@@ -143,6 +144,14 @@ def create_abci_chains_indexes(conn, dbname):
...
 
+
+def create_elections_secondary_index(conn, dbname):
+    logger.info('Create `elections` secondary index.')
+
+    conn.conn[dbname]['elections'].create_index('election_id',
+                                                name='election_id',
+                                                unique=True,)
+
     logger.info('Create `abci_chains.chain_id` secondary index.')
```
ldmberman: These 2 lines belong to the function above.

bigchaindb/common/vote.py
kansi: I still see `cls.VOTE`

bigchaindb/lib.py
kansi: I still see the function name as `store_election`

bigchaindb/common/election.py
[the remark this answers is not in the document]
ldmberman: It also costs nothing to add `VOTE_TYPE` when we need it. It's not part of the current design and we do not know how custom votes would look like if we introduce them at all.

bigchaindb/common/vote.py
kansi: I don't see `OPERATION = VOTE` specified anywhere

bigchaindb/upsert_validator/validator_election.py
z-bowen: Done

kansi: We can try removing it and see if the tests fail. If they do fail then we can defer removing `skip_id` to a separate PR as that is not the context of this PR.

bigchaindb/common/election.py
```diff
@@ -0,0 +1,249 @@
+# Copyright BigchainDB GmbH and BigchainDB contributors
```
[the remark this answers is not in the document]
z-bowen: Done

bigchaindb/backend/localmongodb/schema.py
z-bowen: Done

bigchaindb/common/election.py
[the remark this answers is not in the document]
z-bowen: Actually, now that I take a second look at the code, we do need custom vote types for each election. On line 235, in the `approved_update` method, we check to see if each transaction is of type `VOTE_TYPE`. While we don't need any custom logic for the different `Vote`s, this loop is going to be more efficient if we only look at votes for the type of election that we're concluding.

I'm going to leave `VOTE_TYPE` in and reinstate the `ValidatorVote` as a subclass of `Vote`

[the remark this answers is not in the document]
z-bowen: The change that will implement different vote types for each election is going to come as part of the PR that generalizes the CLI methods. `run_election_approve` is going to have to get the `VOTE_TYPE` from the `election_id` when it casts its vote.

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,253 @@
...
+
+        * A valid election is one where voters are validators and votes are
+          allocated according to the voting power of each validator node.
+
+        Args:
+            :param bigchain: (BigchainDB) an instantiated bigchaindb.lib.BigchainDB object.
+            :param current_transactions: (list) A list of transactions to be validated along with the election
+
+        Returns:
+            ValidatorElection object
```
kansi: Return value should also have a generalized name

bigchaindb/elections/vote.py
```diff
@@ -28,7 +30,7 @@ def validate(self, bigchain, current_transactions=[]):
             bigchain (BigchainDB): an instantiated bigchaindb.lib.BigchainDB object.
```
kansi: Maybe reference to BEP-21 should be removed from the doc string? /cc @ldmberman

bigchaindb/common/election.py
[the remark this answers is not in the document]
ldmberman: You need to change the vote schema to accept a type; extend the approve command to infer the type, and for every new election we would need to make up a type. I do not think it's worth doing it only to optimise for the conclusion blocks that contain a lot of votes of a different type. If the number of votes of a different type is small, there would be no tangible boosts.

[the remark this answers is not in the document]
ldmberman: There are actually should be no performance gains anyway since you need to iterate over all txs of the block in any case.

tests/tendermint/test_core.py
```diff
@@ -366,7 +366,7 @@ def test_end_block_return_validator_updates(b, init_chain_request):
     resp = app.end_block(RequestEndBlock(height=99))
     assert resp.validator_updates[0] == encode_validator(validator)
 
-    updates = b.get_validator_update()
+    updates = b.is_approved()
```
kansi: There is no `is_approved` I think. It seems to be changed to `on_approved`

tests/tendermint/test_lib.py
```diff
@@ -164,11 +164,11 @@ def test_validator_updates(b, validator_pub_key):
                         'update_id': VALIDATOR_UPDATE_ID}
     query.store_validator_update(b.connection, validator_update)
 
-    updates = b.get_validator_update()
+    updates = b.is_approved()
```
kansi: There is no `is_approved`. It seems to be changed to on_approved

bigchaindb/common/election.py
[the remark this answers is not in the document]
z-bowen: Fixed

bigchaindb/elections/election.py
z-bowen: Fixed

bigchaindb/upsert_validator/validator_election.py
```diff
@@ -2,264 +2,48 @@
...
+        curr_validator_set = bigchain.get_validators(new_height)
+        updated_validator_set = new_validator_set(curr_validator_set,
+                                                  validator_updates)
 
-        if latest_change_height >= election_height:
-            return self.INCONCLUSIVE
-        else:
-            return self.ONGOING
+        updated_validator_set = [v for v in updated_validator_set if v['voting_power'] > 0]
+        bigchain.store_validator_set(new_height+1, updated_validator_set, election.id)
```
kansi: As we decided in our discussion we will not store `election.id` in the validators collection instead store the fact that `election.id` concluded in the `elections` collections

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,251 @@
...
+        Returns:
+            Election: a Election object or an object of the derived Election subclass.
+
+        Raises:
+            ValidationError: If the election is invalid
+        """
+        input_conditions = []
+
+        duplicates = any(txn for txn in current_transactions if txn.id == self.id)
+        if bigchain.get_transaction(self.id) or duplicates:
```
kansi: It would be better to use `bigchain.get_commited` instead of `bigchain.get_transaction`

bigchaindb/upsert_validator/validator_election.py
z-bowen: Done

bigchaindb/upsert_validator/validator_election.py
z-bowen: We're calling `validate_schema` with `skip_id=True` in the `Election.generate()` code. I'm guessing we do this because we've just computed the `tx_id` when generating the election, so no need to repeat this when validating.

In any case, it looks like the `skip` parameter serves a purpose, so let's leave it in.

bigchaindb/elections/election.py
kansi: Sorry, I was talking about [this](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/lib.py#L223). Fixed the comment.

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            return self.ONGOING
+
+    def get_election(self, election_id, bigchain):
+        result = bigchain.get_election(election_id)
+        return result
+
+    @classmethod
+    def store_election_results(cls, bigchain, election, height):
+        bigchain.store_election_results(height, election)
```
kansi: Is nothing being returned here?

bigchaindb/elections/election.py
z-bowen: Done

bigchaindb/elections/election.py
z-bowen: I'll have it return a status response so I can do some error handling in the surrounding function.

tests/upsert_validator/test_upsert_validator_vote.py
```diff
@@ -298,19 +299,22 @@ def test_get_validator_update(b, node_keys, node_key, ed25519_node_keys):
...
+    assert not ValidatorElection.approved_update(b, 4, [tx_vote0, tx_vote1])
 
-    update = ValidatorElection.get_validator_update(b, 4, [tx_vote0, tx_vote1, tx_vote2])
-    update_public_key = codecs.encode(update[0].pub_key.data, 'base64').decode().rstrip('\n')
-    assert len(update) == 1
+    update = ValidatorElection.approved_update(b, 4, [tx_vote0, tx_vote1, tx_vote2])
+    update_public_key = None
+    if update:
+        update_public_key = codecs.encode(update.pub_key.data, 'base64').decode().rstrip('\n')
+    assert update
```
kansi: We could do the following and avoid  `if` and `update_public_key = None`

```python
update = ValidatorElection.approved_update(b, 4, [tx_vote0, tx_vote1, tx_vote2]) 
assert update 
update_public_key = codecs.encode(update.pub_key.data, 'base64').decode().rstrip('\n') 
```

bigchaindb/elections/election.py
z-bowen: No need for a return value here. The return value for the surrounding function is handled by `on_approval`, which is specific to the type of election being called.

tests/upsert_validator/test_upsert_validator_vote.py
```diff
@@ -298,19 +299,22 @@ def test_get_validator_update(b, node_keys, node_key, ed25519_node_keys):
...
 
     b.store_bulk_transactions([tx_vote0, tx_vote1])
 
-    update = ValidatorElection.get_validator_update(b, 4, [tx_vote2])
-    update_public_key = codecs.encode(update[0].pub_key.data, 'base64').decode().rstrip('\n')
-    assert len(update) == 1
+    update = ValidatorElection.approved_update(b, 4, [tx_vote2])
+    if update:
+        update_public_key = codecs.encode(update.pub_key.data, 'base64').decode().rstrip('\n')
+    assert update
```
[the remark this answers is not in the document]
z-bowen: Done

tests/upsert_validator/test_upsert_validator_vote.py
z-bowen: Done

bigchaindb/upsert_validator/validator_election.py
ldmberman: > I'm guessing we do this because we've just computed the tx_id when generating the election, so no need to repeat this when validating.

The parameter was introduced so that we can reuse the same validation function for `CREATE` and `TRANSFER` transactions. When a `CREATE` transaction is built, its ID is computed as a hash of all of its data so there is nothing to validate in the beginning. While for `TRANSFER` its ID is `asset.id` and it can be validated.

In this PR, after we generalised the election process, there are two distinct classes - `Election` and `Vote`, and each of them implements `validate_schema` differently so `skip_id` does not serve any purpose now - let's remove it.

bigchaindb/backend/localmongodb/schema.py
```diff
@@ -142,8 +143,14 @@ def create_abci_chains_indexes(conn, dbname):
     conn.conn[dbname]['abci_chains'].create_index('height',
                                                   name='height',
                                                   unique=True,)
-
-    logger.info('Create `abci_chains.chain_id` secondary index.')
```
ldmberman: There is a conflict resolution leftover.

bigchaindb/backend/query.py
```diff
@@ -351,6 +351,13 @@ def store_validator_set(conn, validator_update):
     raise NotImplementedError
 
 
+@singledispatch
+def store_election_results(conn, validator_update):
```
ldmberman: The parameter was not renamed.

bigchaindb/common/schema/transaction_v2.0.yaml
```diff
@@ -63,7 +63,7 @@ definitions:
     - CREATE
     - TRANSFER
     - VALIDATOR_ELECTION
```
ldmberman: We also need a generic name instead of `VALIDATOR_ELECTION` now.

bigchaindb/common/schema/transaction_validator_election_v2.0.yaml
```diff
@@ -43,6 +45,7 @@ properties:
           - node_id
           - public_key
           - power
+          - seed
```
ldmberman: We don't have to make it required.

kansi: Should it be required for the validator election?

bigchaindb/common/schema/transaction_v2.0.yaml
kansi: What purpose would that serve since the name would always be overriden by the specific election type?

bigchaindb/common/schema/transaction_validator_election_v2.0.yaml
ldmberman: We do not have to require it there either.

bigchaindb/common/schema/transaction_v2.0.yaml
ldmberman: True, there is no purpose in such an operation..

But extending the base schema every time a new election is introduced does not sound necessary either.

I would not require the operation in the base spec. Every specific spec can require the particular operation itself.

Does not have to be done in this PR though.

bigchaindb/common/schema/transaction_validator_election_v2.0.yaml
z-bowen: It's generated automatically on creation of the election, so it will always be present, regardless. But I can remove the required tag, anyway.

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,256 @@
...
+    ELECTION_THRESHOLD = 2 / 3
+
+    @classmethod
+    def get_validator_change(cls, bigchain, height=None):
+        """Return the latest change to the validator set
+
+        :return: {
+            'height': <block_height>,
+            'validators': <validator_set>,
+            'election_id': <election_id_that_approved_the_change>
```
ldmberman: `election_id` is not there anymore.

bigchaindb/common/schema/transaction_v2.0.yaml
z-bowen: I don't think we do...

Since `Election` is a base class, and will be extended with distinct election classes, shouldn't each of those get their own `operations`? The `ValidatorElection` is still using `VALIDATOR_ELECTION` as it's operation.

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @classmethod
+    def generate(cls, initiator, voters, election_data, metadata=None):
+        # Break symmetry in case we need to call an election with the same properties twice
+        uuid = uuid4()
+        election_data['seed'] = str(uuid)
+
+        (inputs, outputs) = cls.validate_create(initiator, voters, election_data, metadata)
+        election = cls(cls.OPERATION, {'data': election_data}, inputs, outputs, metadata)
+        cls.validate_schema(election.to_dict(), skip_id=True)
```
ldmberman: We can also remove `skip_id` from here.

bigchaindb/backend/query.py
z-bowen: Done

bigchaindb/elections/vote.py
```diff
@@ -39,20 +41,20 @@ def validate(self, bigchain, current_transactions=[]):
     @classmethod
     def generate(cls, inputs, recipients, election_id, metadata=None):
         (inputs, outputs) = cls.validate_transfer(inputs, recipients, election_id, metadata)
-        election_vote = cls(cls.VALIDATOR_ELECTION_VOTE, {'id': election_id}, inputs, outputs, metadata)
+        election_vote = cls(cls.OPERATION, {'id': election_id}, inputs, outputs, metadata)
         cls.validate_schema(election_vote.to_dict(), skip_id=True)
```
ldmberman: We can also remove `skip_id` from here.

bigchaindb/elections/election.py
```diff
@@ -0,0 +1,254 @@
...
+            raise InvalidProposer('Public key is not a part of the validator set')
+
+        # NOTE: Check if all validators have been assigned votes equal to their voting power
+        if not self.is_same_topology(current_validators, self.outputs):
+            raise UnequalValidatorSet('Validator set much be exactly same to the outputs of election')
+
+        return self
+
+    @classmethod
+    def generate(cls, initiator, voters, election_data, metadata=None):
```
ldmberman: What do you think about adding an additional parameter `generate_random_seed=True`? - it will simplify those tests that currently have to mock `uuid4` and will clearly display to the reader that generating a seed is optional.

bigchaindb/backend/localmongodb/schema.py
z-bowen: Restored

bigchaindb/common/schema/transaction_v2.0.yaml
ldmberman: Yes. An extra point (we can address it later) is that `bigchaindb/common/schema/transaction_v2.0.yaml` does not have to require the operation to be from the given set. It just makes us extend this list every time a new operation is introduced.

bigchaindb/elections/vote.py
kansi: why?

ldmberman: Because `Vote` is always a `TRANSFER`, and we always validate how identifiers are built for transfers.

PR #2502, opened 2018-08-31
Problem: Need better exception handling for `store_unspent_outputs`

z-bowen:
Solution: Added an error message in case of a `DuplicateKeyError`

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -228,7 +228,9 @@ def store_unspent_outputs(conn, *unspent_outputs):
                 )
             )
         except DuplicateKeyError:
-            # TODO log warning at least
+            raise DuplicateKeyError(
+                f'Duplicate keys in transactions [{unspent_outputs}].'
+            )
             pass
 
 


bigchaindb/backend/localmongodb/query.py
```diff
@@ -228,7 +228,9 @@ def store_unspent_outputs(conn, *unspent_outputs):
                 )
             )
         except DuplicateKeyError:
-            # TODO log warning at least
+            raise DuplicateKeyError(
```
ldmberman: The UTXO implementation is incomplete and currently [disabled](https://github.com/bigchaindb/bigchaindb/commit/8c32ae798bfbad2c580035a8b3bfc9e1959d56d0). There does not seem to be a need to change anything here.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -228,7 +228,9 @@ def store_unspent_outputs(conn, *unspent_outputs):
...
+                f'Duplicate keys in transactions [{unspent_outputs}].'
+            )
             pass
```
kansi: I guess there is no need for `pass`?

PR #2508, opened 2018-09-03
Problem: Public key format not standardized

kansi:
Solution: Add `type` field to public key so that it can be decoded properly

NOTE: This PR doesn't acutally change the format in which public key is stored in an election proposal but rather add a new `type` field which can have values `ed25519-base58` or `ed25519-base64`. The code still assumes that the election asset field has the public key in base58 format but now there is another additional information field `type` that can be used later on to know about the encoding of the public key.


ldmberman:
`type` is assigned but not used anywhere, what is the purpose of the field?

Perhaps, you want to do something like this:
```
def encode_public_key(pk, encoding):
    if pk['encoding'] == encoding:
        return pk
    return pk.encode(pk['encoding']).decode(encoding)

# a place, where base64 is required
encode_public_key(pk, 'base64')

# a place, where base58 is required
encode_public_key(pk, 'base58')
```

kansi:
We are not supporting multiple encodings right now but in case we have to support these encodings we will need the `type` field. So, in future if we wish to migrate to a different encoding scheme we would still be able to validate older transactions using the `type` field.

NOTE: As mentioned in the desription, I didn't add the code to perform such encoding/decoding operation based on `type` field but that can be done too if you insist. As of now I just assumed the encoding/decoding operations.

ldmberman:
Why do not we introduce the field when we need to support multiple encodings? I struggle to see how the field is useful if we do not enforce the compliance.

bigchaindb/common/exceptions.py
```diff
@@ -112,3 +112,7 @@ class UnequalValidatorSet(ValidationError):
 
 class InvalidPowerChange(ValidationError):
     """Raised if proposed power change in validator set is >=1/3 total power"""
+
+
+class InvalidPublicKey(ValidationError):
+    """Raised if proposed power change in validator set is >=1/3 total power"""
```
ldmberman: Copypaste is evil.

tests/commands/test_commands.py
```diff
@@ -429,6 +429,7 @@ def mock_write(tx, mode):
         assert caplog.records[0].msg.__class__ == InvalidPowerChange
 
 
+@pytest.mark.google
```
ldmberman: What is this mark for?

vrde:
Picking up @ldmberman suggestion, I this that at this point we should use a more generic encode/decode function. Would it make sense to split `type` into `type` and `encoding`, and having something like
```
{
    'value': '...',
    'type': 'ed25519',
    'encoding': 'base16'
}
```

It's more clear if we normalize the data and put `type` and `encoding` in two different values, otherwise we need to remember that the `type` string should be split using `-` and add extra code. In this format we have everything ready. Moreover, we can have a nice generic function to decode public keys, like:

```python
def decode_public_key(value, type, encoding):
    ... do the decoding part ...
    return public_key

public_key = {
    'value': '...',
    'type': 'ed25519',
    'encoding': 'base16'
}

decoded_public_key = decode_public_key(**public_key)
```

kansi:
> At this point we should use a more generic encode/decode function. Would it make sense to split type into type and encoding, and having something like

I thought about doing so but then it occurred to me that any public key is always going to be encoded in some format which implies that public key type and the base in which it is encoded are closely tied. So I thought it would be better to have them both in one place.

> It's more clear if we normalize the data and put type and encoding in two different values, otherwise we need to remember that the type string should be split using - and add extra code.

Is there a situation you anticipate that the we would need to spilt the type into curve and base encoding? If so then could you please give an example?

vrde:
> Is there a situation you anticipate that the we would need to spilt the type into curve and base encoding? If so then could you please give an example?

It's a matter of separation of concerns, or normalization. A public key has a value, a curve type, and an encoding. I find it weird that the value is stored in a field, and curve+encoding in another field. We can just add another field and call it encoding.

I can bring a couple of example to the table. We talked about adding other curve types, like `secp256k1`. Plus, a commonly used encoding is `base48`. If we want to add this curve and encoding, then would end up with a really long list of combos in the function `get_public_key_decoder` and eventually in other schema files.

kansi:
I would urge you to note that I am trying to discuss a very specific scenario here i.e. I agree that on a general day we should separate the type and encoding of the public key but in this particular case, we are trying to return a properly encoded validator to Tendermint. In this case, we need to always return `base16` encoded `ed25519` public key. So whenever we are in a position to send back the public key we need to know both the things simultaneously so imho it would make sense to de-normalize.

Having said the above I don't think the change you are suggesting is huge and I can implement the same if you would like me to. So please let me know either way.

PR #2510, opened 2018-09-04
Problem: The commit mode behaves incorrectly.

ldmberman:
Solution: Parse the Tendermint response properly. The functionality was disabled in https://github.com/bigchaindb/bigchaindb/pull/2235/files#diff-c6511560546a7dc577e7e647b5bfdaceL68 and was not fixed since then.

### DIFF
--- a/bigchaindb/lib.py
+++ b/bigchaindb/lib.py
@@ -99,7 +99,7 @@ class BigchainDB(object):
         result = response['result']
         if mode == self.mode_commit:
             check_tx_code = result.get('check_tx', {}).get('code', 0)
-            deliver_tx_code = result['deliver_tx'].get('code', 0)
+            deliver_tx_code = result.get('deliver_tx', {}).get('code', 0)
             error_code = check_tx_code or deliver_tx_code
         else:
             error_code = result.get('code', 0)
--- a/bigchaindb/lib.py
+++ b/bigchaindb/lib.py
@@ -96,6 +96,10 @@ class BigchainDB(object):
     def _process_post_response(self, response, mode):
         logger.debug(response)
 
+        error = response.get('error')
+        if error:
+            return (500, error)
+
         result = response['result']
         if mode == self.mode_commit:
             check_tx_code = result.get('check_tx', {}).get('code', 0)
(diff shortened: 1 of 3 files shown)


bigchaindb/lib.py
```diff
@@ -94,29 +95,23 @@ def write_transaction(self, transaction, mode):
...
-        if response.get('error') is not None:
-            return (500, 'Internal error')
+
+        result = response['result']
+        if mode == self.mode_commit:
+            check_tx_code = result.get('check_tx', {}).get('code', 0)
+            deliver_tx_code = result['deliver_tx'].get('code', 0)
+            error_code = check_tx_code or deliver_tx_code
+        else:
+            error_code = result.get('code', 0)
```
kansi: Can we have test case for this response object?

bigchaindb/lib.py
```diff
@@ -94,29 +95,23 @@ def write_transaction(self, transaction, mode):
 
     def _process_post_response(self, response, mode):
         logger.debug(response)
-        if response.get('error') is not None:
-            return (500, 'Internal error')
+
+        result = response['result']
+        if mode == self.mode_commit:
+            check_tx_code = result.get('check_tx', {}).get('code', 0)
```
vrde: If you omit the second parameter, `get` returns `None` if the lookup is unsuccessful, so you can just write `check_tx_code = result.get('check_tx', {}).get('code')` if you want.

ldmberman: `0` is the success code the API uses so I put it there but `None` would work here as well. I do not have a preference.

bigchaindb/lib.py
```diff
@@ -94,29 +95,23 @@ def write_transaction(self, transaction, mode):
 
     def _process_post_response(self, response, mode):
         logger.debug(response)
-        if response.get('error') is not None:
-            return (500, 'Internal error')
+
+        error = response.get('error')
+        if error:
+            return (500, error)
```
ttmc: Shouldn't we return an HTTP 400 "Bad Request" rather than 500 "Internal Server Error"? That's what [the HTTP API docs](http://docs.bigchaindb.com/projects/server/en/master/http-client-server-api.html#post--api-v1-transactions?mode=mode) say happens:

screenshot from 2018-09-06 14-38-03

If not, we need to update the HTTP API docs.

ldmberman: This is how Tendermint docs explain the error field:

> only returns error if mempool.BroadcastTx errs (ie. problem with the app) or if we timeout waiting for tx to commit

This is thus a server error and it can occur even if the posted tx is valid.

ttmc: Ah, okay.

PR #2514, opened 2018-09-05
Problem: Txs posted in async/sync can be lost.

ldmberman:
Solution: Document when the transactions posted in the async and sync modes can be lost. Describe how they are stored. Mention that BigchainDB does not store or expose rejected transactions.

### DIFF
--- a/docs/server/source/http-client-server-api.rst
+++ b/docs/server/source/http-client-server-api.rst
@@ -154,7 +154,15 @@ Transactions
    is in a committed block.
 
    .. note::
-   
+       In the async and sync modes, after a successful HTTP response is returned, the transaction may still be rejected later on. All the transactions are recorded internally by Tendermint in WAL (Write-Ahead Log) before the HTTP response is returned. Nevertheless, the following should be noted:
+
+       - Transactions in WAL including the failed ones are not exposed in any of the BigchainDB or Tendermint APIs.
+       - Transactions are never fetched from WAL. WAL is never replayed.
+       - A critical failure (e.g. the system is out of disk space) may occur preventing transactions from being stored in WAL, even when the HTTP response indicates a success.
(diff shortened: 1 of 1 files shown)


docs/server/source/http-client-server-api.rst
```diff
@@ -154,7 +154,15 @@ Transactions
    is in a committed block.
 
    .. note::
-   
+       In the async and sync modes, after a successful HTTP response is returned, the transaction may still be rejected later on. All the transactions are recorded internally before the HTTP response is returned. Nevertheless, the following should be noted:
+
+       - Recorded transactions including the failed ones are not exposed in any of the BigchainDB or Tendermint APIs.
```
ttmc: Surely there must be _some_ Tendermint API to get the block at a given height, including its list of transactions? What about this one:

https://tendermint.github.io/slate/?shell#block

ldmberman: I am referring to WAL records here. I do want to mention that they are recorded but do not want to provide any details or use the WAL term since those are not well documented and can change any time. Perhaps, you can think about a better way to call those transactions?

docs/server/source/http-client-server-api.rst
```diff
@@ -154,7 +154,15 @@ Transactions
    is in a committed block.
 
    .. note::
-
```
ttmc: I think a blank line might be necessary to keep Spinx happy.

docs/server/source/http-client-server-api.rst
ttmc: Oh, sorry, I got confused by the phrase "recorded transactions" thinking it meant something more broad. I think it's probably okay to mention the WAL (but spell it out), maybe with a note that Tendermint might change how tat works in the future.

docs/server/source/http-client-server-api.rst
ldmberman: It rendered fine for me. Other notes here also have a blank line.

ttmc: ok

PR #2520, opened 2018-09-06
Problem: New collections are not created in existing DB.

ldmberman:
Solution: Do not abort the initialisation if a collection exists. Unify the index creation.

### DIFF
--- a/bigchaindb/backend/localmongodb/schema.py
+++ b/bigchaindb/backend/localmongodb/schema.py
@@ -7,9 +7,9 @@
 import logging
 
 from pymongo import ASCENDING, DESCENDING, TEXT
+from pymongo.errors import CollectionInvalid
 
 from bigchaindb import backend
-from bigchaindb.common import exceptions
 from bigchaindb.backend.utils import module_dispatch_registrar
 from bigchaindb.backend.localmongodb.connection import LocalMongoDBConnection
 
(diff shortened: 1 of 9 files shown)


bigchaindb/backend/localmongodb/schema.py
```diff
@@ -32,118 +64,22 @@ def create_database(conn, dbname):
...
     for table_name in backend.schema.TABLES:
-        logger.info('Create `%s` table.', table_name)
         # create the table
         # TODO: read and write concerns can be declared here
-        conn.conn[dbname].create_collection(table_name)
+        try:
+            logger.info(f'Create `{table_name}` table.')
+            conn.conn[dbname].create_collection(table_name)
+            create_indexes(conn, dbname, table_name, INDEXES[table_name])
+        except CollectionInvalid:
```
kansi: The class name of this sample seems generic, is there a way to specifically determine that the error is in fact because of the fact that the collection already exists?

ldmberman: I did not manage to find anything more specific. `pymongo` docs explain:

> CollectionInvalid will be raised if the collection already exists.

bigchaindb/backend/localmongodb/schema.py
```diff
@@ -32,118 +64,22 @@ def create_database(conn, dbname):
...
 def create_tables(conn, dbname):
     for table_name in backend.schema.TABLES:
-        logger.info('Create `%s` table.', table_name)
         # create the table
         # TODO: read and write concerns can be declared here
-        conn.conn[dbname].create_collection(table_name)
+        try:
+            logger.info(f'Create `{table_name}` table.')
+            conn.conn[dbname].create_collection(table_name)
+            create_indexes(conn, dbname, table_name, INDEXES[table_name])
```
kansi: What if we just add new index?

ldmberman: We can. But we would need to change logging then.  Unconditional `Creating indexes for ..` messages are disorienting.

tests/commands/test_commands.py
```diff
@@ -78,26 +78,6 @@ def test_bigchain_show_config(capsys):
     assert output_config == config
 
 
-@pytest.mark.tendermint
-def test_bigchain_run_init_when_db_exists(mocker, capsys):
```
kansi: I guess this test could re-purposed to asset a success message?

ldmberman: This particular test does not use a real database but asserts that one function catches an exception raised by the other.

bigchaindb/backend/localmongodb/schema.py
kansi: Then we should change logging and ensure all the indexes exist too

ldmberman: Why?

ldmberman: I was thinking about it. Most of our indexes come with a unique constraint. Unique constraints are not backwards-compatible and thus can not be introduced by a new `bigchaindb start` or `bigchaindb init`.

On the other hand, we are losing a good means of adding those rare indexes without constraints.. So, I think it's better to arrange it how you suggest.

tests/commands/test_commands.py
ldmberman: There is a pattern of not using the real database in this file so I would leave it to be handled by  future test (re)organisation efforts.

tests/conftest.py
```diff
@@ -270,7 +264,6 @@ def signed_create_tx(alice, create_tx):
     return create_tx.sign([alice.private_key])
 
 
-@pytest.mark.abci
```
vrde: Is it something related to this PR?

bigchaindb/backend/localmongodb/schema.py
```diff
@@ -18,12 +18,47 @@
 register_schema = module_dispatch_registrar(backend.schema)
 
 
+INDEXES = {
```
vrde: I really like this!

PR #2522, opened 2018-09-06
Problem: There are unnecessary markers

codegeschrei:
Solution: remove the tendermint marker for tests

I deleted the test `test_get_blocks_by_txid_endpoint_returns_400_bad_query_params` since we already have it in `test_block`


tests/commands/test_commands.py
```diff
@@ -42,22 +40,21 @@ def test_main_entrypoint(mock_start):
...
+@patch('bigchaindb.commands.bigchaindb._run_init')
+@patch('bigchaindb.config_utils.autoconfigure')
+def test_bigchain_run_start(mock_setup_logging, mock_run_init,
+                            mock_autoconfigure, mock_processes_start):
     from bigchaindb.commands.bigchaindb import run_start
     args = Namespace(config=None, yes=True,
                      skip_initialize_database=False)
     run_start(args)
-    mocked_setup_logging.assert_called_once_with(user_log_config=config['log'])
+    assert mock_setup_logging.called
```
z-bowen: What's the motivation for the weaker assertion?

tests/commands/test_commands.py
```diff
@@ -235,22 +223,19 @@ def test_run_start_when_db_already_exists(mocker,
...
+    mocked_start = mocker.patch('bigchaindb.start.start')
 
     def mock_run_init():
         raise DatabaseAlreadyExists()
     monkeypatch.setattr('builtins.input', lambda: '\x03')
     monkeypatch.setattr(
         'bigchaindb.commands.bigchaindb._run_init', mock_run_init)
     run_start(run_start_args)
-    mocked_setup_logging.assert_called_once_with(user_log_config=config['log'])
     assert mocked_start.called
```
z-bowen: Same question here: why are we no longer asserting the `args`

tests/commands/test_commands.py
codegeschrei: The removed line does not work like this anymore. Before `setup_logging` had parameters you could provide but that's not the case anymore.

tests/commands/test_commands.py
codegeschrei: see above

tests/web/test_block_tendermint.py
```diff
@@ -73,29 +71,3 @@ def test_get_blocks_by_txid_endpoint_returns_empty_list_not_found(client):
...
-    res = client.get(BLOCKS_ENDPOINT + '?transaction_id=123&foo=123')
-    assert res.status_code == 400
-    assert res.json == {
-        'message': 'Unknown arguments: foo'
-    }
-
-    res = client.get(BLOCKS_ENDPOINT + '?transaction_id=123&status=123')
-    assert res.status_code == 400
-    assert res.json == {
-        'message': 'Unknown arguments: status'
```
z-bowen: Why is the error response message different between these two tests? The one in `test_blocks` asserts 'Unknown arguments: status'. How did these ever both pass?

PR #2528, opened 2018-09-10
Problem: No log rotation when BigchainDB and Tendermint started with Monit

muawiakh:
## Solution
Currently, BigchainDB server, takes care of [log rotation using](https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/log.py) `logging` module.
This enforces log rotation if the file size exceeds `209715200` i.e. for `bigchaindb-benchmark.log`, `bigchaindb.log`, `bigchaindb-errors.log`.

- Remove the extra `bigchaindb.err.log` and `bigchaindb.out.log` redirection, because BigchainDB also stores logs and has log rotation for the relevant files instead redirect to `/dev/null`
- Introduce `logrotate` script.
- Fix a minor log message.
- Add tendermint log rotation for `tendermint.err.log` and `tendermint.out.log` if the size exceeds `200 MB`.
- Add docs for log rotation

### DIFF
--- a/docs/server/source/appendices/log-rotation.md
+++ b/docs/server/source/appendices/log-rotation.md
@@ -22,12 +22,13 @@ please refer to [MongoDB docs](https://docs.mongodb.com/v3.6/tutorial/rotate-log
 
 ## Log rotation for BigchainDB
 
-Log rotation is baked into BigchainDB server using the `logging` module. If you followed the [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide. BigchainDB processes are monitored using [Monit]( https://www.mmonit.com/monit). BigchainDB server logs information into the following files:
+Log rotation is baked into BigchainDB server using the `logging` module. BigchainDB server logs information into the following files:
  - `bigchaindb.log`
  - `bigchaindb-errors.log`
  - `bigchaindb-benchmark.log`
 
-These log files are created by default in the directory from you run `bigchaindb start`, if you are using `monit` the default directory is: `$HOME/.bigchaindb-monit/logs`
+These log files are created by default in the directory from where you run `bigchaindb start`, if you are using `monit`, from 
+[How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide, the default directory is: `$HOME/.bigchaindb-monit/logs`
 
 The logs for BigchainDB server are rotated when any of the above mentioned file exceeds `209715200 bytes (i.e. approximately 209 MB).`.
 
--- a/docs/server/source/appendices/log-rotation.md
+++ b/docs/server/source/appendices/log-rotation.md
@@ -1,3 +1,4 @@
+<!---
 Copyright BigchainDB GmbH and BigchainDB contributors
 SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
 Code is Apache-2.0 and docs are CC-BY-4.0
@@ -26,14 +27,14 @@ Log rotation is baked into BigchainDB server using the `logging` module. If you 
  - `bigchaindb-errors.log`
  - `bigchaindb-benchmark.log`
 
-Default location for these log files is: `$HOME/.bigchaindb-monit/logs`
+These log files are created by default in the directory from you run `bigchaindb start`, if you are using `monit` the default directory is: `$HOME/.bigchaindb-monit/logs`
 
-The logs for BigchainDB server are rotated when any of the above mentioned file exceeds `209715200 bytes`.
+The logs for BigchainDB server are rotated when any of the above mentioned file exceeds `209715200 bytes (i.e. approximately 209 MB).`.
 
 
 ## Log rotation for Tendermint
 
-Log rotation for Tendermint is handled by [Monit]( https://www.mmonit.com/monit). When we start BigchainDB and Tendermint using [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide.
+In order to set up log rotation of Tendermint, you will need to use the [Monit]( https://www.mmonit.com/monit) scripts provided by us. Covered in the [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide.
 
 ```bash
 $ monit -d 1
--- a/docs/server/source/appendices/log-rotation.md
+++ b/docs/server/source/appendices/log-rotation.md
@@ -1,0 +1,46 @@
+Copyright BigchainDB GmbH and BigchainDB contributors
+SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+Code is Apache-2.0 and docs are CC-BY-4.0
+--->
+
+# Log rotation for a BigchainDB node
+
+Each BigchainDB node comprises of 3 main services:
+  - BigchainDB server
+  - Tendermint
+  - MongoDB
+
+To run a BigchainDB test network/dev node, that is expected to run for relatively longer periods
+of time, we need to consider the log rotation of these services i.e. we do not want the logs taking
+up large amounts of storage and making the node unresponsive or get into bad state.
+
+## Log rotation for MongoDB
+
+Currently, we leave the log rotation of MongoDB to the BigchainDB administrator. For more notes on MongoDB log rotation
+please refer to [MongoDB docs](https://docs.mongodb.com/v3.6/tutorial/rotate-log-files/).
+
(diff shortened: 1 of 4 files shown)


pkg/scripts/bigchaindb-monit-config
```diff
@@ -110,6 +110,7 @@ case \$1 in
   start_tendermint)
 
     pushd \$4
+
       nohup tendermint node --consensus.create_empty_blocks=false >> \$3/tendermint.out.log 2>> \$3/tendermint.err.log &
```
ttmc: I'm not sure we need the `--consensus.create_empty_blocks=false` in the command to start Tendermint, because that's ensured by the settings we tell them to put in `config.toml`, i.e. see http://docs.bigchaindb.com/projects/server/en/master/simple-deployment-template/network-setup.html#member-connect-to-the-other-members

pkg/scripts/bigchaindb-monit-config
```diff
@@ -93,7 +93,7 @@ case \$1 in
   start_bigchaindb)
 
     pushd \$4
-      nohup bigchaindb -l DEBUG start >> \$3/bigchaindb.out.log 2>> \$3/bigchaindb.err.log &
+      nohup bigchaindb -l DEBUG start > /dev/null 2>&1 &
```
shahbazn: Why set logging level to debug? I think its better to leave logging flag out of the command here and let it pick up logging configs from the config file.

pkg/scripts/bigchaindb-monit-config
```diff
@@ -161,9 +178,17 @@ check process bigchaindb
 check process tendermint
     with pidfile ${monit_pid_path}/tendermint.pid
     start program "${monit_script_path} start_tendermint ${monit_pid_path}/tendermint.pid ${monit_log_path} ${monit_log_path}"
-    restart program "${monit_script_path} start_bigchaindb ${monit_pid_path}/bigchaindb.pid ${monit_log_path} ${monit_log_path}"
+    restart program "${monit_script_path} start_tendermint ${monit_pid_path}/tendermint.pid ${monit_log_path} ${monit_log_path}"
```
shahbazn: should a restart first stop and then start?

muawiakh: Not necessarily, restart just happens when the service is not running.

pkg/scripts/bigchaindb-monit-config
muawiakh: It is a test network, but I can change if that is necessary.

docs/server/source/appendices/log-rotation.md
```diff
@@ -0,0 +1,46 @@
+Copyright BigchainDB GmbH and BigchainDB contributors
```
ttmc: The block at the top should be an HTML comment (so it doesn't render on ReadTheDocs). To do that, you will have to add `<!--` or `<!---` before it, maybe as a new line 1. (All HTML markup is valid Markdown markup.)

docs/server/source/appendices/log-rotation.md
```diff
@@ -0,0 +1,46 @@
...
+## Log rotation for BigchainDB
+
+Log rotation is baked into BigchainDB server using the `logging` module. If you followed the [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide. BigchainDB processes are monitored using [Monit]( https://www.mmonit.com/monit). BigchainDB server logs information into the following files:
+ - `bigchaindb.log`
+ - `bigchaindb-errors.log`
+ - `bigchaindb-benchmark.log`
+
+Default location for these log files is: `$HOME/.bigchaindb-monit/logs`
+
+The logs for BigchainDB server are rotated when any of the above mentioned file exceeds `209715200 bytes`.
```
ttmc: I wonder why that number of bytes. I guess that is the default inherited from the `logging` module?

pkg/scripts/bigchaindb-monit-config
ttmc: Some people might use our Monit script for their production node.

pkg/scripts/bigchaindb-monit-config
```diff
@@ -129,21 +130,37 @@ exit 0
...
 
+cat >${monit_script_path}_logrotate <<EOF
+#!/bin/bash
+case \$1 in
+
+  rotate_tendermint_logs)
+    /bin/mv \$2 \$2.\$(date +%y-%m-%d)
+    /bin/tar -cvf \$2.\$(date +%Y%m%d_%H%M%S).tar.gz \$2.\$(date +%y-%m-%d)
+    /bin/rm \$2.\$(date +%y-%m-%d)
+    /bin/kill -2 \$(cat \$3)
```
ldmberman: I think, you can clear the file without removing it, then Tendermint does not have to be restarted. One way of clearing it is `cp /dev/null \$2`.

docs/server/source/appendices/log-rotation.md
```diff
@@ -0,0 +1,46 @@
...
+up large amounts of storage and making the node unresponsive or get into bad state.
+
+## Log rotation for MongoDB
+
+Currently, we leave the log rotation of MongoDB to the BigchainDB administrator. For more notes on MongoDB log rotation
+please refer to [MongoDB docs](https://docs.mongodb.com/v3.6/tutorial/rotate-log-files/).
+
+## Log rotation for BigchainDB
+
+Log rotation is baked into BigchainDB server using the `logging` module. If you followed the [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide. BigchainDB processes are monitored using [Monit]( https://www.mmonit.com/monit). BigchainDB server logs information into the following files:
```
ldmberman: Should we mention Monit in the BigchainDB section?

Also, makes sense to add that `bigchaindb.log` and other log files are by default created in the directory from where `bigchaindb` is run.

docs/server/source/appendices/log-rotation.md
```diff
@@ -0,0 +1,46 @@
...
+ - `bigchaindb-benchmark.log`
+
+Default location for these log files is: `$HOME/.bigchaindb-monit/logs`
+
+The logs for BigchainDB server are rotated when any of the above mentioned file exceeds `209715200 bytes`.
+
+
+## Log rotation for Tendermint
+
+Log rotation for Tendermint is handled by [Monit]( https://www.mmonit.com/monit). When we start BigchainDB and Tendermint using [How to Set Up a BigchainDB Network](../simple-deployment-template/network-setup.md) guide.
```
ldmberman: > Log rotation for Tendermint is handled by Monit.

It kinda suggests that Monit is always there. I would say something like "you need to use Monit setup to use the log rotation scripts prepared by us".

docs/server/source/appendices/log-rotation.md
muawiakh: Its already part of the BigchainDB code. https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/log.py#L49

Not entirely sure why.

docs/server/source/appendices/log-rotation.md
```diff
@@ -0,0 +1,46 @@
+Copyright BigchainDB GmbH and BigchainDB contributors
+SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+Code is Apache-2.0 and docs are CC-BY-4.0
+--->
+
+# Log rotation for a BigchainDB node
```
ttmc: This is a new docs page so it must be added to an `index.rst` file otherwise Sphinx will say, for example:
```bash
Warning, treated as error:
/home/troy/repos/bigchaindb/docs/server/source/appendices/log-rotation.md:document isn't included in any toctree
```

It's in the appendices directory, so that's the `index.rst` file that should be updated, i.e.

`docs/server/source/appendices/index.rst`

docs/server/source/appendices/log-rotation.md
ttmc: Maybe write "209715200 bytes (i.e. approximately 209 MB)." since 209 MB is easier to understand for humans.

docs/server/source/appendices/log-rotation.md
muawiakh: I rephrased the sentence, i think it makes sense now for a `monit` based user as well as a normal user.

docs/server/source/appendices/log-rotation.md
muawiakh: forgot to add it will add in next commit.

PR #2529, opened 2018-09-10
Problem: Events API is unreliable

### LINKED ISSUE #2434: tests/tendermint/test_event_stream.py::test_subscribe_events is brittle
Every once in a while the test fails on Travis CI.

Attaching the logs from such a failure.

[test_subscribe_events.log](https://github.com/bigchaindb/bigchaindb/files/2253116/test_subscribe_events.log)

### LINKED ISSUE #2442: Problem: Reading from event stream API is highly unreliable
When the system is under heavy load reading from bigchaindb event stream API is highly unreliable. This behaviour was observed while doing benchmark testing using this [tool](https://github.com/bigchaindb/benchmark). For comparison when I updated the benchmark tool [here](https://github.com/shahbazn/benchmark/pull/2) to read directly from tendermint instead of bigchaindb event stream, the results became much more reliable.

To reproduce the issue run benchmark test with high number of requests on the network:
`bigchaindb-benchmark --csv 18.csv -p http://<node-1-ip>:9984 -p http://<node-1-ip>:9984 -p http://<node-1-ip>:9984 -p http://<node-1-ip>:9984 --processes 16 send -r 100000 --mode sync`

### LINKED ISSUE #2506: Problem: Events API returns all txs recorded by Tendermint in blocks, including invalid ones
We recently realized that if BigchainDB's response to DeliverTx tells Tendermint that a transaction is invalid, Tendermint still records that transaction in its block. (BigchainDB _doesn't_ record invalid transactions in MongoDB.) That's not a bug on Tendermint's part: invalid transactions are needed as proof of Byzantine behavior, so Byzantine nodes can be punished later on. There's a good discussion of this and related ideas in the Tendermint issue https://github.com/tendermint/tendermint/issues/2175 .

## The Problem

Currently, the Events API reports the latest block of transactions _according to Tendermint_, so there might be invalid transactions in that block.

## A Solution Idea

MongoDB now has something like RethinkDB's "changefeeds" called "[Change Streams](https://docs.mongodb.com/manual/changeStreams/)." They were new as of MongoDB 3.6. See "[An Introduction to Change Streams](https://www.mongodb.com/blog/post/an-introduction-to-change-streams)" in the MongoDB Blog. BigchainDB could get a feed of all the latest valid transactions by subscribing to a change stream on [the transactions collection](https://bigchaindb.readthedocs.io/en/latest/query.html#example-documents-from-transactions) in MongoDB.

BigchainDB could return the same info as what the Events API currently returns.

Question: [The docs](http://docs.bigchaindb.com/projects/server/en/master/events/websocket-event-stream-api.html#id3) say that what gets returned is something like:

```js
{
    "transaction_id": "<sha3-256 hash>",
    "asset_id": "<sha3-256 hash>",
    "block_id": "<int>"
}
```
Is that right? If so, shouldn't it be giving the "block_height" rather than the "block_id"?

vrde:
Solution: Stop getting events from the Tendermint events API. Get the
event right before returning from the COMMIT phase of Tendermint and
publish it to the events queue.

Fixes #2442 
Fixes #2506
Fixes #2434


bigchaindb/web/websocket_server.py
```diff
@@ -50,13 +50,13 @@ def _multiprocessing_to_asyncio(in_queue, out_queue, loop):
...
-        except KeyError:
-            asset_id = tx['id']
+        if tx.asset:
+            asset_id = tx.asset.get('id', tx.id)
+        else:
+            asset_id = tx.id
         yield {'height': block['height'],
                'asset_id': asset_id,
-               'transaction_id': tx['id']}
+               'transaction_id': tx.id}
```
ttmc: While we're busy fixing the events API, we may as well fix the docs about it. If you look in the docs subsection

http://docs.bigchaindb.com/projects/server/en/master/events/websocket-event-stream-api.html#id3

it gives this "Example message:"

```js
{
    "transaction_id": "<sha3-256 hash>",
    "asset_id": "<sha3-256 hash>",
    "block_id": "<int>"
}
```

Clearly, `'block_id'` should be replaced by `'height'`.

ttmc: The sentence right before that also refers to the "containing block’s ID." but it should say "containing block's height."

vrde: I'll use `int`. Given that the biggest integer in JavaScript is `9007199254740991`, and if a BigchainDB network creates a block every second, we will have problems in 286 million years.

At that point in time we would have other problems to deal with, since _all the continents on Earth may fuse into a supercontinent_ (from [Wikipedia](https://en.wikipedia.org/wiki/Timeline_of_the_far_future)).

vrde:
> Seems good. It would be good to test how reliable its the events api now under stress

I'm implicitly testing this with the performance tests. I'm able to push 1M transactions and get them through the WebSocket API without any problem. Before it was not possible. I also think this way of getting events removes some load from both BigchainDB and Tendermint (with the previous version of the code, I saw some delay between committing a block and having the transactions through the WebSocket).

PR #2532, opened 2018-09-12
Updated the Python Driver version in acceptance/python/Dockerfile

ttmc:
because we just released a new version of the Python Driver: [v0.5.3](https://github.com/bigchaindb/bigchaindb-driver/releases/tag/v0.5.3)

### DIFF
--- a/acceptance/python/Dockerfile
+++ b/acceptance/python/Dockerfile
@@ -5,5 +5,5 @@ RUN pip install --upgrade \
 	pycco \
 	websocket-client~=0.47.0 \
 	pytest~=3.0 \
-	bigchaindb-driver==0.5.3 \
+	bigchaindb-driver~=0.5.0 \
 	blns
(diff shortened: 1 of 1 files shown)


acceptance/python/Dockerfile
```diff
@@ -5,5 +5,5 @@ RUN pip install --upgrade \
 	pycco \
 	websocket-client~=0.47.0 \
 	pytest~=3.0 \
-	bigchaindb-driver==0.5.2 \
+	bigchaindb-driver==0.5.3 \
```
vrde: If we use the `~` operator we can let the package manager handle minor versions. What about `bigchaindb-driver~=0.5.0`?

ttmc: Good idea.

ttmc:
@vrde I made the change you suggested.

PR #2535, opened 2018-09-12
Create migration election class

z-bowen:
**Problem: We need a way to synchronize a halt to block production to manage migrations**

## Solution

Created the `MigrationElection` class


bigchaindb/commands/election_types.py
```diff
@@ -16,5 +16,14 @@
                 'help': 'Path to the private key of the election initiator.'
             }
         }
+    },
+    'migration': {
```
ldmberman: This particular migration is connected to the ABCI (Tendermint) chain upgrades. We might introduce other migrations in the future. So I would call it `tendermint-chain-migration` or `abci-chain-migration`.

bigchaindb/common/schema/transaction_migration_election_v2.0.yaml
```diff
@@ -0,0 +1,44 @@
+# Copyright BigchainDB GmbH and BigchainDB contributors
+# SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+# Code is Apache-2.0 and docs are CC-BY-4.0
+
+---
+"$schema": "http://json-schema.org/draft-04/schema#"
```
ldmberman: This schema has a lot of unnecessary details. We should just validate the transaction agains the base election schema [documented here](https://github.com/bigchaindb/BEPs/pull/82/files) and additionally assert the operation is `TENDERMINT_MIGRATION_ELECTION`.

tests/upsert_validator/test_upsert_validator_vote.py
```diff
@@ -243,15 +243,15 @@ def test_upsert_validator(b, node_key, node_keys, ed25519_node_keys):
                                           new_validator, None).sign([node_key.private_key])
     code, message = b.write_transaction(election, 'broadcast_tx_commit')
     assert code == 202
-    time.sleep(3)
+    time.sleep(1)
```
ldmberman: Why do you have a `sleep` here? The commit mode ensures the data is persisted when you get a response.

`sleep`s are inefficient because you always end up waiting for some extra time. If you instead continuously poll the resource in question and fail after a certain timeout, the test only wastes extra time in the unsuccessful case.

bigchaindb/commands/election_types.py
z-bowen: The keys from this `dict` directly translate into the CLI methods, as in `election new tendermint-chain-migration`. I'd prefer to keep them short if possible, for user convenience. How about `chain-migration` as a compromise?

tests/upsert_validator/test_upsert_validator_vote.py
z-bowen: I tried pulling the `sleep` entirely, and it broke the test. There must be some lag before the tx shows up in the database...

ldmberman: Yeah, we need to investigate.

bigchaindb/commands/bigchaindb.py
```diff
@@ -158,6 +159,40 @@ def run_election_new_upsert_validator(args, bigchain):
...
+        election = MigrationElection.generate([key.public_key],
+                                              voters,
+                                              {}, None).sign([key.private_key])
+        election.validate(bigchain)
+    except ValidationError as e:
+        logger.error(e)
+        return False
+    except FileNotFoundError as fd_404:
+        logger.error(fd_404)
+        return False
```
kansi: I think the above logic is same for `election new upsert-validator`? It would be good to abstract

bigchaindb/commands/bigchaindb.py
```diff
@@ -158,6 +159,40 @@ def run_election_new_upsert_validator(args, bigchain):
...
+
+    resp = bigchain.write_transaction(election, 'broadcast_tx_commit')
+    if resp == (202, ''):
+        logger.info('[SUCCESS] Submitted proposal with id: {}'.format(election.id))
+        return election.id
+    else:
+        logger.error('Failed to commit election proposal')
+        return False
```
kansi: I think the above logic is same for `election new upsert-validator`? It would be good to abstract

bigchaindb/common/schema/transaction_migration_election_v2.0.yaml
kansi: I would suggest merging (if possible) the two yaml files i.e. the base election spec and migration election spec once they are loaded in `_load_schema` function. This we would only need one additional schema validation instead of two but we will still retain different base election spec and migration election sepc.

ldmberman: I do not see the yaml file to be a necessity but does not hurt to have it either. The important bit - the only thing it should describe is custom operation.

tests/upsert_validator/test_upsert_validator_vote.py
ldmberman: It fails because the test calls Tendermint, not BigchainDB (`b.endpoint` points to Tendermint's RPC API). To fix it, fetch validators via `b.get_validators()`.

bigchaindb/core.py
```diff
@@ -209,15 +209,14 @@ def end_block(self, request_end_block):
         else:
             self.block_txn_hash = block['app_hash']
 
-        # Check if the current block concluded any validator elections and
-        # update the locally tracked validator set
-        validator_update = ValidatorElection.approved_update(self.bigchaindb,
-                                                             self.new_height,
-                                                             self.block_transactions)
+        # Check the current block to see if any elections have concluded.
+        concluded_elections = Election.approved_elections(self.bigchaindb,
```
ldmberman: We should directly return the validator set update from `approved_elections` since preparing such an update based on the given block is its concern.

bigchaindb/elections/election.py
```diff
@@ -242,24 +242,32 @@ def show_election(self, bigchain):
...
             election = cls.has_concluded(bigchain, election_id, election_votes, new_height)
-            # Once an election concludes any other conclusion for the same
-            # or any other election is invalidated
             if election:
+                # Once we conclude an election, we store the result in the db
                 cls.store_election_results(bigchain, election, new_height)
-                return cls.on_approval(bigchain, election, new_height)
-        return None
+                # And keep the transaction filed under the election_type
+                elections[election_operation] = election.on_approval(bigchain, election, new_height)
```
ldmberman: We need to be able to foresee if a particular election updates the validator set and only persist (conclude) one of those if there are more than one. Elections that do not update the validator set are concluded too.

ldmberman: This is roughly how the conclusion can look like:
```
votes_by_election = defaultdict(list)
for txn in txns:
    if not isinstance(txn, Vote):
        continue
    election_id = txn.asset['id']
    votes_by_election[election_id].append(tx)

updated_validator_set = False  # only one validator set update is allowed per block
for election_id, votes in votes_by_election:
    election = Election.from_dict(bigchain.get_transaction(election_id))

    if not election.has_concluded(bigchain, election_id, votes, new_height):
        continue

    if election.get_validator_set() != bigchain.get_validators():
        if updated_validator_set:
            continue
        updated_validator_set = True

    election.on_approval(bigchain, new_height)
```
What do you think?

bigchaindb/commands/bigchaindb.py

bigchaindb/common/schema/transaction_migration_election_v2.0.yaml
kansi: > I do not see the yaml file to be a necessity 
I would like to point out that the yaml files for a particular election transaction define the `asset.data` field. So imho it seems necessary

bigchaindb/elections/election.py
z-bowen: I wanted to talk to you about this tomorrow.

The logic as I have it now allows for only one concluded election of a given type per block. This is desirable for `ValidatorElection`s, and doesn't really matter for `MigrationElection`s, but what do we want the generalized behavior to be?

Also, I think we can't easily avoid type-checking the `election`s. `Election.get_validators()` just gets the validator set from `bigchain`. The only place a `ValidatorElection` stores the new validator set is in its asset data. So it's probably easiest just to check the `OPERATION` or the class `.__name__`.

bigchaindb/core.py
z-bowen: I'm trying to avoid logic specific to a given election type in that method. My goal here is to have one pass over the block txs where I check for all concluded elections, and then let `end_block` do what it needs to with those results. If I have `approved_elections` return a validator set for an approved `ValidatorElection`, then I need to add custom logic to handle that election class differently than the others.

bigchaindb/elections/election.py
ldmberman: We have no reason to have a one per block restriction when it's not about the validator set update.

A way to set elections that update validator set apart is to introduce a generic method that reports the new validator set. It can return an empty list by default whereas `ValidatorElection` will infer it from the asset.

bigchaindb/common/schema/transaction_migration_election_v2.0.yaml
ldmberman: Discussed offline - we do not strictly need a schema for the migration election. Transactions follow the base election schema described in [BEP-18](https://github.com/bigchaindb/BEPs/pull/82) and additionally have a custom `TENDERMINT_MIGRATION` operation.

bigchaindb/elections/election.py
```diff
@@ -35,7 +35,7 @@ class Election(Transaction):
     ELECTION_THRESHOLD = 2 / 3
 
     @classmethod
-    def get_validator_change(cls, bigchain, height=None):
+    def get_validator_change(cls, bigchain):
         """Return the latest change to the validator set
```
ldmberman: It's a good time to update this docstring:

- there is no `election_id` in the collection anymore
- it does not return the latest change but the change applied at the latest block height

bigchaindb/commands/election_types.py
kansi: The election transaction `operation` would also need to be changed to `CHAIN_MIGRATION_ELECTION`

tests/upsert_validator/test_upsert_validator_vote.py
z-bowen: Fixed

bigchaindb/commands/election_types.py
z-bowen: Done. I also refactored the Class name etc.

bigchaindb/common/schema/transaction_migration_election_v2.0.yaml
z-bowen: Deferred to another PR. https://github.com/bigchaindb/bigchaindb/issues/2542

bigchaindb/elections/election.py
```diff
@@ -242,24 +244,41 @@ def show_election(self, bigchain):
...
+
+            election.on_approval(bigchain, election, new_height)
+
+        return validator_set_change
+
+    def makes_validator_set_change(self):
+        return self.CHANGES_VALIDATOR_SET
+
+    def get_validator_set_change(self, bigchain, new_height):
+        if self.makes_validator_set_change():
```
kansi: This `if` seems unnecessary i.e. looking at the logic from `approved_elections` this function will only be called if `makes_validator_set_change()` returns `True`?

z-bowen: The idea is that `get_validator_set_change` is a generic function, that is safe to call on any approved election. If the election has `CHANGES_VALIDATOR_SET == False`, then nothing happens. If the election has `CHANGES_VALIDATOR_SET == True`, then we execute the custom code to change the validator set.

The reason I built it this way is so that I can set the base `Election` class with `CHANGES_VALIDATOR_SET == True` and have the `change_validator_set` function raise a `NotImplementedError`. This addressed the concern Lev raised that someone could extend the base class and not understand how validator updates work. Anyone who makes a custom election class needs to either explicitly set `CHANGES_VALIDATOR_SET` to `False`, or deal with implementing the `change_validator_set` function, so in either case, they can't ignore the validator set issue.

bigchaindb/elections/election.py
```diff
@@ -242,24 +244,41 @@ def show_election(self, bigchain):
...
-            if election:
-                cls.store_election_results(bigchain, election, new_height)
-                return cls.on_approval(bigchain, election, new_height)
-        return None
+        validator_set_updated = False
+        validator_set_change = []
+        for election_id, votes in elections.items():
+            election = bigchain.get_transaction(election_id)
+
+            if not election.has_concluded(bigchain, election_id, votes, new_height):
```
kansi: `has_concluded` also called `bigchain.get_transaction(election_id)`. It would be good to avoid this call.

bigchaindb/elections/election.py
ldmberman: @z-bowen such an interface might be helpful indeed but one does not expect the function named `get_validator_set_change` to update the database. What about doing the same before calling `on_approval`?

```
if self.CHANGES_VALIDATOR_SET:
    self.change_validator_set()
election.on_approval(...)
```

bigchaindb/elections/election.py
```diff
@@ -242,24 +244,41 @@ def show_election(self, bigchain):
...
+                if validator_set_updated:
+                    continue
+                validator_set_change.append(election.get_validator_set_change(bigchain, new_height))
+                validator_set_updated = True
+
+            election.on_approval(bigchain, election, new_height)
+
+        return validator_set_change
+
+    def makes_validator_set_change(self):
```
ldmberman: There is no need to wrap the field in a method. We can make it a `@property` in the future if we need to transform it on read.

bigchaindb/elections/election.py
kansi: I agree but it would be much easy to understand if the change to the validator set or any state mutation is done in `on_approval` method and nowhere else.

ldmberman: Yeah, that's a tradeoff. I am fine with doing everything in `on_approval` too.

tests/elections/test_election.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_approved_elections_no_elections(b):
+    txns = []
+    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
+    mock_chain_migration.assert_not_called()
+    mock_store_validator.assert_not_called()
+
+
+def run_approved_elections(bigchain, txns):
```
kansi: I would suggest to create new fixture as follows,

```python
@fixutre
def b_mock(b):
    mock_chain_migration = MagicMock()
    mock_store_validator = MagicMock()
    b.migrate_abci_chain = mock_chain_migration
    b.store_validator_set = mock_store_validator
    return b
```

and then use this `b_mock` in all the tests,

```python

def test_election_something(b_mock):
   txns = ['a_lot_of_election_votes']
    Election.approved_elections(bigchain, 1, txns) 
    b_mock.migrate_abci_chain.assert_not_called()
    b_mock.store_validator.assert_not_called()
```

ldmberman: I am actually thinking about using the test database instead of mocks here.

tests/elections/test_election.py
```diff
@@ -0,0 +1,64 @@
...
+        ongoing_validator_election, validator_election_votes,
+        ongoing_validator_election_2, validator_election_votes_2,
+        ongoing_chain_migration_election, chain_migration_election_votes
+):
+    txns = validator_election_votes + \
+           validator_election_votes_2 + \
+           chain_migration_election_votes
+    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
+    mock_chain_migration.assert_called_once()
+    mock_store_validator.assert_called_once()
```
kansi: This test should be done without mocking i.e. the test should assert that,
- First validator election concludes and updates the validator collection
- Second validator election doesn't update the validator collection
- The chain migration is concluded and the new chain is read and assert form `abci_chains` collection

tests/elections/test_election.py
```diff
@@ -0,0 +1,64 @@
...
+        b,
+        ongoing_validator_election, validator_election_votes,
+        ongoing_chain_migration_election, chain_migration_election_votes,
+        ongoing_chain_migration_election_2, chain_migration_election_votes_2
+):
+    txns = validator_election_votes + \
+           chain_migration_election_votes + \
+           chain_migration_election_votes_2
+    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
+    assert mock_chain_migration.call_count == 2
```
kansi: This is an interesting test. Can two migration elections be concluded knowing that the first migration was never applied?

tests/elections/test_election.py
```diff
@@ -0,0 +1,64 @@
...
+def test_approved_elections_one_migration_one_upsert(
+        b,
+        ongoing_validator_election, validator_election_votes,
+        ongoing_chain_migration_election, chain_migration_election_votes
+):
+    txns = validator_election_votes + \
+           chain_migration_election_votes
+    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
+    mock_chain_migration.assert_called_once()
+    mock_store_validator.assert_called_once()
```
kansi: I would suggest to re-write this test and exclude mocking.

PR #2538, opened 2018-09-13
Update the Quickstart page

ttmc:
- The URL of the "Get Started" page changed so I updated that.
- The testnet.bigchaindb.com site is going away (because it's hosted by 3scale) and the BigchainDB "Get Started" page will have all the information they'll need to connect to the BigchainDB Testnet, so just link there.
- The "Old Quickstart" subsection isn't really needed any more because:
  - The BigchainDB "Get Started" page will have a prominent link to "Install BigchainDB Server" and it will link to the [Simple Deployment Template](http://docs.bigchaindb.com/projects/server/en/master/simple-deployment-template/index.html).
  - The [Simple Deployment Template](http://docs.bigchaindb.com/projects/server/en/master/simple-deployment-template/index.html) follows soon after the Quickstart page in the table of contents.

### DIFF
--- a/docs/server/source/quickstart.md
+++ b/docs/server/source/quickstart.md
@@ -51,26 +51,18 @@ Code is Apache-2.0 and docs are CC-BY-4.0
 Create a BigchainDB transaction and post it to a BigchainDB network in 20 seconds or less:
 
 <div class="buttondiv">
-    <a class="button" href="https://www.bigchaindb.com/getstarted/">Try BigchainDB Now</a>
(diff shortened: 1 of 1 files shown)


ttmc:
On second thought, this PR could be merged now, since it's consistent with how things work now, and how things _will_ work in the future.

vrde:
Looks good to me. I think we should keep a link on how to run a local node as a secondary option.

ttmc:
Okay I added links to docs about how to deploy a local node.

ttmc:
Codecov#$!$%&###!!

PR #2541, opened 2018-09-13
Problem: Users trying to use an incompatible version of Tendermint

### LINKED ISSUE #2525: Problem: Users trying to use an incompatible version of Tendermint
Solution:

- [ ] Improve the docs. PR #2524 is a step, but the following may also imply changes in docs.
- [ ] Hard-wire the supported Tendermint version as a constant right in the code of BigchainDB Server. Check the version of Tendermint, if possible, when connecting to it, and respond appropriately: maybe give a helpful error, maybe disconnect, maybe quit?
- [ ] Add a new command-line command `bigchaindb tendermint-version` to get the supported version of Tendermint. Useful for scripts etc.

muawiakh:
## Solution

- Hard-wire the supported Tendermint version(s) right in the code of BigchainDB Server. Check the version of Tendermint and disconnect if Tendermint version is an unsupported one.
- Expose a CLI command `bigchaindb tendermint-version` to display the supported Tendermint versions.
- PR also takes care the long list of warnings we get when we run tests.
  - Updated deprecated pymongo methods
  - Do not call pytest fixtures directly.
- Also added the doc for the new cli command

## Issues Resolved

Resolves #2525


docs/server/source/server-reference/bigchaindb-cli.md
```diff
@@ -158,4 +158,8 @@ $ bigchaindb election show ELECTION_ID
 status=<status>
 ```
 
-The election data is the same set of arguments used in the `election new` command that originally triggered the election. `status` takes three possible values, `ongoing`, if the election has not yet reached a 2/3 majority, `concluded`, if the election reached the 2/3 majority needed to pass, or `inconclusive`, if the validator set changed while the election was in process, rendering it undecidable.
\ No newline at end of file
+The election data is the same set of arguments used in the `election new` command that originally triggered the election. `status` takes three possible values, `ongoing`, if the election has not yet reached a 2/3 majority, `concluded`, if the election reached the 2/3 majority needed to pass, or `inconclusive`, if the validator set changed while the election was in process, rendering it undecidable.
+
+## bigchaindb tendermint-version
+
+Show the Tendermint versions supported by BigchainDB server.
```
kansi: Please add an example with a sample output listed

tests/backend/localmongodb/test_queries.py
```diff
@@ -58,7 +58,7 @@ def test_write_assets():
     cursor = conn.db.assets.find({}, projection={'_id': False})\
                            .sort('id', pymongo.ASCENDING)
 
-    assert cursor.count() == 3
+    assert cursor.collection.count_documents({}) == 3
```
kansi: Is there a specific reason for this change? CI seems to build fine regardless

muawiakh: I know, but if you see here, there is a tremendous amount of noise after the tests pass. This is just resolving those deprecation warnings.

docs/server/source/server-reference/bigchaindb-cli.md
muawiakh: done.

tests/backend/localmongodb/test_queries.py
muawiakh: e.g. here: https://travis-ci.org/bigchaindb/bigchaindb/jobs/428168863

bigchaindb/core.py
```diff
@@ -115,6 +117,17 @@ def info(self, request):
...
+        # Check if BigchainDB supports the Tendermint version
+        try:
+            if check_tendermint_version(request.version):
+                logger.info(f"Tendermint version: {request.version}")
+            else:
+                logger.error(f"Error: Tendermint version {request.version} not supported."
+                             f" Currently, BigchainDB only supports {__tm_supported_versions__} .Exiting!")
+                sys.exit(1)
+        except AttributeError:
+            logger.debug("Skipping Tendermint version check.")
```
ldmberman: According to the docs, the version field is not optional, so I would as well panic when there is no version.

bigchaindb/utils.py
```diff
@@ -185,3 +186,23 @@ def load_node_key(path):
...
+
+def check_tendermint_version(running_tm_ver):
+    """
+    Check Tendermint compatability with BigchainDB server
+
+    :param running_tm_ver: Version number of the connected Tendermint instance
+    :type running_tm_ver: str
+    :return: True/False depending on the compatability with BigchainDB server
+    :rtype: bool
+    """
```
ldmberman: A neatpick:
```
if not tm_ver:
    return False

for ver in __tm_supported_versions__:
    if version.parse(ver) == version.parse(tm_ver[0]):
        return True

return False
```
Also, in this version it does not fail if the first match fails but only if it does not match any version in the list.

bigchaindb/core.py
```diff
@@ -115,6 +117,17 @@ def info(self, request):
 
         self.abort_if_abci_chain_is_not_synced()
 
+        # Check if BigchainDB supports the Tendermint version
+        try:
+            if check_tendermint_version(request.version):
+                logger.info(f"Tendermint version: {request.version}")
+            else:
+                logger.error(f"Error: Tendermint version {request.version} not supported."
+                             f" Currently, BigchainDB only supports {__tm_supported_versions__} .Exiting!")
```
ttmc: The spacing is slightly off near the end of this line. It should end with `}. Exiting!")`

bigchaindb/core.py
```diff
@@ -115,6 +117,17 @@ def info(self, request):
 
         self.abort_if_abci_chain_is_not_synced()
 
+        # Check if BigchainDB supports the Tendermint version
+        try:
+            if check_tendermint_version(request.version):
```
ttmc: This function (currently named `check_tendermint_version()`) returns a Boolean, so it could be renamed so that lines like this read more like plain English, e.g.

```python
if tendermint_version_is_compatible(request.version):
```

In other words, I'm suggesting renaming the function `check_tendermint_version()` as `tendermint_version_is_compatible()`.

bigchaindb/commands/bigchaindb.py
```diff
@@ -280,6 +281,16 @@ def run_start(args):
     start(args)
 
 
+def run_tendermint_version(args):
+    """Show the supported Tendermint version(s)"""
+    supported_tm_ver = {
+        "description": "BigchainDB supports the following Tendermint version(s)",
+        "tendermint": [],
+    }
+    [supported_tm_ver["tendermint"].append(ver) for ver in __tm_supported_versions__]
```
ldmberman: It may be slightly more concise:
```
supported_tm_ver = {
    'description': 'BigchainDB supports the following Tendermint version(s)',
    'tendermint': __tm_supported_versions__
}
```
On a side note, PEP8 recommends to stick to a single type of quotes, and we generally use `'`.

bigchaindb/core.py
ttmc: Maybe we should say _why_ the Tendermint version check is being skipped: apparently the `request` from Tendermint didn't include a `version` attribute.

bigchaindb/utils.py
```diff
@@ -185,3 +186,23 @@ def load_node_key(path):
...
+    tm_ver = running_tm_ver.split('-')
```
ttmc: Maybe put a comment with an example value of `running_tm_ver` so it's clear why you're doing this split.

bigchaindb/core.py
```diff
@@ -115,6 +117,18 @@ def info(self, request):
 
         self.abort_if_abci_chain_is_not_synced()
 
+        # Check if BigchainDB supports the Tendermint version
```
ldmberman: One more neatpick:

```
if not hasattr(request, 'version') or not tendermint_version_is_compatible(request.version):
    logger.error('Unsupported Tendermint version: {getattr(request, "version", "no version")} ..')
    sys.exit(1)
```

bigchaindb/utils.py
```diff
@@ -185,3 +186,24 @@ def load_node_key(path):
...
+    """
+
+    # Splitting because version can look like this e.g. 0.22.8-40d6dc2e
+    tm_ver = running_tm_ver.split('-')
+    if not tm_ver:
+        return False
+    for ver in __tm_supported_versions__:
+        if version.parse(ver) == version.parse(tm_ver[0]):
+            return True
+    else:
```
ldmberman: `return False` belongs to the outer scope, not to `if-else`.

bigchaindb/core.py
ldmberman: A slight correction: `not (hasattr(request, 'version') and tendermint_version_is_compatible(request.version))`

tests/commands/test_commands.py
```diff
@@ -468,11 +471,21 @@ def test_election_approve_called_with_bad_key(caplog, b, bad_validator_path, new
...
+    assert len(output_config["tendermint"]) == len(__tm_supported_versions__)
+    assert sorted(output_config["tendermint"]) == sorted(__tm_supported_versions__)
+
+
 def mock_get_validators(height):
-    keys = node_keys()
-    pub_key = list(keys.keys())[0]
     return [
-        {'public_key': {'value': pub_key,
+        {'public_key': {'value': "zL/DasvKulXZzhSNFwx4cLRXKkSM9GPK7Y0nZ4FEylM=",
```
kansi: What is the need for this change?

PR #2547, opened 2018-09-17
Problem: send function has a deprecation warning in the driver

codegeschrei:
Solution: replace send with the matching new functions

As you might have noticed when running the test, the driver has a deprecation warning for the `send()` function to send a transaction. We use it in the acceptance tests. Since we want to remove the `send()` in the next driver release, I'm preparing the server for the changes.


ttmc:
@codegeschrei I guess you did a search for all "send(" strings, or something like that?

codegeschrei:
@ttmc I went through the files of the acceptance tests and searched for `send`
Did I miss one?

PR #2553, opened 2018-09-18
Adjust multiple elections conclusion.

ldmberman:
- Do not conclude migration election if there is a migration in progress.
- Rewrite election tests to not use mocks and assert many different things.
- Record concluded elections in the `election` collection.

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -283,12 +283,11 @@ def store_validator_set(conn, validators_update):
 
 @register_query(LocalMongoDBConnection)
 def store_election_results(conn, election):
-    height = election['height']
     return conn.run(
         conn.collection('elections').replace_one(
-            {'height': height},
+            {'election_id': election['election_id']},
             election,
-            upsert=True
+            upsert=True,
         )
     )
 
--- a/tests/elections/test_election.py
+++ b/tests/elections/test_election.py
@@ -1,64 +1,143 @@
-from unittest.mock import MagicMock
-
 import pytest
 
+from tests.utils import generate_election, generate_validators
+
+from bigchaindb.lib import Block
 from bigchaindb.elections.election import Election
+from bigchaindb.migrations.chain_migration_election import ChainMigrationElection
+from bigchaindb.upsert_validator.validator_election import ValidatorElection
 
 
 @pytest.mark.bdb
-def test_approved_elections_one_migration_one_upsert(
-        b,
-        ongoing_validator_election, validator_election_votes,
-        ongoing_chain_migration_election, chain_migration_election_votes
-):
-    txns = validator_election_votes + \
-           chain_migration_election_votes
-    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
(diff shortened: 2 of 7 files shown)


bigchaindb/backend/localmongodb/query.py
```diff
@@ -283,12 +283,9 @@ def store_validator_set(conn, validators_update):
 
 @register_query(LocalMongoDBConnection)
 def store_election_results(conn, election):
-    height = election['height']
     return conn.run(
-        conn.collection('elections').replace_one(
-            {'height': height},
+        conn.collection('elections').insert_one(
```
kansi: I would recommend using `replace_one` because in an event of a crash the replay won't be successful.

tests/elections/test_election.py
```diff
@@ -1,64 +1,143 @@
...
+                                        ChainMigrationElection,
+                                        public_key, private_key,
+                                        {})
+
+    txs += [election]
+    total_votes += votes
+
+    b.store_abci_chain(1, 'chain-X')
+    b.store_block(Block(height=1,
+                        transactions=[tx.to_dict() for tx in txs],
```
kansi: `transactions` is `Block` only store the transaction ids.

tests/elections/test_election.py
```diff
@@ -1,64 +1,143 @@
...
+
+    election, votes = generate_election(b,
+                                        ValidatorElection,
+                                        public_key, private_key,
+                                        another_validator['election'])
+    txs += [election]
+    total_votes += votes
+
+    b.store_block(Block(height=1,
+                        transactions=[tx.to_dict() for tx in txs],
```
kansi: `transactions` is `Block` only store the transaction ids.

tests/elections/test_election.py
```diff
@@ -1,64 +1,143 @@
...
+                                        ChainMigrationElection,
+                                        public_key, private_key,
+                                        {})
+
+    txs += [election]
+    total_votes += votes
+
+    b.store_abci_chain(1, 'chain-X')
+    b.store_block(Block(height=1,
+                        transactions=[tx.to_dict() for tx in txs],
```
kansi: `transactions` in `Block` only store the transaction ids.

tests/elections/test_election.py
```diff
@@ -1,64 +1,143 @@
...
-        b,
-        ongoing_validator_election, validator_election_votes,
-        ongoing_chain_migration_election, chain_migration_election_votes
-):
-    txns = validator_election_votes + \
-           chain_migration_election_votes
-    mock_chain_migration, mock_store_validator = run_approved_elections(b, txns)
-    mock_chain_migration.assert_called_once()
-    mock_store_validator.assert_called_once()
+def test_approved_elections_concludes_all_elections(b):
```
kansi: I would also suggest writing a test wherein the order of validator election and migration witnessed is reversed. This will allow us to assert the rule that a block with validator and migration election concludes both regardless of their order.

ldmberman: Good point! But I would do it in the next PR where I implement [this](https://github.com/bigchaindb/BEPs/pull/82#discussion_r218057405).

PR #2556, opened 2018-09-18
Separate pending and effective validator updates.

ldmberman:
- Pending validator updates do not prevent elections from concluding.
- ValidatorElection overrides `has_concluded` to not conclude when there is a pending update for the matching height.
- Effective validator updates deem past elections inconclusive.

### DIFF
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -282,11 +282,14 @@ def store_validator_set(conn, validators_update):
 
 
 @register_query(LocalMongoDBConnection)
-def store_election_results(conn, election):
+def store_election(conn, election_id, height, is_concluded):
     return conn.run(
         conn.collection('elections').replace_one(
-            {'election_id': election['election_id']},
-            election,
+            {'election_id': election_id,
+             'height': height},
+            {'election_id': election_id,
+             'height': height,
+             'is_concluded': is_concluded},
             upsert=True,
         )
     )
@@ -315,6 +318,7 @@ def get_election(conn, election_id):
     cursor = conn.run(
         conn.collection('elections')
         .find(query, projection={'_id': False})
+        .sort([('height', DESCENDING)])
     )
 
     return next(cursor, None)
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -296,6 +296,13 @@ def store_election(conn, election_id, height, is_concluded):
 
 
 @register_query(LocalMongoDBConnection)
+def store_elections(conn, elections):
+    return conn.run(
+        conn.collection('elections').insert_many(elections)
+    )
+
+
+@register_query(LocalMongoDBConnection)
 def get_validator_set(conn, height=None):
     query = {}
     if height is not None:
--- a/bigchaindb/backend/localmongodb/query.py
+++ b/bigchaindb/backend/localmongodb/query.py
@@ -322,13 +322,11 @@ def get_validator_set(conn, height=None):
 def get_election(conn, election_id):
     query = {'election_id': election_id}
 
-    cursor = conn.run(
+    return conn.run(
         conn.collection('elections')
-        .find(query, projection={'_id': False})
-        .sort([('height', DESCENDING)])
+        .find_one(query, projection={'_id': False},
+                  sort=[('height', DESCENDING)])
     )
-
-    return next(cursor, None)
 
 
 @register_query(LocalMongoDBConnection)
--- a/bigchaindb/elections/election.py
+++ b/bigchaindb/elections/election.py
@@ -207,8 +207,7 @@ class Election(Transaction):
         votes_committed = self.get_commited_votes(bigchain, election_pk)
         votes_current = self.count_votes(election_pk, current_votes)
 
-        current_validators = self.get_validators(bigchain)
-        total_votes = sum(current_validators.values())
+        total_votes = sum(output.amount for output in self.outputs)
         if (votes_committed < (2/3) * total_votes) and \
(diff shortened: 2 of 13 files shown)


bigchaindb/elections/election.py
```diff
@@ -199,28 +204,38 @@ def has_concluded(self, bigchain, current_votes=[], height=None):
...
         return False
 
     def get_status(self, bigchain):
         concluded = self.get_election(self.id, bigchain)
         if concluded:
             return self.CONCLUDED
 
+        return self.INCONCLUSIVE if self.has_validator_set_changed(bigchain) else self.ONGOING
+
+    def has_validator_set_changed(self, bigchain):
```
kansi: Why is this approach preferable over checking the election outputs?

ldmberman: It is simpler. I am fine with doing it differently, but we should use the same approach in `get_status` and `has_concluded`.

kansi: I don't deem either of the approaches simple or difficult but this approach has database calls which can be avoided with the other approach. Since this code will be triggered during `end_block` it would prefer to have the approach with the least amount of database calls.

> we should use the same approach in `get_statu` and `has_concluded`

agreed.

ldmberman: Ah, I almost forgot! The output approach does not fit here cause the validator set can be changed and then restored to the original value while we treat elections as inconclusive as soon as the first validator set change happens.

bigchaindb/elections/election.py
```diff
@@ -199,36 +204,43 @@ def has_concluded(self, bigchain, current_votes=[], height=None):
...
         latest_change_height = latest_change['height']
-        election_height = bigchain.get_block_containing_tx(self.id)[0]
 
-        if latest_change_height >= election_height:
-            return self.INCONCLUSIVE
-        else:
-            return self.ONGOING
+        election = self.get_election(self.id, bigchain)
+        if not election:
+            return True
```
kansi: It would good to write a code comment describing this `if` condition

bigchaindb/elections/election.py
```diff
@@ -243,45 +255,53 @@ def show_election(self, bigchain):
...
+
+           `on_approval` methods are implemented by elections of particular type.
+           The method may contain side effects but should be idempotent. To account
+           for other concluded elections, if it requires so, the method should
+           rely on the database state.
+        """
+        elections = OrderedDict()
         for tx in txns:
+            if isinstance(tx, Election):
+                tx.store(bigchain, new_height, is_concluded=False)
```
kansi: I would recommend to aggregate and store

bigchaindb/elections/election.py
```diff
@@ -199,36 +204,43 @@ def has_concluded(self, bigchain, current_votes=[], height=None):
...
         votes_current = self.count_votes(election_pk, current_votes)
-        current_validators = self.get_validators(bigchain, height)
 
-        if self.is_same_topology(current_validators, self.outputs):
-            total_votes = sum(current_validators.values())
-            if (votes_committed < (2/3) * total_votes) and \
-                    (votes_committed + votes_current >= (2/3)*total_votes):
-                return True
+        if self.has_validator_set_changed(bigchain):
+            return False
```
kansi: It would be better to move this check to the beginning of the function so that the votes are not counted unnecessarily.

ldmberman:
Election records are inserted in bulk now.

There is also the crash recovery issue (already in master) which is not only connected with election records but all database artifacts we create during `end_block`. In order to address it, we need to extend the crash recovery which happens on startup to account for all the new data that was introduced. I will make it in a separate PR.

bigchaindb/elections/election.py
```diff
@@ -186,49 +191,55 @@ def get_commited_votes(self, bigchain, election_pk=None):
...
         votes_current = self.count_votes(election_pk, current_votes)
-        current_validators = self.get_validators(bigchain, height)
 
-        if self.is_same_topology(current_validators, self.outputs):
-            total_votes = sum(current_validators.values())
-            if (votes_committed < (2/3) * total_votes) and \
-                    (votes_committed + votes_current >= (2/3)*total_votes):
-                return True
+        current_validators = self.get_validators(bigchain)
+        total_votes = sum(current_validators.values())
```
kansi: It occured to me that one could sum the `"amount"` of the outputs of the election to get the `total_votes` and avoid the query.

bigchaindb/backend/localmongodb/query.py
```diff
@@ -315,6 +325,7 @@ def get_election(conn, election_id):
     cursor = conn.run(
         conn.collection('elections')
         .find(query, projection={'_id': False})
+        .sort([('height', DESCENDING)])
```
kansi: It would be better to use `find_one` and `sort` as an argument to the query.

bigchaindb/elections/election.py
```diff
@@ -186,49 +191,54 @@ def get_commited_votes(self, bigchain, election_pk=None):
...
 
-        if latest_change_height >= election_height:
-            return self.INCONCLUSIVE
-        else:
-            return self.ONGOING
+        election = self.get_election(self.id, bigchain)
+        if not election:
+            return True
+
+        return latest_change_height > election['height']
```
kansi: The domain in which the above condition with return `false` is as follows,
- `latest_change_height == election['height']`
- `latest_change_height < election['height']`

I understand that the second point will never occur but if we do `return latest_change_height != election['height']` then it would not return `false` for the second point and the logic becomes more transparent.

kansi:
This code is a dependency for fixing crash recovery so I will merge it. @vrde Please leave your comments and @ldmberman will address them in a separate a PR.

PR #2580, opened 2018-10-10
Edit nginx/nginx.conf to enable CORS

ttmc:
I'm not sure if this will work. I based it on:

https://github.com/bigchaindb/bigchaindb/blob/master/k8s/nginx-http/container/nginx.conf.template

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -82,6 +82,22 @@ http {
       # max client request body size: average transaction size.
       client_max_body_size 15k;
 
+      # GET requests are forwarded to BDB.
+      if ($request_method = GET) {
+        proxy_pass http://localhost:9984;
+      }
+
+      # POST requests: Enable CORS then forward to BDB.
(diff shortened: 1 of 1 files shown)


ttmc:
Before I merge this PR, I will test the change in the testnet.

ttmc:
I tested it and it didn't work. Then I looked at the nginx.conf for the web proxy, at:

and noticed this line in the 'OPTIONS' block had more stuff:

`add_header 'Access-Control-Allow-Headers' 'DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range,app_key,app_id';`

https://github.com/bigchaindb/bigchaindb/blob/master/k8s/nginx-https-web-proxy/container/nginx.conf.template

So I changed that line in the live testnet, and then the "Get Started" page worked, yay! I will make a commit to update this PR accordingly.

PR #2594, opened 2018-11-03
Update acceptance Dockerfile to use latest Python driver

ttmc:


ttmc:
Codecov is complaining but I don't care. Merging now!

PR #2600, opened 2018-11-06
Fix links in network setup readme file

sd189:
## Solution

Updating the links in the [network setup](https://github.com/bigchaindb/bigchaindb/blob/master/docs/server/source/simple-deployment-template/network-setup.md)

### DIFF
--- a/docs/server/source/simple-deployment-template/network-setup.md
+++ b/docs/server/source/simple-deployment-template/network-setup.md
@@ -197,7 +197,7 @@ If you want to start and manage the BigchainDB and Tendermint processes yourself
 
 ## How Others Can Access Your Node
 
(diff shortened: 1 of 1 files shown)


ttmc:
Those links _are_ broken, but changing the extension from `.rst` to `.html` won't fix them. It comes down to how Sphinx works when rendering links in Markdown files. Markdown isn't very sophisticated; you basically have to put the link to the final HTML file (which can be relative). It seems the problem here is that the relative path of the linked pages changed, but the links weren't updated. I'll do a quick fix now, but the "latest" (default) BigchainDB docs won't be updated online until we do a new release of BigchainDB.

ttmc:
Thanks for pointing this out, @sd189 . Merging now!

PR #2625, opened 2019-01-23
pip 19.1 has deprecated --process-dependency-links

Chuseuiti:
## Problem 
PIP 19.1 has deprecated --process-dependency-links 
## Solution
To fix the deprecation problem it was needed to fix the version on pip to 18.1. I would recommend to fix the other versions not to run into other similar bugs.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -4,7 +4,7 @@ LABEL maintainer "devs@bigchaindb.com"
 
 RUN apt-get update \
     && apt-get install -y vim \
-    && pip install -U pip \
+    && pip install -U pip==18.1 \
     && pip install pynacl \
     && apt-get autoremove \
     && apt-get clean


Dockerfile-dev
```diff
@@ -4,7 +4,7 @@ LABEL maintainer "devs@bigchaindb.com"
 
 RUN apt-get update \
     && apt-get install -y vim \
-    && pip install -U pip \
+    && pip install -U pip==18.1 \
```
ttmc: Maybe we could just remove the `--process-dependency-links` option below (currently in line 35) instead?

ttmc: I did a search for `dependency` in this repo and it looks like that `--process-dependency-links` option shows up in many other places as well.

Maybe all of them could be removed, _and_ we could require `pip>=18.1`?

Chuseuiti: I agree, to remove the flag is a better option if it is not used, on my side I wasn't sure if there are  non-PyPi packages during installation that may be affected by removing  `--process-dependency-link`.

ttmc:
I ended up removing all `--process-dependency-links` as part of pull request https://github.com/bigchaindb/bigchaindb/pull/2627

Thanks for bringing this issue to our attention!

PR #2628, opened 2019-01-29
Remove --no-cache-dir from all pip install cmds

ttmc:
because it broke in pip 19

See https://github.com/pypa/pip/issues/6158


ttmc:
I agree to license this contribution, and all my contributions going forward, under an open source license.

PR #2667, opened 2019-10-04
Transcation API extension: last_tx parameter

eckelj:
IoT devices do have limited memory capacities but need to be able to inspect the latest transaction for a given asset very quickly. This is not possible with the given API and long asset/token lifecycles in mind.

## Solution

Very often only the lastest transaction of an asset is of interest to the IoT device. Therefore, we extended the transaction API so that only the latest transaction is returned for a given assset_id.


bigchaindb/web/views/transactions.py
```diff
@@ -48,12 +48,22 @@ def get(self):
...
         args = parser.parse_args()
-
         with current_app.config['bigchain_pool']() as bigchain:
             txs = bigchain.get_transactions_filtered(**args)
-
-        return [tx.to_dict() for tx in txs]
+            if args['last_tx'] and args['last_tx'] is True:
+                lastTX = None
+
+                for x in txs:
```
davie-lab: What is going on here? Are you just trying to get the last element of the iterable like that?

davie0 (changes requested):
This implementation is wasting computing time. It will iterate over every single transaction each time someone wants to access last_tx. I think we might need to dig deeper and implement it in query module.

PR #2689, opened 2020-04-13
Updated bigchaindb-abci

zzappie:
Reason: gevent API change breaks BigchainDB installation
Signed-off-by: David Dashyan <mail@davie.li>


eckelj:
Updated bigchaindb-abci, Reason: gevent API change breaks BigchainDB installation

